PR #19, opened 2023-11-02
Change authenticator_maps for UI

john-westcott-iv:
Adds name field to authenticator_maps.
Also adds role and org authenticator_maps in addition to teams.


ansible_base/serializers/authenticator_map.py
```diff
@@ -23,10 +23,15 @@ def validate(self, data) -> dict:
         map_type = data.get('map_type', None)
         team = data.get('team', None)
         org = data.get('organization', None)
+        role = data.get('role', None)
         if map_type == 'team' and not team:
             errors["team"] = "You must specify a team with the selected map type"
         if map_type == 'team' and not org:
             errors["organization"] = "You must specify an organization with the selected map type"
+        if map_type == 'organization' and not org:
```
relrod, line 30: I am guessing these need changes/additions in authentication.common too?

john-westcott-iv: Yea, I was thinking I would do it later but I suppose now is as good of time as any.

ansible_base/authentication/common.py
```diff
@@ -101,16 +106,26 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
...
 
     return {
         "access_allowed": access_allowed,
         "is_superuser": is_superuser,
         "is_system_auditor": is_system_auditor,
-        "claims": org_team_mapping,
+        "claims": {
+            "team_membership": org_team_mapping,
+            "roles": roles_mapping,
+            "org_membership": org_membership,
```
relrod: Nitpick: can we standardize on "org" or "organization"? The map type is "organization" but here it's "org" - I want to make it easy for API users to intuit which one to use.

ansible_base/authentication/common.py
```diff
@@ -307,4 +322,21 @@ def update_user_claims(user, database_authenticator, groups):
     if results['access_allowed'] is not True:
         logger.warning(f"User {user.username} failed an allow map and was denied access")
         return None
+
+    # We have allowed access so now we need to make the user within the system
+    reconcile_class = getattr(settings, 'ANSIBLE_BASE_AUTHENTICATOR_RECONCILE_CLASS', 'ansible_base.authentication.common')
```
relrod: Probably `ANSIBLE_BASE_AUTHENTICATOR_RECONCILE_MODULE` instead, since it's not actually specifying the class.

ansible_base/authentication/common.py
john-westcott-iv: Anything user facing I try to do `organization` but internally I typically use `org`. That being said, we've talked about exposing the claims field in such a way that it could be user facing so I'll update this one too.

ansible_base/authentication/common.py
```diff
@@ -101,16 +106,26 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
...
+            logger.error(f"Map type {auth_map.map_type} of rule {auth_map.name} does not know how to be processed")
 
     return {
         "access_allowed": access_allowed,
         "is_superuser": is_superuser,
         "is_system_auditor": is_system_auditor,
-        "claims": org_team_mapping,
+        "claims": {
+            "team_membership": org_team_mapping,
+            "roles": roles_mapping,
```
newswangerd: I would remove roles here for the following reasons:

1. RBAC is still a work in progress, and I'm not sure if we'll have global roles. Roles may be required to be assigned to a resource such as an Organization.
2. Role associations can be created for the user via their org and team mappings. Users should set up their teams with the level of access that each team requires and then be grouped into those teams at login.

I can see how this would create problems for EDA since doesn't have teams or organizations yet, but I think it makes more sense for them to at least create teams for this feature.

relrod (approved):

PR #20, opened 2023-11-03
Fixing authetnticator_map for v1 page

john-westcott-iv:
On the api/gateway/v1 page the endpoint `authenticator_maps` was missing. 
Fixes issue induced in https://github.com/ansible/django-ansible-base/pull/12 which I approved.


relrod (approved):
oops

PR #28, opened 2023-11-17
Add SAML AuthenticatorPlugin

john-westcott-iv:


ansible_base/authenticator_plugins/saml.py
```diff
@@ -0,0 +1,191 @@
...
+        # pull the cert_info out of the existing object (if we have one)
+        cert_info = {
+            "SP_PRIVATE_KEY": getattr(self.instance, 'configuration', {}).get('SP_PRIVATE_KEY', None),
+            "SP_PUBLIC_CERT": getattr(self.instance, 'configuration', {}).get('SP_PUBLIC_CERT', None),
+        }
+
+        # Now get the cert_info out of the passed in attrs (if there is any)
+        for cert_type in cert_info.keys():
+            item = attrs.get(cert_type, None)
+            if item and item == ENCRYPTED_STRING and not self.instance:
```
relrod: But `SP_PUBLIC_CERT` isn't encrypted, right? At least, it's not in `configuration_encrypted_fields` below, so I think we can avoid looping here and just check `SP_PRIVATE_KEY`?

ansible_base/authenticator_plugins/saml.py
```diff
@@ -0,0 +1,191 @@
...
+                # Catch a case where we got input but it was ENCRYPTED and we don't have an object yet
+                errors[cert_type] = f"Can not be {ENCRYPTED_STRING} on creation"
```
relrod: Does the authenticator serializer already handle this case? It seems like maybe we can drop this check entirely?

https://github.com/ansible/django-ansible-base/blob/b2509574cd6b17a1e9c095c412a248278877129a/ansible_base/serializers/authenticator.py#L88-L90

ansible_base/authenticator_plugins/saml.py
```diff
@@ -0,0 +1,191 @@
...
+        # so there is an inefficiency here but it should be trivial
+        try:
+            validate_cert_with_key(cert_info['SP_PUBLIC_CERT'], cert_info['SP_PRIVATE_KEY'])
+        except ValidationError as e:
+            errors['SP_PRIVATE_KEY'] = e
+
+        idp_data = attrs.get('ENABLED_IDPS', {}).get(idp_string, {})
+        # TODO: Check to make sure either ID or Perminant ID is set
+        if not idp_data['attr_user_permanent_id'] and not idp_data['attr_username']:
+            errors['IDP_ATTR_USERNAME'] = "Either IDP_ATTR_USERNAME or IDP_ATTR_USER_PERMANENT_ID needs to be set"
```
relrod: Both the `IDP_ATTR_USERNAME` and `IDP_ATTR_USER_PERMANENT_ID` fields default to required, so I think this is dead code currently?

Should one or both of those be optional?

ansible_base/authenticator_plugins/saml.py
relrod: In fact this happens before the authenticator config's validate is called, so I think this check in saml.py is dead code currently.

ansible_base/authenticator_plugins/saml.py
```diff
@@ -0,0 +1,191 @@
...
+            raise serializers.ValidationError(errors)
+
+        response = super().validate(attrs)
+        return response
+
+    def to_internal_value(self, data):
+        resp = super().to_internal_value(data)
+        idp_data = {}
+        for field in self.settings_to_enabled_idps_fields:
+            if field in resp:
```
relrod, line 172: Always true right now since every field is required.

ansible_base/authenticator_plugins/saml.py
```diff
@@ -0,0 +1,191 @@
...
+                idp_data[self.settings_to_enabled_idps_fields[field]] = resp[field]
+                del resp[field]
+        resp['ENABLED_IDPS'] = {idp_string: idp_data}
+        return resp
+
+    def to_representation(self, configuration):
+        if 'ENABLED_IDPS' in configuration:
+            for config_setting_name in self.settings_to_enabled_idps_fields:
+                enabled_idp_field_name = self.settings_to_enabled_idps_fields[config_setting_name]
+                if enabled_idp_field_name in configuration['ENABLED_IDPS'][idp_string]:
```
relrod, line 182: Here too.

ansible_base/authenticator_plugins/saml.py
john-westcott-iv: Yes, initially I was doing more in this loop but since that got moved out we can just check the PRIVATE_KEY.

ansible_base/authenticator_plugins/saml.py
john-westcott-iv: Agreed, removed. I wonder if I started my local testing before that change merged? 🤷

ansible_base/authenticator_plugins/saml.py
john-westcott-iv: They shouldn't have been, fixed.

ansible_base/authenticator_plugins/saml.py
john-westcott-iv: Allowed username/permanent_id and groups to be null, I think those are the ones that don't have to be set but will confirm when testing.

ansible_base/authenticator_plugins/local.py
```diff
@@ -3,22 +3,30 @@
...
-#       Change the validator to not allow it to be deleted or a second one added
+# TODO: Change the validator to not allow it to be deleted or a second one added
+
+
+class LocalConfiguration(BaseAuthenticatorConfiguration):
+    documentation_url = "https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.backends.ModelBackend"
+
+    def validate_configuration(self, data: dict, instance: object) -> None:
+        if data != {}:
+            raise ValidationError({"configuration": "Can only be {} for local authenticators"})
```
relrod: So, `validate_configuration` is only on the plugin superclass, not the configuration superclass.

So I think this either needs to be:

```python
    def validate(self, data: dict) -> None:
        if data != {}:
            raise ValidationError({"configuration": "Can only be {} for local authenticators"})
        return data
```

or moved back to the plugin, though just moving it back to the plugin seems to cause weird SQL constraint issues that I don't fully understand. I guess the {} gets cast to None/null and SQL hates it.

john-westcott-iv: 100% refactored.

newswangerd (approved):

PR #29, opened 2023-11-21
Fix stacks raised by get_authenticator_plugin

john-westcott-iv:
When testing with an authenticator created from a plugin and the plugin gets removed or corrupted we got stacks. This PR makes some of them friendlier.


relrod (approved):
I think the overall change is fine, so approving. But it would be nice to have a few tests that would throw previously but pass after this change.

PR #30, opened 2023-11-21
Add UI labels for authenticator_plugin serializers

john-westcott-iv:
Introduce a MixIn class which will store the `ui_field_label` from the plugins' serializer fields.
Show the `ui_field_label` on the authenticator_plugin screen and default the value to `Undefined` if it was not properly added.
This field is for consumption by the UI to display the field names when configuring an authenticator_plugin.

### DIFF
--- a/ansible_base/authenticator_plugins/base.py
+++ b/ansible_base/authenticator_plugins/base.py
@@ -3,9 +3,10 @@ import logging
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
 from rest_framework.fields import empty
-from rest_framework.serializers import JSONField, ValidationError
+from rest_framework.serializers import ValidationError
 
 from ansible_base.models import Authenticator
+from ansible_base.serializers.fields import JSONField
 
 logger = logging.getLogger('ansible_base.authentication.authenticator_lib')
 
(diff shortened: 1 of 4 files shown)


ansible_base/authenticator_plugins/base.py
```diff
@@ -16,6 +17,7 @@ class BaseAuthenticatorConfiguration(serializers.Serializer):
         help_text="Any additional fields that this authenticator can take, they are not validated and passed directly back to the authenticator",
         required=False,
         allow_null=True,
+        ui_field_label='Additional Authenticator Fields',
```
relrod: Can/should we start _()ing new user-facing strings? We already import `_` above.

ansible_base/authenticator_plugins/keycloak.py
```diff
@@ -18,15 +17,29 @@ class KeycloakConfiguration(BaseAuthenticatorConfiguration):
         help_text=_("Location where this app can fetch the user's token from."),
         default="https://keycloak.example.com/auth/realms/<my_realm>/protocol/openid-connect/token",
         allow_null=False,
+        ui_field_label='Keycloak Access Token URL',
```
relrod: And these. We already do it for `help_text` :sweat_smile:

ansible_base/authenticator_plugins/base.py
```diff
@@ -29,7 +31,16 @@ def get_configuration_schema(self):
...
 
-            schema.append({"name": f, "help_text": field.help_text, "required": not field.allow_null, "default": default, "type": field.__class__.__name__})
+            schema.append(
+                {
+                    "name": f,
+                    "help_text": field.help_text,
+                    "required": not field.allow_null,
+                    "default": default,
+                    "type": field.__class__.__name__,
+                    "ui_field_label": getattr(field, 'ui_field_label', 'Undefined'),
```
relrod: And here

```suggestion
                    "ui_field_label": getattr(field, 'ui_field_label', _('Undefined')),
```

relrod (approved):
This looks good to me. You can merge as-is if you like. But as mentioned inline, I'm wondering if it's a good time to start `_()`ing new strings.

We can go back and get the existing ones later on, but maybe new ones should start doing it so we're moving in the right direction.

john-westcott-iv:
> But as mentioned inline, I'm wondering if it's a good time to start `_()`ing new strings.

Yea, we should probablly start doing that.

PR #31, opened 2023-11-24
[CI] Avoid using pyfakefs's fs fixture

relrod:
This took roughly a work day to track down. This issue only presented itself in certain conditions, based on how pytest-xdist split up the tests being run, into workers (e.g. a certain number of tests, and a certain value for -n). Originally the issue only showed up on GHA, but after much experimenting and shelling into the GHA Runner and playing around, I finally found that I was able to reproduce it locally as well by forcing -n to 4. But only on a branch that had more than a certain number of tests so that the grouping worked out to trigger the issue.

The issue manifested itself with this, seemingly completely unrelated error:

```
>       assert renderer, ".accepted_renderer not set on Response"
E       AssertionError: .accepted_renderer not set on Response

.tox/py311/lib/python3.11/site-packages/rest_framework/response.py:55: AssertionError
```

What is happening here is that the 'fs' fixture makes global changes to mock out the filesystem, and ultimately the issue is that it doesn't seem to clean up after itself (tear down) properly.

The exact details are a bit fuzzy to me still - and I've spent enough time trying to track this down - but what seems to be happening is that when it doesn't teardown properly, every other test in the same pytest-xdist worker that is expecting an error response fails spectacularly. In the end, Django tries to render an error template but it can't even find that, because the mocked filesystem is still in play.

The seemingly easy solution here, which seems to work fine, is to stop using the 'fs' fixture, and instead use a context manager to set up the files we need for this particular test. The context manager properly cleans up after itself, and things end up being green again.

### DIFF
--- a/ansible_base/tests/functional/authentication/test_authenticator_plugins_view.py
+++ b/ansible_base/tests/functional/authentication/test_authenticator_plugins_view.py
@@ -1,6 +1,7 @@
 import sys
 
 from django.urls import reverse
+from pyfakefs.fake_filesystem_unittest import Patcher
 
 
 def test_plugin_authenticator_view(admin_api_client):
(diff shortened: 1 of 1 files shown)


Zokormazo:
I wonder if it's not going to be easier to maintain long term with a real tmpdir with [tmp_path](https://docs.pytest.org/en/6.2.x/tmpdir.html)  than with a mocked one

Zokormazo (approved):

relrod:
> I wonder if it's not going to be easier to maintain long term with a real tmpdir with [tmp_path](https://docs.pytest.org/en/6.2.x/tmpdir.html) than with a mocked one

This is really testing "what happens if we try to import an authenticator plugin but can't?" - and really we don't even need to do it this way anymore, since now we can import plugins from other places. Might be a good change in the future, we could have some fake module and plugin, somewhere under `ansible_base/tests` and let it try to import that instead.

Zokormazo:
Makes sense. Thankx for the explanation

PR #32, opened 2023-11-25
[CI] Drop pyfakefs, use custom auth fixture

relrod:
- Introduces a custom auth fixture with a few broken "plugins" and a "working" plugin to test.
- Tests that the broken ones fail properly and that the working one can be used to auth.


slemrmartin (approved):
:+1: looks good

PR #35, opened 2023-11-28
Add login_redirect_override to ui_auth endpoint

john-westcott-iv:


ansible_base/tests/unit/utils/test_authentication.py
```diff
@@ -0,0 +1,57 @@
...
+    response = generate_ui_auth_data()
+    assert response == {
+        'login_redirect_override': settings.LOGIN_REDIRECT_OVERRIDE,
+        'passwords': [],
+        'show_login_form': False,
+        'ssos': [],
+    }
+
+
+redirect_url = 'www.monster.com'
```
relrod: Just because I'm paranoid about referencing company names, let's use fake URLs instead (example.com and example.org are reserved by IANA for this purpose).

```suggestion
redirect_url = 'example.com'
```

ansible_base/utils/authentication.py
```diff
@@ -0,0 +1,56 @@
...
+            response['login_redirect_override'] = settings.LOGIN_REDIRECT_OVERRIDE
+        except ValidationError:
+            # Maybe we are pointing to function
+            try:
+                module_name, _, function_name = settings.LOGIN_REDIRECT_OVERRIDE.rpartition('.')
+                the_function = getattr(importlib.import_module(module_name), function_name)
+                redirect_url = the_function()
+                response['login_redirect_override'] = redirect_url
+            except Exception:
+                logger.error("LOGIN_REDIRECT_OVERRIDE was set but was neither a URL nor a function, ignoring")
```
relrod: Maybe a better error message here, since we'll hit this if _anything_ goes wrong in `the_function()`.

And maybe `logger.exception()`?

relrod (approved):
Approved with two minor inlines.

ansible_base/tests/unit/utils/test_authentication.py
john-westcott-iv: There was also a reference to google.com that I changed to example.com.

ansible_base/utils/authentication.py
john-westcott-iv: Converted to exception and changed message.

PR #40, opened 2023-12-04
Catch exception in urlparse

john-westcott-iv:
In our URLValidator we called urlpase in a specific scenario which could raise an exception.
We were not catching that exception but this PR will now catch the exception.
Additional testing added for this function (may not be 100%).

### DIFF
--- a/ansible_base/utils/validation.py
+++ b/ansible_base/utils/validation.py
@@ -33,7 +33,10 @@ def validate_url(url: str, schemes: list = ['https'], allow_plain_hostname: bool
         raise ValidationError(VALID_STRING)
     if allow_plain_hostname:
         # The default validator will not allow names like https://junk so, if we are ok with simple hostnames we are going to munge up the URL for the validator
-        url_parts = urlparse(url)
+        try:
+            url_parts = urlparse(url)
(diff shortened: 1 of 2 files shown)


ansible_base/utils/validation.py
```diff
@@ -33,7 +33,10 @@ def validate_url(url: str, schemes: list = ['https'], allow_plain_hostname: bool
         raise ValidationError(VALID_STRING)
     if allow_plain_hostname:
         # The default validator will not allow names like https://junk so, if we are ok with simple hostnames we are going to munge up the URL for the validator
-        url_parts = urlparse(url)
+        try:
+            url_parts = urlparse(url)
+        except ValueError as e:
+            raise ValidationError(e)
```
relrod: can we raise this `from e`?

relrod (approved):

ansible_base/utils/validation.py
relrod: Also, should this be `ValidationError(str(e))` or something? Passing in an exception feels weird.

slemrmartin (commented):
:+1:

PR #41, opened 2023-12-06
Resource registry

newswangerd:
The Resource Registry allows for ansible services to register shared resources that are required to be maintained in sync between each other. This PR implements the Resource Registry and it's corresponding API endpoints.


ansible_base/resource_registry/registry.py
```diff
@@ -0,0 +1,95 @@
...
+        model_label = model._meta.label
+
+        parent_map = {}
+        if parent_resources:
+            for parent in parent_resources:
+                parent_map[parent["model"]._meta.label] = parent
+
+        self.registry[model_label] = {
+            "model": model,
+            "externally_managed": externally_managed,
```
AlanCoding: The big need that I see is to configure which models are externally managed. For example, AWX upstream will have the organization model as a normal editable model, but downstream deployments will want this to be externally managed. The pattern of `resource_registry.register(Organization, externally_managed=settings.SOMETHING)` isn't good, because we don't want to reference settings on import. So I think the `externally_managed` property needs to be set some other way... I assume a method that references settings.

AlanCoding:
When getting into the serializers needed to do things using uuids, like manage permissions involving objects in different components, we need to look up related objects by something other than the id. The closest I have come to do that is the slug related field in DRF.

https://www.django-rest-framework.org/api-guide/relations/#slugrelatedfield

However, that tends to allow looking up an item with a different field on the model itself, like looking up a user by last name. But if looking up an object by resource uuid, you would be looking up via a field on a related model... and would probably require 2 lookups because you can't traverse a reverse GenericForeignKey.

So that is the extent of my thinking on that problem, and it is definitely a problem I want to run away from.

ansible_base/views/resource.py
```diff
@@ -0,0 +1,145 @@
...
+    resource_type = filters.CharFilter(field_name="content_type__resource_type__resource_type")
+
+    class Meta:
+        model = Resource
+        fields = ["name", "object_id"]
+
+
+class TransactionViewSet(
+    mixins.RetrieveModelMixin,
+    mixins.ListModelMixin,
```
AlanCoding: I'm not necessarily convinced that a list view of transactions is needed. I just don't want this to become harder than it needs to.

ansible_base/models/resource.py
```diff
@@ -0,0 +1,132 @@
...
+        return self.content_type.resource_type.resource_type
+
+    class Meta:
+        unique_together = ('content_type', 'object_id')
+        indexes = [
+            models.Index(fields=["content_type", "object_id"]),
+        ]
+
+
+class PostgresTransaction(models.Model):
```
AlanCoding: I can't figure out right now where this model is being used... but it feels a little dangerous.

Requests in often in a transaction by default. Certainly, we can't create this model while _inside_ the transaction we haven't committed, or else it won't show on subsequent requests right? Try to avoid this - make the actions non-atomic on the whole, and it might go okay. But I also worry that this will go down the route of complex cursor / connection management. If the model isn't _desperately_ needed for functionality, I wouldn't want to add it.

AlanCoding:
Thanks for spearheading the effort. I would like to try to get the transaction management bits into an Minimum Viable Product demo with AWX as soon as possible. We have ansible_base added as a requirement now https://github.com/ansible/awx/pull/14705 so we can just switch to the branch and experiment. Code is needed on the AWX side, but showing this would really help shore up the contract needed with the app that's using this. At the first level, we could mock the requests coming from Gateway. I'm really interested to see what that API contract will be looking like, and get broader feedback on it.

ansible_base/models/resource.py
newswangerd: This model is for a postgres view that tracks prepared transactions, which are different from regular transactions (https://www.postgresql.org/docs/current/sql-prepare-transaction.html). Prepared transactions get ejected from the current connection session and tracked separately.

AlanCoding: Thanks, I just want to be sure that we can clearly separate what is in the request transaction and what is in the prepared transaction. Because AWX is somewhat notoriously bad about side effects that happen from requests. If the pending changes in this transaction are just those intended by the Resource API then that should go fine.

chrismeyersfsu: Are we OK increasing our Postgres surface area dependence?

I thought we wanted to avoid Postgres specific features to allow for supporting new DBs in the future.

ansible_base/resource_registry/registry.py
chrismeyersfsu: @AlanCoding your question makes the presumption that AWX (vs. Controller) will _want_ to register it's models in the `resource_registery`. The code lives in `django_ansible_base`, which is open source, but does it make sense to expose the resource registry to open source users? The answer may be no currently, but I could see a future where the answer is yes. So I think we need to define a north star and my vote is to invoke the same AWX/Controller code path as much as possible. So we _would_ want to pivot `externally_managed` for AWX vs. Controller.

TL;DR I agree with @AlanCoding .

chrismeyersfsu:
I'm a bit confused by `service_id`. Currently `service_id` looks like what AWX calls `INSTALL_UUID`; it uniquely identifies an AWX installation cluster, lives in the DB across upgrades/migrations. 

I maybe expected `service_id` to be stable per `service_type`. Since `service_type` is a string and unique I imagined the `service_id` being generated via `hash(service_type)`. This would limit 1 service of any given type in a registry, but that reflects the reality of this being a library. This also aligns with gateways restriction of 1 service of any given type.

^^ I actually like `service_id` be being like `INSTALL_UUID`. But shouldn't each service tell gateway their chosen `service_id`? I reckon it's not strictly a requirement. Gateway can reach into the service's api to discover their service_id once the service is registered.

ansible_base/models/resource.py
```diff
@@ -0,0 +1,132 @@
...
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE, related_name="resources")
+
+    # this has to accommodate integer and UUID object IDs
+    object_id = models.TextField(null=False)
+    content_object = GenericForeignKey('content_type', 'object_id')
+
+    _service_id = models.CharField(null=True, max_length=8)
+
+    # we're not using this as the primary key because the ansible_id can change if the object is
+    # externally managed.
```
chrismeyersfsu: `ansible_id can change if the object is externally managed.` <-- Can you mention some cases where the ansible_id can change?
Only case I can think of is the initial case where a Resource exists without a resource_id BUT, once set, will never be changed.

chrismeyersfsu:
I think I understand the usage of postgres prepared transactions in this PR. But I don't think I understand the greater problem they are attempting to solve. Let me see if I can describe the problem.

Gateway will be the source of truth for Organizations. An Organization that is created in Gateway will be synced to other services (Controller, and Hub for example). The semantics for "synced" is where the problem comes in. What is the "synced" guarantee? We have decided that we prefer **C**onsistency**AP**. This is the problem statement.

Now for solutioning. We decided that Gateway will ensure consistency. For the Org creation example, Gateway will use the Resource API to create the Org in both Controller and Hub. Doing so will generate a transaction in each service. At this point I see a problem. For the happy path, Gateway needs to do two things

1. Create the new Org in Gateway
2. COMMIT the pending transaction on Controller and Hub

Which does Gateway do first, which does it do second? Either way, you can get into an inconsistent state when a failure happens.

ansible_base/resource_registry/registry.py
```diff
@@ -0,0 +1,102 @@
...
+        parent_map = {}
+        if parent_resources:
+            for parent in parent_resources:
+                parent_map[parent.model._meta.label] = parent
+
+        self.registry[model_label] = {
+            "model": model,
+            "externally_managed": externally_managed,
+            "managed_serializer": managed_serializer,
+            "parent_resources": parent_map,
```
chrismeyersfsu: Is `parent_resources` for a feature that doesn't yet exist?

ansible_base/models/resource.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class PostgresTransaction(models.Model):
+    transaction = models.BigIntegerField(primary_key=True)
+    gid = models.CharField(max_length=200)
+    prepared = models.DateTimeField()
+
+    class Meta:
+        managed = False
+        db_table = "pg_prepared_xacts"
```
chrismeyersfsu: Something that can be done later. We should limit the surface area of this model/api endpoint. The set of transactions here should be limited to the ones that were created by our service/app. Thinking about users/customers with shared database servers. This might be a non-issue and Postgres users/roles might be "the way".

ansible_base/signals/handlers.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@receiver(post_delete)
+def remove_resource(sender, instance, **kwargs):
+    if sender in get_resource_models():
+        Resource.objects.filter(object_id=instance.pk, content_type=ContentType.objects.get_for_model(instance).pk).delete()
+
+
+@receiver(post_save)
+def update_resource(sender, instance, created, **kwargs):
```
chrismeyersfsu: On the surface, I see that these signals keep the `Resource` object in sync with the registered models. But I'm failing to see the bigger picture. This feels only necessary for 2-way syncing.

ansible_base/utils/transactions.py
```diff
@@ -0,0 +1,34 @@
+from django.db import connection
+
+
+def create_transaction(transaction_id):
+    cursor = connection.cursor()
+    cursor.execute(
+        "PREPARE TRANSACTION %s;",
+        [
+            str(transaction_id),
```
chrismeyersfsu: Picky, but can we wrap these in a bit more protection? Something like `str(int(transaction_id))`. I'm worried about sql injection. Currently, I don't see a vector with the usage but I worry about anyone using them in the future.

chrismeyersfsu:
In the aap-gateway PR associated with this PR that shows the usage of this new resource registry, you register a `ServiceCluster` without an associated `shared_resource`. What is the use case for this?

ansible_base/resource_registry/registry.py
AlanCoding: My comment is a week old and my thinking is changing a lot in a week right now...

To basics - AWX needs a way to disable editing orgs/teams/users and their relationships. This needs to switch on and off for upstream/downstream but we don't need to frame it in those terms. The syncing thing is happening upstream, you disable editing of the stuff is syncing is turned on. Nothing should sync from AWX to Gateway for now... although credentials/projects may get synced somewhere soon.

Maybe you're right that the default upstream behavior doesn't register any of these models. On the reverse, in my RBAC proposal, `permission_registry` would register all models in upstream... and maybe downstream too, I'm not totally sure yet.

newswangerd:
> I maybe expected `service_id` to be stable per `service_type`. Since `service_type` is a string and unique I imagined the `service_id` being generated via `hash(service_type)`. This would limit 1 service of any given type in a registry, but that reflects the reality of this being a library. This also aligns with gateways restriction of 1 service of any given type.

Service ID is it's own unique identifier in case we need to support multiple instances of each service type. Even if we never support this, I think having a unique ID can be helpful for debugging so that we know exactly where a resource lives.
 
> ^^ I actually like `service_id` be being like `INSTALL_UUID`. But shouldn't each service tell gateway their chosen `service_id`? I reckon it's not strictly a requirement. Gateway can reach into the service's api to discover their service_id once the service is registered.

Yes! That's exactly what will happen. The service's metadata endpoint is intended to be used by the Gateway to pick up the service's ID, which will be used to route requests to individual service's API endpoints.

ansible_base/models/resource.py
newswangerd: This can happen when merging resources across several services. For example if two services have the same group/team name, we are going to offer the admin to merge them into a single team. In this case, if the local instance already has an ansible_id for the resource, the ID will have to change to match the Gateway's ID.

I would kind of like to make the resource API available upstream, so that it can be used for RBAC, which is why an ansible ID might already exist.

With that said, I have been thinking about this problem this morning. It would be kind of nice to have the ansible ID be the PK for this table. It would allow us to use a one to one field on the resource model, and would let us serialize the ansible ID for resources without having to join on the Resource table or make a separate DB query, sp I'm open to the idea of changing this.

ansible_base/resource_registry/registry.py
newswangerd: yeah, that will be used for RBAC. I might get rid of it for the time being, depending on what @AlanCoding thinks.

newswangerd:
> I think I understand the usage of postgres prepared transactions in this PR. But I don't think I understand the greater problem they are attempting to solve. Let me see if I can describe the problem.
> 
> Gateway will be the source of truth for Organizations. An Organization that is created in Gateway will be synced to other services (Controller, and Hub for example). The semantics for "synced" is where the problem comes in. What is the "synced" guarantee? We have decided that we prefer **C**onsistency**AP**. This is the problem statement.
> 
> Now for solutioning. We decided that Gateway will ensure consistency. For the Org creation example, Gateway will use the Resource API to create the Org in both Controller and Hub. Doing so will generate a transaction in each service. At this point I see a problem. For the happy path, Gateway needs to do two things
> 
>     1. Create the new Org in Gateway
> 
>     2. COMMIT the pending transaction on Controller and Hub
> 
> 
> Which does Gateway do first, which does it do second? Either way, you can get into an inconsistent state when a failure happens.

@chrismeyersfsu This was intended for use by the two phase commit protocol (https://en.wikipedia.org/wiki/Two-phase_commit_protocol). Basically:

1. Create a new org in each service in a transaction.
2. Wait for each service to ACK the transaction.
3. If all the services ACK the transaction, send a separate API request to commit it.
4. If any service fails to ACK, rollback the transaction on each service.

Sadly we can't use this because prepared transactions aren't enabled by default in Postgres and per @matburt we don't have a guarantee that we can change the postgres config in every environment, so I will be removing this feature before we merge the PR.

chrismeyersfsu:
> @chrismeyersfsu This was intended for use by the two phase commit protocol (https://en.wikipedia.org/wiki/Two-phase_commit_protocol). Basically:
> 
> 1. Create a new org in each service in a transaction.
> 2. Wait for each service to ACK the transaction.
> 3. If all the services ACK the transaction, send a separate API request to commit it.
> 4. If any service fails to ACK, rollback the transaction on each service.
> 
> Sadly we can't use this because prepared transactions aren't enabled by default in Postgres and per @matburt we don't have a guarantee that we can change the postgres config in every environment, so I will be removing this feature before we merge the PR.

Two phase commit requires the second phase to have an atomic commit. The separate API request(s) are where you want/need an atomic commit and it doesn't exist.

Let's say you have Gateway, Hub, and Controller.

> 3. If all the services ACK the transaction, send a separate API request to commit it.

* All services ACK
* commit API request SUCCCESS to Hub
* commit API request FAIL Controller

> 4. If any service fails to ACK, rollback the transaction on each service.

* Abort Controller
* Rollback Hub <-- Say hub isn't contactable by Gateway. The system is now in a state where Hub has the resource created and thinks everything is OK. 

The root of the problem is that there isn't a system wide ATOMIC commit NEEDED for the 2-phase commit to work.

ansible_base/resource_registry/registry.py
newswangerd: My hope is that we'll eventually use the resource API for managing RBAC upstream as well as downstream.

I think that disabling apis in controller can be based on a different flag, such as whether or not a gateway url is configured.

ansible_base/models/resource.py
AlanCoding: I would try to avoid handing the `ansible_id can change` scenario ever, because I don't believe there's a business need for it. Yes, I think some teams will need to be merged. If the user doesn't want to merge them... then maybe we can offer a flag to control this in the migration. Before a service starts its own migration, couldn't it get a list of teams & ansible_ids from other services? I don't think that's technically problematic. Then we only need to define precedence of which services migrates first (or does this part of the migration) and passes on data to the others.

ansible_base/models/resource.py
```diff
@@ -0,0 +1,122 @@
...
+    content_object = GenericForeignKey('content_type', 'object_id')
+
+    _service_id = models.CharField(null=True, max_length=8)
+
+    # we're not using this as the primary key because the ansible_id can change if the object is
+    # externally managed.
+    resource_id = models.UUIDField(default=uuid.uuid4, db_index=True, unique=True)
+
+    # human readable name for the resource
+    name = models.CharField(max_length=256, null=True)
```
john-westcott-iv: Could we pull this in via NamedCommonModel? Or do we not want the created/deleted fields from CommonModel?

ansible_base/resource_registry/resource_urls.py
```diff
@@ -0,0 +1,19 @@
...
+service_router = routers.SimpleRouter()
+
+service_router.register(r'resources', views.ResourceViewSet)
+service_router.register(r'resource-types', views.ResourceTypeViewSet)
+
+service = [path('metadata/', views.ServiceMetadataView.as_view()), path('', include(service_router.urls))]
+
+resource_api_urls = [
+    path('service/', include(service)),
+]
```
john-westcott-iv: This feels kind of wrid to have URLs here instead of at the top level. Can you at least load this from urls.py if the feature is "enabled"?

ansible_base/resource_registry/shared_types.py
```diff
@@ -0,0 +1,24 @@
...
+
+class UserType(serializers.Serializer):
+    RESOURCE_TYPE = "user"
+
+    username = serializers.CharField()
+    email = serializers.EmailField()
+    first_name = serializers.CharField()
+    last_name = serializers.CharField()
+    is_superuser = serializers.BooleanField()
+    # is_system_auditor = serializers.BooleanField()
```
john-westcott-iv, line 13: Can we add more of a comment as to why this is commented out (or remove it)?

ansible_base/resource_registry/shared_types.py
```diff
@@ -0,0 +1,24 @@
+from rest_framework import serializers
+
+
+class UserType(serializers.Serializer):
```
john-westcott-iv, line 4: Is there a reason why we have the "resource_registry" folder and are not putting things like these serializers into the serializers directory?

ansible_base/serializers/resource.py
```diff
@@ -0,0 +1,161 @@
...
+    if not ANSIBLE_ID_REGEX.match(val):
+        raise serializers.ValidationError(message)
+
+    (s_id, r_id) = val.split(":")
+
+    try:
+        if str(uuid.UUID(r_id, version=4)) != r_id:
+            raise serializers.ValidationError(message)
+    except ValueError:
+        raise serializers.ValidationError(message)
```
john-westcott-iv: Can this function be simplified?
```
def ansible_id_validator(val):
    matches = re.match("^[0-9a-f]{8}:(.*)$", val, re.IGNORECASE)
    try:
        if not matches:
            raise ValueError
        print("Checking UUID")
        print(matches.group(1))
        if str(uuid.UUID(matches.groups(1), version=4)) != matches.groups(1):
            raise ValueError
    except (AttributeError, ValueError):
        raise serializers.ValidationError(_(
            "Ansible ID must in the format of <service_id>:<resource_id>, "
            "where service_id is the first 8 characters of the service's ID "
            "and resource_id is a valid UUID 4."
        ))
```

john-westcott-iv: Can this also be moved into utils/validation.py or somewhere else besides a serializer?

ansible_base/management/commands/update_resources.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Command(BaseCommand):
+    help = "Reset resource index"
+
+    def handle(self, *args, **options):
+        ResourceType.objects.all().delete()
+
+        ResourceType.update_resource_types_from_registry()
+        Permission.update_permissions()
+        Resource.update_index()
```
john-westcott-iv: Why would we want to reset this during normal operation?

ansible_base/models/service_id.py
```diff
@@ -0,0 +1,24 @@
...
+        if self.objects.exists():
+            raise RuntimeError("This service already has a ServiceID")
+
+        return super().save()
+
+
+@lru_cache(maxsize=1)
+def service_id():
+    system_id_obj = ServiceID.objects.first()
+    return str(system_id_obj.pk)
```
john-westcott-iv: Could this be moved out of the model? Maybe into something like utils/resource_registry.py or or something along those lines?

ansible_base/apps.py
```diff
@@ -1,8 +1,56 @@
...
 from django.apps import AppConfig
+from django.db.models.signals import post_migrate
 
 import ansible_base.checks  # noqa: F401 - register checks
 
+logger = logging.getLogger('ansible_base.apps')
+
+
+def initialize_resources(sender, apps, **kwargs):
+    from ansible_base.models.resource import get_registry
```
john-westcott-iv: Maybe move this into utils/resource_registry.py as well?

ansible_base/models/resource.py
```diff
@@ -0,0 +1,122 @@
...
+
+# Circular imports make this really hard to initialize correctly.
+# I don't know why, but I can't put this function anywhere else without inducing
+# circular imports.
+def get_registry():
+    global _resource_registry
+    if _resource_registry is None:
+        from django.conf import settings
+
+        if hasattr(settings, "RESOURCE_REGISTRY_CONFIG"):
```
john-westcott-iv: Should we prefix this setting with ANSIBLE_BASE? There are a couple of "settings" without the prefix but these were mostly copied from AWX.

ansible_base/models/resource.py
```diff
@@ -0,0 +1,122 @@
...
+    manager automatically annotates ansible_id as <service_short_id>:<resource_id>
+    for easier lookups.
+    """
+
+    def get_queryset(self):
+        # this needs to wait for migrations to run before it becomes available.
+        try:
+            s_id = service_id().split('-')[0]
+        except ProgrammingError:
+            s_id = "00000000"
```
john-westcott-iv: Why the 0000000 here instead of raising?

ansible_base/models/resource.py
```diff
@@ -0,0 +1,122 @@
...
+
+        return (
+            super()
+            .get_queryset()
+            .annotate(
+                _computed_service_id=models.Case(
+                    models.When(_service_id__isnull=False, then=models.F('_service_id')), default=models.Value(s_id), output_field=models.CharField()
+                ),
+                _ansible_id=Concat("_computed_service_id", models.Value(":"), models.F("resource_id"), output_field=models.CharField()),
```
john-westcott-iv: I'd be really nice if we could make this a constant from somewhere. I can't image we'd want to change it but if we had to searching for `:` in the source is gonna be cringe.

john-westcott-iv (commented):
Can you add a requirements/requirements_<feature>.in file for this (even if its empty). Can we also add unit/functional tests for as much of this as possible?

ansible_base/management/commands/update_resources.py
newswangerd: Oh, I need to get rid of this. I added that for testing. It's horribly out of date.

ansible_base/models/resource.py
newswangerd: > Or do we not want the created/deleted fields from CommonModel?

This

ansible_base/resource_registry/resource_urls.py
newswangerd: The problem is that we don't have a defined place for authenticators or the resource api to live. This is why I think a separate app for the resource api and authenticators makes a lot of sense.

ansible_base/resource_registry/shared_types.py
newswangerd: This is a good question. The idea here is that these serializers form the basis of the API contract for shared resources (kinda like how we use serializers for authenticator config validation), so putting them alongside the normal serializers didn't make sense to me.

ansible_base/resource_registry/shared_types.py
newswangerd: +1. I did this so that it would work with automation hub (which doesn't have a system auditor yet)

ansible_base/models/resource.py
```diff
@@ -0,0 +1,72 @@
...
+    content_object = GenericForeignKey('content_type', 'object_id')
+
+    service_id = models.CharField(default=short_service_id, max_length=8)
+
+    # we're not using this as the primary key because the ansible_id can change if the object is
+    # externally managed.
+    resource_id = models.UUIDField(default=uuid.uuid4, db_index=True, unique=True)
+
+    # human readable name for the resource
+    name = models.CharField(max_length=256, null=True)
```
AlanCoding: The common `name` in AWX has max length of 512

https://github.com/ansible/awx/blob/30d4df788ff8d4be4fd4e503854cd432e1490586/awx/main/models/base.py#L358

same with Gateway I think. I think it's safe to just go with that.

john-westcott-iv: Should we abstract this into a constant in DAB that this and the other services can reference so we all stay in-sync?

AlanCoding: A common named model already exists in DAB. If inheriting isn't your cup of tea then there are ways to get reuse. I don't feel too particular about any of this, other than that this field can contain names from everything else.

https://github.com/ansible/django-ansible-base/blob/cdf5fb0e6879bd268f1f29926466cc631276af69/ansible_base/models/common.py#L111

ansible_base/resource_registry/registry.py
```diff
@@ -0,0 +1,131 @@
...
+        if parent_resources:
+            for parent in parent_resources:
+                parent_map[parent.model._meta.label] = parent
+
+        self.model = model
+        self.externally_managed = externally_managed
+        self.managed_serializer = managed_serializer
+        self.parent_resources = parent_map
+        self.actions = ResourceInspector().model_map.get(self.model_label, {})
+        self.name_field = name_field
```
john-westcott-iv, line 88: Should we validate that the model actually has this name_field? We don't catch an exception or default the name when we try to get the name_field up above where we get the name_field from the model:
```
resource.name = getattr(obj, resource_config.name_field)
```

ansible_base/models/resource.py
john-westcott-iv: Actually, this is a good point, not everything inherits form NamedCommonModel in the first place. What if we needed/wanted to share a resource without a specific name? i.e. a token or an authenticator_map. Or are we making a hard requirement that anything that will be shared has to have a name field?

slemrmartin:
@newswangerd are we going to implement ansible collection for resources? Does it make sense?

ansible_base/resource_registry/registry.py
newswangerd: I refactored this so that the update code is cleaner, but if the attribute doesn't exist, we'll just set the name to None. I like defaulting this to "name" since that's the most common field name, but not all resources will necessarily have a name.

ansible_base/models/resource.py
newswangerd: Updated the max length to 512 and set it to truncate anything larger than that.

benthomasson:
I have concerns about the resource API in general.   I am not sure a new API is needed that abstracts over the existing data types.   If we can add ansible_id UUIDs to the existing APIs then we might not need a separate API at all and we can use the existing APIs.

AlanCoding:
I'm still undecided if using the existing APIs is a good idea, because we still need to distinguish between requests coming with certain JWT data versus a general user (for whom the request may be denied with a 400 error), and if we're taking action conditionally based on the user or auth details, a separate view with its own unique auth policy is probably better.

What I feel like I'm missing here is docs. Say I'm trying to run this with a proxying server and AWX. How do I set that up? How does the ResourceType entries get populated with the content types from AWX? There's no management command to do this.

newswangerd:
@AlanCoding I'll fix up the documentation, and then I think this should be good to merge.

test_app/resource_api.py
```diff
@@ -0,0 +1,20 @@
...
+    service_type = "aap"
+
+
+RESOURCE_LIST = (
+    ResourceConfig(get_user_model(), shared_resource=SharedResource(serializer=UserType, is_provider=False), name_field="username"),
+    ResourceConfig(
+        Group,
+        shared_resource=SharedResource(serializer=TeamType, is_provider=True),
+    ),
+    ResourceConfig(Authenticator),
```
AlanCoding, line 26: authenticators are not shared, right? Each component can have its own authenticators which may be different from all the others, and they might not even have the same plugins on the file system available, intentionally. But this system may, hypothetically, allow for sharing metadata for `aap.authenticators` and `awx.authenticators`.

test_app/resource_api.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class APIConfig(ServiceAPIConfig):
+    service_type = "aap"
+
+
+RESOURCE_LIST = (
+    ResourceConfig(get_user_model(), shared_resource=SharedResource(serializer=UserType, is_provider=False), name_field="username"),
+    ResourceConfig(
+        Group,
```
AlanCoding: the test_app does have a team and organization model. I tested this and I can see a resources entry created when creating a Group, but seeing them for organization & team would be more instructive.

ansible_base/resource_registry/models/resource.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class ResourceType(models.Model):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        from ansible_base.resource_registry.registry import get_registry
+
+        self.resource_registry = get_registry()
+
+    content_type = models.OneToOneField(ContentType, on_delete=models.CASCADE, related_name="resource_type", unique=True)
```
AlanCoding, line 23: Here `content_type` isn't nullable. What about external resources? We need to create a `ResourceType` object to track some type on another server, like say I'm test_app and I need to know about `awx.workflow`. There can be no ContentType for workflow, because it's on the other system.

test_app/resource_api.py
newswangerd: They aren't. I was just looking for existing models that I could hook in here for testing.

test_app/resource_api.py
newswangerd: oooh, good to know. I can update the tests to use those.

ansible_base/resource_registry/models/resource.py
newswangerd: Shared resources should always have a local version of the shared model (ie organizations). I'm not sure that services will need to track resources in other services, outside of maybe the Gateway.

ansible_base/resource_registry/views.py
```diff
@@ -0,0 +1,74 @@
...
+    mixins.ListModelMixin,
+    GenericViewSet,
+):
+    """
+    Index of all the resources in the system.
+    """
+
+    queryset = Resource.objects.select_related("content_type__resource_type").all()
+    serializer_class = ResourceSerializer
+    permission_classes = [permissions.IsAdminUser]
```
AlanCoding, line 28: What I believe we need here is a permission class that corresponds exactly to something set by `JWTAuthentication`. I don't even expect that superusers can make changes here. It has to be made by a request that carries valid JWT authentication. It would make sense for superusers and auditors to be able to GET.

https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/jwt_consumer/common/auth.py

@john-westcott-iv does this align with your thinking?

newswangerd: yes. I want to add this later though. I really want to get this merged.

ansible_base/resource_registry/models/resource.py
newswangerd: From slack: lets investigate caching resource types in the gateway later

AlanCoding: I want to integrate `ResourceType` specifically into the RBAC app, and this is near-term, at least for doing it in test_app. After giving this a lot of thought, it's just not possible in this form.

Let me make a pitch here - instead of linking `ResourceType` to `contenttypes.ContentType`, what if it _subclassed_ the ContentType model? There are good reasons to let some related models be defined in settings, obviously we're both already working with different user models, but only provided that they share certain common fields/functionality. I really want this to be something that I work with _as if_ it were the ContentType model in some cases. Looking through the rest of this patch, I don't see any hard reasons that we need this to reference the canonical contenttypes app. If you customized the RBAC content type model, you also have to customize the `Permission` model and have that link to your custom content type. Many projects already use a custom permission model, and I think this suits us just fine, because we actually really want a service identifier with the permission model because its the same naming scheme, like an `awx.inventory` should have an `awx.change_inventory` permission.

ansible_base/resource_registry/apps.py
```diff
@@ -0,0 +1,76 @@
...
+    registry = get_registry()
+    if registry:
+        # Create resource types
+        for key, resource in registry.get_resources().items():
+            content = ContentType.objects.get_for_model(resource.model)
+
+            if serializer := resource.managed_serializer:
+                resource_type = f"shared.{serializer.RESOURCE_TYPE}"
+            else:
+                resource_type = f"{registry.api_config.service_type}.{content.model}"
```
AlanCoding, line 48: This seems a good bit less intuitive than having a `service_type` field on `ResourceType`. I realize that "shared" may be the only supported type right now.

ansible_base/resource_registry/registry.py
```diff
@@ -0,0 +1,131 @@
...
+        if model_label:
+            return self.registry[model_label]
+
+        raise AttributeError("Must include either model or model_label arg.")
+
+
+def get_registry() -> ResourceRegistry:
+    from django.conf import settings
+
+    if hasattr(settings, "ANSIBLE_BASE_RESOURCE_CONFIG_MODULE"):
```
AlanCoding, line 123: I've seen a lot of Django apps implement a registry of some sort, but I've never seen this. I don't quite get it, you're using the apps `.ready` method as a hook to do finalization/setup of the registry. Why does this need the project using it to give an import location with this setting? Instead of having an expected value in that module like `RESOURCE_LIST`, can't you just pass whatever value that is to your registry?

newswangerd: I was sort of modelling it after this: https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-ROOT_URLCONF

This is one area I'm not so sure about. If you have another suggestion, I'd love to hear it.

newswangerd: So looking back over the code, one of the things that made me really nervous about having a gobal `ResourceRegistry` object that the app imports (similar to how you have `PermissionRegistry`) is that it is a global variable that could get modified. Defining a static list of resources (similar to `urlpatterns`) seemed safer, since you can't accidentally modify the configuration.

ansible_base/resource_registry/apps.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Create resources
+        for r_type in ResourceType.objects.filter(migrated=False):
+            resource_model = apps.get_model(r_type.content_type.app_label, r_type.content_type.model)
+
+            logger.info(f"adding unmigrated resources for {r_type.name}")
+
+            data = []
+            for obj in resource_model.objects.all():
+                data.append(Resource.init_from_object(obj, resource_type=r_type))
```
AlanCoding, line 60: One concern we may want to note for later is the case where an object is deleted inside of a migration, which could leave around an orphan `Resource` object. Minor I just want to say to be aware of it.

ansible_base/resource_registry/models/service_id.py
```diff
@@ -0,0 +1,27 @@
...
+
+from django.db import models
+
+
+class ServiceID(models.Model):
+    """
+    Provides a globally unique ID for this service.
+    """
+
+    id = models.UUIDField(default=uuid.uuid4, primary_key=True, null=False, editable=False)
```
AlanCoding, line 11: Services do have a name right? In the RBAC branch I introduced a setting `ANSIBLE_BASE_SERVICE_PREFIX` which defaults to "local" for the test_app mainly. I fully expect that apps need to declare a setting that gives their service name like "awx". It seems like that would get added to the ServiceID model.

We're not supporting multiple services of the same type, still fine to have the uuid, but even if we were that's not inconsistent with tracking the name as an informational property. Right now the name only gets added in the ResourceType.name combined with other strings, but I am absolutely sure the service name part by itself is going to get reused for other things.

ansible_base/resource_registry/registry.py
AlanCoding: Ah, I think I faced a form of that problem, and added a flag which got flipped when the apps `.ready` method was called.

AlanCoding (approved):
I want to continue filing issues for this, but right now we should be set to merge and allow downstream to move forward using devel

AlanCoding:
Sonar is broke

PR #46, opened 2023-12-11
Settings changes to allow database reuse with tests

AlanCoding:
I found these changes were needed to get running locally with the tests. With this, you can do things like:

```
DJANGO_SETTINGS_MODULE=ansible_base.tests.settings_overrides python manage.py showmigrations
```

This will not show the migration status stored from the last test run, because you're using 2 different databases in your test settings for normal versus for running tests, which is cool. Totally fine.

But what I really wanted to establish is this:

```
time py.test ansible_base/tests/functional/authentication/test_ldap.py -k test_ldap_auth_failed
```

This will finish the test in 1.8 seconds on my machine. From Django Con @chrismeyersfsu and others and I talked a lot about options, and we converged on reusing the database as being the most practical to make development really fast.

### DIFF
--- a/ansible_base/tests/settings_overrides.py
+++ b/ansible_base/tests/settings_overrides.py
@@ -64,3 +64,19 @@ AUTHENTICATION_BACKENDS = [
 ]
 
 ANSIBLE_BASE_AUTHENTICATOR_CLASS_PREFIXES = ['ansible_base.authenticator_plugins']
+
+# Must be defined to use management commands with test settings
+TEMPLATES = [
+    {
+        "BACKEND": "django.template.backends.django.DjangoTemplates",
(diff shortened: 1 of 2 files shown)


AlanCoding:
To more fully clarify my frustration here, if I run the following:

```
time py.test ansible_base/tests/functional/authentication/test_ldap.py -k test_ldap_auth_failed --reuse-db
```

It takes 10 seconds. This is still a young project! Don't expect it to ever get any faster! It also shows stuff like:

```
...
  Applying contenttypes.0002_remove_content_type_name... OK (0.130s)
  Applying auth.0002_alter_permission_name_max_length... OK (0.113s)
...
```

And why should it? I used the `--reuse-db` flag.

PR #47, opened 2023-12-11
Fixing PATCHing issue in authenticators when configuration is not in payload

john-westcott-iv:
We want to be able to patch fields on authentications (like enabled) but we would get exceptions if configuration was not defined. This PR circumvents the code looking at the configuration if it was not provided.


Dostonbek1 (approved):
tested, was able to send PATCH request without defining `configuration`  field in payload.

PR #48, opened 2023-12-12
Add M2M relationship between authenticator and users

john-westcott-iv:


ansible_base/models/authenticator.py
```diff
@@ -23,6 +26,12 @@ class Authenticator(UniqueNamedCommonModel):
     )
     slug = fields.SlugField(max_length=1024, default=None, editable=False, unique=True, help_text="An immutable identifier for the authenticator")
     category = fields.CharField(max_length=30, default=None, help_text="The base type of this authenticator")
+    users = ManyToManyField(
+        User,
```
AlanCoding: isn't it better to use `settings.AUTH_USER_MODEL`? here? That's standard-ish and seems less bad than calling `get_user_model()` on import.

john-westcott-iv: The Django docs seem to say to use the `get_user_model()` function: https://docs.djangoproject.com/en/5.0/topics/auth/customizing/#referencing-the-user-model as it will try settings.AUTH_USER_MODEL and then return the built in User if its not set. We used this pattern in other places in DAB. If we think it would be better to just references settings directly we can but we should do it everywhere.

AlanCoding: That link says

> When you define a foreign key or many-to-many relations to the user model, you should specify the custom model using the [AUTH_USER_MODEL](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-AUTH_USER_MODEL) setting

john-westcott-iv: I agree with changing this, done.

AlanCoding (approved):

PR #53, opened 2023-12-18
Add DRF middleware for channels authentication

bzwei:
This is an extension of django channels AuthMiddleWare to support authentication by settings.AUTHENTICATION_BACKEND.

It is originally need by EDA for websocket authentication. An internal discussion suggested to submit this code to django-ansible-base.

Source code mainly contributed by @mkanoor


john-westcott-iv (approved):
This looks ok to me but I haven't done a lot with web sockets. @AlanCoding @relrod any thoughts on this PR?

ansible_base/channels/middleware.py
```diff
@@ -0,0 +1,53 @@
...
+
+logger = logging.getLogger(__name__)
+User = get_user_model()
+
+
+@database_sync_to_async
+def _get_authenticated_user(scope: dict) -> Optional[User]:
+    request = HttpRequest()
+    request.META = {_http_key(k.decode()): v.decode() for (k, v) in scope["headers"]}
+    auth_classes = [auth() for auth in api_settings.DEFAULT_AUTHENTICATION_CLASSES]
```
benthomasson, line 22: @bzwei how does this work?

benthomasson:
See https://github.com/ansible/eda-server/pull/540 for an example of how this works.

Specifically https://github.com/ansible/eda-server/pull/540/files#diff-91351935cfeb144ac4f3f6eb42f70e7060fde1386c9b7f70dff5cd1b8b5c8165

ansible_base/channels/middleware.py
bzwei: The caller places any authentication data in headers. We attempt to use the authentication classes from settings such as `rest_framework.authentication.BasicAuthentication` or `rest_framework_simplejwt.authentication.JWTAuthentication` to resolve the authentication header into a user. Authentication succeeds if a valid user can be identified.

chrismeyersfsu (approved):

docs/channels_authentication.md
```diff
@@ -0,0 +1,43 @@
...
+    ),
+
+})
+```
+
+While you can wrap the middleware around each consumer individually, it’s recommended you wrap it around a higher-level application component, like in this case the URLRouter.
+
+Note that the DrfAuthMiddleware will only work on protocols that provide HTTP headers in their scope - by default, this is HTTP and WebSocket.
+
+If the user can be retreieved from the stored session or by any backend in `settings.AUTHENTICATION_BACKEND`, the user is stored in `scope["user"]`. Othwerwise the websocket connection is denied and closed with return code 403.
```
fosterseth:
```suggestion
If the user can be retrieved from the stored session or by any backend in `settings.AUTHENTICATION_BACKEND`, the user is stored in `scope["user"]`. Otherwise the websocket connection is denied and closed with return code 403.
```

ansible_base/channels/middleware.py
```diff
@@ -0,0 +1,53 @@
...
+        return await self.inner(scope, receive, send)
+
+
+# Handy shortcut for applying all three layers at once
+def DrfAuthMiddlewareStack(inner):  # noqa: N802
+    return CookieMiddleware(SessionMiddleware(DrfAuthMiddleware(inner)))
+
+
+def _http_key(key: str) -> str:
+    return f"HTTP_{key.replace('-','_').upper()}"
```
fosterseth, line 53: the request META values are just underscored versions of the http headers?

fosterseth (approved):

fosterseth:
dropped minor comments, but looks good to me

ansible_base/channels/middleware.py
bzwei: yes. it was discovered and tested.

PR #44, opened 2023-12-07
[WIP] Oauth2 provider

john-westcott-iv:
This PR allows for services to offer an oauth2 provider similar to how AWX does this work. This adds the /applications and /tokens endpoints.

PR #54, opened 2023-12-19
Add abstract organization model

cutwater:


john-westcott-iv:
Thanks for this PR. Please let me know if you need any assistance to get the CI checks passing. If not, ping me here when you are ready for merge.

cutwater:
@john-westcott-iv Thanks, I've fixed failing tests and added some more. Please review.

ansible_base/models/organization.py
```diff
@@ -0,0 +1,30 @@
...
+        settings.AUTH_USER_MODEL,
+        related_name="organizations",
+        help_text="The list of users in this organization.",
+    )
+
+    teams = models.ManyToManyField(
+        settings.ROLE_TEAM_MODEL,
+        related_name="organizations",
+        help_text="The list of teams in this organization.",
+    )
```
john-westcott-iv: Can you add a doc in the docs folder talking about this AbstractModel and how to implement it? I'm noting here that applications using DAB will need to implement ROLE_TEAM_MODEL, what happens if they don't implement that? Do they get an error in the migrations?

cutwater: It will fail at import, because `settings.ROLE_TEAM_MODEL` is not defined.

ansible_base/models/organization.py
```diff
@@ -0,0 +1,30 @@
...
+        help_text="The organization description.",
+    )
+
+    users = models.ManyToManyField(
+        settings.AUTH_USER_MODEL,
+        related_name="organizations",
+        help_text="The list of users in this organization.",
+    )
+
+    teams = models.ManyToManyField(
```
AlanCoding, line 27: In Gateway / AWX this is a reverse ForeignKey. That is, the `Team.organization` field is just a normal ForeignKey.

ansible_base/models/organization.py
AlanCoding: We might want to rename that setting. Just about all settings in the RBAC stuff I was working on could stand to get better names.

I also don't think it will matter unless someone actually inherits from the abstract model. The generated migration, IIRC, tends to put in the hard-coded value, although you can use the Django swappable_dependency thing, I have not seen it do anything helpful.

ansible_base/models/organization.py
```diff
@@ -0,0 +1,30 @@
...
+    class Meta:
+        abstract = True
+
+    description = models.TextField(
+        null=False,
+        default="",
+        help_text="The organization description.",
+    )
+
+    users = models.ManyToManyField(
```
AlanCoding, line 21: This is starting to be somewhat duplicated work with https://github.com/ansible/aap-gateway/pull/164

At least you and @relrod seem to be on the same page about the name of this field.

ansible_base/models/organization.py
```diff
@@ -0,0 +1,30 @@
...
+class AbstractOrganization(UniqueNamedCommonModel):
+    """An abstract base class for organizations."""
+
+    class Meta:
+        abstract = True
+
+    description = models.TextField(
+        null=False,
+        default="",
+        help_text="The organization description.",
```
AlanCoding: @john-westcott-iv do these help_text strings need to be translated?

AlanCoding:
Right now I do not see a corresponding PR in the gateway repo to adopt this common model. I would hold off on that until Rick's PR merges at least. Also, when that is done I would also like to see the `admins` relationship added.

I would like to get AWX migrated to this, but the biggest issue I see with that is the `modified` vs `modified_on` type of naming discrepancy. This ties into the Resource API, where I do not know if they intend to sync these meta-fields or not. This is another thing I would like to get clarification on.

ansible_base/models/organization.py
john-westcott-iv: Yes, we have a ton of strings that are going to need translation. That is likely going to be a large task at somepoint.

ansible_base/models/organization.py
cutwater: I confirmed it with @newswangerd and according to the information I got, this is supposed to be many to many relationship. Can we all synchronize on this?

ansible_base/models/organization.py
cutwater: And it will be even more duplication, since I'm working on adding organizations to Hub.

ansible_base/models/organization.py
newswangerd: @AlanCoding I was working under the assumption that a team could live in two orgs, but if that's not the case in AWX, then I don't see the need to complicate it. @cutwater you can change this to a foreign key.

newswangerd:
@AlanCoding it would be nice if we could merge this to unblock the organizations work in hub

AlanCoding:
@newswangerd I'd like to also get a patch that gets Gateway, itself, using this as a parent class. This is prioritized as a thing to merge before https://github.com/ansible/django-ansible-base/issues/61

AlanCoding:
In https://github.com/ansible/django-ansible-base/pull/44 @john-westcott-iv is using `ANSIBLE_BASE_ORGANIZATION_MODEL`. So in these PRs, I'd like to go ahead and have the conversation and settle on setting names. Using `ANSIBLE_BASE_TEAM_MODEL` seems fine to me.

docs/organizations.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Configuration
+
+The `ROLE_TEAM_MODEL` setting is mandatory and must be defined in django settings. 
+It must be of the format `<app_label>.<model_name>`.
+
+Example:
+
+```python
+ROLE_TEAM_MODEL = 'example.Team'
```
john-westcott-iv: Getting this in as an official opened item. As @AlanCoding  suggested, can we convert `ROLE_TEAM_MODEL` into `ANSIBLE_BASE_TEAM_MODEL`? This will align with the work being done in #44 which introduces `ANSIBLE_BASE_ORGANIZATION_MODEL`.

AlanCoding (approved):

ansible_base/tests/functional/test_organization_model.py
```diff
@@ -0,0 +1,46 @@
+from datetime import datetime
```
relrod, line 1: Can we move this under unit instead of functional? Since it's not using the django test client to call into the system.

Generally we've been separating the tests as: If it's just testing individual functions or behaviors, or creating things directly, then it goes under unit; if it's using the client to make actual (fake) http calls into the API, then it goes under functional.

ansible_base/tests/functional/test_organization_model.py
relrod, line 1: Can we move this test file under unit instead of functional? Since it's not using the django test client to call into the system.

Generally we've been separating the tests as: If it's just testing individual functions or behaviors, or creating things directly, then it goes under unit; if it's using the client to make actual (fake) http calls into the API, then it goes under functional.

AlanCoding: In `ansible_base/tests/settings_overrides.py` the app `ansible_base.tests.functional` is added. I think that psuedo-migrates (all tests are psuedo-migrations) to have these models in the test database. If you look at the tests, it creates an `Organization` object. This file tells django-pytest to create a table for organizations.

Understanding the spirit of your comment, I think you'll agree that functional tests better than unit tests for this. Maybe another division, separate from functional, would be better, but not unit.

(EDIT: before we jump on this great idea of separated functional-with-custom-models tests, realize that this doesn't make sense unless those tests use separate _settings_, and would probably have to be invoked separately, which gets messy)

relrod: It just feels weird to me to have 'functional' tests which are directly creating objects (i.e. using the internal project API and django calls rather than the publicly exposed REST API). We do have some that create objects, but only for the purpose of calling the (public) API path to make sure they show up correctly.

(I realize that AWX groups them together and has functional tests which create objects directly too and then tests things about those created objects without using the public API - in my mind that is more of a unit test than a functional test, but it's probably a blurry line).

My opinion isn't super strong, so I'm fine with leaving it as is.

cutwater: > Understanding the spirit of your comment, I think you'll agree that functional tests better than unit tests for this. Maybe another division, separate from functional, would be better, but not unit.

I think you have a good point. I acknowledge that might be well established practice in AWX, the tests organization feels weird to me. And adding third layer perfectly makes sense here, in my opinion.

Personally I've always treated unit tests as a tests for an isolated piece of code (a unit), which is somewhat distorted in Django, since database is not considered an external dependency, and thus tests for the code that directly accesses the database still lands into "unit" tests category.

In my past experience we used to have "unit" and "integration" tests separation. Integration tests were usually used to test the code that involves either has external dependencies or multiple components of the application.

What you you think about it?

AlanCoding: Much of what I wrote and other bits of the conversation are outdated as of https://github.com/ansible/django-ansible-base/pull/27 merging. Given the new test app, seems the appropriate place for test models would be in the test app. I believe that's the implied intent with that merge. In fact, this is why the PR has conflicts now.

AlanCoding: @relrod correct me if I'm wrong, but because this is a library, there is no publicly exposed API at all, unless we make use of a test app. Testing stuff which uses this as a dependency, AWX, galaxy_ng, etc. is a totally different category in my head. The benefit of a test app is to clearly separate concerns about what's django-ansible-base responsibility and what's not. Something like AWX has tons of baggage.

ansible_base/models/organization.py
john-westcott-iv: Closing this since net-new files are translated and we will have to go back through and translate remaining files.

docs/organizations.md
john-westcott-iv: Rename completed, closing this convo.

AlanCoding (approved):
Looks like all outstanding conversations are resolved with the rebase.

john-westcott-iv (approved):

PR #55, opened 2024-01-03
Optional requirements

john-westcott-iv:
Breaking out requirements.in into features and adding optional features. This will allow components to only add requirements to their project for whichever features they use.

The requirements/requirements.txt represents `all` features.


AlanCoding (approved):
Certainly looks promising, and people can still install [all] in any case

PR #58, opened 2024-01-04
Introduce custom user model

AlanCoding:
Still putting this up as a draft. I have tested AWX with this, and have a corresponding patch for aap-gateway.


john-westcott-iv:
Likely to be closed per @AlanCoding

AlanCoding:
yeah, b/c I am throwing in the towel on doing a custom user model in AWX

PR #59, opened 2024-01-08
Equal created and modified datetime for new records

cutwater:
New instances of model classes inherited from `CommonModel` are created with `created_on` and `modified_on` fields values that are not equal. Typically the `modified_on` field of a new record is either expected to be `None` or equal to `created_on` field value, so that one can determined if the record has been ever modified after creation.

This patch fixes function `now()` beeing called twice when saving a new record.


john-westcott-iv (approved):

AlanCoding (approved):
AWX has this flaw as well, and I would rather it not.

slemrmartin (approved):
:+1:

PR #62, opened 2024-01-10
Update ui_auth response to include custom login info and logo

Dostonbek1:
* change `/ui_auth/` endpoint to return `custom_login_info` and `custom_logo` settings so the UI can consume it

### DIFF
--- a/ansible_base/utils/validation.py
+++ b/ansible_base/utils/validation.py
@@ -113,12 +113,12 @@ def validate_image_data(data: str) -> None:
 
     match = CUSTOM_LOGO_RE.match(data)
     if not match:
-        raise ValidationError("Invalid format for custom logo. Must be a data URL with a base64-encoded GIF, PNG or JPEG image.")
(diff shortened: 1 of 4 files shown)


ansible_base/utils/validation.py
```diff
@@ -101,6 +104,23 @@ def validate_cert_with_key(public_cert_string, private_key_string):
...
+def validate_image_data(data: str) -> None:
+    # in case we are passed an empty string, we can skip validation
+    if not data:
+        return None
+
+    CUSTOM_LOGO_RE = re.compile(r'^data:image/(?:png|jpeg|gif);base64,([A-Za-z0-9+/=]+?)$')
+
+    match = CUSTOM_LOGO_RE.match(data)
+    if not match:
+        raise ValidationError("Invalid format for custom logo. Must be a data URL with a base64-encoded GIF, PNG or JPEG image.")
```
john-westcott-iv: Wrap this string and the one below in the translation module `_()`. See line 95 as an example.

john-westcott-iv (approved):

PR #63, opened 2024-01-10
Add get_setting function

john-westcott-iv:
This function will check the setting ANSIBLE_BASE_SETTINGS_FUNCTION and call that as a function if set so that a service can implement whatever kind of preferences/settings it wants

### DIFF
--- a/ansible_base/utils/settings.py
+++ b/ansible_base/utils/settings.py
@@ -1,0 +1,31 @@
+import importlib
+import logging
+from typing import Any
+
+from django.conf import settings
+
+logger = logging.getLogger('ansible_base.settings')
+
+
+class SettingNotSetException(Exception):
+    pass
+
+
+def get_setting(name: str, default: Any = None) -> Any:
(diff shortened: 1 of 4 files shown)


ansible_base/utils/settings.py
```diff
@@ -0,0 +1,28 @@
...
+        try:
+            module_name, _, function_name = settings_function.rpartition('.')
+            the_function = getattr(importlib.import_module(module_name), function_name)
+            setting = the_function(name)
+            return setting
+        except SettingNotSetException:
+            # If the setting was not set thats ok, we will fall through to trying to get it from the django setting or the default value
+            pass
+        except Exception:
+            logger.exception('ANSIBLE_BASE_SETTINGS_FUNCTION but calling it as a function failed (see exception), ignoring')
```
Dostonbek1: Does the exception message seem a bit unclear? Thoughts?

john-westcott-iv: Yea, that got messed up, thanks for checking. Updated.

Dostonbek1 (approved):

PR #69, opened 2024-01-15
Add mandatory configuration to Installation doc

slemrmartin:
#54 defines new mandatory setting. All initial configurations should be mentioned on one place. I'm adding it to Installation doc


AlanCoding:
It's only required if you have a model that inherits from the abstract organization model here.

AlanCoding:
Oh, I see that's not quite working as intended, thus these docs.

slemrmartin:
closed in favor of #75

PR #72, opened 2024-01-16
Prevent error-prone logic from running on import

AlanCoding:
The Django docs say to use the settings for AUTH_USER_MODEL when defining fields, but `get_user_model` basically when running code. Having the user model available in the imports and for type annotations is not addressed, so I'm saying that's unsupported, thus we shouldn't do it.

The other bit with `SimpleLazyObject` is addressing a report that `awx-manage collectstatic` could give an error. Reproducer with test_app:

```
python manage.py --help
```

This will error if I insert `SECRET_KEY = ""` in the test settings. But with this change it works.

### DIFF
--- a/ansible_base/channels/middleware.py
+++ b/ansible_base/channels/middleware.py
@@ -6,17 +6,15 @@ from channels.auth import get_user as get_session_user
 from channels.db import database_sync_to_async
 from channels.security.websocket import WebsocketDenier
 from channels.sessions import CookieMiddleware, SessionMiddleware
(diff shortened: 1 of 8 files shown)


AlanCoding:
Also, we are concretely hitting circular imports due to `get_user_model`, and I think this relates to problems with the `swappable_dependency` usage like what @john-westcott-iv mentioned, although his problems may be tangential.

slemrmartin (approved):
It looks good :+1:

john-westcott-iv (approved):

PR #73, opened 2024-01-16
Make team-organization relationship a ForeignKey

AlanCoding:
I think this already came up somewhere? But it didn't get done, so I figured I should do it now. After all, the whole point is to get these all aligned.


ansible_base/models/organization.py
```diff
@@ -24,8 +24,30 @@ class Meta:
         help_text=_("The list of users in this organization."),
     )
 
-    teams = models.ManyToManyField(
-        settings.ANSIBLE_BASE_TEAM_MODEL,
-        related_name="organizations",
-        help_text=_("The list of teams in this organization."),
+
+class AbstractTeam(NamedCommonModel):
+    """
```
john-westcott-iv: Can we move this to models/team.py?

cutwater (approved):

ansible_base/models/organization.py
AlanCoding: I guess my logic was that in AWX we lumped those all into organization.py. But yeah, just because it happened there doesn't mean it was a good idea (it was not).

john-westcott-iv (approved):

PR #74, opened 2024-01-16
Avoid importing ansible_encryption at start up

TheRealHaoLiu:
Importing of ansible_encryption trigger code that requires settings.SECRET_KEY which is not present in CI environment.

Example CI failure https://github.com/ansible/awx/actions/runs/7546412726/job/20544110452

CI test PR for awx with the code change https://github.com/ansible/awx/pull/14773

### DIFF
--- a/ansible_base/models/common.py
+++ b/ansible_base/models/common.py
@@ -8,8 +8,6 @@ from django.utils import timezone
 from inflection import underscore
 from rest_framework.reverse import reverse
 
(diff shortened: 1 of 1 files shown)


relrod (approved):

AlanCoding:
This was duplicated effort with https://github.com/ansible/django-ansible-base/pull/72, but I still want that in.

PR #81, opened 2024-01-19
Adding jwt_auth feature

john-westcott-iv:


ansible_base/lib/jwt_auth/common/auth.py
```diff
@@ -0,0 +1,197 @@
...
+        if not token:
+            logger.info("X-DAB-JW-TOKEN header not set for JWT authentication")
+            return None, None
+        logger.debug(f"Received token: {token}")
+
+        jwt_key_setting = None
+        try:
+            jwt_key_setting = settings.ANSIBLE_BASE_JWT_KEY
+        except NameError:
+            logger.info("Failed to get the setting ANSIBLE_BASE_JWT_KEY")
```
AlanCoding: Well if it requires a setting, does that make it a feature? I would rather not, but maybe this needs to be documented somewhere or maybe have a default in the dynamic settings?

john-westcott-iv: No, this is a good callout. We documented that if it requires a feature its an app and not a lib. That being said, I don't think we want to set a default for this because I wouldn't want someone using a "known" JWT key just because they forgot the setting and we catch if its not set and return an actual info message.

AlanCoding (approved):

ansible_base/lib/jwt_auth/common/auth.py
```diff
@@ -0,0 +1,197 @@
...
+    def __init__(self, user_fields=default_mapped_user_fields) -> None:
+        self.mapped_user_fields = user_fields
+
+    def parse_jwt_token(self, request):
+        logger.debug("Starting JWT Authentication")
+        token = request.headers.get("X-DAB-JW-TOKEN", None)
+        if not token:
+            logger.info("X-DAB-JW-TOKEN header not set for JWT authentication")
+            return None, None
+        logger.debug(f"Received token: {token}")
```
AlanCoding: For log hygiene, eventually, I would like to either remove some of these logs, or add more contextual information, since there are many "token" objects this could refer to. Also, could this token be really long? Maybe we could combine this with the prior `logger.debug` because it's better to have all the context on the same line.

ansible_base/lib/jwt_auth/common/auth.py
```diff
@@ -0,0 +1,197 @@
...
+            return None, None
+
+        decryption_key = self.get_decryption_key(
+            jwt_key_setting,
+            validate_certs=getattr(settings, "ANSIBLE_BASE_JWT_VALIDATE_CERT", True),
+            timeout=getattr(settings, "ANSIBLE_BASE_JWT_URL_TIMEOUT", 30),
+        )
+        validated_body = self.validate_token(token, decryption_key)
+        user_model = get_user_model()
+        user, created = user_model.objects.update_or_create(username=validated_body["sub"])
```
AlanCoding: I'm a little confused here by the lack of `defaults` and `create_defaults`. Sure, this could work if you used those, but the way it looks like it could make a new user with the desired username, but blank first_name, email, etc. and that can't possibly be intended.

AlanCoding:
Just to clarify - this has nothing to do with https://github.com/ansible/eda-server/pull/540 because that's some kind of special-purpose JWT for websocket connections?

ansible_base/lib/jwt_auth/common/auth.py
john-westcott-iv: This probably feels weird at first. 
In here we just map the sub because thats the actual username. But if you look above in this objects init you will see it takes a parameter `user_fields=default_mapped_user_fields` and default_mapped_user_fields is:
```
    "first_name", 
    "last_name", 
    "email", 
    "is_superuser", 
    "is_system_auditor", 
``` 

So this function only extracts the `sub` field and maps that into the `username` creating a user with just a username. However, this object `JWTCommonAuth` is called from `JWTAuthentication` below. Here we initialize with our `JWTCommonAuth` with `map_fields` (which is `default_mapped_fields`) and then, after we auth and get a user with just a username we call `map_user_fields` which will populate the remaining fields.

I did this because we don't have a common user model. That being said, its likely safe to assume that we will all at a minimum inherit form django.contri.auth.models.User which means we could add: first_name, last_name, email, is_superuser to the user create and truncate the `default_mapped_fields` to just `is_system_auditor`. 

LMK if you agree with that assumption or if we should leave it as is.

ansible_base/lib/jwt_auth/common/auth.py
john-westcott-iv: I'm torn on this because there are likely many tokens but there should only be one given the context of:
```
logger = logging.getLogger("ansible_base.lib.jwt_auth.common.auth")
```

ansible_base/lib/jwt_auth/common/auth.py
AlanCoding: > its likely safe to assume that we will all at a minimum inherit form django.contri.auth.models.User

I strongly believe that to be true. I have been somewhat trying to police the model definitions. eda-server adds `modified_at` which is weird but ok. Of course it uses the AbstractUser base. galaxy_ng uses a custom user model but _does no customizations_ which is for the better, since AWX, much to my chagrin uses `auth.User` itself. In fact, the only sever that does potentially meaningful or dangerous customization is Gateway which adds `is_system_auditor` which has a somewhat unclear/concerning support status in other servers.

So yes I would much prefer to set all AbstractUser fields on creation. You could get it from the abstract model's _meta concrete_fields or however makes sense.

chrismeyersfsu (approved):

kdelee:
fwiw right now getting 500 error from awx when make an unauthenticated request like `curl -k https://GATEWAY/api/controller/v2/`

chrismeyersfsu:
@kdelee I can maybe help get your env healthy. 

```
curl -k https://localhost/api/controller/v2/me/
{"detail":"Authentication credentials were not provided. To establish a login session, visit /api/controller/login/."}
```

ansible_base/lib/jwt_auth/common/auth.py
john-westcott-iv: Changed message slightly but resolving since it has its own logger.

ansible_base/lib/jwt_auth/hub/auth.py
```diff
@@ -0,0 +1,15 @@
+from ansible_base.lib.jwt_auth.common.auth import JWTAuthentication
+from ansible_base.lib.jwt_auth.common.exceptions import InvalidService
+
+
+class HubJWTAuth(JWTAuthentication):
```
chrismeyersfsu: Interesting. I expected this sort of code to live in hub, not in the shared library.

ansible_base/lib/jwt_auth/common/auth.py
```diff
@@ -0,0 +1,203 @@
...
+            return None, None
+
+        decryption_key = self.get_decryption_key(
+            jwt_key_setting,
+            validate_certs=getattr(settings, "ANSIBLE_BASE_JWT_VALIDATE_CERT", True),
+            timeout=getattr(settings, "ANSIBLE_BASE_JWT_URL_TIMEOUT", 30),
+        )
+        validated_body = self.validate_token(token, decryption_key)
+        user_model = get_user_model()
+        user, created = user_model.objects.update_or_create(
```
chrismeyersfsu: I reckon this assumes the User object to be a certain "shape" (have certain fields). Wondering if the resource sync system User model comes into play here.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -0,0 +1,203 @@
...
+]
+
+
+class JWTCommonAuth:
+    def __init__(self, user_fields=default_mapped_user_fields) -> None:
+        self.mapped_user_fields = user_fields
+
+    def parse_jwt_token(self, request):
+        logger.debug("Starting JWT Authentication")
+        token = request.headers.get("X-DAB-JW-TOKEN", None)
```
chrismeyersfsu, line 31: This is hard-coded, but a certain other project has a different default. I'm thinking this expected header should be configurable.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -0,0 +1,203 @@
...
+        if not token:
+            logger.info("X-DAB-JW-TOKEN header not set for JWT authentication")
+            return None, None
+        logger.debug(f"Received JWT auth token: {token}")
+
+        jwt_key_setting = get_setting("ANSIBLE_BASE_JWT_KEY")
```
chrismeyersfsu, line 37: Oh, maybe disregard my comment above. This seems to be the setting. Something doesn't seem quite right here though. Maybe we clean up this code after this merges.

ansible_base/lib/jwt_auth/common/auth.py
john-westcott-iv: We provide ways to customize what fields are in the user model but those fields would also have to be present in the JWT.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -0,0 +1,203 @@
...
+        if not jwt_key_setting:
+            logger.info("Failed to get the setting ANSIBLE_BASE_JWT_KEY")
+            return None, None
+
+        decryption_key = self.get_decryption_key(
+            jwt_key_setting,
+            validate_certs=get_setting("ANSIBLE_BASE_JWT_VALIDATE_CERT", True),
```
chrismeyersfsu, line 44: `ANSIBLE_BASE_JWT_VALIDATE_CERT` didn't _exactly_ do what I expected. I expected _all_ of the jwt validation to be short-circuited when I set this. This flag _does_ indicate that it's just about validating the cert. I guess I'm looking for another flag that would do the above.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -0,0 +1,203 @@
...
+        else:
+            logger.info(f"User {user.username} authenticated from JWT auth")
+
+        return user, validated_body
+
+    def log_and_raise(self, details):
+        logger.error(details)
+        raise AuthenticationFailed(details)
+
+    def get_decryption_key_from_url(self, url, timeout, validate_certs):
```
chrismeyersfsu, line 67: I triggered this code path by setting `ANSIBLE_BASE_JWT_KEY = 'https://aap_gw_proxy_1:9080'` BUT Django got pissy because the header `Host: aap_gw_proxy_1` violated some RFC and wouldn't serve my request. I've hit this problem in awx too with container named `tools_awx_1`.

We should bite the bullet and change our container names `s/_/-/g`

```
import socket, requests
ip_add = socket.gethostbyname("aap_gw_proxy_1")
ANSIBLE_BASE_JWT_KEY = requests.get(url=f'https://{ip_add}:9080/api/gateway/v1/jwt_key/', verify=False).text
```
^^ workaround

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # Decrypt the token
+        try:
+            logger.info("Decrypting token")
+            validated_body = jwt.decode(
+                token,
+                decryption_key,
+                audience="ansible-services",
+                options={"require": local_required_field},
+                issuer="ansible-issuer",
```
chrismeyersfsu, line 162: I had to change hard-coded gateway values to match in order to pass the jwt verification phase. Maybe these should be configurable?

chrismeyersfsu:
I was able to get gateway to send a JWT to AWX & verify the JWT in AWX with the below diff. I'm for merging this and following up with fixes. 

AWX mods

```
diff --git a/awx/api/authentication.py b/awx/api/authentication.py
index 48fc00db44..939acdec47 100644
--- a/awx/api/authentication.py
+++ b/awx/api/authentication.py
@@ -14,6 +14,9 @@ from rest_framework import authentication
 # Django-OAuth-Toolkit
 from oauth2_provider.contrib.rest_framework import OAuth2Authentication
 
+from ansible_base.lib.jwt_auth.common.auth import JWTAuthentication
+
+
 logger = logging.getLogger('awx.api.authentication')
 
 
@@ -49,3 +52,17 @@ class LoggedOAuth2Authentication(OAuth2Authentication):
             )
             setattr(user, 'oauth_scopes', [x for x in token.scope.split() if x])
         return ret
+
+
+class AwxJWTAuthentication(JWTAuthentication):
+    def authenticate(self, request):
+        print(f"Request headers {request.headers}")
+        res = super(AwxJWTAuthentication, self).authenticate(request)
+        print(f"Going to return {res}")
+        if res and len(res) == 2 and res[0] is None and res[1] is None:
+            return
+        else:
+            return res
+
+    def process_permissions(self, user, claims, token):
+        logger.error("Processing claims for {}".format(user.username))
diff --git a/awx/settings/defaults.py b/awx/settings/defaults.py
index f3f8f1881b..9b5135d399 100644
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -362,6 +362,7 @@ REST_FRAMEWORK = {
     'DEFAULT_PAGINATION_CLASS': 'awx.api.pagination.Pagination',
     'PAGE_SIZE': 25,
     'DEFAULT_AUTHENTICATION_CLASSES': (
+        'awx.api.authentication.AwxJWTAuthentication',
         'awx.api.authentication.LoggedOAuth2Authentication',
         'awx.api.authentication.SessionAuthentication',
         'awx.api.authentication.LoggedBasicAuthentication',
diff --git a/awx/settings/development.py b/awx/settings/development.py
index 25ec408ad9..887205ea37 100644
--- a/awx/settings/development.py
+++ b/awx/settings/development.py
@@ -117,3 +117,14 @@ from .application_name import set_application_name
 set_application_name(DATABASES, CLUSTER_HOST_ID)  # NOQA
 
 del set_application_name
+
+
+LOGGING['loggers']['ansible_base'] = {
+    'handlers': ['console'],
+    'level': 'DEBUG'
+}
+
+import socket, requests
+ip_add = socket.gethostbyname("aap_gw_proxy_1")
+ANSIBLE_BASE_JWT_KEY = requests.get(url=f'https://{ip_add}:9080/api/gateway/v1/jwt_key/', verify=False).text
+ANSIBLE_BASE_JWT_VALIDATE_CERT = False
\ No newline at end of file
diff --git a/requirements/requirements_git.txt b/requirements/requirements_git.txt
index 0d2bef4430..62aaf2a8df 100644
--- a/requirements/requirements_git.txt
+++ b/requirements/requirements_git.txt
@@ -5,4 +5,4 @@ git+https://github.com/ansible/ansible-runner.git@devel#egg=ansible-runner
 # specifically need https://github.com/robgolding/django-radius/pull/27
 git+https://github.com/ansible/django-radius.git@develop#egg=django-radius
 git+https://github.com/ansible/python3-saml.git@devel#egg=python3-saml
-django-ansible-base @ git+https://github.com/john-westcott-iv/django-ansible-base@app_split#egg=django-ansible-base[rest_filters]
\ No newline at end of file
+django-ansible-base @ git+https://github.com/john-westcott-iv/django-ansible-base@service_lib_to_dab#egg=django-ansible-base[rest_filters,api_documentation,authentication]
```

Gateway modifications

https://localhost/api/gateway/v1/settings/all/ <-- set `"gateway_token_name": "X-DAB-JW-TOKEN"`

```
diff --git a/aap_gateway_api/utils/jwt_token.py b/aap_gateway_api/utils/jwt_token.py
index 6a83a73..59218b2 100644
--- a/aap_gateway_api/utils/jwt_token.py
+++ b/aap_gateway_api/utils/jwt_token.py
@@ -17,9 +17,9 @@ def create_signed_jwt(user):
     due_date = datetime.now() + timedelta(seconds=get_preference_value("proxy", "gateway_access_token_expiration"))
     teams = [{"name": team.name, "organization": team.organization.name} for team in user.teams.select_related("organization").all()]
     payload = {
-        "iss": "aap-gateway",
+        "iss": "ansible-issuer",
         "exp": int(due_date.timestamp()),
-        "aud": "aap-services",
+        "aud": "ansible-services",
         "sub": user.username,
         "first_name": user.first_name,
         "last_name": user.last_name,
diff --git a/requirements/requirements_git.txt b/requirements/requirements_git.txt
index 797391e..9c0825d 100644
--- a/requirements/requirements_git.txt
+++ b/requirements/requirements_git.txt
@@ -1,2 +1,2 @@
 # If you change the optional dependencies please also change them in tools/scripts/ci/tox-reinstall-django-ansible-base.sh
-django-ansible-base @ git+https://github.com/ansible/django-ansible-base@devel#egg=django-ansible-base[rest_filters,api_documentation,authentication]
+django-ansible-base @ git+https://github.com/john-westcott-iv/django-ansible-base@service_lib_to_dab#egg=django-ansible-base[rest_filters,api_documentation,authentication]
```

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -0,0 +1,203 @@
...
+        common_auth = JWTCommonAuth(self.map_fields)
+        user, token = common_auth.parse_jwt_token(request)
+
+        if user:
+            self.process_user_data(user, token)
+            self.process_permissions(user, token.get("claims", None), token)
+
+            return user, None
+        else:
+            return None, None
```
chrismeyersfsu, line 196: I can't exactly remember the contract with Django auth middleware. I _think_ you are supposed to return `None` (not a tuple of 2 nones) if your auth is not being used. If that is the case, I'd think another return case is needed in this function to short-circuit with a `return` when/if the JWT header is NOT present.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def authenticate(self, request):
+        common_auth = JWTCommonAuth(self.map_fields)
+        user, token = common_auth.parse_jwt_token(request)
+
+        if user:
+            self.process_user_data(user, token)
+            self.process_permissions(user, token.get("claims", None), token)
+
+            return user, None
```
chrismeyersfsu, line 194: awx oauth authentication class returns the oauth token as the second tuple value. I wonder if we should return the token here?

rcarrillocruz (approved):

ansible_base/jwt_consumer/common/auth.py
john-westcott-iv: This is rest_framework authenticaiton and those are all returning tuples: https://github.com/encode/django-rest-framework/blob/master/rest_framework/authentication.py

ansible_base/jwt_consumer/common/auth.py
john-westcott-iv: It looks like the only one from the rest framework that returns the token is the token auth https://github.com/encode/django-rest-framework/blob/master/rest_framework/authentication.py. We have a JWT token but I don't think it would be super useful to return that since they are pretty short lived.

ansible_base/jwt_consumer/common/auth.py
john-westcott-iv: I'm not sure I'd want a setting to ignore the decryption of the JWT.

PR #84, opened 2024-01-23
Fix summary_fields for the User model, document expectations

AlanCoding:
So this doesn't really fix anything at this point, but it gets us to at least identifying the problem. Here, as you can see, I define a customer user model that defines `summary_fields`. Maybe this is a sufficient argument to put in a new abstract user model (beyond the `is_system_auditor` which I closed, but probably shouldn't have).

### DIFF
--- a/ansible_base/lib/utils/models.py
+++ b/ansible_base/lib/utils/models.py
@@ -2,6 +2,8 @@ from itertools import chain
 
 from inflection import underscore
 
+from django.contrib.auth import get_user_model
+
 
 def get_all_field_names(model):
     # Implements compatibility with _meta.get_all_field_names
@@ -42,3 +44,17 @@ def prevent_search(relation):
     """
     setattr(relation, '__prevent_search__', True)
     return relation
+
+
+def user_summary_fields(user):
+    sf = {}
+    # field names come from from AWX awx.api.serializers
+    for field_name in ('id', 'username', 'first_name', 'last_name'):
+        sf[field_name] = getattr(user, field_name)
+    return sf
+
+
+def decorate_user_model():
+    user_cls = get_user_model()
+    if not hasattr(user_cls, 'summary_fields'):
+        user_cls.add_to_class('summary_fields', user_summary_fields)
--- a/ansible_base/lib/utils/models.py
+++ b/ansible_base/lib/utils/models.py
@@ -47,7 +47,6 @@ def prevent_search(relation):
 
 def user_summary_fields(user):
     sf = {}
-    # field names come from from AWX awx.api.serializers
     for field_name in ('id', 'username', 'first_name', 'last_name'):
         sf[field_name] = getattr(user, field_name)
     return sf
--- a/ansible_base/lib/utils/models.py
+++ b/ansible_base/lib/utils/models.py
@@ -1,8 +1,7 @@
 from itertools import chain
 
-from inflection import underscore
-
 from django.contrib.auth import get_user_model
+from inflection import underscore
 
 
 def get_all_field_names(model):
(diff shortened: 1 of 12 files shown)


AlanCoding:
Update - I went ahead and put in logic such that the issue I filed will be completely resolved.

ansible_base/lib/utils/models.py
```diff
@@ -42,3 +43,17 @@ class AuthToken(BaseModel):
     """
     setattr(relation, '__prevent_search__', True)
     return relation
+
+
+def user_summary_fields(user):
```
john-westcott-iv, line 48: Does this need to be a separate method here?

ansible_base/lib/utils/models.py
```diff
@@ -42,3 +43,17 @@ class AuthToken(BaseModel):
...
+    sf = {}
+    # field names come from from AWX awx.api.serializers
```
john-westcott-iv: Can we drop this comment?

test_app/urls.py
```diff
@@ -2,9 +2,11 @@
 from django.urls import include, path, re_path
 
 from ansible_base.lib.dynamic_config.dynamic_urls import api_urls, api_version_urls, root_urls
+from test_app.views import router as user_router
 
 urlpatterns = [
     path('api/v1/', include(api_version_urls)),
+    path('api/v1/', include(user_router.urls)),
```
john-westcott-iv: Hmmm.... we don't say this in the documentation but we had indirectly said before that a service should puil in all of the URLs from DAB first before adding any of their own. That way if something got hosed up in DAB and `api_urls` contained an entry like api/v1 it wouldn't be able to overwrite the services urls. So this line should be after `path('', include(root_urls)),` and we should probably formally note this better in Installation.md.

john-westcott-iv: Also, we should think about how we want to use routers. In other services that use them we have a routers.py file at the root instead of in the views. It would be nice to be consistent across services.

test_app/settings.py
```diff
@@ -75,6 +75,8 @@
     }
 }
 
+AUTH_USER_MODEL = 'test_app.User'
```
john-westcott-iv, line 80: I really wish this could come out of DAB :(

AlanCoding: I don't follow what this comment means. Like, an `AbstractUser` model? AWX couldn't use it, and all other apps already have a custom user model so not having an `AbstractUser` makes the most sense to me right now.

ansible_base/lib/utils/models.py
AlanCoding: separate from what? This is the method we attach to `AUTH_USER_MODEL` (if needed), I don't see how that can be reduced.

test_app/urls.py
AlanCoding: Yes to the prior thing, to have test_app model the intended use. Pushed.

ansible_base/lib/utils/models.py
john-westcott-iv: I was thinking merge this logic into decorate_user_model since its only called from there. i.e. build the `sf` variable in `decorate_user_model` and pass that into the `add_to_class` call there.

test_app/settings.py
john-westcott-iv: Yea, I was just lamenting about not having an Abstract (or even non-Abstract) user model in DAB. No changes needed to this PR.

john-westcott-iv (approved):

ansible_base/lib/utils/models.py
AlanCoding: That doesn't make sense to me because `sf` is computed from the function argument `user` object. `decorate_user_model` doesn't have the user properties needed for it.

PR #86, opened 2024-01-23
[AAP-19247] Unify Authenticator User auth_time field

Dostonbek1:
Update `auth_time` field in AuthenticatorUser model to be consistent across all authenticator types.

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/local.py
+++ b/ansible_base/authentication/authenticator_plugins/local.py
@@ -2,6 +2,8 @@ import logging
 
 from django.contrib.auth.backends import ModelBackend
 from django.core.exceptions import ValidationError
+from django.utils.timezone import now
+from rest_framework.serializers import DateTimeField
 
 from ansible_base.authentication.authenticator_plugins.base import AbstractAuthenticatorPlugin, BaseAuthenticatorConfiguration
 from ansible_base.authentication.models import AuthenticatorUser
@@ -37,7 +39,15 @@ class AuthenticatorPlugin(ModelBackend, AbstractAuthenticatorPlugin):
         # This auth class doesn't create any new local users, so we just need to make sure
         # it has an AuthenticatorUser associated with it.
         if user:
-            AuthenticatorUser.objects.get_or_create(uid=username, user=user, provider=self.database_instance)
+            user_attrs = {
+                "username": username,
+                "first_name": user.first_name,
+                "last_name": user.last_name,
+                "email": user.email,
+                "is_superuser": user.is_superuser,
+                "auth_time": DateTimeField().to_representation(now()),
+            }
+            AuthenticatorUser.objects.update_or_create(uid=username, provider=self.database_instance, defaults={"extra_data": user_attrs})
 
         # TODO, we will need to return attributes and claims eventually
         return user
--- a/ansible_base/authentication/authenticator_plugins/local.py
+++ b/ansible_base/authentication/authenticator_plugins/local.py
@@ -2,11 +2,8 @@ import logging
 
 from django.contrib.auth.backends import ModelBackend
 from django.core.exceptions import ValidationError
-from django.utils.timezone import now
-from rest_framework.serializers import DateTimeField
 
 from ansible_base.authentication.authenticator_plugins.base import AbstractAuthenticatorPlugin, BaseAuthenticatorConfiguration
-from ansible_base.authentication.models import AuthenticatorUser
 from ansible_base.authentication.utils.claims import get_or_create_authenticator_user
 
 logger = logging.getLogger('ansible_base.authentication.authenticator_plugins.local')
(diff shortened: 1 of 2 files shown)


ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -37,7 +38,21 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
         if user:
-            AuthenticatorUser.objects.get_or_create(uid=username, user=user, provider=self.database_instance)
+            logger.info(f"user 2: {user.__dict__}")
+            user_attrs = {
+                "username": username,
+                "first_name": user.first_name,
+                "last_name": user.last_name,
+                "email": user.email,
+                "is_superuser": user.is_superuser,
+                "auth_time": now().isoformat().replace('+00:00', 'Z')
```
AlanCoding: This does raise eyebrows. What are you doing and why? I think you want to save the datetime in the JSON content in the same way it is presented in the API. What is the existing logic that the API uses to do this? I think the origin comes to this:

https://github.com/encode/django-rest-framework/blob/552a67acde5eb13f84ec9dbdf4363313cdfda2f0/rest_framework/fields.py#L1212-L1214

So my code preference would be to be deferential to the source logic. So instead of

```python
In [4]: now().isoformat().replace('+00:00', 'Z')
Out[4]: '2024-01-23T16:39:16.693724Z'
```

I would find this preferable

```python
In [7]: DateTimeField().to_representation(now())
Out[7]: '2024-01-23T16:44:02.038764Z'
```

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -37,7 +38,21 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+            }
+            try:
+                auth_user = AuthenticatorUser.objects.get(uid=username, provider=self.database_instance)
+                auth_user.extra_data = user_attrs
+                auth_user.save()
+            except AuthenticatorUser.DoesNotExist:
+                AuthenticatorUser.objects.create(uid=username, provider=self.database_instance, extra_data=user_attrs)
```
john-westcott-iv: We should be able to use `update_or_create` here right?

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -37,7 +38,21 @@ def authenticate(self, request, username=None, password=None, **kwargs):
         # This auth class doesn't create any new local users, so we just need to make sure
         # it has an AuthenticatorUser associated with it.
         if user:
-            AuthenticatorUser.objects.get_or_create(uid=username, user=user, provider=self.database_instance)
+            logger.info(f"user 2: {user.__dict__}")
```
john-westcott-iv: If this is debug log make sure it gets removed, if not make it more descriptive.

john-westcott-iv (changes requested):
Love where this is going. It will be really nice to have this implemented.

AlanCoding (approved):

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -37,7 +39,15 @@ def authenticate(self, request, username=None, password=None, **kwargs):
         # This auth class doesn't create any new local users, so we just need to make sure
         # it has an AuthenticatorUser associated with it.
         if user:
-            AuthenticatorUser.objects.get_or_create(uid=username, user=user, provider=self.database_instance)
```
AlanCoding, line 40: so I guess my reading of the test failure is that `user=user` is still needed to satisfy the non-null constraint of the field.

Once that's fixed, what you're doing looks good.

Dostonbek1: @AlanCoding Switched to re-using the existing `get_or_create_authenticator_user` function that we are already using for LDAP authenticator, as per @john-westcott-iv 's suggestion.

ansible_base/authentication/utils/claims.py
```diff
@@ -261,6 +262,10 @@ def update_user_claims(user, database_authenticator, groups):
 
     needs_save = False
     authenticator_user, _ = AuthenticatorUser.objects.get_or_create(provider=database_authenticator, user=user)
+    # update the auth_time field to align with the general format used for other authenticators
+    authenticator_user.extra_data = {**authenticator_user.extra_data, "auth_time": DateTimeField().to_representation(now())}
+    authenticator_user.save()
+
```
Dostonbek1, line 268: @AlanCoding This might not be very efficient but couldn't figure out how to include it in the initial call on the AuthenticatorUser on line 264. I'm open for any feedback on this, tbh.

AlanCoding: The reason for this makes sense to me. It would be good to add `update_fields=["extra_data"]` to help a little with efficiency/isolation.

Dostonbek1: @AlanCoding Updated the PR.

john-westcott-iv (approved):

PR #68, opened 2024-01-15
Add explicit packages list to pyproject.toml

bpedersen2:
fixes: #67

PR #89, opened 2024-01-24
Exclude test_app from distribution

cutwater:


john-westcott-iv (approved):

relrod (approved):

AlanCoding (approved):

slemrmartin (commented):
@cutwater it looks like revert of #68.
In that case, please prepare a revert for https://github.com/ansible/aap-gateway/pull/189 to ensure gateway can be started in Docker

cutwater:
@slemrmartin I'm not sure why do you think that's a revert of #68.
This PR specifies explicitly the packages that are included into a distribution package.

I will prepare a revert for https://github.com/ansible/aap-gateway/pull/189, but just for awareness, the revert of that change shouldn't change the behavior of docker image build process, therefore this PR can be merged independently.

PR #90, opened 2024-01-24
Exclude redundant files from source distribution 

cutwater:


john-westcott-iv:
I think I'd prefer to try and keep this in the pyproject.toml file if possible. If you scroll down in this [section of setuptools docs](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration) they have a green tip where they show a way to include/exclude files based on patterns.

cutwater:
I don't think it's possible to control source distribution contents via `pyproject.toml`. The section you're referring to if my understanding is correct refers to the package data, not sdist contents.

AlanCoding:
I don't really have any preference or opinions about this PR.

john-westcott-iv (approved):

PR #91, opened 2024-01-24
fix: update jwt auth user model method

Dostonbek1:


AlanCoding:
From the documentation it's not obvious what you're doing

https://docs.djangoproject.com/en/5.0/ref/models/querysets/

If you intend these to be the defaults kwarg, I would prefer to see `defaults=`

But then I'm still confused. The model is username-unique, so I would expect `username` to be its own kwarg which is used for lookups.

Dostonbek1:
> From the documentation it's not obvious what you're doing
> 
> https://docs.djangoproject.com/en/5.0/ref/models/querysets/
> 
> If you intend these to be the defaults kwarg, I would prefer to see `defaults=`
> 
> But then I'm still confused. The model is username-unique, so I would expect `username` to be its own kwarg which is used for lookups.

This was to fix a bug that I hit when I was trying to run EDA with JWT auth where we had a local user with `username=admin` and JWT Auth had a user with the same username but instead of updating the existing user, it was trying to create a new user resulting in the following error
```
docker-eda-api-1                | django.db.utils.IntegrityError: duplicate key value violates unique constraint "core_user_username_key"
docker-eda-api-1                | DETAIL:  Key (username)=(admin) already exists.
```

AlanCoding:
Yeah, I think you're correct that moving fields from kwargs to defaults is the proper solution. Just not all fields. Example:

```python
User.objects.update_or_create(username='alice', defaults={"email": "alice@teaparty.com"})
```

Tested this pretty easily, if "alice" exists, then it grabs that user and sets the email address. If that user doesn't exist it creates a new "alice" user with that email.

john-westcott-iv:
Lets implement what @AlanCoding suggested and rearrange this like:
```
        user, created = user_model.objects.update_or_create(
            "username": validated_body["sub"],
            defaults = {
                "first_name": validated_body["first_name"],
                "last_name": validated_body["last_name"],
                "email": validated_body["email"],
                "is_superuser": validated_body["is_superuser"],
            }
        )
  ```
  
  I think that is a little clearer.

john-westcott-iv (approved):
Pre-approving pending the move of username outside of the dict and the dict being set to `defaults=`

PR #95, opened 2024-01-25
Add boostrap script for test_app

AlanCoding:


test_app/management/commands/create_demo_data.py
```diff
@@ -0,0 +1,30 @@
...
+from test_app.models import EncryptionModel, Organization, Team, User
+
+
+class Command(BaseCommand):
+    help = 'Creates demo data for development.'
+
+    def handle(self, *args, **kwargs):
+        if Organization.objects.exists():
+            print('It looks like you already have data loaded, great!')
+            return
```
john-westcott-iv, line 13: Instead of this do you want to just do `update_or_create` calls below?

john-westcott-iv (approved):
One comment is totally optional. If you just want to merge I'm good with that too.

test_app/management/commands/create_demo_data.py
AlanCoding: I kind of don't want to do it because it creates more work about figuring out what lookup fields are. I want to keep this scrappy right now.

PR #96, opened 2024-01-25
fix: change authenticate to return None instead of tuple

Dostonbek1:
When a user is not authenticated, we get a response with 403 Forbidden instead of 401 Unauthorized. This PR fixes this issue. 

NOTE: we still need to write tests around this but this is a quick fix to unblock [my PR on EDA](https://github.com/ansible/eda-server/pull/622) to fix QE pipelines.

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -196,7 +196,7 @@ class JWTAuthentication(BaseAuthentication):
 
             return user, None
         else:
-            return None, None
+            return None
 
     def process_user_data(self, user, token):
         common_auth = JWTCommonAuth(self.map_fields)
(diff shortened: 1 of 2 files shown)


ansible_base/jwt_consumer/common/auth.py
```diff
@@ -196,7 +196,7 @@ def authenticate(self, request):
 
             return user, None
         else:
-            return None, None
+            return None
```
mkanoor, line 199: @Dostonbek1 This is the difference between DRF Authentication and Django Authentication. The DRF authentication returns a Tuple (user, token) while the Django Authentication returns a single value(user) . Is the context for JWTAuthentication DRF or Django?

john-westcott-iv:
@chrismeyersfsu FYI

chrismeyersfsu (approved):

PR #98, opened 2024-01-26
Allow system user in attribution fields

relrod:
Also:
- Provide m2m link in related fields
- Rework some test fixtures around logging to have a nicer API
- Add tests for the above
- Regenerate test_app migrations to squash them into one.
- Make test_app use custom user model that extends CommonModel


test_app/models.py
```diff
@@ -19,8 +20,11 @@ class Organization(AbstractOrganization):
     pass
 
 
-class User(AbstractUser):
-    pass
+class User(AbstractUser, CommonModel):
+    # This doesn't get injected automatically by utils.models.decorate_user_model
+    # because we extend CommonModel which already inserts a summary_fields attribute.
+    def summary_fields(self):
+        return user_summary_fields(self)
```
AlanCoding, line 25: I already added logic to attach this method, I flied an issue for it and merged the fix which adds summary_fields when apps are ready

https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/authentication/apps.py#L19

AWX will not have a custom user model. I tried very hard to get a solution, and it was a nightmare, and I don't have the time to complete that migration. The reason for dong the solution in apps ready methods is because that will work for services that don't have a custom model, which is going to be reality because that problem is nigh unfixable.

test_app/settings.py
```diff
@@ -46,7 +46,7 @@
     'test_app',
 ]
 
-AUTH_USER_MODEL = 'auth.User'
+AUTH_USER_MODEL = 'test_app.User'
```
AlanCoding: this is defined twice in this file. That was a mistake and the earlier assignment (this) should be deleted.

AlanCoding (changes requested):
We discussed and have concrete TODO items for merge.
 - remove duplicate AUTH_USER_MODEL setting from test_app
 - delete my existing solution for User.summary_fields, file an issue to implement AWX-specific solution
 
I've looked through the rest of the content, no blockers.

AlanCoding (approved):

PR #101, opened 2024-01-26
Implement TACACS Authenticator

djyasin:
Add plugin for TACACS auth 

Testing steps:

- In `container-startup.yml`, for `container_reference:` provide your IP address or `host.docker.internal`
- Uncomment `tacacs_enabled: True`
- You should see `aap_gw_tacacs_1`
- Then you should be able to log in with username `iosadmin` and password `cisco`


ansible_base/authentication/authenticator_plugins/tacacs.py
```diff
@@ -0,0 +1,140 @@
...
+
+from ansible_base.authentication.authenticator_plugins.base import AbstractAuthenticatorPlugin, BaseAuthenticatorConfiguration
+from ansible_base.authentication.models import AuthenticatorUser
+from ansible_base.lib.serializers.fields import CharField, IntegerField, ChoiceField, BooleanField
+
+logger = logging.getLogger('ansible_base.authentication.authenticator_plugins.tacacs')
+
+
+class TacacsConfiguration(BaseAuthenticatorConfiguration):
+    documentation_url = "https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.backends.ModelBackend"
```
gundalow: Do we pin the version to 4.2 in the code somewhere?

djyasin:
@john-westcott-iv  I was able to get rid of all of the errors I was getting when trying to test `django-ansible-base/ansible_base/authentication/authenticator_plugins/tacacs.py` by removing the `category`, `category_slug`, and `encrypted` fields. I didn't see any of those anywhere in keyloack or LDAP, so it seemed like an okay move. Let me know if I should've resolved this differently. The `SECRET` field had "encrypted = true`. It looks like that will be done elsewhere. Please let me know if I should add that back and declare it somewhere.

Dostonbek1 (approved):

PR #102, opened 2024-01-26
Allow opting out of nonexistent sys user warning

relrod:
Follow-up to #98.

This makes it so that apps setting up their system user have a way to call .save() without getting warned.

Also drops the severity from error to warning.

### DIFF
--- a/ansible_base/lib/abstract_models/common.py
+++ b/ansible_base/lib/abstract_models/common.py
@@ -55,8 +55,7 @@ class CommonModel(models.Model):
         help_text="The user who last modified this resource",
     )
 
-    def save(self, *args, warn_nonexistent_system_user=True, **kwargs):
-        update_fields = list(kwargs.get('update_fields', []))
+    def _attributable_user(self):
         user = get_current_user()
         if user is None:
             # If no user is logged in, we try attributing the action to the system user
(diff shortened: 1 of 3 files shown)


ansible_base/lib/abstract_models/common.py
```diff
@@ -66,8 +65,14 @@ def save(self, *args, **kwargs):
                 try:
                     user = get_user_model().objects.get(username=system_username)
                 except get_user_model().DoesNotExist:
-                    logger.error(f"SYSTEM_USERNAME is set to {system_username} but no user with that username exists. User attribution will be None.")
+                    if warn_nonexistent_system_user:
+                        logger.warn(f"SYSTEM_USERNAME is set to {system_username} but no user with that username exists. User attribution will be None.")
                     user = None
+        return user
+
+    def save(self, *args, warn_nonexistent_system_user=True, **kwargs):
```
AlanCoding, line 73: My strange first-impression reaction to this PR was that since `SYSTEM_USERNAME` is defined in settings, the opt-out of the log should also be defined in settings. I'd like a little better sense of what code is expected to use this. I can't think of what would.

Like... I imagine you might not want the log when calling `awx-manage createsuperuser` because as it tries to set `created_by` on the admin user. But... that's Django's code, and it wouldn't be practical to add this kwarg to the save.

AlanCoding (approved):

ansible_base/lib/abstract_models/common.py
relrod: So the main motivation for this change is: Say an app wants to create a system user on install (perhaps as part of a migration). The SYSTEM_USERNAME will be defined, and when the user is created, we don't want to warn, since we're creating the user, so it doesn't make sense.

So for that particular instance a setting doesn't really work.

PR #103, opened 2024-01-26
Use routers in the authentication app

AlanCoding:


AlanCoding:
welp, don't know why it doesn't work

PR #104, opened 2024-01-27
Add GitHub workflow "Release django-ansible-base"

john-westcott-iv:
This workflow will create a GitHub release of django-ansible-base using calver for its versioning.
Calver uses versioning in the format 'yyyy.m.d'. 
If we need to release twice in a day we can simply delete the release and the the tag and run the build again.
The release name will be `v<version>` and additionally a tag of `<version>` will be created.
During the build process we will generate a version file in DAB and then build the release generating the `tar.gz` and `whl` files. These files are uploaded to GitHub release as assets.

An example release created from this action can be seen [here](https://github.com/john-westcott-iv/django-ansible-base/releases/tag/2024.1.31).

Currently when you install from a github release like:
```
pip install git+https://github.com/john-westcott-iv/django-ansible-base@2024.01.27#egg=django-ansible-base[filtering,swagger,authentication]
```

The version comes out as the dev version because the updated `_version.py` file is not committed to the tag. But if you install from the uploaded `tar.gz` or `whl` file like:
```
pip install https://github.com/john-westcott-iv/django-ansible-base/releases/download/2024.01.27/django-ansible-base-2024.1.27.tar.gz
```

The package will have the correct version (2024.1.27).

We could add a task to the `tools/ansible/release.yml` file to commit the updated `_version.py` file before cutting the release which would allow a generic get install to have the correct version but unless we also cut a branch the dev version will always be the latest version. However, we decided this was an acceptable issue given that we are performing this install from a requirements_git file. In the future, if we release to pypi and install from there this issue will be mute. 

I also noted that around 8pm est I started seeing the builds come out as the next day... perhaps the GH action server is running in UTC?


ansible_base/__init__.py
```diff
@@ -0,0 +1,18 @@
+try:
+    from ._version import __version__, __version_tuple__  # noqa: F401
+except Exception:
+    import datetime
+    import subprocess
+    from typing import Tuple, Union
+
+    calver_now = datetime.datetime.now().strftime("%Y.%m.%d")
+    shaw = subprocess.check_output(['git', 'rev-parse', '--short', 'HEAD']).decode('ascii').strip()
```
shanemcd: This doesn't feel right. And after reading further I am left wondering why we are doing any of the stuff in this file.

shanemcd (approved):

relrod (approved):

webknjaz:
@john-westcott-iv any reason this is a separate workflow that does not depend on the tests passing?

PR #105, opened 2024-01-29
Fix login redirect override error on empty string

Dostonbek1:
This PR aims to resolve the following exception error stack when LOGIN_REDIRECT_OVERRIDE is an empty string:
```
  File "/usr/lib/python3.9/site-packages/ansible_base/authentication/views/ui_auth.py", line 57, in generate_ui_auth_data
    validate_url(url=login_redirect_override, allow_plain_hostname=True)
  File "/usr/lib/python3.9/site-packages/ansible_base/lib/utils/validation.py", line 64, in validate_url
    raise ValidationError(e.message)
rest_framework.exceptions.ValidationError: [ErrorDetail(string='Enter a valid URL.', code='invalid')]
```

### DIFF
--- a/ansible_base/authentication/views/ui_auth.py
+++ b/ansible_base/authentication/views/ui_auth.py
@@ -57,6 +57,7 @@ def generate_ui_auth_data():
         # ignore validation if login_redirect_override is None or empty string
         if login_redirect_override is not None and login_redirect_override != '':
             validate_url(url=login_redirect_override, allow_plain_hostname=True)
+            response['login_redirect_override'] = login_redirect_override
     except ValidationError:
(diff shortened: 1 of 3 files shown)


john-westcott-iv (commented):
Do we also add the change in the ui_auth view to go from `exception` to `error`? If something was invalid I don't think the exceptions are very helpful for troubleshooting. But perhaps if one of the settings is not `''` than maybe its helpful?

john-westcott-iv (approved):

Dostonbek1:
@john-westcott-iv Just FYI, I moved the None and empty string check out of the URL validator function because 1) it seemed a specific case for LOGIN_REDIRECT_OVERRIDE that we allow an empty string and 2) Neither None nor empty string is technically a valid URL.

PR #116, opened 2024-02-01
Let model fields be flexible with related views

relrod:
- Allow model fields to have a way to get out of showing related views
- Allow model fields to specify their own related view instead of the guessed one
- Also change the default to use the field name instead of the related model to generate the "guessed route".


ansible_base/lib/abstract_models/common.py
```diff
@@ -54,6 +55,7 @@ class Meta:
         on_delete=models.DO_NOTHING,
         help_text="The user who last modified this resource",
     )
+    modified_by.related_view = 'user-detail'
```
AlanCoding, line 58: I don't love that the specification is on the field level. You're matching to the `basename=` given in routers most of the time... hopefully. That would better map to the `User` model. AWX had `get_absolute_url` on models, and I don't want to copy bad practices from AWX, but that has some background general to Django.

So basically I would rather set the router basename in the model when it's not obvious, and do a really good job of documenting

ansible_base/lib/abstract_models/common.py
```diff
@@ -123,6 +125,29 @@ def get_summary_fields(self):
...
+
+        if reverse_view is None or reverse_view:
+            # Give fields a chance to opt out of showing up here by forcing related_view to None
+            return reverse_view
+
+        # otherwise reverse_view is False, meaning we should try to guess the view name
+        if isinstance(field, models.ManyToManyField):
+            return f"{underscore(self.__class__.__name__)}-{field.name}-list"
+
+        return f"{underscore(field.name)}-detail"
```
AlanCoding, line 149: This line eats at me. Previously with `reverse_view = f"{underscore(field.related_model.__name__)}-detail"` in place of this, we could have relied on the guess instead of adding customization.

You've made really good progress in DAB standardizing models so that we don't need to write extra view and serializer logic for everything, and I'd like to continue that. I think this goes in the wrong direction.

ansible_base/lib/abstract_models/common.py
AlanCoding: I'm only just starting to get the problem this would solve. The test_app model does the best job of articulating a scenario:

```
    more_teams = models.ManyToManyField(Team, related_name='related_fields_test_model_more_teams')
    more_teams.related_view = "related_fields_test_model-more_teams-list"
```

Yeah, there could be a _related_ endpoint with a very custom name we need to link to. However, this isn't ever seen in a real model. Nor would I expect it to be, because the only app we're trying to use this for creates related views by a factory... so the guess should almost always be sufficient. In cases like AWX this will probably be relevant, but when that happens we likely won't be able to use this feature because the models will be defined in DAB and AWX can't easily set this (well it can, but I think you might balk at it yourself).

I still think this should never be used for ForeignKey fields or for detail views.

PR #119, opened 2024-02-02
Fix resource registry migrations

newswangerd:
Fix two bugs:
- Resource post_save and post_delete signals were getting called in the migrations
- The post migration signal to add resources to the system was broken


ansible_base/resource_registry/apps.py
```diff
@@ -52,25 +54,34 @@ def initialize_resources(sender, **kwargs):
...
             r_type.save()
 
 
+def connect_signals(*args, **kwargs):
+    from django.db.models.signals import post_delete, post_save
+
+    from ansible_base.resource_registry.signals import handlers
+
+    post_save.connect(handlers.update_resource)
+    post_delete.connect(handlers.remove_resource)
```
AlanCoding, line 76: Trying to understand why this is necessary - Signals should not fire in migrations.

ansible_base/resource_registry/apps.py
```diff
@@ -52,25 +54,34 @@ def initialize_resources(sender, **kwargs):
...
             r_type.migrated = True
             r_type.save()
 
 
+def connect_signals(*args, **kwargs):
+    from django.db.models.signals import post_delete, post_save
+
+    from ansible_base.resource_registry.signals import handlers
+
+    post_save.connect(handlers.update_resource)
```
AlanCoding, line 75: What model does this signal connect to? Expectation is that the post_save signal fires whenever a registered resource is saved.

AlanCoding: Oh dear. Can we please have this do:

```python
for cls in get_resource_models():
    post_save.connect(handlers.update_resource, sender=cls)
```

ansible_base/resource_registry/signals/handlers.py
```diff
@@ -18,10 +15,8 @@ def get_resource_models():
     return resource_models
 
 
-@receiver(post_delete)
+# @receiver(post_delete)
```
AlanCoding: Delete comment. Moving the connection to the app's `.ready` call is the right thing to do.

test_app/migrations/0002_set_up_resources_test_data.py
```diff
@@ -0,0 +1,43 @@
...
+        2. That the post migration signal to initialize resource instances from
+           existing objects works correctly.
+    """
+    ResourceMigrationTestModel = apps.get_model("test_app", "ResourceMigrationTestModel")
+    Resource = apps.get_model("dab_resource_registry", "Resource")
+
+    ResourceMigrationTestModel.objects.create(name="migration resource")
+    r2 = ResourceMigrationTestModel.objects.create(name="migration resource 2")
+
+    assert Resource.objects.all().count() == 0
```
AlanCoding, line 21: You could do this with an existing model like Organization. Having assertions here is otherwise fine by me.

ansible_base/resource_registry/apps.py
```diff
@@ -31,6 +31,8 @@ def initialize_resources(sender, **kwargs):
     if apps is None:
         from django.apps import apps
 
+    from ansible_base.resource_registry.models import init_resource_from_object
+
     Resource = apps.get_model("dab_resource_registry", "Resource")
```
AlanCoding, line 36: Just thoughts here:

You should be free to import `Resource` from models when this is called. What you _actually need_ is a heuristic to know if the migration is complete. Practically you only need this for `dab_resource_registry`. That's why moving out of the `@classmethod` is not hitting directly at the issue you're trying to fix.

this is code from AWX which does that

https://github.com/ansible/awx/blob/devel/awx/main/middleware.py#L203-L205

I wish there was something more direct, like a parameter from the signal.

ansible_base/resource_registry/apps.py
newswangerd: beats me. The sender needs to be the model, so i figured that leaving this null would set the sender to the model (which seems to be the case). I can try setting it to self and see what happens

test_app/migrations/0002_set_up_resources_test_data.py
newswangerd: I did this with orgs first, but it broke other tests that assume they are getting a blank slate, so I figured that creating a new model specifically for this test was the best course of action.

ansible_base/resource_registry/apps.py
newswangerd: Interesting, I'll look into this. I tried importing resource directly, but it complained that ContentType was invalid.

AlanCoding: I've worried about DAB post_migrate signals having ordering problems relative to permissions / contenttypes, but so far I have never seen that happen. The resource registry does list contenttypes migration as a dependency so it's not an ordering problem on that layer.

AlanCoding (approved):
We should get this moving, but delete the commented code first.

PR #120, opened 2024-02-02
Add ViewWithHeaders which add custom headers

john-westcott-iv:
This adds a base View called `ViewWithHeaders` that adds additional headers to responses. It also make all DAB django views use a specific parent view whose parent can be specified by a service.

### DIFF
--- a/ansible_base/lib/utils/views.py
+++ b/ansible_base/lib/utils/views.py
@@ -1,21 +1,84 @@
 import logging
+import time
 
 from django.conf import settings
+from django.db import connection
+from django.utils.translation import gettext_lazy as _
 from rest_framework import views
+from rest_framework.exceptions import AuthenticationFailed, ParseError, PermissionDenied, UnsupportedMediaType
+
+from ansible_base.lib.utils.settings import get_function_from_setting
 
 logger = logging.getLogger('ansible_base.lib.utils.views')
 
 
-setting = 'ANSIBLE_BASE_EXTRA_HEADERS'
+class AnsibleBaseView(views.APIView):
+    def initialize_request(self, request, *args, **kwargs):
+        """
+        Store the Django REST Framework Request object as an attribute on the
+        normal Django request, store time the request started.
+        """
+        self.time_started = time.time()
+        if getattr(settings, 'SQL_DEBUG', False):
+            self.queries_before = len(connection.queries)
 
+        # If there are any custom headers in REMOTE_HOST_HEADERS, make sure
+        # they respect the allowed proxy list
+        if all(
+            [
+                settings.PROXY_IP_ALLOWED_LIST,
+                request.environ.get('REMOTE_ADDR') not in settings.PROXY_IP_ALLOWED_LIST,
+                request.environ.get('REMOTE_HOST') not in settings.PROXY_IP_ALLOWED_LIST,
+            ]
+        ):
+            for custom_header in settings.REMOTE_HOST_HEADERS:
+                if custom_header.startswith('HTTP_'):
+                    request.environ.pop(custom_header, None)
 
-class ViewWithHeaders(views.APIView):
+        drf_request = super().initialize_request(request, *args, **kwargs)
+        request.drf_request = drf_request
+        try:
+            request.drf_request_user = getattr(drf_request, 'user', False)
+        except AuthenticationFailed:
+            request.drf_request_user = None
+        except (PermissionDenied, ParseError) as exc:
+            request.drf_request_user = None
+            self.__init_request_error__ = exc
+        except UnsupportedMediaType as exc:
+            exc.detail = _(
+                'You did not use correct Content-Type in your HTTP request. If you are using our REST API, the Content-Type must be application/json'
+            )
+            self.__init_request_error__ = exc
+        return drf_request
+
     def finalize_response(self, request, response, *args, **kwargs):
         response = super().finalize_response(request, response, *args, **kwargs)
-        for name, value in getattr(settings, setting, {}).items():
-            if isinstance(name, str) and isinstance(value, str):
-                response[name] = value
-            else:
-                logger.error(f"When adding a custom header expected (str,str) but got ({type(name)}, {type(value)}) please check your {setting} settings")
+
+        if request.user.is_authenticated:
+            version = 'Unknown'
+            try:
+                the_function = get_function_from_setting('ANSIBLE_BASE_PRODUCT_VERSION_FUNCTION')
+                version = the_function()
+            except Exception:
+                logger.exception(_('ANSIBLE_BASE_PRODUCT_VERSION_FUNCTION was set but calling it as a function failed (see exception).'))
+
+            response['X-API-Product-Version'] = version
+
+        response['X-API-Product-Name'] = getattr(settings, 'ANSIBLE_BASE_PRODUCT_NAME', _('Unnamed'))
+        response['X-API-Node'] = getattr(settings, 'CLUSTER_HOST_ID', _('Unknown'))
+
+        time_started = getattr(self, 'time_started', None)
+        if time_started:
+            time_elapsed = time.time() - self.time_started
+            response['X-API-Time'] = '%0.3fs' % time_elapsed
+
+        if getattr(settings, 'SQL_DEBUG', False):
+            queries_before = getattr(self, 'queries_before', 0)
+            q_times = [float(q['time']) for q in connection.queries[queries_before:]]
+            response['X-API-Query-Count'] = len(q_times)
+            response['X-API-Query-Time'] = '%0.3fs' % sum(q_times)
+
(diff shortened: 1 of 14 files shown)


ansible_base/lib/utils/views.py
```diff
@@ -0,0 +1,84 @@
...
+        if getattr(settings, 'SQL_DEBUG', False):
+            queries_before = getattr(self, 'queries_before', 0)
+            q_times = [float(q['time']) for q in connection.queries[queries_before:]]
+            response['X-API-Query-Count'] = len(q_times)
+            response['X-API-Query-Time'] = '%0.3fs' % sum(q_times)
+
+        if getattr(self, 'deprecated', False):
+            response['Warning'] = _('This resource has been deprecated and will be removed in a future release.')
+
+        return response
```
AlanCoding: Thanks, I'm very strongly aligned with this general approach. You just need to replace references to `ViewWithHeaders`.

Eventually, I want us to re-parent all AWX views with this.

ansible_base/lib/utils/views.py
```diff
@@ -0,0 +1,84 @@
...
+        time_started = getattr(self, 'time_started', None)
+        if time_started:
+            time_elapsed = time.time() - self.time_started
+            response['X-API-Time'] = '%0.3fs' % time_elapsed
+
+        if getattr(settings, 'SQL_DEBUG', False):
+            queries_before = getattr(self, 'queries_before', 0)
+            q_times = [float(q['time']) for q in connection.queries[queries_before:]]
+            response['X-API-Query-Count'] = len(q_times)
+            response['X-API-Query-Time'] = '%0.3fs' % sum(q_times)
```
AlanCoding: I'm skeptical that we should continue these headers about query count, because there is so much other tooling that does similar things. It might be better to axe it just to reduce the amount of future commitments.

ansible_base/lib/utils/views.py
```diff
@@ -0,0 +1,84 @@
...
+        if getattr(settings, 'SQL_DEBUG', False):
+            self.queries_before = len(connection.queries)
+
+        # If there are any custom headers in REMOTE_HOST_HEADERS, make sure
+        # they respect the allowed proxy list
+        if all(
+            [
+                settings.PROXY_IP_ALLOWED_LIST,
+                request.environ.get('REMOTE_ADDR') not in settings.PROXY_IP_ALLOWED_LIST,
+                request.environ.get('REMOTE_HOST') not in settings.PROXY_IP_ALLOWED_LIST,
```
AlanCoding: This isn't general, this is AWX-specific, and because of that we probably need some more eyes on this.

ansible_base/lib/utils/views.py
```diff
@@ -0,0 +1,84 @@
...
+            ]
+        ):
+            for custom_header in settings.REMOTE_HOST_HEADERS:
+                if custom_header.startswith('HTTP_'):
+                    request.environ.pop(custom_header, None)
+
+        drf_request = super().initialize_request(request, *args, **kwargs)
+        request.drf_request = drf_request
+        try:
+            request.drf_request_user = getattr(drf_request, 'user', False)
```
AlanCoding: I this this exists to pass data to a method `get_current_user_from_drf_request` which is integrated deeply with the CRUM library. The functionality is working without it so hopefully we don't need it.

docs/lib/views.md
```diff
@@ -1,5 +1,50 @@
...
+
+class MyServiceModelViewSet(ModelViewSet, MyServiceBaseApiView):
+    pass
+```
+
+## Changing the parent view
+
+All views in django-ansible-base actually inherit from a class called `ansible_base.lib.utils.views.AnsibleBaseDjanoAppApiView` which inherits, by default from `AnsibleBaseView`.
+
+However, if you already have an existing parent view with additional features you can override the view that `AnsibleBaseDjangoAppApiView` inherits from by setting the django setting `ANSIBLE_BASE_CUSTOM_VIEW_PARENT`.
```
AlanCoding, line 22: Yeah, I think AWX should use this. But when we do that, our base should inherit from `AnsibleBaseDjangoAppApiView`, just to add extra stuff.

ansible_base/lib/utils/views.py
```diff
@@ -0,0 +1,105 @@
...
+            exc.detail = _(
+                'You did not use correct Content-Type in your HTTP request. If you are using our REST API, the Content-Type must be application/json'
+            )
+            self.__init_request_error__ = exc
+        return drf_request
+
+    def finalize_response(self, request, response, *args, **kwargs):
+        response = super().finalize_response(request, response, *args, **kwargs)
+
+        if request.user.is_authenticated:
```
AlanCoding: I saw the test failures about the request object not having a user, and thought it might be some fatal flaw with the crum library. But no, seems it's just because the mock_request fixture doesn't have a user.

ansible_base/lib/utils/views.py
```diff
@@ -0,0 +1,105 @@
...
+            version = 'Unknown'
+            try:
+                the_function = get_function_from_setting('ANSIBLE_BASE_PRODUCT_VERSION_FUNCTION')
+                version = the_function()
+            except Exception:
+                logger.exception(_('ANSIBLE_BASE_PRODUCT_VERSION_FUNCTION was set but calling it as a function failed (see exception).'))
```
AlanCoding: This message isn't always right. You didn't give `ANSIBLE_BASE_PRODUCT_VERSION_FUNCTION` a default value. In the case that it's not set, I recon that it's best to do nothing. Someone using this library might not want that header.

ansible_base/lib/utils/views.py
john-westcott-iv: Yea, the tests need a full rewrite based on the changes I've made. Working on that currently.

ansible_base/lib/utils/views.py
john-westcott-iv: Although I feel like the X headers are a quick way to gage something I agree that other things do this job so removing.

ansible_base/lib/utils/views.py
john-westcott-iv: Removed.

ansible_base/lib/utils/views.py
john-westcott-iv: Good callout. Fixed so that this message should only appear if the setting was set.

ansible_base/lib/utils/views.py
```diff
@@ -0,0 +1,82 @@
...
+
+# Determine and load the parent view
+# If anything fails in here its a pretty low level exception so we don't catch anything and just let it raise
+parent_view = getattr(settings, 'ANSIBLE_BASE_CUSTOM_VIEW_PARENT', None)
+parent_view_class = AnsibleBaseView
+if parent_view:
+    try:
+        module_name, junk, class_name = parent_view.rpartition('.')
+        logger.debug(f"Trying to import parent view {class_name} from package {module_name}")
+        module = importlib.import_module(module_name, package=class_name)
```
AlanCoding: This isn't circular?

Say I set `ANSIBLE_BASE_CUSTOM_VIEW_PARENT`, and that should be a class that inherits from `AnsibleBaseView`. Python imports my class, that imports this module, now on line 73 it imports my class again.

If that is a problem, or just to make it less circular, I could see it making sense to put ln 65 and after in another module. That way, when it imports my class, my class imports `AnsibleBaseView` and fully completes the import of the entire module its in and can return control here without additional smarts.

ansible_base/lib/utils/views.py
john-westcott-iv: Removed for now. We can think about pulling it back in later if needed.

ansible_base/lib/utils/views.py
john-westcott-iv: You'd think it would be but we can setup a quick test does not show a circular import. Lets meet up tomorrow to discuss. The brain is too tired tonight :)

john-westcott-iv: Split into seperate files.

ansible_base/lib/utils/views/ansible_base.py
```diff
@@ -0,0 +1,53 @@
...
+
+            if the_function:
+                try:
+                    version = the_function()
+                except Exception:
+                    logger.exception(_('{setting} was set but calling it as a function failed (see exception).'.format(setting=setting)))
+
+            response['X-API-Product-Version'] = version
+
+        response['X-API-Product-Name'] = get_setting('ANSIBLE_BASE_PRODUCT_NAME', _('Unnamed'))
```
AlanCoding, line 42: After this merges I need to think about the setting name being introduced in https://github.com/ansible/django-ansible-base/pull/127

AlanCoding (approved):
Great, I'd love to see this integrated into AWX. After changes like this (or my change about the router_basename for models) it might be good to file issues to track the follow-on downstream changes expected.

PR #121, opened 2024-02-03
Standard way to link to model with multi-word name

AlanCoding:
This is a counterpart to https://github.com/ansible/django-ansible-base/pull/116 and needs to be combined with that. EDIT: done

Trying to develop some new views on my own, I hit this. There are conflicting expectations about how the view names will be named with a model name has more than 1 word, and whether you use "-" "_" or nothing is actually hard.

Yeah this needs a unit test, but here it is done in the API

```json
{
    "id": 2,
    "url": "/api/v1/teams/2/",
    "related": {
        "created_by": "/api/v1/users/2/",
        "modified_by": "/api/v1/users/1/",
        "organization": "/api/v1/organizations/1/",
        "encryptioner": "/api/v1/encrypted_models/1/"
    },
    "summary_fields": {
        "created_by": {
            "id": 2,
            "username": "angry_spud",
            "first_name": "",
            "last_name": ""
        },
        "modified_by": {
            "id": 1,
            "username": "admin",
            "first_name": "",
            "last_name": ""
        },
        "organization": {
            "id": 1,
            "name": "AWX_community"
        },
        "encryptioner": {
            "id": 1,
            "name": "foo"
        }
    },
    "created_on": "2024-02-03T03:13:22.225324Z",
    "modified_on": "2024-02-03T03:13:59.036461Z",
    "name": "awx_devs",
    "description": "",
    "created_by": 2,
    "modified_by": 1,
    "organization": 1,
    "scrum_master": null,
    "encryptioner": 1
}
```

The change allows it to have the "encryptioner" in the related & summary_fields, and I legitimately couldn't figure out a way to do that before.

We may need to apply this solution to authenticator_maps as well... I remember seeing the warning that it couldn't reverse the view for that.


ansible_base/lib/abstract_models/common.py
```diff
@@ -14,6 +15,27 @@
...
 
+def get_cls_view_basename(cls):
+    # This gives the expected base Django view name of cls
+    if hasattr(cls, 'router_basename'):
+        return cls.router_basename
+    return underscore(cls.__name__)
+
+
+def get_url_for_object(obj, request=None):
+    # get_absolute_url mainly exists to support AWX
```
relrod, line 26: This feels a little odd to me, having project-specific exceptions in DAB.

I know AWX doesn't use DAB `CommonModel` right now, but I'm wondering if maybe in the future AWX does something like having an `AWXCommonModel` that extends `CommonModel` and overrides this method, to account for the AWX specific bits, rather than doing it here. 🤔

docs/lib/default_models.md
```diff
@@ -12,3 +12,82 @@ Models include:
...
 If you are using either of these models as a base class you can use their corresponding default serializers as well:
 `ansible_base.lib.serializers.common.CommonModelSerializer` or `ansible_base.lib.serializers.common.NamedCommonModelSerializer`
+
+## Managing Related Links
+
+The common serializers expect certain things from your models in order to reduce code
+needed to populate standard patterns in the returned API data. In particular:
+
+ - "url" - the URL of the object itself
+ - "related" - linked to related objects
```
relrod:
```suggestion
 - "related" - links to related objects
```

docs/lib/default_models.md
```diff
@@ -12,3 +12,82 @@ Models include:
...
+the "url" field is use to put an entry in "related". The key in the related
+dictionary is the field name.
+
+### Many-to-Many
+
+It is expected that models have a sublist for many-to-many relationships.
+For example, `/api/v2/organizations/1/admins/` would list users who are admins
+of the organization.
+
+If this doesn't apply, and it is not intended to have a view for a realationship
```
relrod:
```suggestion
If this doesn't apply, and it is not intended to have a view for a relationship
```

ansible_base/lib/abstract_models/common.py
```diff
@@ -124,31 +146,36 @@ def get_summary_fields(self):
...
-                else:
-                    reverse_view = f"{underscore(field.related_model.__name__)}-detail"
-                    pk = getattr(self, field.name).pk
-                try:
-                    response[field.name] = reverse(reverse_view, kwargs={'pk': pk})
-                except NoReverseMatch:
-                    logger.debug(f"Model {self.__class__.__name__} wanted to reverse view to {reverse_view} but said view is not defined")
+            if obj := getattr(self, field.name):
+                if related_url := get_url_for_object(obj):
+                    response[field.name] = related_url
```
relrod, line 161: This is clean, I like this. :+1:

ansible_base/lib/abstract_models/common.py
AlanCoding: This isn't for linking AWX serializers to DAB models. This is for linking DAB serializers to AWX models.

So for RBAC, we have these roles, and users given a role for an object. I'll call that an assignment, a user is assigned a role, for an object. The role-assignment needs to link the object it applies for. The assignment is a DAB model, and the object is an AWX model. That object has `get_absolute_url` method on it. Otherwise we _have no way_ of identifying or reasonably guessing what the view name is.

I'm all ears for if you have another solution. But DAB models are expected to link to local models over and over again if you look at what the current features are.

It wouldn't totally surprise me if we need another solution for EDA, and then for galaxy_ng. But that's why we want to mark these as temporary and non-standard. Converting those other apps to standard DAB patterns is a much bigger thing.

AlanCoding:
Now I've become familiar enough with this to quote specific logs that demonstrate the problem.

```
DEBUG    ansible_base.lib.abstract_models.common:common.py:146 Model RoleDefinition wanted to reverse view to role_definition-permissions-list but said view is not defined
```

In this case I have a model that's RoleDefinition. It tries to reverse a many-to-many relation, to permissions. Ignore the fact that this view doesn't exist. The router `basename=` I'm using is "roledefinition", and this is the best value to use. Yeah, I could set the serializer `reverse_url_name` to "roledefinition-detail" to get the url self-link to work, but then the permissions related link would still expect underscores, and these come from the same router basename.

relrod (approved):

relrod:
I think this is the way. And I think it's *amazing* to have documentation for this now.

PR #122, opened 2024-02-05
Fix exception under specific circumstances in ui_auth

john-westcott-iv:
While working with authenticators and testing I found an exception when loading ui_auth if the user has an ebaled authenticator but the authenticator class is unable to be loaded.


slemrmartin (approved):
:+1:

PR #125, opened 2024-02-07
Alter detection of user on common model save

john-westcott-iv:
Currently the code is looking for a user of None to decide that it should put the `_system` user into the created/modified by fields. However, AnonymousUser is not None. We see AnonymousUser as a new user is logging in from an Authenticator on the initial save of a new User object. This modification allows None and AnonymousUser to both trigger the save of an item as `_system` and adds better handeling in the actual save function if somehow we still got a user of `None`.


ansible_base/lib/abstract_models/common.py
```diff
@@ -72,22 +72,34 @@ def _attributable_user(self, warn_nonexistent_system_user):
...
+        user = self._attributable_user(warn_nonexistent_system_user)
+        # Its possible for _attributable_user to return None in which case we can't save the created/modified by fields so warn
+        if not user:
+            logger.error("Unable to set create/modified by attributes on {self.__class__}/{self.pk} because user was not set. Save will proceed")
+        else:
+            if 'created_by' not in update_fields or not self.created_by:
+                self.created_by = user
                 update_fields.append('created_by')
+                if not self.pk:
+                    logger.warning(f"Existing model {self.__class__}/{self.pk} needed created_on updated!")
```
AlanCoding: > needed created_on updated

Copypasta. "created_by" updated.

Also, what if object was created by the system, the system user was not set, and then later a user modifies the user in the API? I don't think `created_by` should be filled in for that case. It should be left as `None`.

Perhaps if the object is saved with `obj.save(update_fields=['created_by'])` then I can agree, but that's a different criteria from this.

ansible_base/lib/abstract_models/common.py
```diff
@@ -57,7 +57,7 @@ class Meta:
 
     def _attributable_user(self, warn_nonexistent_system_user):
         user = get_current_user()
-        if user is None:
+        if user is None or user.is_anonymous:
```
AlanCoding: Isn't it very concerning if an object is modified by an anonymous user? Wouldn't it be good to log that?

ansible_base/lib/abstract_models/common.py
slemrmartin: - "created_by" instead of "created_on"
- it's in condition `not self.pk`, so it can't be used in log message

ansible_base/lib/abstract_models/common.py
```diff
@@ -72,22 +72,34 @@ def _attributable_user(self, warn_nonexistent_system_user):
...
+
+            if 'modified_by' not in update_fields or not self.modified_by:
+                self.modified_by = user
+                update_fields.append('modified_by')
+
+        if 'created_on' not in update_fields or not self.created_on:
+            self.created_on = now
+            update_fields.append('created_on')
+            if not self.pk:
+                logger.warning(f"Existing model {self.__class__}/{self.pk} needed created_on updated!")
```
slemrmartin: the same: `self.pk` is empty here

ansible_base/lib/abstract_models/common.py
```diff
@@ -72,22 +72,34 @@ def _attributable_user(self, warn_nonexistent_system_user):
...
+                self.created_by = user
                 update_fields.append('created_by')
+                if not self.pk:
+                    logger.warning(f"Existing model {self.__class__}/{self.pk} needed created_on updated!")
+
+            if 'modified_by' not in update_fields or not self.modified_by:
+                self.modified_by = user
+                update_fields.append('modified_by')
+
+        if 'created_on' not in update_fields or not self.created_on:
```
slemrmartin: should it be allowed to change "created_on"?

AlanCoding: I see an argument that some system-level logic may need to update `created_by` after the fact. Imagine it has to create an object before it knows the attributable user.

But that argument doesn't apply to `created_on`, because there's no valid reason you wouldn't be able to get a timestamp when creating the object.

ansible_base/lib/abstract_models/common.py
john-westcott-iv: It also dosen't help that I didn't save changes in vscode before pushing :(

ansible_base/lib/abstract_models/common.py
john-westcott-iv: Yea, this is the debate because we have one spot where we know this happens. When a user is logging in from an authenticator they are AnonymousUser when it tries to save the new User for them. I am going to push a change where I try to better catch just that exception and error otherwise.

ansible_base/lib/abstract_models/common.py
john-westcott-iv: Yea, this should be updated now.

ansible_base/lib/abstract_models/common.py
```diff
@@ -66,27 +81,46 @@ def _attributable_user(self, warn_nonexistent_system_user):
                     user = get_user_model().objects.get(username=system_username)
                 except get_user_model().DoesNotExist:
                     if warn_nonexistent_system_user:
-                        logger.warn(f"SYSTEM_USERNAME is set to {system_username} but no user with that username exists. User attribution will be None.")
+                        logger.warn(f"SYSTEM_USERNAME is set to {system_username} but no user with that username exists.")
```
slemrmartin: I think `warn` is deprecated, `warning` should be used

ansible_base/lib/abstract_models/common.py
```diff
@@ -66,27 +81,46 @@ def _attributable_user(self, warn_nonexistent_system_user):
...
 
         # Manually perform auto_now_add and auto_now logic.
         now = timezone.now()
-        if not self.pk and not self.created_on:
-            self.created_on = now
+
+        user = self._attributable_user(warn_nonexistent_system_user)
+
+        if self.created_by is None:
+            self._check_user(user, warn_nonexistent_system_user)
```
slemrmartin: what is the change related to state before? The variable "warn" doesn't cause warning (which is quite confusing), but raises exception. 
And is True by default - is there some `except ValueError` I don't see which makes the difference?

john-westcott-iv: There is no except. Previously it would try to save if the user got into the `save` function as `None` or `AnonymousUser`. This was the root of the bug we are fixing, `None` is technically ok but if `AnonymousUser` got here we would get an exception because `AnonymousUser` has no ID. We now have a framework for converting `AnonymousUser` to the `SYSTEM_USER` for specific valid cases. So, in general, for cases where we get `None` or `Anonymous` we raise an exception and prevent a save because we expect legtimite users to be managing the objects in the system. However, if you had to do something programatic and the system user was not defined (like when we create the system user initially in a migration) `None` user may be legitimate for that use case so we provide the `warn` variable to get around the exception and allow the save (with `None`, `AnonymousUser` will still fail with an exception). Note, you cant do:
```
system_user = User.objects.create(username=settings.SYSTEM_USER)
```

Because you can't currently pass the variable through the create function. Instead of you have to be explicit (like we do in the tests):
```
system_user = User()
system_user.username = settings.SYSTEM_USER
system_user.save(warn_nonexistent_system_user=False)
```

This will bypass the logic for raising the exception if we get `None` or `AnonymousUser` and allow a save with `None`. As mentioned you would still get an exception for `AnonymousUser` so if we run into another condition where `AnonymousUser` is legitimate for a save than we need to change the logic which sets `convert_anonymous_user_to_system` above.

john-westcott-iv: Good call on the name. Before it just warned but now it actually raises so I'll update the name.

ansible_base/lib/abstract_models/common.py
john-westcott-iv: Yea, this was actually biting me in the tests yesterday because I was trying to test a `warning` log but using the logic that was used to test this `warn`. I'll update this reference.

ansible_base/lib/abstract_models/common.py
slemrmartin: thanks for explanation, it's clear to me now

ansible_base/lib/abstract_models/common.py
```diff
@@ -77,40 +78,70 @@ class Meta:
...
-        user = self._attributable_user(warn_nonexistent_system_user)
 
         # Manually perform auto_now_add and auto_now logic.
         now = timezone.now()
-        if not self.pk and not self.created_on:
+        user = self._attributable_user(non_existent_user_fatal)
+
+        if self.created_by is None:
+            self._check_user(user, non_existent_user_fatal)
+            self.created_by = user
```
AlanCoding: I still think this should be left as None if it's not a new object.

john-westcott-iv: Fixed. We will only set created_on/created_by if there is no pk on the object.

slemrmartin (approved):

ansible_base/lib/abstract_models/common.py
```diff
@@ -74,40 +75,69 @@ class Meta:
...
+        if 'modified_on' not in update_fields:
             self.modified_on = now
-            if user is not None and not user.is_anonymous:
-                self.modified_by = user
             update_fields.append('modified_on')
+
+        if 'modified_by' not in update_fields:
+            self._check_user(user, non_existent_user_fatal)
+            self.modified_by = user
             update_fields.append('modified_by')
```
AlanCoding, line 147: Might be good to document in a comment or method docstring that adding a field like "modified_by" to update_fields will be taken as a signal that the saver is providing the value. This is quite unlike the field `auto_now` type functions of datetime fields.

AlanCoding (approved):

ansible_base/lib/abstract_models/common.py
john-westcott-iv: Docstring added.

PR #127, opened 2024-02-07
Move the local service name into settings

AlanCoding:
The RBAC app will also need the service name setting, and the `get_registry` method will wind up importing more things as a side effect, and by design, isn't generally available on import.

### DIFF
--- a/ansible_base/lib/dynamic_config/dynamic_settings.py
+++ b/ansible_base/lib/dynamic_config/dynamic_settings.py
@@ -26,6 +26,12 @@ except NameError:
     REST_FRAMEWORK = {}
 
 
+try:
+    ANSIBLE_BASE_SERVICE_PREFIX
+except NameError:
+    ANSIBLE_BASE_SERVICE_PREFIX = 'local'
+
+
 if 'ansible_base.api_documentation' in INSTALLED_APPS:
     if 'drf_spectacular' not in INSTALLED_APPS:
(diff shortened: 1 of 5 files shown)


ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -26,6 +26,12 @@
     REST_FRAMEWORK = {}
 
 
+try:
+    ANSIBLE_BASE_SERVICE_PREFIX
+except NameError:
+    ANSIBLE_BASE_SERVICE_PREFIX = 'local'
```
john-westcott-iv, line 32: Would it be an issue if there were "multiple" locals? i.e. should this be a UUID or some other unique string?

AlanCoding: Anything using resource registry or centralized RBAC should set this. This will prefix the type like "local.inventory", and I hate the idea of a UUID there. My hot take here is that "local" works for something that is absolutely not going to be doing any clustering with another system via DAB stuff.

john-westcott-iv (approved):

AlanCoding:
I've updated my RBAC branch to address the issue this was addressing. It's functional with the current resource registry.

I believe the import circularity was a temporary problem, which I solved by another method, and this method is no longer needed. Definitely time to abandon this.

PR #129, opened 2024-02-08
All github auth plugins

jctanner:
No-Issue


john-westcott-iv (commented):
One of the goals we have for authenticators is splitting them out into their own repos. I know you mentioned sonar wanted you put the configs in a separate file to reduce duplication but this would be counter productive to our plan. 

@shanemcd any thoughts on this?

A couple options I can think of:

We could duplicate the configs in each file and just tell sonar it’s fine 

We could move the configs in the authenticator_plugin files like they are and just import from the base GitHub authenticator in the others that depend on it. That would allow us to keep these authenticators in a single file and we can just have a python dependency on the main GitHub authenticator from the “child” ones. 

We could have “another” python dependence that all GitHub authenticators depend on. But that feels excessive.


Any other ideas?


I don’t think this decision needs to hold up this PR as it can be handled in a follow up, when we split out the authenticators.

john-westcott-iv:
Can we add some simple conversion from the awx adapters to GitHub authenticators in the tools/scripts/comvert_awx_adaper.py script? One thing to think about with this is that I think there are two places in awx that the social maps can exist. Idk if one overrides the other or if they are additive; we will have to review the awx code.

john-westcott-iv (approved):

ansible_base/authentication/authenticator_configurators/github.py
```diff
@@ -0,0 +1,109 @@
+from django.utils.translation import gettext_lazy as _
+
+from ansible_base.authentication.authenticator_plugins.base import BaseAuthenticatorConfiguration
+from ansible_base.lib.serializers.fields import CharField, ListField, URLField
+
+
+class GithubConfiguration(BaseAuthenticatorConfiguration):
+
+    documenation_url = "https://python-social-auth.readthedocs.io/en/latest/backends/github.html"
```
rochacbruno:
```diff
-	documenation_url = "https://python-social-auth.readthedocs.io/en/latest/backends/github.html"
+   documentation_url = "https://python-social-auth.readthedocs.io/en/latest/backends/github.html"
```

PR #130, opened 2024-02-09
Declare missing migration dependency

AlanCoding:
Above in this file it does `Resource = apps.get_model("dab_resource_registry", "Resource")`. This ordering doesn't normally happen, but if the initial "dab_resource_registry" first migration hasn't happened yet, then this line will error because that model in that app doesn't exist in that apps state.


john-westcott-iv (approved):
Cool hack,TIL there was `__first__` and `__latest__` that could be referenced like this.

PR #131, opened 2024-02-09
Add AssociationRouter

john-westcott-iv:


test_app/views.py
```diff
@@ -1,30 +1,35 @@
...
-class TestAppViewSet(ModelViewSet):
+class TestAppViewSet(ModelViewSet, AnsibleBaseView):
     permission_classes = [permissions.IsAuthenticated]
 
     def get_queryset(self):
         return self.serializer_class.Meta.model.objects.all()
 
 
 class OrganizationViewSet(TestAppViewSet):
+    queryset = Organization.objects.all()
```
AlanCoding, line 17: this might be a rebase conflict, but you don't need a `queryset` because the parent viewset defines `get_queryset`.

docs/lib/routers.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Breaking down related views
+
+Related views are expressed in the following format:
+
+```
+    '<entry in API related field>': (<ViewSet for relation>, '<related field name>')
+```
+
+NOTE: Often times the `<entry in API related field>` will be the same as `<related field name>` but this is not always the case.
```
AlanCoding, line 45: I thought this wasn't the plan. Here I think you're talking about building the URLs, right? Like `/api/v2/organizations/2/users/`, where "entry in API related field" is "users" here. But that's not the full story. If you view the organization detail view at `/api/v2/organizations/2/` then it should have

```python
"related": {
  "users": "/api/v2/organizations/2/users/"
}
```

We construct this related dictionary inside an instance method for a common abstract _model_. If you were to register a different related name like `/api/v2/organizations/2/member_users/`, then we may be able to coherently link to it, but it would be very difficult to change the key in the "related" dictionary to match. I'm also concerned about the mechanic of the internal view name.

For a complex example, consider "notification_templates_success". The related name _is_ non-trivial. But the link name, field name, and view suffix all seem consistent

https://github.com/ansible/awx/blob/519fd22bece0e33fa17e5aa6591fc28cf48f565b/awx/main/models/base.py#L381

https://github.com/ansible/awx/blob/519fd22bece0e33fa17e5aa6591fc28cf48f565b/awx/api/urls/project.py#L41

https://github.com/ansible/awx/blob/519fd22bece0e33fa17e5aa6591fc28cf48f565b/awx/api/serializers.py#L1342

So those are 3 things that all have correspondence. The view name (the 4th naming) is "project_notification_templates_success_list", so it breaks down at that point.

My view is that we should start out declaring that all 4 of these things must have correspondence. Ping @relrod  If there is a need for something to break this correspondence, I would really like to press the issue and ask why.

docs/lib/routers.md
```diff
@@ -0,0 +1,69 @@
...
+
+
+If a model does not specify a related field name one will automatically be created by django in the format `<model._meta.model_name>_set`.
+See: https://docs.djangoproject.com/en/dev/topics/db/queries/#backwards-related-objects
+
+For example, in the AuthenticatorMap model we define a foreign key to authenticator like:
```
AlanCoding, line 73: I want to put a pause here. Consider examples up until this point - members of a team, members of an organization. Those are many-to-many. Historically, AWX took relationships like authenticator->authenticator_maps and munged them together with the tooling for many-to-many views. This was a mistake I have tried to correct.

My belief is that a read-only view suits these reverse relationships better. This would require creating a new router factory and having a separate registration mechanism. Whereas m2m is specified as `related_views={`, I would like one-to-many (reverse ForeignKeys) to specify these separately like `reverse_views={`. The reason I want this read-only is because the right way to move an AuthenticatorMap to a new Authenticator is to PATCH to the authenticator_map endpoint `{"authenticator": 2}`, for example. Introducing a new API mechanism introduces new surface area for access control checks, in particular. It has also led to the ability to bypass validation, like moving a host from one inventory to another in AWX (and keeping the same groups, which is deeply wrong).

test_app/router.py
```diff
@@ -12,3 +13,14 @@
...
+associative_router = AssociationResourceRouter()
+associative_router.register(
+    r'related_model',
+    views.RelatedFieldsTestModelViewSet,
+    related_views={
+        'teams': (views.TeamViewSet, 'more_teams'),
+        'user': (views.UserViewSet, 'users'),
+    },
+    basename='related_model',
+)
```
AlanCoding, line 31: This doesn't appear to follow the guidance from the docs. Why is test_app managing two different routers? Why can't we simply replace `router = SimpleRouter()` with the new association resource router?

Also, we already registered `RelatedFieldsTestModelViewSet` at a different endpoint. The canonical way to do this would be to just add `related_views` to that registration.

docs/lib/routers.md
relrod: Internally the only thing `<entry in API related field>` is used for at this point is for constructing the URL. So `'foo': (FooViewSet, 'some_m2m')` would stick it under `.../foo/...`.

It's badly named "related_name" in the code, but it's not really that. It's "URL component".

The view names are (now, as of the work @AlanCoding did in #121 and related PRs) always generated using the field name - the second component of the 2-tuple.

> If you were to register a different related name like `/api/v2/organizations/2/member_users/`, then we may be able to coherently link to it, but it would be very difficult to change the key in the "related" dictionary to match.

Correct, the key under the related dict in the API response is generated by the field name, too.

I've previously thought about letting the router expose the data passed in, so that in `CommonModel#related_fields` we could pull out that data. And that's feasible (you'd need some way to get at the router instance from the model, but it can likely be done) if we want to support this particular case, but I'm not sure how critical it is.

> My view is that we should start out declaring that all 4 of these things must have correspondence.

I think they are? The URL component can be customized as I said, but that doesn't matter much in terms of actually getting at the route by some predetermined name. You can change the key from `users` to `people` and everything would continue working except the URL would change.

Where do we deviate from the pattern?

ansible_base/lib/routers/associative_resource_router.py
```diff
@@ -0,0 +1,131 @@
...
+                if is_associate_viewset and action_str not in associate_actions:
+                    continue
+                bound_methods[method] = action_str
+        return bound_methods
+
+    def register(self, prefix, viewset, basename=None, related_views={}):
+        if basename is None:
+            basename = self.get_default_basename(viewset)
+
+        for related_name, (related_view, fk) in related_views.items():
```
relrod: In the interest of future-proofing, do we want to make the value here _not_ be a tuple, and instead be a dataclass or namedtuple?

For example, piggybacking off of @AlanCoding's discussion - if we ever want to be able to customize the name of the related field key, from the router, and expose that data, maybe we could store that here.

So basically instead of:

```python
related_views={
    'users': (UserViewSet, 'users'),
}
```

we could have

```python
related_views={
    'users': RouteAssocation(viewset=UserViewSet, field='users', related_name='people'),
}
```

and we'd be able to add optional kwargs later on if the need ever came up.

Or we could even make `related_views` be a list and store the url component as a kwarg too:

```python
related_views=[
    RouteAssocation(url_component='users', viewset=UserViewSet, field='users', related_name='people'),
]
```

I'm mostly trying to future proof "oh we need to support some case and we need extra data for that case, and we know that data really should belong with the router, but we have no way to inject that data at route registration time".

But if we're deciding we're going to completely standardize _everything_ and not allow for this kind of customization, maybe it's fine to keep it as-is.

AlanCoding: I'm strongly in favor of standardizing everything, unless we can prove the existence of a specific requirement that prevents that.

I expect the problem will not be with many-to-many fields, and I believe that standardization of those is acheivable.

It's the reverse ForeignKeys that will be the problem because of the "_set" auto-naming. Looking at AWX, consider instance group jobs, which is the `unifiedjob_set` relationship. The URL ends with `/jobs/`, the "related" entry is "jobs" and the reverse view name is "instance_group_unified_jobs_list". But this is not a many-to-many, so I don't want to address these cases in the current PR.

docs/lib/routers.md
john-westcott-iv: Ok, take a look at the updated code. `related_views` should now happen auto-magically and be read only.

test_app/router.py
john-westcott-iv: I left the original router in place because the comments (on some of the items) made it seem like there are already some tests against the router that I didn't want to break. We can reevaluate it.

test_app/views.py
john-westcott-iv: Removed.

ansible_base/lib/abstract_models/common.py
```diff
@@ -175,11 +172,22 @@ def related_fields(self, request):
...
+            try:
+                response[field_name] = reverse(reverse_view, kwargs={'pk': self.pk})
+            except NoReverseMatch:
+                logger.error(f"Wanted to add {reverse_view} for {self.__class__} but view was missing")
+
+        sorted_response = OrderedDict()
+        sorted_keys = list(response.keys())
+        sorted_keys.sort()
+        for key in sorted_keys:
+            sorted_response[key] = response[key]
```
AlanCoding, line 177: so... just going alphabetically?

ansible_base/lib/abstract_models/common.py
```diff
@@ -37,13 +37,10 @@ def get_url_for_object(obj, request=None):
 
 
 class CommonModel(models.Model):
-    # These are fields that should be reversed lookup as related fields.
-    # For example, an environment has related organizations so environment might specify reverse_foreign_key_fields = ['organizations']
-    # This would end up with a view like environment/1/organizations
-    reverse_foreign_key_fields = []
```
AlanCoding, line 43: Ok this is interesting. This is the kind of thing a changelog would be needed for. I take it that this needs a simultaneous merge in aap-gateway now, because it's pinned to devel, right?

ansible_base/lib/abstract_models/common.py
john-westcott-iv: Yea, this makes sense to my brain.

ansible_base/lib/abstract_models/common.py
```diff
@@ -37,13 +37,10 @@ def get_url_for_object(obj, request=None):
...
-
     # Any field marked as encrypted will automatically be stored in an encrypted fashion
     encrypted_fields = []
+    # Any field set in here will not be used in the views
+    ignore_relations = []
```
AlanCoding, line 43: This conflicts with the solution to the same problem

https://github.com/ansible/django-ansible-base/blob/devel/test_app/models.py#L42

```python
    teams_with_no_view = models.ManyToManyField(Team, related_name='related_fields_test_model_teams_with_no_view')
    teams_with_no_view.related_view = None
```

So that test_app model should be updated, and maybe it wasn't being tested correctly to begin with.

AlanCoding: And down below, just in this file, delete:

```python
        for field in self._meta.many_to_many:
            if getattr(field, 'related_view', False) is None:
                # Give fields a chance to opt out of showing up here by forcing related_view to None
                continue
```

I think your solution will work for both ForeignKey and ManyToMany, which the prior solution did not, so :+1: on that solution, just delete the old approach.

docs/lib/routers.md
```diff
@@ -0,0 +1,69 @@
+# Associative Resource Router
+
+django-ansible-base provides an `AssociationResourceRouter` which will auto-construct /associate and /disassociate endpoints for related ManyToMany fields for your models.
```
AlanCoding, line 3: Based on your other comments, do we need to update this statement to include reverse ForeignKey (many-to-one) relationships as well?

john-westcott-iv: Good callout, I need to update that file.

test_app/router.py
AlanCoding: sounds good, we can punt for later.

docs/lib/routers.md
AlanCoding: Had a look at the code, and it looks great, but wasn't exactly what I expected. I see that `reverse_views` is a separate option that does mostly what I expect. I just don't see any docs or tests around this, and I just want to be sure we're agreed on what the contract looks like.

If you're registering Organization, then you would like "teams" in `reverse_views`. Is that right?

AlanCoding (changes requested):
This is amazing, I just want to clear up our understanding. So my pending requested changes are:

 - Remove the prior solution to set _on a field_ `related_view = None`, both its use in test_app and its actual implementation in the common models
 - Add some example of how apps would use your current `reverse_views` solution

test_app/views.py
john-westcott-iv: Actually, upon reviewing the code more thoroughly, the ViewSets will require queryset because the router itself is utilizing that field. And we can't call get_queryset because we have a View class, not an instance.

AlanCoding: Ultimately, as per RBAC core requirements, we can't produce a queryset without a view and a user.

ansible_base/lib/routers/associative_resource_router.py
```diff
@@ -0,0 +1,179 @@
...
+        if basename is None:
+            basename = self.get_default_basename(viewset)
+
+        for reverse_name, (reverse_view, fk) in reverse_views.items():
+
+            def related_serializer_factory(reverse_view=reverse_view):
+                class RelatedSerializer(serializers.Serializer):
+                    instances = serializers.PrimaryKeyRelatedField(
+                        queryset=reverse_view.queryset,
+                        many=True,
```
AlanCoding: from the docs

https://www.django-rest-framework.org/api-guide/relations/

> queryset - The queryset used for model instance lookups when validating the field input. Relationships must either set a queryset explicitly, or set read_only=True.

Shouldn't this be `read_only=True`, because the POST to associate is a totally separate and unrelated action?

ansible_base/lib/abstract_models/common.py
john-westcott-iv: Yea, its not pinned. But this wouldn't require a simultaneous change (although I have one) because it just removes it off the base model. The models that inherit from it might set it as a field it just wouldn't be used anywhere.

ansible_base/lib/routers/associative_resource_router.py
john-westcott-iv: The screens render w/o the post options on them the way it is. From a quick read it sounds like adding `read_only: True` wouldn't hurt is probably better. However, when I added it on the POST option showed up. There may be some other black magic going on here. @relrod you have any ideas?

AlanCoding: Whether the POST option shows up is more the business of the viewset, not the serializer. I'd expect that the preferences of `reverse_view` dictates whether an object can be created, which isn't really right.

I'm getting confused by a number of things. The view `ReverseViewMixin` defines `get_queryset`, but... this doesn't fully add up. Is it the `instances` queryset that defines what list of related objects to show? Or is it the view `get_queryset`?

john-westcott-iv: Lets see what @relrod says, I mostly dug in deep enough to "make it work".

john-westcott-iv: Also, this function moved into another mixin to try and reduce untested duplicated code.

ansible_base/lib/routers/association_resource_router.py
```diff
@@ -0,0 +1,187 @@
...
+    def register(self, prefix, viewset, **kwargs):
+        basename = kwargs.get('basename', None)
+        related_views = kwargs.get('related_views', {})
+        reverse_views = kwargs.get('reverse_views', {})
+        if basename is None:
+            basename = self.get_default_basename(viewset)
+
+        for reverse_name, (reverse_view, fk) in reverse_views.items():
+
+            def related_serializer_factory(reverse_view=reverse_view):
```
AlanCoding: code style nit: better not to define the factory in-line. Yes, you absolutely need to define `RelatedSerializer` in-line. What I'm saying is that `def related_serializer_factory` does not need to be created inside the method here. It is much better to make that an instance method, class method, or totally stand-alone static method.

I realize you need `reverse_view`, but you made it an arg, so pass it as an arg. Making it the kwarg default is eyebrow-raising.

relrod: > Making it the kwarg default is eyebrow-raising.

You probably know this already, but for others wondering why it does that: It's actually really important to do that here, with the way this is currently defined due to Python's late binding rules. The code is _incorrect_ if the arg is removed, it will end up always referencing the last thing `reverse_view` was set to (in the last iteration of the loop) in every serializer it creates:

```python
In [5]: funcs = []

In [6]: for i in range(5):
   ...:     def foo():
   ...:         return i
   ...:     funcs.append(foo)
   ...: 

In [7]: for func in funcs:
   ...:     print(func())
   ...: 
4
4
4
4
4
```

AlanCoding:
Maybe it would be easier for you to just throw what you have in a feature branch and have us go from there?

docs/lib/routers.md
john-westcott-iv: Yes, docs added around this. Open to other implementations as well if we want to make this work differently.

ansible_base/lib/routers/association_resource_router.py
```diff
@@ -0,0 +1,187 @@
...
+        parent_pk = self.kwargs['pk']
+        parent_model = self.backward_related_qs.model
+
+        try:
+            parent_instance = parent_model.objects.get(pk=parent_pk)
+        except parent_model.DoesNotExist:
+            return parent_model.objects.none()
+
+        if issubclass(self.__class__, AssociateMixin):
+            child_queryset = getattr(parent_instance, self.association_fk).all()
```
relrod: nit: Can we let the subclass itself determine whether to use `self.association_fk` or `self.related_fk`? Like have it be a class attribute or something, rather than introspecting the subclass of this base class.

ansible_base/lib/routers/association_resource_router.py
```diff
@@ -0,0 +1,187 @@
...
+            if association_class is None:
+                raise RuntimeError('You must set association_serializer on the viewset')
+            return association_class
+        return super().get_serializer_class()
+
+
+class ReverseViewMixin(QuerySetMixinBase):
+    def get_serializer_class(self):
+        related_class = getattr(self, 'related_serializer', None)
+        if self.action in ('associate', 'disassociate'):
```
relrod: This feels weird to me. These actions are defined in `AssociateMixin` which (in our code that subclasses the mixins) is mutually exclusive with `ReverseViewMixin` right?

So `self.action` will never be these actions because these actions (presumably) don't exist.

ansible_base/lib/routers/association_resource_router.py
john-westcott-iv: Function moved out.

ansible_base/lib/routers/association_resource_router.py
john-westcott-iv: Yea, I notice this code as well and stripped it out since its effectively dead.

ansible_base/lib/routers/association_resource_router.py
john-westcott-iv: Fixed... ish. Has a bug that I need to track down but it now it uses class property.

ansible_base/lib/routers/association_resource_router.py
john-westcott-iv: I merged the related and the reverse views and also got rid of the kwargs (I only added it to make a test easier but that went away). Take a look and let me know if you are good with the direction I'm going. There is a bug that I need to track down later this afternoon.

ansible_base/lib/routers/association_resource_router.py
```diff
@@ -0,0 +1,167 @@
...
+    def register(self, prefix, viewset, related_views={}, basename=None):
+        if basename is None:
+            basename = self.get_default_basename(viewset)
+
+        for related_name, (related_view, fk) in related_views.items():
+            try:
+                for view in [viewset, related_view]:
+                    if getattr(view, 'queryset', None) is None:
+                        logger.error(f"ViewSet {view.__class__} does not have an associated queryset. Unable to add related view {related_name}")
+                        raise MissingQuerySetException()
```
AlanCoding: https://github.com/ansible/django-ansible-base/blob/devel/test_app/views.py#L12

If it has a queryset, then sure, you can use that. If not, can you use this method to attempt `view.serializer_class.Meta.model`? Do you need anything other than the model?

john-westcott-iv: That might be a way to get around some things. We do use the model from the query set. But the associate/disassociate do:
```
instance = self.backward_related_qs.get(pk=kwargs['pk'])
```

But maybe instead of that we could just do a `view.serializer_class.Meta.model.objects.get` instead?

👀

john-westcott-iv: That totally worked and gets us around the queryset issue.

ansible_base/lib/routers/association_resource_router.py
```diff
@@ -0,0 +1,156 @@
...
+        except AttributeError:
+            # If the field is FK and non-nullable, .remove() won't exist.
+            # https://docs.djangoproject.com/en/dev/ref/models/relations/#django.db.models.fields.related.RelatedManager.remove
+            # "For ForeignKey objects, this method only exists if null=True."
+            # Return something stating that the operation is not supported.
+            raise serializers.ValidationError(
+                {
+                    'instances': _('Cannot disassociate these objects because there must be a related object'),
+                }
+            )
```
relrod: Is this relevant anymore? Aren't we guaranteed to not be FK now?

john-westcott-iv: Yea, I couldn't figure out how to get into this logic from a test. It may just be dead code at this point but I wasn't sure if there was some fringe use case you found during initial development that might bight us if we removed it.

relrod (approved):
Approving, but I'd just drop the dead code. I think it's not an issue anymore with this approach.

AlanCoding (approved):

PR #134, opened 2024-02-12
Update security definition for EDA JWT Auth

Dostonbek1:
EDA e2e tests are failing due to EDAJWTAuthentication class security type being invalid. This PR sets the type to 'http'.


john-westcott-iv (approved):
From a code perspective it looks fine but I'm not familiar with the usage of `get_security_definition`.

PR #135, opened 2024-02-12
Add flag to skip logging during get_setting

slemrmartin:
It's needed to have a possibility to skip exception logging during "get_setting" because it can cause a exception loop.

DAB is using `logger.debug` several times during modules import phase, i.e. here: https://github.com/ansible/django-ansible-base/blob/25d39c407e43fc93dd79edd4a37e45d8c09963d6/ansible_base/lib/dynamic_config/dynamic_urls.py#L16

In first run, it tries to debug **before** the database is initialized and migrated.

With external logging implemented, it requires to get the setting if it's enabled - for gateway it's a dynamic preference. 
So logger wants the db settings, which calls get_setting, which raises exception, which wants to log ==> which wants the db settings etc. etc. 

So we need to avoid the loop.

* prerequisity for: [external loggers](https://github.com/ansible/aap-gateway/pull/226)

### DIFF
--- a/ansible_base/lib/utils/settings.py
+++ b/ansible_base/lib/utils/settings.py
@@ -12,7 +12,7 @@ class SettingNotSetException(Exception):
     pass
 
 
-def get_setting(name: str, default: Any = None) -> Any:
+def get_setting(name: str, default: Any = None, log_exception: bool = True) -> Any:
     try:
         the_function = get_function_from_setting('ANSIBLE_BASE_SETTINGS_FUNCTION')
         if the_function:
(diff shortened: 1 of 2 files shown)


ansible_base/lib/utils/settings.py
```diff
@@ -12,7 +12,7 @@ class SettingNotSetException(Exception):
     pass
 
 
-def get_setting(name: str, default: Any = None) -> Any:
+def get_setting(name: str, default: Any = None, log_exception: bool = True) -> Any:
```
john-westcott-iv, line 15: Do we plan on using this somewhere?

john-westcott-iv (commented):
Can you add a coverage test in https://github.com/ansible/django-ansible-base/blob/devel/test_app/tests/lib/utils/test_settings.py?

ansible_base/lib/utils/settings.py
slemrmartin: it's linked in the description - https://github.com/ansible/aap-gateway/pull/226, that's the only way I found, how to avoid exception loop in the early phase, before migrations - edge case

john-westcott-iv (approved):
Just add a quick test and this should be good to go.

PR #136, opened 2024-02-12
feat: Add resource manifest endpoint

rochacbruno:
Issue: AAP-20218

- [x] Add `resource_hash` to Resource model
- [x] Add `hash_serializer_data` utils function
- [x] Compute `resource_hash` field on `.save` on Resource 
- [x] Compute `resource_hash` on resource bulk* operations
- [x] Surface the /manifest/ endpoint on `ResourceTypeViewset`
- [ ] Compute `resource_hash` on save,update,bulk on Team, Org, User models
- [ ] Add API tests
- [ ] Add a management command to recompute all null resource hashes

```console
$curl  -X GET \
  'https://localhost/api/gateway/v1/service-index/resource-types/shared.user/manifest/' \
  --header 'Accept: */*' \
  --header 'Authorization: Basic YWRtaW46YWRtaW4='
```
```json
{
  "shared.user": {
    "863b5744-bc79-49fb-95cb-bb171a372c50": "6fe6021f948f23a378d338e5aae048b05bbf2a796101e6e5b10cf15dd0917a2a",
    "4c9ddcb4-080f-4d8c-a0d2-9dbffee1b3ce": "a9ca17c1dbb2ed5f2ddf24e16fc5ca635168c8c6bfe0d4502d167c2b9a0a11ef"
  }
}
```


AlanCoding:
Why can't we use the `modified_on` / `modified` timestamp to serve the same purpose that this hash would?

rochacbruno:
@AlanCoding modified can lead to a false positive of entries that requires sync, we are interested in only a specific set of fields and those come from resource_data, other fields may be changed without demanding a service sync.

AlanCoding:
Do you have any examples? Because the only fields I know of that have a consequence on that are AWX's organization `max_hosts` and `default_environment`. In the first place, AWX will never be the source of truth for organizations, and I really hope that we are never actively syncing changes from AWX organizations somewhere else. Even if we do, and if we had a false-positive for this particular reason, it would seem to have little to no practical consequence. And I'm perfectly find to move `max_hosts` to the shared model anyway.

AlanCoding:
After understanding the poll mechanism proposed, I'm in favor of the `modified` approach even more. The DAB base models will guarantee a time stamp exists. A server will poll another server which will be its source of truth. With this hashing method, all resources will have to be transmitted every poll interval, from all services using this.

While a server might add some fields in addition to the common model, the provider of the shared model should _never_ have unique fields not present in the other servers.

rochacbruno:
replaced by #148

PR #138, opened 2024-02-14
oidc authentication plugin

jctanner:
No-Issue


test_app/tests/conftest.py
```diff
@@ -179,6 +179,33 @@ def github_enterprise_team_authenticator(github_enterprise_team_configuration):
...
+        name="Test OIDC Authenticator",
+        enabled=True,
+        create_objects=True,
+        users_unique=False,
+        remove_users=True,
+        type="ansible_base.authentication.authenticator_plugins.oidc",
+        configuration=oidc_configuration,
+    )
+    yield authenticator
+    authenticator.delete()
```
john-westcott-iv, line 207: We found some issues around the deletion of authentications. If a test associates a user with the authenticator when the test gets back here and tries to delete it you will get an error because there are related users. Most of the authenticator fixtures are incorrectly missing this line right before delete:
```
authenticator.authenticator_user.all().delete()
```

That will remove all associated uses which then allows for the deletion of the authentication. Can you add this line in?

test_app/tests/conftest.py
```diff
@@ -179,6 +179,33 @@ def github_enterprise_team_authenticator(github_enterprise_team_configuration):
     authenticator.delete()
 
 
+@pytest.fixture
+def oidc_configuration():
+    return {
+        "OIDC_ENDPOINT": "https://localhost/api/gateway/callback/github_test/",
```
john-westcott-iv: NIT: Looks like there is a github reference here can we change it to `oidc_test`?

ansible_base/authentication/authenticator_plugins/oidc.py
```diff
@@ -0,0 +1,47 @@
...
+logger = logging.getLogger('ansible_base.authentication.authenticator_plugins.oidc')
+
+
+class OpenIdConnectConfiguration(BaseAuthenticatorConfiguration):
+    documentation_url = "https://python-social-auth.readthedocs.io/en/latest/backends/oidc.html"
+
+    OIDC_ENDPOINT = URLField(
+        help_text=_("The URL for your OIDC provider including the path up to /.well-known/openid-configuration"),
+        allow_null=False,
+        ui_field_label=_('OIDC Provider URL'),
```
john-westcott-iv, line 19: Can we call the URL validator on this and add a test or two for invalid URLs?

john-westcott-iv (commented):
What are we doing for testing going forward on this? Are we going to create a new Keycloak realm this can be plumbed into or using a new side car?

john-westcott-iv (approved):

PR #139, opened 2024-02-14
Get rid of in-line imports

AlanCoding:
I feel pretty confident that this is fine. There was an issue which was fixed by `SimpleLazyObject` so that we don't reference settings on import, and we don't. The imported file has nothing else that causes concerns of circular imports.


john-westcott-iv (approved):

AlanCoding:
D on a patch to improve code quality...

PR #141, opened 2024-02-14
Add related link between users and authenticators

john-westcott-iv:


djyasin (approved):

ansible_base/authentication/urls.py
```diff
@@ -23,19 +27,35 @@
...
+except Exception:
+    pass
+
+router = AssociationResourceRouter()
+router.register('authenticators', views.AuthenticatorViewSet, related_views=authenticator_related_views)
+router.register(
+    'authenticator_maps',
+    views.AuthenticatorMapViewSet,
+    related_views={
+        'authenticators': (views.AuthenticatorViewSet, 'authenticators'),
```
AlanCoding, line 51: I've wanted to see this change happen, but until now it couldn't.

ansible_base/authentication/models/authenticator_map.py
```diff
@@ -27,6 +27,7 @@ class Meta:
         null=False,
         on_delete=models.CASCADE,
         help_text="The authenticator this mapping belongs to",
+        related_name="authenticator_map",
```
AlanCoding: It's weird for you to declare a `related_name` that's singular. "authenticator_maps" would make much more sense.

john-westcott-iv: Good callout, fixed.

Issue #142, opened 2024-02-14
Ability to set attributable user to None


AlanCoding:
Right now AWX leaves `created_by` as None for objects created by the task system and in migrations. Due to recent changes we have problems creating DAB objects in migrations, like:

```
    role_definition, created = RoleDefinition.objects.get_or_create(
  File "/awx_devel/testing/django-ansible-base/ansible_base/rbac/models.py", line 62, in get_or_create
    return super().get_or_create(defaults=defaults, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/models/query.py", line 923, in get_or_create
    return self.create(**params), True
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/models/query.py", line 658, in create
    obj.save(force_insert=True, using=self.db)
  File "/awx_devel/testing/django-ansible-base/ansible_base/lib/abstract_models/common.py", line 132, in save
    self._check_user(user, non_existent_user_fatal)
  File "/awx_devel/testing/django-ansible-base/ansible_base/lib/abstract_models/common.py", line 115, in _check_user
    raise ValueError("Unable to save model without user!")
ValueError: Unable to save model without user!
```

As a library, we don't necessarily want to turn this off. But as AWX, there is clear precedent to outright flip the off switch and simply allow `None` values. Even setting

```python
SYSTEM_USERNAME = None
```

Does not fix the problem. The DAB interpretation seems to be that a None value requires the context to always have a user available. This is contrary to AWX needs, which _actually expects_ it to be None.

AlanCoding:
patch merged

PR #143, opened 2024-02-14
Allow None user for created_by if requested

AlanCoding:
Link issue https://github.com/ansible/django-ansible-base/issues/142


relrod (approved):
Clearly a bug, we explicitly say we want to allow `None`/no system user on line 96:

```
# If there is no system username defined, we just leave the user as None
```

PR #155, opened 2024-02-21
Activity Stream

relrod:
This is an activity stream. It's implemented via the use of signals and in that regard take some inspiration from AWX. In about every other aspect, it does its own thing, ignoring how AWX works.

This has a few implications: Some of the implementation is cleaner, we don't have a bunch of null FK fields on every activity stream row which is a huge win (we use generic foreign keys instead). But it also means we might (and probably do) miss out on some lessons AWX learned over the years with its implementation. That's why I call this an "Initial go" - because I'm sure there are cases where things will need to change a bit or performance won't be up to par.

That said, I think the foundation is relatively okay, if the goal is to provide an audit log. At the heart of this implementation, we have a new `diff` util function, which works quite a bit of logic to diff the fields of two models (which may or may not be the same type) and returns a structure containing their added, removed, and changed fields. The signals just make use of this utility function, and do some logic of their own for dealing with things like m2m associations.

One noteworthy thing w.r.t. m2m associations is that all relations happen _only_ on the forward relation. This can be good or bad. It means we're not needlessly duplicating data by creating entries in _both_ directions, and it means that if you look for entries from a model, you'll see _all_ the entries from that model. But it also means that you need to know the direction to do that lookup.

This is somewhat unfortunate because it ultimately means we leak implementation details - to filter, the user needs to know if they need to look from the perspective of model A or model B, because only in one direction will they find the entries.

Speaking of filtering - this is an area this PR doesn't really handle right now. There's a very simple serializer for getting activity stream entries. There's no RBAC integration right now, so all entries are exposed - that needs to be fixed. General filtering, though, also hasn't been tested/implemented. This is something that needs to be done in a followup PR.

But this PR broke 1100 lines and it's time to split off and make incremental changes once this merges.


ansible_base/activitystream/models/entry.py
```diff
@@ -0,0 +1,81 @@
...
+    content_object = GenericForeignKey('content_type', 'object_id')
+    operation = models.CharField(max_length=12, choices=OPERATION_CHOICES)
+    changes = models.JSONField(null=True, blank=True)
+
+    # This is used for m2m (dis)associations
+    related_content_type = models.ForeignKey(ContentType, on_delete=models.DO_NOTHING, null=True, blank=True, related_name='related_content_type')
+    related_object_id = models.TextField(null=True, blank=True)
+    related_content_object = GenericForeignKey('related_content_type', 'related_object_id')
+
+    # TODO: AWX stores denormalized actor data to account for deleted users, we should do the same
```
AlanCoding: Since you're using `DO_NOTHING`, I think it's actually a much better policy to leave it as it is and document what's possible. Get the user id, and then look at the activity stream filtered for the user type and id. If you want to really hold their hand, then you could add  "related" link to the activity stream for content_object and related_content_object, which should still be valid for deleted objects.

Something felt wrong about what AWX did, and there's legitimacy to that feeling. The deletion record for the user will show the username, and that will be saved according to the user's own retention policy. Saving a serialized version of related objects doesn't really obey the user's intent properly.

ansible_base/activitystream/models/entry.py
```diff
@@ -0,0 +1,83 @@
...
+    content_type = models.ForeignKey(ContentType, on_delete=models.DO_NOTHING)
+    object_id = models.TextField(null=True, blank=True)
+    content_object = GenericForeignKey('content_type', 'object_id')
+    operation = models.CharField(max_length=12, choices=OPERATION_CHOICES)
+    changes = models.JSONField(null=True, blank=True)
+
+    # This is used for m2m (dis)associations
+    related_content_type = models.ForeignKey(ContentType, on_delete=models.DO_NOTHING, null=True, blank=True, related_name='related_content_type')
+    related_object_id = models.TextField(null=True, blank=True)
+    related_content_object = GenericForeignKey('related_content_type', 'related_object_id')
```
AlanCoding, line 39: A question I am asking myself on the RBAC side - for any GFK, should we link to the `Resource` entry (dab resource registry) for the object? I'm talking about a nullable ForeignKey (to accommodate models it is not enabled for). I'm not selling this as the replacement for the GFK, but if we have any asks for showing the resource `ansible_id` or anything, it become _dead simple_ with this method, and the other approaches we have available work but are much more questionable.

relrod: > I'm not selling this as the replacement for the GFK

So you are envisioning an optional/nullable FK field _in addition_ to the GFK, that points to the `Resource` entry?

So, I think I'm okay with that, but what implications does this come with? If an app doesn't enable `dab_resource_registry` in `INSTALLED_APPS` - will referencing the model in an FK still work?

And then - assuming that does work fine - where does the value come from? Like, how do the signals know what the FK should point to? I suppose we could have a class var that models can set and points to the name of the field that contains it or something?

AlanCoding: > So, I think I'm okay with that, but what implications does this come with? If an app doesn't enable `dab_resource_registry` in `INSTALLED_APPS` - will referencing the model in an FK still work?

This is _exactly_ what I'm struggling with myself. It would make the resource registry a hard dependency of the app doing this. Migrations would have to list it as a dependency so that the Resource model exists. Up until this point the RBAC lib is a general Django RBAC list, the activity stream is a general Django activity stream, and I'd hate to start crippling that.

> where does the value come from?

The UI needs the ansible_id in order to navigate related objects between multiple systems. This mostly impacts shared resources like organizations. For a specific case... you can see an organization's activity stream which gives CRUD for the org itself, but maybe that links to the activity stream for AWX resources in the organization or something like that.

> Like, how do the signals know what the FK should point to?

The `Resource` model, itself, has a GFK to the object, so right before creating the activity stream entry, it's simple to just look up that id and add it to the creation data.

relrod: > The Resource model, itself, has a GFK to the object, so right before creating the activity stream entry, it's simple to just look up that id and add it to the creation data

So then I guess my question becomes: what purpose does denormalizing the `Resource` id serve here? If the UI (for example) needs to get the ansible_id, it's still going to have to query the resource table with `id=<the stored fk>`. Assuming our indexes are right on the resource registry side, then why not just query with `content_type=<activity stream> AND object_id=<my id>`?

In other words, why can't we just expose this interface (from the resource registry):

```python
class ResourceModel:
    @property
    def resource(self):
        # ignoring the not found case, for now
        return Resource.objects.get(content_object=self)
```

And then models that are part of the resource registry can mix this in, and clients that need the ansible_id can do `activity_stream_entry.content_object.resource.ansible_id`

Assuming the indexes are set up correctly, I would expect this to be just as fast as filtering by the id/pk, and it avoids the hard dependency on the resource registry (since if you're using the mix-in, you're already using the resource registry... and if you're not, then you're not.)

AlanCoding: Doing `Resource.objects.get` incurs an extra query for each serializer wheras a related `self.resource` can be a table JOIN. This would also let clients GET `/activity_stream/?resource__resource_id=123-232-232`

ansible_base/activitystream/models/entry.py
```diff
@@ -0,0 +1,83 @@
...
+    create/update/delete events sent to the activity stream.
+    """
+
+    class Meta:
+        abstract = True
+
+    activity_stream_excluded_field_names = []
+
+    @classmethod
+    def connect_signals(cls):
```
AlanCoding: I'd rather not have this as a class method, since there's no need for the nonsense call of `Organization.connect_signals`. That doesn't even clarify _what_ signals. If someone does `dir(organization)` this adds cruft to what they see. You could leave this as a plain function lying around wherever in this activity stream app.

ansible_base/activitystream/models/entry.py
```diff
@@ -0,0 +1,83 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+from django.db.models.signals import m2m_changed, post_save, pre_delete, pre_save
+from django.utils.translation import gettext_lazy as _
+
+from ansible_base.activitystream.signals import activitystream_create, activitystream_delete, activitystream_update, activitystream_m2m_changed
+from ansible_base.lib.abstract_models import CommonModel
+
+
+class Entry(CommonModel):
```
AlanCoding: If this inherits from `CommonModel` that adds a lot of fields that might not make sense here.

For the RBAC assignments, I wanted assignments to be immutable so I used the common model and set `modified_by = None` and `modified_on = None`. That's worked well, I like it.

Even so, I feel like this stretches the definition of `created_by` a little bit? Does that then become the attributable user for this entry? I saw you ordered by `created_on`, so it leads me to think that. That would still be okay if intended and documented, but I'd also expect immutability.

relrod: I feel like using `CommonModel` is the right idea here, it gets us some stuff for almost-free (the system user logic, etc.).

I agree that I don't expect activity stream entries to ever be modified, so I'm fine forcing immutability in `save()`, and nullifying the `modified_*` fields.

ansible_base/activitystream/models/entry.py
relrod: @AlanCoding Where did you land on this for RBAC? Do you still think we should have a resource field here?

AlanCoding: I don't know if there's usable cross-over to RBAC or resource registry right now. For one, no system using DAB RBAC is planned to use this. Then, we don't track a resource object for items other than users/teams/organizations. While integrating the resource could be used for those... that doesn't seem helpful in any way.

ansible_base/activitystream/signals.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    This signal is registered via the activity stream AuditableModel abstract
+    model/class. It is called before delete() of any model that inherits from
+    AuditableModel. (It is registered as a pre_delete signal.)
+    """
+    if instance.pk is None:
+        return
+
+    try:
+        old = sender.objects.get(pk=instance.pk)
```
AlanCoding: Why? Just pass in `instance`. This is `pre_delete`

ansible_base/activitystream/apps.py
```diff
@@ -0,0 +1,38 @@
...
+    def ready(self):
+        # Iterate models and look for ones that inherit from AuditableModel.
+        # If found, connect the signal to the model by calling its static connect_signals().
+        from django.apps import apps
+
+        from ansible_base.activitystream.models import AuditableModel
+
+        for model in apps.get_models():
+            if issubclass(model, AuditableModel):
+                connect_activitystream_signals(model)
```
AlanCoding, line 43: It's not that there is anything wrong with this, but I would like to raise the fact that DAB RBAC & resource registry do use a "registry" which is a competing approach to do the same thing. I also agree that the RBAC & resource registry should eventually (perhaps partially) merged into a single registry. That argues for doing this as a registry for no other reason that there could be more consistency.

ansible_base/activitystream/apps.py
```diff
@@ -0,0 +1,38 @@
...
+    post_save.connect(activitystream_create, sender=cls, dispatch_uid=f'dab_activitystream_{cls.__name__}_create')
+    pre_save.connect(activitystream_update, sender=cls, dispatch_uid=f'dab_activitystream_{cls.__name__}_update')
+    pre_delete.connect(activitystream_delete, sender=cls, dispatch_uid=f'dab_activitystream_{cls.__name__}_delete')
+
+    # Connect to m2m_changed signal for all m2m fields
+    for field in cls._meta.many_to_many:
+        if field.name in cls.activity_stream_excluded_field_names:
+            continue
+
+        fn = partial(activitystream_m2m_changed, field_name=field.name)
```
AlanCoding, line 19: If you're doing this as a partial, this advice may be applicable

> Note also that Django stores signal handlers as weak references by default, so if your handler is a local function, it may be garbage collected. To prevent this, pass weak=False when you call the signal’s [connect()](https://docs.djangoproject.com/en/5.0/topics/signals/#django.dispatch.Signal.connect).

Another alternative to doing a partial is to make your method a method in a class, and initialize the class in this code with `field.name` in the arguments. You would also need to save a reference (like from the AppConfig).

Looking at the call arguments of m2m_changed, I agree that manually saving the field name (in general) is a good idea, particularly if a model might use `through_fields` or some other voodoo.

ansible_base/activitystream/serializers.py
```diff
@@ -0,0 +1,8 @@
+from ansible_base.activitystream.models import Entry
+from ansible_base.lib.serializers.common import ImmutableCommonModelSerializer
+
+
+class EntrySerializer(ImmutableCommonModelSerializer):
+    class Meta:
+        model = Entry
+        fields = ImmutableCommonModelSerializer.Meta.fields + ['operation', 'changes']
```
AlanCoding: Isn't this missing a lot of fields? What about including `content_object` in the summary_fields (I did that in RBAC). Also for you, `related_content_object`.

ansible_base/lib/utils/models.py
```diff
@@ -97,3 +99,101 @@ def current_user_or_system_user():
...
+
+    :param old: The old instance for comparison
+    :param new: The new instance for comparison
+    :param require_type_match: If True, old and new must be of the same type of
+        model. (default: True)
+    :param json_safe: If True, the diff will be made JSON-safe by converting
+        all non-JSON-safe values to strings using Django's smart_str function.
+        (default: True)
+    :param exclude_fields: A list of field names to exclude from the diff.
+        (default: [])
```
AlanCoding, line 137: I can't find uses in the actual functional code that use these options.

ansible_base/lib/utils/models.py
```diff
@@ -97,3 +99,101 @@ def current_user_or_system_user():
...
+                make_json_safe(new_value) if json_safe else new_value,
+            )
+
+    for field in get_all_field_names(new, concrete_only=True):
+        if skip_field(new, field):
+            continue
+        if not hasattr(old, field):
+            diff_dict["added_fields"][field] = make_json_safe(getattr(new, field)) if json_safe else getattr(new, field)
+
+    return diff_dict
```
AlanCoding: What about encrypted fields?

ansible_base/lib/utils/models.py
```diff
@@ -97,3 +99,101 @@ def current_user_or_system_user():
...
+            continue
+
+        new_value = getattr(new, field)
+        if old_value != new_value:
+            diff_dict["changed_fields"][field] = (
+                make_json_safe(old_value) if json_safe else old_value,
+                make_json_safe(new_value) if json_safe else new_value,
+            )
+
+    for field in get_all_field_names(new, concrete_only=True):
```
AlanCoding: For developer sanity, I think it would help if there was just 1 field list, because old & new are established to be the same type.

EDIT: this makes more sense to put in the same loop.

ansible_base/lib/utils/models.py
```diff
@@ -97,3 +99,101 @@ def current_user_or_system_user():
...
+    """
+
+    def skip_field(source, field):
+        if field in exclude_fields:
+            return True
+        if limit_fields and field not in limit_fields:
+            return True
+        if not include_m2m and source._meta.get_field(field).many_to_many:
+            return True
+        return False
```
AlanCoding: prefer that `def` is moved out of this method.

ansible_base/lib/utils/models.py
```diff
@@ -97,3 +99,101 @@ def current_user_or_system_user():
...
+        raise TypeError('old and new must be of the same type')
+
+    for field in get_all_field_names(old, concrete_only=True):
+        if skip_field(old, field):
+            continue
+
+        old_value = getattr(old, field)
+
+        if not hasattr(new, field):
+            diff_dict["removed_fields"][field] = make_json_safe(old_value) if json_safe else old_value
```
AlanCoding: When does this line ever get called? I know `type(old) == type(new)` and I know that `field` (field name) is a concrete model field. Under those conditions `hasattr` should always be true. Can you delete this `if` block and still have all your tests pass? If so, then I would push for that change.

ansible_base/activitystream/admin.py
```diff
@@ -0,0 +1,5 @@
+from django.contrib import admin
+
+from ansible_base.activitystream.models import Entry
+
+admin.site.register(Entry)
```
AlanCoding: I have a ReadOnlyAdmin here, might be good to use here.

https://github.com/ansible/django-ansible-base/blob/19a3e55bc70a060f49972ec45aa4024ea22cca3c/ansible_base/rbac/admin.py#L6

ansible_base/lib/utils/models.py
relrod: > old & new are established to be the same type.

Only if `require_type_match` is True. I intentionally wanted this diff function to be able to handle models that aren't exactly the same type.

ansible_base/lib/serializers/common.py
```diff
@@ -71,8 +71,18 @@ def update(self, instance, validated_data):
...
 class NamedCommonModelSerializer(CommonModelSerializer):
     class Meta(CommonModelSerializer.Meta):
         fields = [
             'name',
         ] + CommonModelSerializer.Meta.fields
+
+
+class ImmutableCommonModelSerializer(AbstractCommonModelSerializer):
+    class Meta(AbstractCommonModelSerializer.Meta):
+        fields = ['id', 'url', 'created', 'created_by', 'related', 'summary_fields']
```
AlanCoding: I suggest putting `['id', 'url', 'created', 'created_by', 'related', 'summary_fields']` in a tuple defined outside of these classes, and then have `CommonModelSerializer` add to it to avoid duplication.

ansible_base/lib/utils/models.py
relrod: >  I know type(old) == type(new)

We don't know that, though. This invariant is guaranteed iff `require_type_match`.

ansible_base/lib/utils/models.py
relrod: Yes, and that's a bug, good find. The class variable `activity_stream_excluded_field_names` should work, but we never pass that to this function from the signals.

ansible_base/lib/utils/models.py
relrod: I didn't see any reason to handle them differently. Store their changes and if need be someone can decrypt the values in an audit, right?

ansible_base/lib/utils/models.py
relrod: I don't think this inner function is very useful as a general function, and it's only because I was repeating this logic in this function a lot, so I wanted to make it less annoying to read and write.

ansible_base/activitystream/admin.py
relrod: Yep. I'll pull this out to a util class and use it in both places.

ansible_base/activitystream/apps.py
relrod: I intentionally didn't want a registry here, just because it's another layer and it doesn't really buy us anything. I mean, I suppose it gives you a way to programmatically "back out" of having a model in the activity stream, but I can't imagine a scenario where that would be wanted.

If we had a "registry" I'd still want it to be automated and work like it does now (just that inheriting from `AuditableModel` would auto-register the model with the registry instead of directly connect the signals). But I just don't see the advantage.

It's also something we could do later on if there becomes a more compelling argument for it. The calling code shouldn't have to change at all if we do it like I said above.

ansible_base/activitystream/admin.py
relrod: @AlanCoding Do you intentionally not have `has_change_permission` there? Can we override that to return `False` too and make it truly read-only? Or would that break your expectation for how it works in RBAC?

ansible_base/activitystream/apps.py
```diff
@@ -0,0 +1,43 @@
+from functools import partial
+
+from django.apps import AppConfig
+from django.db.models.signals import m2m_changed, post_save, pre_delete, pre_save
+
+from ansible_base.activitystream.signals import activitystream_create, activitystream_delete, activitystream_m2m_changed, activitystream_update
+
+
+def connect_activitystream_signals(cls):
+    post_save.connect(activitystream_create, sender=cls, dispatch_uid=f'dab_activitystream_{cls.__name__}_create')
```
john-westcott-iv, line 10: Is there a reason we can't do pre_save for the create as well? I see we key off the created field but could we do something `if not pk` or is there a case like get_or_create where that would fall through?

ansible_base/activitystream/serializers.py
```diff
@@ -0,0 +1,43 @@
+from rest_framework import serializers
+
+from ansible_base.activitystream.models import Entry
+from ansible_base.lib.serializers.common import ImmutableCommonModelSerializer
+
+
+class EntrySerializer(ImmutableCommonModelSerializer):
```
john-westcott-iv, line 7: This will be in interesting serializer when we implement RBAC because we will have to look at what type each thing is and then determine if the user can see "that thing"  (non blocking).

ansible_base/activitystream/apps.py
relrod: We need the PK of the newly created object when we make the activity stream entry. If we did this in pre_save, we'd have no way to get that.

ansible_base/activitystream/signals.py
```diff
@@ -0,0 +1,154 @@
...
+    if operation not in ('create', 'update', 'delete'):
+        raise ValueError("Invalid operation: {}".format(operation))
+
+    old_excluded = getattr(old, 'activity_stream_excluded_field_names', [])
+    new_excluded = getattr(new, 'activity_stream_excluded_field_names', [])
+    excluded = set(old_excluded + new_excluded)
+
+    old_limit = getattr(old, 'activity_stream_limit_field_names', [])
+    new_limit = getattr(new, 'activity_stream_limit_field_names', [])
+    limit = set(old_limit + new_limit)
```
john-westcott-iv: Is there a valid case were the ContentType of an object can change? I think that is the only way the old/new objects would ever have different exclude_field_names or limit_field_name values right? If we aren't sure maybe we just add a comment like "Just incase we ever provide a way to alter a content type in a save".

ansible_base/activitystream/signals.py
```diff
@@ -0,0 +1,154 @@
...
+
+    delta = diff(old, new, exclude_fields=excluded, limit_fields=limit)
+
+    if delta["added_fields"] == {} and delta["changed_fields"] == {} and delta["removed_fields"] == {}:
```
john-westcott-iv: I'm wondering if it would be worth having `diff` return a changed boolean either in the dict or as a tuple of (changed: bool, {}) so if we ever use it somewhere else we don't need to duplicate the `if` logic like we have here.

ansible_base/activitystream/signals.py
```diff
@@ -0,0 +1,154 @@
...
+        # No changes to store
+        return
+
+    content_object = new
+
+    # If only one of old or new is None, then use the existing one as content_object
+    # The case where both are None is handled above (no changes to store)
+    if old is None:
+        content_object = new
```
john-westcott-iv: Do we need this part of the if since you have like 28 above?

ansible_base/activitystream/signals.py
```diff
@@ -0,0 +1,154 @@
...
+# pre_save
+def activitystream_update(sender, instance, raw, using, update_fields, **kwargs):
+    """
+    This signal is registered via the activity stream AuditableModel abstract
+    model/class. It is called before save() of any model that inherits from
+    AuditableModel. (It is registered as a pre_save signal.)
+
+    This signal only handles creation of new objects (created=True). For
+    updates, use the activitystream_update signal, where we can compare the
+    old and new objects to determine what has changed.
```
john-westcott-iv: Is this wrong? Shouldn't this be the opposite?

ansible_base/activitystream/signals.py
```diff
@@ -0,0 +1,154 @@
...
+    This signal is registered via the activity stream AuditableModel abstract
+    model/class. It is called when a many-to-many relationship is changed
+    (added or removed) for any model that inherits from AuditableModel. (It is
+    registered as a m2m_changed signal.)
+    """
+    if action not in ('post_add', 'post_remove', 'pre_clear'):
+        return
+
+    if 'field_name' not in kwargs:
+        raise ValueError("Missing field_name in kwargs")
```
john-westcott-iv: Would it be helpful to add the action (or anything else) to this error incase the code ever gets here and we need to troubleshoot how?

ansible_base/activitystream/signals.py
```diff
@@ -0,0 +1,154 @@
...
+    field_name = kwargs['field_name']
+    operation = 'associate' if action == 'post_add' else 'disassociate'
+
+    if action == 'pre_clear':
+        if reverse:
+            # Okay. We need to talk. Just you - the reader trying to understand this code - and I.
+            # Look. We want to always store the activity stream entry on the forward relation.
+            # Let's assume we have an Animal model with a 'people_friends' m2m. This is the forward relation.
+            # If we do: user.animal_friends.clear() - the reverse relation - we need to get the PKs of
+            # every animal that is being removed from the user's animal_friends.
```
john-westcott-iv: I love the comment but might expend it even more. In the sentence:
`Let's assume we have an Animal model with a 'people_friends' m2m. This is the forward relation.`
It left me wondering if the forward relation is the Animal Model or the 'people_frields'. Maybe do something like:
`Let's assume we have an Animal model with an m2m relationship to the User model. On the Animal model we will have 'people_friends' which points to the User model and this is considered the forward relation. The reverse relation would be through the user model at user.animal_firends.` Or something along those lines (just clarifing the forward vs reverse relation). But I seriously love this comment block. It will be incredably useful for engineers with less m2m experience.

john-westcott-iv: I might also pull the "metaphor" part of the comment out above the `if reverse` and just leave comments in the if/else explaining what you are doing to the actual models.

ansible_base/activitystream/urls.py
```diff
@@ -0,0 +1,12 @@
...
+from ansible_base.activitystream import views
+from ansible_base.lib.routers import AssociationResourceRouter
+
+router = AssociationResourceRouter()
+router.register(
+    'activitystream',
+    views.EntryReadOnlyViewSet,
+)
+
+api_version_urls = [path('', include((router.urls, 'dab_activitystream'), namespace='activitystream'))]
```
john-westcott-iv: general question, should we adding the "app namespace" (`dab_activitystream`) everywhere? Other django apps in DAB don't seem to pass a tuple to the include function.

ansible_base/lib/serializers/common.py
```diff
@@ -11,16 +11,13 @@
 logger = logging.getLogger('ansible_base.lib.serializers.common')
 
 
-COMMON_FIELDS = ['id', 'url', 'created', 'created_by', 'modified', 'modified_by', 'related', 'summary_fields']
-
-
-class CommonModelSerializer(ValidationSerializerMixin, serializers.ModelSerializer):
+class AbstractCommonModelSerializer(ValidationSerializerMixin, serializers.ModelSerializer):
```
john-westcott-iv, line 14: Do we want the serializer classes to be one-to-one with the structure in `lib/abstract_models/common.py`? So that if someone uses just ModifyableModel (for whatever reason) there is a cooresponding serializer for it?

docs/apps/activitystream.md
```diff
@@ -0,0 +1,105 @@
...
+stream, we are always comparing the same kinds of model.
+
+`deleted_fields`: This is very similar to `added_fields`. In the context
+of the activity stream, a record being deleted is kind of the same as turning it
+into the `None` model, and all fields are stored in `deleted_fields`. Like in
+`added_fields`, the `diff()` function could use this if comparing two different
+kinds of models, but in the context of the activity stream, that doesn't happen.
+
+`changed_fields`: This is a dictionary which has field names as keys and
+2-tuples as values in the form (old_value, new_value).
```
john-westcott-iv, line 58: How does this (the tuple) play out in the JSON format? Thinking about if I wanted to create a SQL JSON query to find anytime some value changed from `X` to something else, would it be easy to construct a query from the resulting JSON?

docs/apps/activitystream.md
```diff
@@ -0,0 +1,105 @@
...
+
+### create
+
+When a new record (instance of a model) is created and saved, we want to act on
+`post_save`. The reason for this is that, if we used `pre_save` which is called
+earlier, we wouldn't yet have a primary key/id to reference the object in the
+activity stream.
+
+Using
+[`post_save`](https://docs.djangoproject.com/en/5.0/ref/signals/#post-save)
```
john-westcott-iv, line 98: I like the link but it seems weird to have it here and not at the first mention of `post_save`.

docs/apps/activitystream.md
```diff
@@ -0,0 +1,105 @@
...
+
+### delete
+
+For deleting, we use
+[`pre_delete`](https://docs.djangoproject.com/en/5.0/ref/signals/#pre-delete).
+
+### (dis)association
+
+For associating and disassociating m2m fields, we use
+[`m2m_changed`](https://docs.djangoproject.com/en/5.0/ref/signals/#m2m-changed).
```
john-westcott-iv, line 122: For things with URLs or settings I usually try to include documentation on how you would add that manually. Like https://github.com/ansible/django-ansible-base/blob/devel/docs/apps/authentication/authentication.md#urls.

pyproject.toml
```diff
@@ -68,13 +68,13 @@ force-exclude = '''
 [tool.isort]
 profile = "black"
 line_length = 160
-extend_skip = [ "ansible_base/authentication/migrations", "test_app/migrations" ]
+extend_skip = [ "ansible_base/authentication/migrations", "ansible_base/activitystream/migrations", "test_app/migrations" ]
```
john-westcott-iv, line 73: For apps we've been trying to add optional deeps (above on lines 35-51) even if you don't need anything. That way, if the app needs a python lib in the future, the deeps are already there.

pyproject.toml
```diff
@@ -68,13 +68,13 @@ force-exclude = '''
...
 
 
 [tool.flake8]
 max-line-length = 160
 extend-ignore = [ "E203" ]
-exclude = [ 'ansible_base/authentication/migrations/*', 'test_app/migrations/*', '.tox', 'build']
+exclude = [ 'ansible_base/*/migrations/*', 'test_app/migrations/*', '.tox', 'build']
```
john-westcott-iv, line 79: Nice usage of a wildcard!

ansible_base/lib/utils/models.py
```diff
@@ -97,3 +99,101 @@ def current_user_or_system_user():
...
+    This function is used in particular by the activitystream application where
+    the changes returned by this function are stored as models change.
+
+    :param old: The old instance for comparison
+    :param new: The new instance for comparison
+    :param require_type_match: If True, old and new must be of the same type of
+        model. (default: True)
+    :param json_safe: If True, the diff will be made JSON-safe by converting
+        all non-JSON-safe values to strings using Django's smart_str function.
+        (default: True)
```
john-westcott-iv, line 135: Do we ever set these to not the defaults? I couldn't find an instance in this PR where that is true. Are we forwarding thinking of a time we would set these to not the defaults? If we don't have anything specific in mind do we want to have this extra code? (yes is a completely ok answer).

ansible_base/lib/utils/models.py
john-westcott-iv: I had a hard time following along some of the ifs/for in here and I felt like there was still duplicated code in regards to the make_json_safe. When I hit something like this I like to use sets because I think its easy to express the added/removed/modified concepts and differences and intersections. Here is a suggestion LMK if you find it easier to read or not. 
```suggestion
    diff_dict = {"added_fields": {}, "removed_fields": {}, "changed_fields": {}}

    # Short circuit if both objects are None
    if old is None and new is None:
        return diff_dict

    # Fail if we are not dealing with None or Model types
    if (old is not None and not isinstance(old, Model)) or (new is not None and not isinstance(new, Model)):
        raise TypeError('old and new must be a Model instance or None')

    # If we have to have matching types and both objects are not None and their types don't match then fail
    if require_type_match and (old is not None and new is not None) and type(old) is not type(new):  # noqa: E721
        raise TypeError('old and new must be of the same type')

    # Extract all of the fields and their values into a dict in the format of:
    #  fields = {
    #     'old': { <field>: <value>, [<field>: <value> ...]},
    #     'new': { <field>: <value>, [<field>: <value> ...]},
    #  }
    fields = {}
    for name, obj in (('old', old), ('new', new)):
        fields[name] = {}
        if obj is None:
            continue

        for field in get_all_field_names(obj, concrete_only=True):
            # Skip the field if needed
            if field in exclude_fields:
                continue
            if limit_fields and field not in limit_fields:
                continue
            if not include_m2m and obj._meta.get_field(field).many_to_many:
                continue

            fields[name][field] = make_json_safe(getattr(obj, field)) if json_safe else getattr(obj, field)

    old_fields_set = set(fields['old'].keys())
    new_fields_set = set(fields['new'].keys())

    # Get any remove fields from the old_fields - new_fields
    for field in old_fields_set - new_fields_set:
        diff_dict['removed_fields'][field] = fields['old'][field]

    # Get any new fields from the new_fields - old_fields
    for field in new_fields_set - old_fields_set:
        diff_dict['added_fields'][field] = fields['new'][field]

    # Find any modified fields from the union of the sets
    for field in new_fields_set & old_fields_set:
        diff_dict['changed_fields'][field] = (
            fields['old'][field],
            fields['new'][field],
        )

    return diff_dict
```

ansible_base/activitystream/signals.py
relrod: I think it's valid that the old and new will always have the same fields. I was thinking about things like polymorphic models (like AWX uses) but I think even with that, you don't have an instance where we go from one type to another during an update.

So we can probably simplify this and just use fields from one of the objects. I'll update this.

ansible_base/activitystream/signals.py
relrod: Changed to return an dataclass instance instead, and implemented `__bool__()` and a `has_changes()` alias.

docs/apps/activitystream.md
relrod: The tuple will be an array in the JSON representation. So each field in `changed_fields` will be of the form `str: [str]`

e.g. `"color": ["red", "brown"]`

I'm not 100% sure how this plays out in SQL JSON queries, as I have pretty much no experience with those, but yes that is a good question w.r.t. filtering.

ansible_base/lib/utils/models.py
relrod: We don't currently, but I wrote this to be a general function, since it's living in utils. It's not 100% meant to be activity stream specific.

ansible_base/activitystream/urls.py
relrod: Removed the namespace stuff. If you specify namespace, you need to also specify the app namespace. So I just removed both for consistency.

ansible_base/lib/serializers/common.py
relrod: So I think we can do this, but I think I'd like to make this a separate patch. Because it's going to need tests to test the whole hierarchy and so on. It's going to add quite a bit to an already big PR. Filed #246 for this.

docs/apps/activitystream.md
john-westcott-iv: I think that should be fine since its being converted to a JSON primitive.

john-westcott-iv (approved):
Approved pending merge meeting (and test fixes).

docs/apps/activitystream.md
```diff
@@ -0,0 +1,122 @@
...
+Django to list all the models it knows about. It will then filter this list for
+ones that inherit from `AuditableModel` and call the
+`AuditableModel::connect_signals` class method which registers the appropriate
+signals.
+
+You can exclude fields from being included in the changes logged by
+`activitystream` by setting the `activity_stream_excluded_field_names` class
+variable in your model.
+
+### What changes look like
```
AlanCoding, line 28: What does it look like when a ForeignKey is changed? Also, a extremely minimal example of this JSON would be nice.

test_app/settings.py
```diff
@@ -14,7 +14,7 @@
     },
     'handlers': {
         'console': {
-            '()': 'logging.StreamHandler',
+            '()': 'logging.NullHandler',
```
relrod: Nix

AlanCoding: TODO - Want to set this back.

AlanCoding (approved):
Got 1 change in test_app/settings.py to change before to merge. My list to file is:
 - add some "changes" JSON examples in the docs (easy)
 - add more info about changed related objects in the detail serializer for the UI
 - handle encrypted fields (highest priority)

docs/apps/activitystream.md
relrod: https://github.com/ansible/django-ansible-base/issues/250

relrod:
>     * add some "changes" JSON examples in the docs (easy)

https://github.com/ansible/django-ansible-base/issues/250

>     * add more info about changed related objects in the detail serializer for the UI

https://github.com/ansible/django-ansible-base/issues/251

>     * handle encrypted fields (highest priority)

https://github.com/ansible/django-ansible-base/issues/252

PR #156, opened 2024-02-21
Remove test_app local_authenticator fixture from test utils

AlanCoding:
Either we need to do this, or we need to move that fixture into this module.

Otherwise other apps can't legitimately use these fixtures.


AlanCoding:
Closing in favor of https://github.com/ansible/django-ansible-base/pull/157

PR #157, opened 2024-02-21
[try2] Move local_authenticator to test lib

AlanCoding:


relrod:
@AlanCoding  what do you think about having apps expose their own fixtures in their own module? To me this fixture is closely related to (and depends on) the authentication app, it feels like it should belong there instead, keeping the testing lib for more general fixtures.

I wanted to have the discussion now because moving it later will be annoying for all apps depending on the fixture.

AlanCoding:
@relrod removing this fixtures from other fixtures like `user` also broke tests related to the resource registry.

Apps can already put their own fixtures in their own module, like rbac, which has its own `test_app/tests/rbac/conftest.py`.

What I need this for is https://github.com/ansible/eda-server/pull/674, which is not an app internal to DAB, but a real project using this as a library. Just like these projects will use DAB base views and other things, it makes sense to me to have shared test utils. A `user` fixture and a user API client fit perfectly well in that scope. I don't care about `local_authenticator` one way or the other, it's just something that got in the way of employing those other useful things.

john-westcott-iv (approved):

PR #160, opened 2024-02-23
Run tests with postgres test database in addition to sqlite3

AlanCoding:
### Text / Integer Mismatch

I was working in eda-server to track down a bug. No tests in DAB was able to reproduce it. However, placing this test in the _EDA_ tests

```python
@pytest.mark.django_db
def test_view_assignments_non_admin(activation, user, user_api_client, view_activation_rd):
    assignment = view_activation_rd.give_permission(user, activation)
    url = reverse('roleuserassignment-list')
    r = user_api_client.get(url)
    assert r.status_code == 200
    assert r.data['count'] == 1
    assert r.data['results'][0]['id'] == assignment.id
```

immediately reproduced the bug. This was because they use a postgres database for testing, and we do not.

This is important for RBAC which often does a version of `pk__in=cls.values('object_id', ...`. But it's also important for activity stream https://github.com/ansible/django-ansible-base/pull/155 and resource registry which use TextField for generic foreign keys which definitely has database-specific behaviors.

I started borrowing some of their CI setup https://github.com/ansible/eda-server/blob/main/.github/workflows/ci.yaml and then eventually that led to this. If someone wants to integrate this into tox... firstly, that doesn't make sense to me because tox isn't what's standing up the postgres container. Secondly, I don't care about running a test matrix of python-version X database. Thirdly, that looks like a nightmare to carry out.

### User deletion issue

The common model has this field:

```python
    created_by = models.ForeignKey(
        settings.AUTH_USER_MODEL,
        related_name='%(app_label)s_%(class)s_created+',
        default=None,
        null=True,
        editable=False,
        on_delete=models.DO_NOTHING,
        help_text="The user who created this resource",
    )
```

I have now reproduced the error (again, I would like to reproduce it in _tests_) where it produces this error type:

```
tools_awx_1 |     raise e.error_from_result(result, encoding=pgconn_encoding(self.pgconn))
tools_awx_1 | django.db.utils.IntegrityError: update or delete on table "auth_user" violates foreign key constraint "dab_rbac_roleteamass_created_by_id_ef815504_fk_auth_user" on table "dab_rbac_roleteamassignment"
tools_awx_1 | DETAIL:  Key (id)=(307) is still referenced from table "dab_rbac_roleteamassignment".
```

So somehow in postgres we get the foreign key constraint even with the `DO_NOTHING` deletion behavior. These are obviously contradictory, and from a quick look it appears that this is a problem shared by all common models. We just need to test it here and figure out next steps from there. Not in AWX.


Makefile
```diff
@@ -51,6 +51,10 @@ check_flake8:
 check_isort:
 	tox -e isort -- --check $(CHECK_SYNTAX_FILES)
 
+## works the same as .github actions checks
+postgres:
+	docker run -itd --rm --name dab_postgres --net=host -e POSTGRES_USER=dab -e POSTGRES_PASSWORD=dabing -e POSTGRES_ADMIN_PASSWORD=dabing -e POSTGRES_DATABASE=dab_db postgres:13
```
chrismeyersfsu, line 56: From the comment on this block I reckon this follows some existing pattern.

I would have expected this to use the existing `docker-compose` commands to start the existing `postgres` service and use that database service/container to run the tests against.

AlanCoding: There are no existing `docker-compose` commands, and that doesn't mean I _want_ them either.

test_app is a beautiful thing, because it runs locally. It does its job well, which is to help test things. I don't want to add container wrappers, in part, because I don't want to maintain a test_app image. I would rather spend time making DAB runnable with fewer dependencies (ping @TheRealHaoLiu), which adds additional value, than creating a new image that no one wants.

The value that `make postgres` provides is that it runs a postgres container with the same postgres config as what the github actions do (username, password, etc). That means you can use the same Django settings for both. I looked into sharing an env file, and it doesn't seem to be possible with the github actions service pattern. Now `make postgres` will start a container in the background, which I justify by looking at the eda-server `task docker:up:minimal`. This is very different from the AWX (and others) workflow, but I believe it better matches DAB which is, and should be, a library first and foremost.

chrismeyersfsu: :facepalm: I thought this was the gateway repo.

:+1:

chrismeyersfsu (approved):

AlanCoding:
Closing because the alternative PR has broader team acceptance, so we should merge that. I can't approve it because I created it.

PR #161, opened 2024-02-24
Get rid of yield pattern in test fixtures

AlanCoding:
When I first saw this I thought it looked weird but not harmful.

But after understanding a bit more about the user deletion issue from https://github.com/ansible/django-ansible-base/pull/160, I now believe it is likely papering over a real issue with cascade deletions. Maybe it made the tests pass at some point, but it's possible it made a test pass that _should_ have been failing.

Also, I need to write tests that do `user.delete()` and with the current fixtures that gives an error on teardown because the user is already deleted.

### DIFF
--- a/ansible_base/lib/testing/fixtures.py
+++ b/ansible_base/lib/testing/fixtures.py
@@ -34,9 +34,7 @@ def local_authenticator(db):
         type="ansible_base.authentication.authenticator_plugins.local",
         configuration={},
     )
-    yield authenticator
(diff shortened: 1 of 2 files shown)


relrod (approved):
I've pondered on this for a few days now, and I think in the end, I'm okay with it. The tests run in a (fake) transaction anyway, so after each test, the database state should be restored regardless of us deleting the objects.

So given that, I think this is fine.

AlanCoding:
Thanks.

PR #165, opened 2024-02-27
Fix login issue with undefined user

john-westcott-iv:
There was an issue where the user already existed in the DB so `not self.pk` was not triggering during a user save. This gets around that by checking the current stack to see we are in a User save from the `authenticate` method.

### DIFF
--- a/ansible_base/authentication/backend.py
+++ b/ansible_base/authentication/backend.py
@@ -1,10 +1,12 @@
 import logging
 from collections import OrderedDict
 
+from crum import impersonate
 from django.contrib.auth.backends import ModelBackend
 
 from ansible_base.authentication.authenticator_plugins.utils import get_authenticator_plugin
 from ansible_base.authentication.models import Authenticator
+from ansible_base.lib.utils.models import get_system_user
 
 logger = logging.getLogger('ansible_base.authentication.backend')
 
@@ -14,6 +16,9 @@ authentication_backends = OrderedDict()
 class AnsibleBaseAuth(ModelBackend):
     def authenticate(self, request, *args, **kwargs):
         logger.debug("Starting AnsibleBaseAuth authentication")
+        system_user = get_system_user()
+        if not system_user:
+            logger.warning("System user is not available, attempting to login but failures could happen")
 
         for database_authenticator in Authenticator.objects.filter(enabled=True):
             # Either get the existing object out of the backends or get a new one for us
@@ -32,7 +37,8 @@ class AnsibleBaseAuth(ModelBackend):
                         continue
             authenticator_object = authentication_backends[database_authenticator.id]
             authenticator_object.update_if_needed(database_authenticator)
-            user = authenticator_object.authenticate(request, *args, **kwargs)
+            with impersonate(system_user):
+                user = authenticator_object.authenticate(request, *args, **kwargs)
             if user:
                 # The local authenticator handles this but we want to check this for other authentication types
                 if not getattr(user, 'is_active', True):
--- a/ansible_base/authentication/backend.py
+++ b/ansible_base/authentication/backend.py
@@ -1,12 +1,10 @@
 import logging
 from collections import OrderedDict
 
-from crum import impersonate
 from django.contrib.auth.backends import ModelBackend
 
 from ansible_base.authentication.authenticator_plugins.utils import get_authenticator_plugin
 from ansible_base.authentication.models import Authenticator
-from ansible_base.lib.utils.models import get_system_user
 
 logger = logging.getLogger('ansible_base.authentication.backend')
 
@@ -16,9 +14,6 @@ authentication_backends = OrderedDict()
 class AnsibleBaseAuth(ModelBackend):
     def authenticate(self, request, *args, **kwargs):
         logger.debug("Starting AnsibleBaseAuth authentication")
-        system_user = get_system_user()
-        if not system_user:
-            logger.warning("System user is not available, attempting to login but failures could happen")
 
         for database_authenticator in Authenticator.objects.filter(enabled=True):
             # Either get the existing object out of the backends or get a new one for us
@@ -37,8 +32,7 @@ class AnsibleBaseAuth(ModelBackend):
                         continue
             authenticator_object = authentication_backends[database_authenticator.id]
             authenticator_object.update_if_needed(database_authenticator)
-            with impersonate(system_user):
-                user = authenticator_object.authenticate(request, *args, **kwargs)
+            user = authenticator_object.authenticate(request, *args, **kwargs)
             if user:
                 # The local authenticator handles this but we want to check this for other authentication types
                 if not getattr(user, 'is_active', True):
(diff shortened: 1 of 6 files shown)


jbradberry (changes requested):
Inspecting the Python execution stack is, IMO, not an acceptable solution for this problem.

ansible_base/lib/utils/models.py
```diff
@@ -49,3 +56,21 @@ def user_summary_fields(user):
...
         sf[field_name] = getattr(user, field_name)
     return sf
+
+
+def get_system_user():
+    system_user = None
+    setting_name = 'SYSTEM_USERNAME'
+    system_username = get_setting(setting_name)
+    if system_username is not None:
+        try:
```
jbradberry: I think that this section would be more compact if you used `.first()` instead of `.get()` and compare the result to `None`.  It would not only get rid of the `try`/`except`, it would remove the need to do `get_user_model()` just to have access to the `DoesNotExist`.

ansible_base/lib/abstract_models/common.py
```diff
@@ -127,7 +99,7 @@ def save(self, *args, non_existent_user_fatal=True, **kwargs):
 
         # Manually perform auto_now_add and auto_now logic.
         now = timezone.now()
-        user = self._attributable_user(non_existent_user_fatal)
+        user = get_current_user() or get_system_user()
```
jbradberry: This was not my intention.  I think model saves that are done via non-web-request processes should be distinguished from non-authenticated web requests.  This does not do that:

| `get_current_user()` | `user` | intention |
|----|----|----|
| some `User()` | `User()` | that `User()` |
| `AnonymousUser()` | `AnonymousUser()` | `User(system)`, since it is (probably) auth |
| `None` | `User(system)` | `None`, since this was non-web |

jbradberry: Unless I'm wrong about `AnonymousUser` evaluating as truthy.  Still, I wouldn't count on it.

ansible_base/authentication/backend.py
```diff
@@ -32,7 +37,8 @@ def authenticate(self, request, *args, **kwargs):
                         continue
             authenticator_object = authentication_backends[database_authenticator.id]
             authenticator_object.update_if_needed(database_authenticator)
-            user = authenticator_object.authenticate(request, *args, **kwargs)
+            with impersonate(system_user):
```
jbradberry: We can probably ditch the `impersonate()`, since you demonstrated that something outside of the block seems also to touch the user.  The update to `last_login`?

ansible_base/lib/abstract_models/common.py
```diff
@@ -77,39 +75,13 @@ class Meta:
...
-
     def _check_user(self, user, non_existent_user_fatal=True):
-        # Its possible for _attributable_user to return None or anonymous user in which case we don't want to save
-        if (not user or user.is_anonymous) and non_existent_user_fatal and get_setting('SYSTEM_USERNAME'):
+        # user might be:
+        #     The logged in user object; save is allowed
+        #     The system user object; save is allowed
+        #     None; Do not allow save (unless non_existent_user_fatal is False)
+        #     AnonymousUser if an anonymous request came in; Do not allow save (unless non_existent_user_fatal)
+        if (not user or user.is_anonymous) and non_existent_user_fatal:
```
jbradberry: I worry about this method in general.  Django ties transaction management to whether or not an exception has been raised.  I can't remember off the top of my head how it goes, but I can easily imagine this rippling out to rollback more objects than intended.

Also looming large in my mind is the question of whether or not non-web actions (shell, management commands, internal processes, etc) should touch these fields for any object.  I kind of lean towards 'no'.

ansible_base/authentication/backend.py
john-westcott-iv: If we ditch this we have to handle more conditions in the User model for when its "ok" for a save to happen with an AnonymousUser. I'm ok with trying that but I also liked the idea of explicity setting this here instead of trying to do something like `if not self.pk and user.is_anonymous` in the `User.save` to try and catch creating a new user during an initial login. I also found that there is another save that happens to "update" the user logging in. So in that case it does not hit the previous `if` because the user has a `pk` at that point.

ansible_base/lib/utils/models.py
john-westcott-iv: Good call, changed to reduce function size.

ansible_base/lib/abstract_models/common.py
john-westcott-iv: This would convert None to system_user but AnonymousUser would fall through and eventually cause an exception. The way I was taking this solution, it's up to the models to grab AnonymousUser and choose if its appropriate to sub out system in place of Anonymous. Let me know if you are good with this approach. I personally had concerns with just converting Anonymous into System incase there was some kind of upstream coding issue that allowed an Anonymous request to attempt to save something.

ansible_base/lib/abstract_models/common.py
john-westcott-iv: If we are in a transaction saving multiple things and one fails because of this I would think the transaction should roll back. Do you feel differently? 

I think non-web actions should still modify these things as they are kind of like "audit trails" in my mind. A management command or internal process I think should do a `with impersonate(system_user)` if applicable or somehow otherwise figure out how to authenticate who the user is (admin presumably if they are running a management command) and update the fields with those fields. 

But I'm open to going with a group concensious on this point.

ansible_base/lib/utils/models.py
jbradberry: unpushed?

ansible_base/lib/abstract_models/common.py
jbradberry: I very strongly do not believe that high-level validation of this sort should be taking place in the `.save()` method.  Yes, Django does have the concept of validation that can take place there, but those are all pretty low level simple things.

jbradberry: `.save()` is not the place to determine if a developer left a view unguarded.

jbradberry:
I'm just going to pull back and do the rest of my comments up here at the top level, since everything that's left basically boils down to the same thing.

- I very strongly believe you should not be doing high-level business logic checks of "did a developer leave a view unguarded?" in a very low-level method like `save()`.
- I think at this point that both `None` and `AnonymousUser` should be converted into `User(username='_system')`, if we insist on having the `created_by` and `updated_by` fks be non-nullable.  Both cases mean that system-level logic (in the `AnonymousUser` case, authentication logic) created or updated the object.
  - That said, your logic on line 102 is incorrect -- you are relying on the validation method to catch the `AnonymousUser` that will (almost certainly) be falling through that line and stored into `user`.

I think something like:
```
user = get_current_user()
if user is None or not user.is_authenticated():
    user = get_system_user()
```

and then do away with the `impersonate()` and `_check_user()`.

ansible_base/lib/utils/models.py
john-westcott-iv: Hmmm... I thought I had pushed before I left let me see what happened.

john-westcott-iv: Its updated now.Updated now.

jctanner (approved):

jbradberry (approved):

PR #166, opened 2024-02-28
Run normal tox checks with postgres database

AlanCoding:
This is another option, as opposed to https://github.com/ansible/django-ansible-base/pull/160


.github/workflows/ci.yml
```diff
@@ -10,6 +10,16 @@ jobs:
...
       packages: write
       contents: read
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_USER: dab
+          POSTGRES_PASSWORD: dabing
+          POSTGRES_ADMIN_PASSWORD: dabing
+          POSTGRES_DATABASE: dab_db
```
AlanCoding: @relrod you corrected this `POSTGRES_DB` in `make postgres`. Should it be changed here as well?

test_app/scripts/bootstrap.sh
```diff
@@ -1,12 +1,34 @@
...
+}
 
-DJANGO_SUPERUSER_PASSWORD=password DJANGO_SUPERUSER_USERNAME=admin DJANGO_SUPERUSER_EMAIL=admin@stuff.invalid python manage.py createsuperuser --noinput
+trap cleanup EXIT
 
-python manage.py authenticators --initialize
+make postgres
 
-python manage.py create_demo_data
+for i in {1..5}; do
```
AlanCoding: I got to 3/5 for this, might want to bump this to 10 or something.

.github/workflows/ci.yml
relrod: good catch

test_app/scripts/bootstrap.sh
relrod: will do

.github/workflows/ci.yml
relrod: also I think `POSTGRES_ADMIN_PASSWORD` doesn't do anything on the official postgres image, so going to nix that too

test_app/scripts/bootstrap.sh
```diff
@@ -1,12 +1,34 @@
+#!/usr/bin/env bash
 
-rm -rf test_app/tests/sqllite_dbs/*.sqlite3
+set -e
 
-python manage.py migrate
+# Always make sure we kill the database when we exit
+function cleanup {
+    echo "Killing dab_postgres container"
```
AlanCoding: add AND THIS WILL DELETE YOUR DATA

.github/workflows/ci.yml
AlanCoding: this was way harder than it needed to be

https://hub.docker.com/_/postgres

 - POSTGRES_PASSWORD
 - POSTGRES_USER
 - POSTGRES_DB

I guess we just kill off the admin password b/c don't need it.

docs/testing.md
```diff
@@ -11,6 +11,29 @@ To run the test suite locally you can use tox. By default, with no arguments, to
...
+Now you can run tests against it, but only using the postgres-specific settings.
+
+```
+DJANGO_SETTINGS_MODULE=test_app.settings py.test test_app/tests
+```
+
+This may not work on an OS that runs docker in a VM.
+In that case, also use the `DOCKER_HOST` env var to point to wherever the VM is exposed.
+This should allow you to reproduce the results reported in the postgres database check.
+Stopping the container with `docker stop dab_postgres` will also remove the container.
```
AlanCoding: Need to update this, and also `test_app/README.md`. Maybe move this to that README and reference it here.

.github/workflows/ci.yml
relrod: Actually do we even need the postgres stuff here? Tox will run it, and CI is using tox.

AlanCoding:
Ran tests locally and got these failures

```
FAILED test_app/tests/jwt_consumer/common/test_auth.py::TestJWTCommonAuth::test_parse_jwt_no_header - AssertionError: assert 'X-DAB-JW-TOKEN header not set for JWT authentication' in ''
FAILED test_app/tests/jwt_consumer/common/test_auth.py::TestJWTCommonAuth::test_map_user_fields[user_fields1-token1-True] - assert "Saving user <Mock name='mock.username' id='139752271742416'>" in ''
FAILED test_app/tests/jwt_consumer/common/test_auth.py::TestJWTCommonAuth::test_map_user_fields[user_fields2-token2-True] - assert "Saving user <Mock name='mock.username' id='139752271633872'>" in ''
FAILED test_app/tests/jwt_consumer/common/test_auth.py::TestJWTCommonAuth::test_map_user_fields[user_fields4-token4-True] - assert "Saving user <Mock name='mock.username' id='139752271974992'>" in ''
FAILED test_app/tests/jwt_consumer/common/test_auth.py::TestJWTCommonAuth::test_map_user_fields[user_fields5-token5-True] - assert "Saving user <Mock name='mock.username' id='139752272272784'>" in ''
FAILED test_app/tests/jwt_consumer/common/test_auth.py::TestJWTAuthentication::test_process_permissions - AssertionError: assert 'process_permissions was not overridden for JWTAuthentication' in ''
```

.github/workflows/ci.yml
AlanCoding: Oh that's crazy. But it looks good to me. Delete the service here.

AlanCoding:
:+1: LGTM

please squash my commits.

requirements/requirements_dev.txt
```diff
@@ -8,3 +8,4 @@ pytest-xdist
 pytest-cov
 pytest-django
 setuptools-scm
+psycopg
```
cutwater: By default `psycopg` uses pure python installation, which dynamically loads `libpq` and calls it via `ctypes`. If compatibility and availability of binary distributions is not an issue, I'd suggest considering using `psycopg[binary]` for development environment.

Please note though, that `psycopg[binary]` should not be used for production environments. For production environments the `psycopg[c]` is the preferred type of installation.

See https://www.psycopg.org/psycopg3/docs/basic/install.html for more details.

Makefile
```diff
@@ -51,6 +51,15 @@ check_flake8:
 check_isort:
 	tox -e isort -- --check $(CHECK_SYNTAX_FILES)
 
+## Starts a postgres container in the background
+postgres:
+	docker run -itd --rm --name dab_postgres -p 55432:5432 -e POSTGRES_USER=dab -e POSTGRES_PASSWORD=dabing -e POSTGRES_DB=dab_db postgres:13
```
cutwater: I don't think `--interactive` and `--tty` flags are actually needed when running postgresql server.

AlanCoding: Confirming - this suggestion is to replace `-itd` with `-d`

PR #167, opened 2024-02-28
Fix resource api signals connection

cutwater:
The `post_save` and `post_delete` signals intended to create resources in DAB tables were connected in the `post_migrate` signal.

The `post_migrate` signal is not executed at the application startup, therefore the signals were not connected.

This fixes `DoesNotExist` exception that occured in the `create_resource` method of `Resource` model, when the resource that was expected to be created by the `post_save` signal was queried:

    resource = cls.objects.get(object_id=content_object.pk, content_type=c_type)


test_app/migrations/0003_create_system_user.py
```diff
@@ -12,16 +10,16 @@ def create_system_user(apps, schema_editor):
     The user is self-referential in its created_by and modified_by fields.
     It is inactive, only used for attributing internal changes to the system.
     """
-    User = get_user_model()
+    User = apps.get_model('test_app', 'User')
```
AlanCoding: This works for test_app, but it might be nice to still write it generally. To do that, I expect

```python
User = apps.get_model(settings.AUTH_USER_MODEL)
```

Would work the same, and still wouldn't encounter the problems you're fixing here about importing a model from current apps in migrations.

cutwater: I'm not sure this is a good idea. The proxy model used in this migration relies on the concrete model defined in the previous migration. The implementation here also relies on the specific set of fields defined in that model.

AlanCoding: I'm lost at "proxy model". I don't know where any proxy models exist in test_app or DAB.

cutwater: Okay, maybe I used an inaccurate term for this, because in Django proxy models is a different very specific thing.

The Django migrations engine tracks state of previous migrations and `apps.get_model` returns a virtual model based on this state. In this specific case the virtual `User` model generated from the model created in the `0001_initial` migration.

This migration relies on implementation details of that model and therefore requests it explicitly.
If the user model ever changes, it will require either introducing new migrations or rebuilding the entire migration tree.

Depending on `settings.AUTH_USER_MODELS` and using `swappable_dependency()` is useful for the library authors, when the specific model defined by the end-user is unknown in advance.

AlanCoding: Ok, I agree with everything you're doing. But I also still think `settings.AUTH_USER_MODEL` is valid, because it will return the prior state of the model. It's the same thing as `"test_app.User"`. It's just a string, and just a minor nit.

AlanCoding (approved):
Strongly agree with both changes here
 - connecting post_save signals in post_migration signal can't possibly be the original intent
 - get_user_model is inappropriate for data migrations

PR #173, opened 2024-02-29
Disconnect resource registry signals during migrations

cutwater:
This doesn't fix any particular immediate issue, just a nice to have change. The `post_save` and `post_delete` signals, that are connected to the `update_resource` and `delete_resource` receivers respectively, are triggered during migrations for the following objects:

1. Migrations themselves (`django.db.migrations.recorder.MigrationRecorder.Migration.<locals>.Migration`).
2. Instances of fake models, dynamically created by Django for migrations (e.g. `__fake__.ContentType').

It's unlikely we want to run resource handlers for these kinds of models, therefore disabling them in `pre_migrate` signal hook.


ansible_base/resource_registry/apps.py
```diff
@@ -67,15 +67,29 @@ def initialize_resources(sender, **kwargs):
             r_type.save()
 
 
+def connect_resource_signals(sender, **kwargs):
```
AlanCoding, line 70: passing `sender` or `**kwargs` doesn't make sense to me here. Even if you wanted to use it later, it's not really the right sender to use for these.

cutwater: It is used because on line 95, the `connect_resource_signals` function is registered in the `post_migrate` signal. I can call it in the `initialize_resources` function, but I decided to register it as a standalone receiver, as it makes code little cleaner in my opinion. Or I can refactor it into something like this:


```
def on_post_migrate(sender, **kwargs):
    initialize_resources()
    connect_resource_signals()
    

class ResourceRegistryConfig(AppConfig):
    # ...
    def ready(self):
        connect_resource_signals()
        signals.pre_migrate.connect(disconnect_resource_signals, sender=self)
        signals.post_migrate.connect(on_post_migrate, sender=self)
```

AlanCoding (approved):
On second look I see the reason for this now. This all makes sense to me now. I'll rebase my PR on top of this.

PR #174, opened 2024-02-29
Only register signals for models in registry

AlanCoding:
I saw https://github.com/ansible/django-ansible-base/pull/173 and it reminded me of something I wanted to do.

Right now, `update_resource` will be called every time any object in the system is saved for all objects. I know it uses cached data to do a no-op, but I don't see the necessity of this, and it makes me nervous to call it for say, an AWX `JobEvent` save.

So this is putting my my preference to only register `post_save` type signals on a per-model basis. This is what I see most commonly in other places, and I do this in the RBAC branch.

Also, we can't directly mock the `update_resource` method, because the way Django registers the signals it will still maintain a reference to the method from before the mock.

### DIFF
--- a/ansible_base/resource_registry/apps.py
+++ b/ansible_base/resource_registry/apps.py
@@ -68,7 +68,7 @@ def initialize_resources(sender, **kwargs):
 
 
 def proxies_of_model(cls):
-    "Return models that are a proxy of cls"
+    """Return models that are a proxy of cls"""
     for sub_cls in cls.__subclasses__():
         if sub_cls._meta.concrete_model is cls:
             yield sub_cls
@@ -80,6 +80,8 @@ def connect_resource_signals(sender, **kwargs):
     for model in handlers.get_resource_models():
         signals.post_save.connect(handlers.update_resource, sender=model)
         signals.post_delete.connect(handlers.remove_resource, sender=model)
+        # On registration, resource registry registers the concrete model
+        # so we connect signals for proxies of that model, and not the other way around
         for sub_cls in proxies_of_model(model):
             signals.post_save.connect(handlers.update_resource, sender=sub_cls)
             signals.post_delete.connect(handlers.remove_resource, sender=sub_cls)
--- a/ansible_base/resource_registry/apps.py
+++ b/ansible_base/resource_registry/apps.py
@@ -70,15 +70,17 @@ def initialize_resources(sender, **kwargs):
 def connect_resource_signals(sender, **kwargs):
     from ansible_base.resource_registry.signals import handlers
 
-    signals.post_save.connect(handlers.update_resource)
-    signals.post_delete.connect(handlers.remove_resource)
+    for model in handlers.get_resource_models():
+        signals.post_save.connect(handlers.update_resource, sender=model)
+        signals.post_delete.connect(handlers.remove_resource, sender=model)
 
 
 def disconnect_resource_signals(sender, **kwargs):
     from ansible_base.resource_registry.signals import handlers
 
-    signals.post_save.disconnect(handlers.update_resource)
-    signals.post_delete.disconnect(handlers.remove_resource)
+    for model in handlers.get_resource_models():
+        signals.post_save.disconnect(handlers.update_resource, sender=model)
+        signals.post_delete.disconnect(handlers.remove_resource, sender=model)
 
 
 class ResourceRegistryConfig(AppConfig):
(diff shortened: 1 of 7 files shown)


newswangerd (commented):
Can you add some testing to make sure this doesn't break with proxy models? The signals should be called if a proxy model or concrete model is updated.

cutwater:
@AlanCoding I really like this idea. Actually it's something we discussed yesterday with @newswangerd and I'm glad that you came up to this idea. Go for it!

AlanCoding:
I can test the proxy model, I already had one for test_app, I'd just have to move it over here. This might be good to discover some hazards of RBAC registration too.

ansible_base/resource_registry/apps.py
```diff
@@ -67,18 +67,33 @@ def initialize_resources(sender, **kwargs):
...
+        if sub_cls._meta.concrete_model is cls:
+            yield sub_cls
+
+
 def connect_resource_signals(sender, **kwargs):
     from ansible_base.resource_registry.signals import handlers
 
-    signals.post_save.connect(handlers.update_resource)
-    signals.post_delete.connect(handlers.remove_resource)
+    for model in handlers.get_resource_models():
```
cutwater, line 80: Minor nit: If `proxies_of_model` (maybe needs to be renamed) returned the model itself, it helpled to avoid code duplication.

cutwater (approved):

ansible_base/resource_registry/apps.py
```diff
@@ -67,18 +67,33 @@ def initialize_resources(sender, **kwargs):
             r_type.save()
 
 
+def proxies_of_model(cls):
+    "Return models that are a proxy of cls"
```
cutwater: Docstrings should be always wrapped in triple double quote:

```
"""This is a docstring."""
```

https://peps.python.org/pep-0008
https://peps.python.org/pep-0257/

ansible_base/resource_registry/apps.py
```diff
@@ -67,18 +67,33 @@ def initialize_resources(sender, **kwargs):
...
+    for sub_cls in cls.__subclasses__():
```
cutwater, line 72: I'm not quite sure I understand the general idea here. This will traverse the inheritance tree down from parent class down to child class. What if a ProxyModel is registered in resource registry? I believe actually it's a default case, at least in Pulp and Galaxy, where proxy models are used by default and base mode is not used directly.

cutwater (changes requested):

ansible_base/resource_registry/apps.py
AlanCoding: > What if a ProxyModel is registered in resource registry?

That's half of the point. See test models, in one case we register the proxy model. In the other case we register the original. In both cases, changes made using both original and proxy model should trigger the signals.

Why does this go _down_ but not _up_? Because if you register a proxy model, the resource registry logic obtains its concrete model and adds that to its registry, not the proxy model. So in all cases the `cls` we get here is the concrete model, because it comes from the registry.

AlanCoding: added a comment to echo my comment from here.

cutwater: I see now, thank you for explaining it to me. I must say this code is confusing, because it is disconnected from the code that determines the real model to be registered in a resource registry.

cutwater (approved):

newswangerd (approved):

ansible_base/resource_registry/apps.py
AlanCoding: I agree in spirit, although I don't want to rename the method, I'll push a commit that still hits the general ideal.

ansible_base/resource_registry/apps.py
```diff
@@ -67,18 +67,31 @@ def initialize_resources(sender, **kwargs):
...
+            yield sub_cls
+
+
 def connect_resource_signals(sender, **kwargs):
     from ansible_base.resource_registry.signals import handlers
 
-    signals.post_save.connect(handlers.update_resource)
-    signals.post_delete.connect(handlers.remove_resource)
+    for model in handlers.get_resource_models():
+        for cls in [model] + list(proxies_of_model(model)):
```
cutwater: IMO This looks cleaner

```suggestion
        for cls in [model, *proxies_of_model(model)]:
```

AlanCoding: pushed in commit.

PR #175, opened 2024-02-29
copy logic to read in additional settings files from awx

kdelee:
This would allow settings to be placed in

```
/etc/ansible_base/settings.py
/etc/ansible_base/conf.d/*.py.
```

now just have to figure out how to test it


AlanCoding (changes requested):
Big picture I don't want to add this environment variable to the configuration needed to correctly run the app.

wouldn't this require adding the `django-split-settings` dependency? That's going to be a mess - does this then become a DAB "app"? We already have resistance to that (@jbradberry I'm using you to make my point), and this become a total wreck.

Right now I say just put this into the apps themselves. I don't like the `/etc/ansible_base/` location either, in the case of AWX that would bring us to too many settings locations and get too confusing.

jbradberry:
I don't love this concept of a `dynamic_settings.py` anyways (though it didn't give me quite the negative reaction that `dynamic_urls.py` did).  Perhaps someone can explain to me exactly what problem it is intended to solve?

AlanCoding:
@jbradberry there needs to be some locations for settings to live that can be mounted into a container. I believe this exists for AWX, but not for other things.

john-westcott-iv:
Closing as this is going to be handled somewhere else.

PR #176, opened 2024-02-29
Add API root view to test_app

fosterseth:
lists endpoints at api/v1/

Also:

- adds a login button on the DRF pages

- enables django-extensions for shell_plus

Screenshot from 2024-02-29 16-14-27

### DIFF
--- a/test_app/settings.py
+++ b/test_app/settings.py
@@ -45,6 +45,7 @@ INSTALLED_APPS = [
     'ansible_base.jwt_consumer',
     'ansible_base.resource_registry',
     'test_app',
+    'django_extensions',
 ]
 
 MIDDLEWARE = [
--- a/test_app/views.py
+++ b/test_app/views.py
@@ -1,4 +1,7 @@
 from rest_framework import permissions
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
+from rest_framework.reverse import reverse
 from rest_framework.viewsets import ModelViewSet
 
 from ansible_base.lib.utils.views.ansible_base import AnsibleBaseView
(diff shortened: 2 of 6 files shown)


AlanCoding (approved):

test_app/views.py
```diff
@@ -32,3 +35,15 @@ class EncryptionModelViewSet(TestAppViewSet):
...
+
+
+# create api root view from the router
+@api_view(['GET'])
+def api_root(request, format=None):
+    return Response(
+        {
+            'organizations': reverse('organization-list', request=request, format=format),
+            'teams': reverse('team-list', request=request, format=format),
+            'users': reverse('user-list', request=request, format=format),
```
AlanCoding, line 47: Actually now I'm wondering if this might be _less_ useful than a 404, because in the event of a 404 it gives you a list of the patterns and names for them. What might be _best_ is to programatically loop over the URLs and auto-generate this. Even if it's a little gnarly, that's fine for test_app.

test_app/settings.py
```diff
@@ -45,6 +45,7 @@
     'ansible_base.jwt_consumer',
     'ansible_base.resource_registry',
     'test_app',
+    'django_extensions',
```
AlanCoding, line 48: I still can't use shell_plus after adding this. What else is needed?

fosterseth: hmm works fine for me, do you have ipython installed? (should be in requirements_dev.txt)

test_app/views.py
fosterseth: yeah would be nice to just collate these endpoints automatically, let me look into it..

test_app/settings.py
AlanCoding: It worked for me only after adding `"django_extensions"` to test_app INSTALLED_APPS like AWX has

AlanCoding: wait, this has that, I guess I was on another branch, this should be good :+1:

PR #177, opened 2024-03-01
Enforce that type hints are not obviously wrong

AlanCoding:
Outcome of day of learning - I've seen type hints become quite popular, but for all they are worth, they might as well be comments.

Since this is a _library_, it seems like the best place to start doing something better. We already know we have at least decent coverage from SonarCloud work, thanks @john-westcott-iv. Given that relatively decent and meaningful coverage, if tests are still passing, that means that at type hint gives relatively good assurance that the type is genuine.

This seems valuable if apps are interacting with DAB through a python interface. Those interface methods, in particular, can offer hardened type hints.

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/base.py
+++ b/ansible_base/authentication/authenticator_plugins/base.py
@@ -65,7 +65,7 @@ class AbstractAuthenticatorPlugin:
         else:
             self.logger = logger
 
-    def validate_configuration(self, data: dict, instance: object) -> None:
+    def validate_configuration(self, data: dict, instance: object) -> dict:
         if not issubclass(self.configuration_class, BaseAuthenticatorConfiguration):
             raise TypeError("self.configuration_class must subclass BaseAuthenticatorConfiguration.")
 
--- a/ansible_base/authentication/authenticator_plugins/ldap.py
+++ b/ansible_base/authentication/authenticator_plugins/ldap.py
@@ -104,7 +104,7 @@ class LDAPSearchField(ListField):
         self.validators.append(validator)
 
 
-def validate_ldap_filter(value: Any, with_user: bool = False) -> bool:
+def validate_ldap_filter(value: Any, with_user: bool = False) -> None:
     if type(value) is not str:
         raise ValidationError(VALID_STRING)
 
--- a/test_app/tests/lib/utils/test_views.py
+++ b/test_app/tests/lib/utils/test_views.py
@@ -1,6 +1,5 @@
 import importlib
 import logging
-import sys
 from unittest import mock
 
 import pytest
@@ -66,7 +65,6 @@ def test_ansible_base_view_parent_view(caplog, setting, log_message, default_par
             assert log_message in caplog.text
 
 
-@pytest.mark.xfail(reason='Type checking calls import_module', condition=lambda: sys.version.startswith('3.11'))
 def test_ansible_base_view_parent_view_exception(caplog):
     with override_settings(ANSIBLE_BASE_CUSTOM_VIEW_PARENT='does.not.exist'):
         with caplog.at_level(logging.ERROR):
--- a/test_app/tests/lib/utils/test_views.py
+++ b/test_app/tests/lib/utils/test_views.py
@@ -1,5 +1,6 @@
 import importlib
 import logging
+import sys
 from unittest import mock
 
 import pytest
@@ -65,6 +66,11 @@ def test_ansible_base_view_parent_view(caplog, setting, log_message, default_par
             assert log_message in caplog.text
 
 
+def is_py_311():
+    return bool(sys.version.startswith('3.11'))
+
+
+@pytest.mark.xfail(reason='Type checking calls import_module', condition=is_py_311)
 def test_ansible_base_view_parent_view_exception(caplog):
     with override_settings(ANSIBLE_BASE_CUSTOM_VIEW_PARENT='does.not.exist'):
         with caplog.at_level(logging.ERROR):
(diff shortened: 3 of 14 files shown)


ansible_base/authentication/authenticator_plugins/base.py
```diff
@@ -65,7 +65,7 @@ def set_logger(self, logger) -> None:
         else:
             self.logger = logger
 
-    def validate_configuration(self, data: dict, instance: object) -> None:
+    def validate_configuration(self, data: dict, instance: object) -> dict:
```
relrod, line 68: Can we specify the type parameters here?

Instead of `dict`, `dict[str, str]` or whatever. I'm guessing it'll be `dict[str, Any]` which I hate for _reasons_ but it's still best to be explicit.

ansible_base/authentication/authenticator_plugins/ldap.py
```diff
@@ -104,7 +104,7 @@ def validator(value):
         self.validators.append(validator)
 
 
-def validate_ldap_filter(value: Any, with_user: bool = False) -> bool:
+def validate_ldap_filter(value: Any, with_user: bool = False) -> None:
```
relrod, line 107: Okay, we're saying `value: Any` and then immediately raising `ValidationError` if it's not a string. If we are actually enforcing types, then theory says it should be safe to drop that check and just accept `value: str`.

But I'm not sure how practical this is, because it probably comes down to the JSON input, and there's no type-level validation at that point.

In other typed languages like Haskell the JSON deserializer would really return a [sum type](https://en.wikipedia.org/wiki/Tagged_union) that is something like `Either[Err, str]` and you'd handle both cases (`Err` being some error type returned by the JSON deserializer that says "hey, your JSON didn't make sense").

More succinctly: I find `value: Any` an the `type(value)` check really odd, but I believe fixing it would require a lot of work.

test_app/tests/lib/utils/test_views.py
```diff
@@ -65,6 +66,11 @@ def test_ansible_base_view_parent_view(caplog, setting, log_message, default_par
             assert log_message in caplog.text
 
 
+def is_py_311():
+    return bool(sys.version.startswith('3.11'))
```
relrod: I think this `bool()` is unnecessary? `startswith` should always return a bool.

test_app/tests/lib/utils/test_views.py
```diff
@@ -65,6 +66,11 @@ def test_ansible_base_view_parent_view(caplog, setting, log_message, default_par
...
+
+
+@pytest.mark.xfail(reason='Type checking calls import_module', condition=is_py_311)
```
relrod: This seems like we're losing this test on 3.11 then? I'm not really a fan of that. Can we just skip type checking for this function on 3.11 instead?

ansible_base/authentication/authenticator_plugins/ldap.py
AlanCoding: > In other typed languages like Haskell the JSON deserializer would really return a [sum type](https://en.wikipedia.org/wiki/Tagged_union) that is something like Either[Err, str] and you'd handle both cases (Err being some error type returned by the JSON deserializer that says "hey, your JSON didn't make sense").

Sounds like what I was reading at https://stackoverflow.com/questions/44282268/python-type-hinting-with-exceptions

tl;dr - no error type hints in python ever.

> If we are actually enforcing types, then theory says it should be safe to drop that check and just accept `value: str`.

That would make this `typeguard` check fail because the test passes in a parameter in violation of the type annotation. This is not where you want it to fail, because you want the _method_ to raise its error, because the entire point of the test to to assert that the method raises that error.

That proves the point of the necessity of the decorator to turn off type enforcement.

But it also proves why this PR to add checks for correctness of type hints should avoid getting into the problem of adding type hints. If someone wanted to change the argument type hint, and thus turn off type checking in the test later, fine.

test_app/tests/lib/utils/test_views.py
AlanCoding: It's more complicated than that. Assume type checking is already turned off. That doesn't prevent `typeguard` _logic_ from running. The problem is that this mocks stuff from `importlib` which is commonly called by other libraries... libraries like `typeguard`. Turning off enforcement doesn't mean disconnecting its logic.

It is hard to co-exist with `mock.patch('importlib.import_module', side_effect=ImportError("Test Exception"))`. And if you wanted a long-term durable solution, I would add another method and mock that method instead of something in `importlib`. I hesitated to go that far here because I didn't want to change the code itself.

ansible_base/authentication/authenticator_plugins/base.py
AlanCoding: Or I could drop the type hint entirely. Better no type hint than a wrong one.

test_app/tests/lib/utils/test_views.py
AlanCoding: I didn't want to do this because it goes so far out of my intended scope of this PR, but I introduced a new utility method so that the test doesn't mock `importlib.import_module` because _it shouldn't do this_.

Ping @john-westcott-iv as well. Since _my last rebase_ more incorrect type hints have been added. I have no issue whatsoever with people adding code with no type hints. What I do not like is code with provably wrong type hints. I've gotten rid of the hack here so that hopefully there should be no more potential blockers. I legitimately have a use case with RBAC, I would add more type hints for the more complex contracts but without regression coverage it's a non-starter.

john-westcott-iv: How are you checking the type hints? I'm wondering if we can create a GitHub action to enforce it on PRs and/or add a pre-commit hook?

AlanCoding: @john-westcott-iv this PR checks type hints at runtime via tests. It is integrated into tox testing, so this enforces type hint correctness (at least for those which are tested) as a CI gate.

john-westcott-iv: Perfect, we won't need a seperate action because it will just be part of the tox tests.

relrod (approved):

PR #180, opened 2024-03-01
Add AnsibleResourceField for reverse lookups to the Resource model

newswangerd:


AlanCoding:
> resolves to sql: "resource_model"."pk" :: text = "dab_resource_registry_resource"."object_id"

What the high-level concern here is, will this operate on _all_ entries in the resource model's table, and can that have a major penalty on queries in the future? (although I know this is fixable in the future, I just don't want to make it hard for ourselves to fix it)

My other concern is - what querysets will this affect? We would add it to the `User` model, but would it affect AWX queries that do `.select_related("created_by")`, where we have no need to for the user ansible_id?

newswangerd:
> > resolves to sql: "resource_model"."pk" :: text = "dab_resource_registry_resource"."object_id"
> 
> What the high-level concern here is, will this operate on _all_ entries in the resource model's table, and can that have a major penalty on queries in the future? (although I know this is fixable in the future, I just don't want to make it hard for ourselves to fix it)

I have no idea. We can do some more testing.

> My other concern is - what querysets will this affect? We would add it to the `User` model, but would it affect AWX queries that do `.select_related("created_by")`, where we have no need to for the user ansible_id?

It should function just like ForeignKey. The Resource queryset should only get resolved if you call `user.resource.<property>`.

I decided to not automatically add this to models in the registry. We'll start off by just adding it manually where it is needed.

newswangerd:
For references here is the sql that select_related spits out:

```
 SELECT "aap_gateway_api_organization"."id",
       "aap_gateway_api_organization"."created_on",
       "aap_gateway_api_organization"."created_by_id",
       "aap_gateway_api_organization"."modified_on",
       "aap_gateway_api_organization"."modified_by_id",
       "aap_gateway_api_organization"."name",
       "aap_gateway_api_organization"."description",
       "aap_gateway_api_organization"."environment_id",
       "dab_resource_registry_resource"."id",
       "dab_resource_registry_resource"."content_type_id",
       "dab_resource_registry_resource"."object_id",
       "dab_resource_registry_resource"."service_id",
       "dab_resource_registry_resource"."resource_id",
       "dab_resource_registry_resource"."name"
FROM   "aap_gateway_api_organization"
       left outer join "dab_resource_registry_resource"
                    ON ((( ( "aap_gateway_api_organization"."id" ) :: text =
       (
       "dab_resource_registry_resource"."object_id" )
       AND "dab_resource_registry_resource"."content_type_id" = 10
       )))
       left outer join "aap_gateway_api_environment"
                    ON ( "aap_gateway_api_organization"."environment_id" =
                                         "aap_gateway_api_environment"."id" )
ORDER  BY "aap_gateway_api_environment"."name" ASC,
          "aap_gateway_api_organization"."name" ASC  
```

AlanCoding:
Borrowing django-extensions dependency from https://github.com/ansible/django-ansible-base/pull/176 by @fosterseth and applying:

```diff
diff --git a/test_app/settings.py b/test_app/settings.py
index 64578f2..da765b2 100644
--- a/test_app/settings.py
+++ b/test_app/settings.py
@@ -39,6 +39,7 @@ INSTALLED_APPS = [
     'django.contrib.staticfiles',
     'rest_framework',
     'social_django',
+    'django_extensions',
     'ansible_base.api_documentation',
     'ansible_base.authentication',
     'ansible_base.rest_filters',
```

I'm able to get some really useful feedback via `python manage.py shell_plus --print-sql`. Let's map out the ORM actions of interest:

```python
orgs = list(Organization.objects.prefetch_related('resource'))
orgs[0]
orgs[0].resource
orgs[0].resource.ansible_id
orgs[1].resource
orgs[1].resource.ansible_id
```

This gives really good feedback, and this shows a lot of progress. However, when I get to `orgs[1].resource` I seem to find an outright bug that it's `None`. Contrast this to `Organization.objects.all()[0].resource`, which finds the correct item.

newswangerd:
@AlanCoding that's interesting. I never bothered testing `prefetch_related` because `select_related` is almost always better.

newswangerd:
@AlanCoding I got pretech_related working.

newswangerd:
I loaded ~10,000 users and orgs into the DB and ran the queries:

Joined query:
```
gateway=> EXPLAIN ANALYZE
SELECT "aap_gateway_api_organization"."id",
       "aap_gateway_api_organization"."created_on",
       "aap_gateway_api_organization"."created_by_id",
       "aap_gateway_api_organization"."modified_on",
       "aap_gateway_api_organization"."modified_by_id",
       "aap_gateway_api_organization"."name",
       "aap_gateway_api_organization"."description",
       "aap_gateway_api_organization"."environment_id",
       "dab_resource_registry_resource"."id",
       "dab_resource_registry_resource"."content_type_id",
       "dab_resource_registry_resource"."object_id",
       "dab_resource_registry_resource"."service_id",
       "dab_resource_registry_resource"."ansible_id",
       "dab_resource_registry_resource"."name"
FROM   "aap_gateway_api_organization"
       LEFT OUTER JOIN "dab_resource_registry_resource"
                    ON ((( ( "aap_gateway_api_organization"."id" ) :: text =
       (
       "dab_resource_registry_resource"."object_id" )
       AND "dab_resource_registry_resource"."content_type_id" = 10
       )))
LIMIT  100;
                                                                                  QUERY PLAN

--------------------------------------------------------------------------------------------------------------------------
----------------------------------------------------
 Limit  (cost=0.29..50.38 rows=100 width=169) (actual time=0.242..13.857 rows=100 loops=1)
   ->  Nested Loop Left Join  (cost=0.29..5078.64 rows=10139 width=169) (actual time=0.166..9.943 rows=100 loops=1)
         ->  Seq Scan on aap_gateway_api_organization  (cost=0.00..246.39 rows=10139 width=85) (actual time=0.048..1.881 r
ows=100 loops=1)
         ->  Index Scan using dab_resourc_content_6d9d9c_idx on dab_resource_registry_resource  (cost=0.29..0.48 rows=1 wi
dth=84) (actual time=0.024..0.024 rows=1 loops=100)
               Index Cond: ((content_type_id = 10) AND (object_id = (aap_gateway_api_organization.id)::text))
 Planning Time: 1.209 ms
 Execution Time: 15.942 ms
(7 rows)
```

Normal Query
```
gateway=> EXPLAIN ANALYZE
SELECT "aap_gateway_api_organization"."id",
       "aap_gateway_api_organization"."created_on",
       "aap_gateway_api_organization"."created_by_id",
       "aap_gateway_api_organization"."modified_on",
       "aap_gateway_api_organization"."modified_by_id",
       "aap_gateway_api_organization"."name",
       "aap_gateway_api_organization"."description",
       "aap_gateway_api_organization"."environment_id"
FROM   "aap_gateway_api_organization"
LIMIT  100;
                                                             QUERY PLAN

--------------------------------------------------------------------------------------------------------------------------
-----------
 Limit  (cost=0.00..2.43 rows=100 width=85) (actual time=0.066..5.858 rows=100 loops=1)
   ->  Seq Scan on aap_gateway_api_organization  (cost=0.00..246.39 rows=10139 width=85) (actual time=0.018..1.896 rows=10
0 loops=1)
 Planning Time: 0.090 ms
 Execution Time: 7.614 ms
(4 rows)
```

AlanCoding:
I made https://github.com/ansible/django-ansible-base/pull/190 to get the explain output from the debug toolbar, which is heavily duplicated with what you just did.

Tests are looking good and explain probably isn't raising any red flags.

test_app/tests/conftest.py
```diff
@@ -14,6 +14,29 @@
 from test_app import models
 
 
+def copy_fixture(copies=1):
```
AlanCoding: This looks like it was copied from here?

https://github.com/ansible/django-ansible-base/blob/5d9db2d0f5470beb1b6fd115ad219d649095a80d/ansible_base/lib/testing/util.py#L1

You could import it if you needed it.

newswangerd: oh, fantastic! I copied it from the gateway repo haha. I didn't realize it was in ansible base

AlanCoding:
Update - I do think this is in good shape. I just want to rebase it on top of the django-toolbar PR and have a quick look for myself. That means I also need to have the _optimizations_ applied in the test_app views, and I think that really seals the deal for this patch. But don't bother until we can merge with the DDT branch, because it introduces other optimizations.

AlanCoding:
Can you rebase this? I'll check it out and test but further changes will need a rebase.

AlanCoding:
I rebased and applied

```diff
diff --git a/test_app/models.py b/test_app/models.py
index e2ad925..93ddaab 100644
--- a/test_app/models.py
+++ b/test_app/models.py
@@ -10,18 +10,30 @@ from ansible_base.resource_registry.fields import AnsibleResourceField
 class Organization(AbstractOrganization):
     resource = AnsibleResourceField(primary_key_field="id")
 
+    def get_summary_fields(self):
+        response = super().get_summary_fields()
+        response["resource"] = self.resource.summary_fields()
+        return response
+
 
 class User(AbstractUser, CommonModel):
     resource = AnsibleResourceField(primary_key_field="id")
 
     def summary_fields(self):
-        return user_summary_fields(self)
+        response = user_summary_fields(self)
+        response["resource"] = self.resource.summary_fields()
+        return response
 
 
 class Team(AbstractTeam):
     resource = AnsibleResourceField(primary_key_field="id")
     encryptioner = models.ForeignKey('test_app.EncryptionModel', on_delete=models.SET_NULL, null=True)
 
+    def get_summary_fields(self):
+        response = super().get_summary_fields()
+        response["resource"] = self.resource.summary_fields()
+        return response
+
 
 class ResourceMigrationTestModel(models.Model):
     name = models.CharField(max_length=255)
diff --git a/test_app/views.py b/test_app/views.py
index 42a07aa..a45e81a 100644
--- a/test_app/views.py
+++ b/test_app/views.py
@@ -25,17 +25,19 @@ class TestAppViewSet(ModelViewSet, AnsibleBaseView):
 
 class OrganizationViewSet(TestAppViewSet):
     serializer_class = serializers.OrganizationSerializer
-    prefetch_related = ('created_by', 'modified_by')
+    prefetch_related = ('created_by', 'modified_by', 'resource')
 
 
 class TeamViewSet(TestAppViewSet):
     serializer_class = serializers.TeamSerializer
     prefetch_related = ('created_by', 'modified_by', 'organization')
+    # done as select_related here for demonstration purposes
+    select_related = ('resource',)
 
 
 class UserViewSet(TestAppViewSet):
     serializer_class = serializers.UserSerializer
-    prefetch_related = ('created_by', 'modified_by')
+    prefetch_related = ('created_by', 'modified_by', 'resource')
 
 
 class EncryptionModelViewSet(TestAppViewSet):
```

I visited http://127.0.0.1:8000/api/v1/organizations/?page_size=25, and there are 153 organizations in total.

It loaded with 66 queries in 36.27ms

devel will load with 6 queries in 6.37ms

The difference is due to an unoptimized content_type reference

Screenshot from 2024-03-04 16-22-21

I was also surprised by the prefetch query, which seemed to load 50 objects for some reason even though the page size was 25. This was a real surprise, and I don't have a good explanation for it.

But whatever the method is intended to be here, I'd like to get it in `test_app`. It is _way_ better to look at performance data of test_app than any other legacy app, because the ambient noise level is an order of magnitude lower.

AlanCoding:
^ disregard above comments about page_size. That is not hooked up in DAB, and it returns 50 items, so 50 is expected.

AlanCoding:
I forgot to mention in my prior comments - the org/user/team serializer is showing a field like `"resource": "42",` in their serializer. This is confusing on multilple levels (it's the resource id, not ansible_id, why is it a string anyway?). I say, just get rid of it. Unfortunately, I believe this means excluding the field from each and every serializer. You could try to hack something in the base serializer, either way let's just have an answer, and file followup issues to apply changes in the respective app serializers.

ansible_base/resource_registry/models/resource.py
```diff
@@ -9,6 +10,11 @@
 from .service_id import service_id
 
 
+@lru_cache(maxsize=None)
+def resource_type_cache(content_type_id):
+    return ContentType.objects.get(pk=content_type_id).resource_type
```
AlanCoding: The `ContentType.objects` manager already has a cache. See its mechanics:

https://github.com/django/django/blob/eff21d8e7a1cb297aedf1c702668b590a1b618f3/django/contrib/contenttypes/models.py#L9

It would obviously be better to use this. Instead of calling `resource_type_cache(id)` the code suggests `ContentType.objects.get_for_id(id)` should work the same.

I know, because I did it the wrong way myself in the RBAC work. I still need to fix that.

Standard pattern should be that, first time you know you're doing something with resource registry, do `ContentType.objects.get_for_models(...)` and pass the full list of models tracked by the resource registry. This will "warm the cache". Likewise for RBAC. And at some point, this will be the same list for both.

AlanCoding:
Rounding out the comments:
 - From my diff above, add `get_summary_fields` to test_app models to establish how to fully implement this. Or do it in the base serialzier.
 - Remove the `resource` field for the test_app models as well
 - (optional) maybe try to re-use the ContentType cache, I think it might be a simple drop-in replacement and if not I don't want to spend more time on it here

With those I'm :+1: ready to ship this.

newswangerd:
> I forgot to mention in my prior comments - the org/user/team serializer is showing a field like `"resource": "42",` in their serializer. This is confusing on multilple levels (it's the resource id, not ansible_id, why is it a string anyway?). I say, just get rid of it. Unfortunately, I believe this means excluding the field from each and every serializer. You could try to hack something in the base serializer, either way let's just have an answer, and file followup issues to apply changes in the respective app serializers.

This is happening because the test app uses `fields = '__all__'` in the serializers, which I don't think we do anywhere else.

AlanCoding:
You don't know if the project using your library is going to build its field list up or start with all and then remove.

https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include

```
class AccountSerializer(serializers.ModelSerializer):
    class Meta:
        model = Account
        exclude = ['users']
```

That is a standard documented pattern, and the logical next-step when an `__all__` comes up with a single field that needs to not be there.

newswangerd:
> You don't know if the project using your library is going to build its field list up or start with all and then remove.
> 
> https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include
> 
> ```
> class AccountSerializer(serializers.ModelSerializer):
>     class Meta:
>         model = Account
>         exclude = ['users']
> ```
> 
> That is a standard documented pattern, and the logical next-step when an `__all__` comes up with a single field that needs to not be there.

Then the author should exclude `resource` 🤷 

I'm going to try to figure out if I can get the serializer to at least show the ansible_id by default, but this really seems like a side quest.

EDIT: I figured out how to turn of serialization completely in the field definition

newswangerd:
@AlanCoding I made the following changes:

- The AnsibleResourceField will be excluded from serializers by default.
- The Resource summary fields will be added to serializers by default.

I left the cached function as it was, because it looked like the content type model wasn't caching the resource field

AlanCoding:
Going over my prior comments :heavy_check_mark: the serializers look to be finished.

```json
{
    "id": 2,
    "url": "/api/v1/organizations/2/",
    "related": {
        "modified_by": "/api/v1/users/1/",
        "teams": "/api/v1/organizations/2/teams/"
    },
    "summary_fields": {
        "created_by": {
            "id": 1,
            "username": "_system",
            "first_name": "",
            "last_name": ""
        },
        "modified_by": {
            "id": 1,
            "username": "_system",
            "first_name": "",
            "last_name": ""
        },
        "resource": {
            "ansible_id": "98087df3-29d4-4581-8f6f-ab789713aa62",
            "resource_type": "shared.organization"
        }
    },
    "created_on": "2024-03-05T16:49:19.902671Z",
    "modified_on": "2024-03-05T16:49:19.902671Z",
    "name": "Galaxy_community",
    "description": "",
    "created_by": 1,
    "modified_by": 1
}
```

This expresses this PR is delivering with the `ansible_id`. Ping @vidyanambiar

AlanCoding:
I put up https://github.com/newswangerd/django-ansible-base/pull/1 because this puts in things to allow optimizing, but I want to see it actually optimized with test_app. The prefetch/select related really needs the extra `__content_type` or else it will incur additional queries. I want to raise this to your awareness, in case this is showing that it's not functioning as you intend.

For my own purposes, I'm fine with these optimizations being applied to AWX, as they are here (in my patch) in test_app.

test_app/tests/resource_registry/models/test_resource_field.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+@pytest.mark.django_db
+def test_resource_field_filtering(organization):
+    """
+    Test that queryset filter works.
+    """
+    resource = Resource.objects.get(object_id=organization.pk, content_type=ContentType.objects.get_for_model(organization))
+
+    org = Organization.objects.get(resource__ansible_id=resource.ansible_id)
```
AlanCoding, line 98: Wild. Absolutely wild.

Manually tested http://127.0.0.1:8000/api/v1/organizations/?resource__ansible_id=98087df3-29d4-4581-8f6f-ab789713aa62 and it works. --> another tip to share with UI developers.

newswangerd: This is why I was so adamant to implement this as a field instead of an annotation :)

AlanCoding: I didn't think it could be done.

AlanCoding (approved):

test_app/tests/resource_registry/models/test_resource_field.py
AlanCoding: For historical record, the SQL

```sql
SELECT DISTINCT "test_app_organization"."id",
       "test_app_organization"."created_on",
       "test_app_organization"."created_by_id",
       "test_app_organization"."modified_on",
       "test_app_organization"."modified_by_id",
       "test_app_organization"."name",
       "test_app_organization"."description"
  FROM "test_app_organization"
 INNER JOIN "dab_resource_registry_resource"
    ON (((("test_app_organization"."id")::text = ("dab_resource_registry_resource"."object_id") AND "dab_resource_registry_resource"."content_type_id" = 19)))
 WHERE "dab_resource_registry_resource"."ansible_id" = '98087df3-29d4-4581-8f6f-ab789713aa62'::uuid
 LIMIT 1
```

That `Cast` becomes `::text` rather nicely without blowing up unexpected parts of the logic.

PR #182, opened 2024-03-01
Use existing dab_postgres container if it is running

AlanCoding:
If you tried to use this, you probably saw this coming.

So I want to keep using this same database with the tests, obviously. I will obviously use the same database forever. Until it breaks.

### DIFF
--- a/test_app/scripts/bootstrap.sh
+++ b/test_app/scripts/bootstrap.sh
@@ -16,7 +16,15 @@ function cleanup {
 
 trap cleanup EXIT
 
-make postgres
+docker_running=$(docker ps -aq -f name=dab_postgres)
+
+if [[ -z "${docker_running// /}" ]]
+then
+    echo "creating new postgres container"
+    make postgres
+else
+    echo "dab_postgres container is already running, will use that container"
+fi
 
 MAX_ATTEMPTS=10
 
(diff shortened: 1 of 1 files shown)


test_app/scripts/bootstrap.sh
```diff
@@ -31,12 +38,12 @@ for i in $(seq 1 $MAX_ATTEMPTS); do
...
 done
 
-python3 manage.py migrate
-
-DJANGO_SUPERUSER_PASSWORD=password DJANGO_SUPERUSER_USERNAME=admin DJANGO_SUPERUSER_EMAIL=admin@stuff.invalid python3 manage.py createsuperuser --noinput
-
-python3 manage.py authenticators --initialize
-
-python3 manage.py create_demo_data
+if [ -z "$docker_running" ]
```
relrod, line 42: Can we standardize on how we're doing conditionals?

Both syntactically (double vs single brackets, `then` on same line vs next line), and also semantically (I don't get why this is one so different from the above one, it seems to be wanting to do the same check? Can we simplify the above one to be like this one?)

relrod (approved):
Approving because I agree with the goal here, but if we can make the patch be a little more consistent, I'd love it more.

test_app/scripts/bootstrap.sh
AlanCoding: I thought the single versus double brackets were meaningful? I'll change the if; then, because I'm sure that doesn't make a difference.

I Google for how to do something and use code from there... I never really got the hang of bash myself and sometimes the syntax does seem like voodoo.

PR #185, opened 2024-03-01
Dynamically create api root for test_app

fosterseth:
Instead of hard-coding root api endpoints, automatically add them.

- Loops over router URL patterns and add an entry for the "*-list" views.

- Checks that the entry is a "root" view by checking for a single "/" in the pattern.
image


AlanCoding (approved):
:tada:

PR #186, opened 2024-03-01
Merge ansible ID and resource ID

newswangerd:
Ansible ID used to be a concatenation of the service ID plus resource UUID. The was originally intended to tell clients what service a resource belongs to, but this wasn't terribly useful and just made the system more complicated than it needed to be.

This change renames `resource_id` to `ansible_id` and removes the computed `ansible_id` property.


AlanCoding (approved):

AlanCoding:
I've rebased the RBAC PR after this merge and adjusted. Now fields like `object_ansible_id` for assigning a permission are DRF UUIDField.

PR #187, opened 2024-03-01
Turn off create-db by default, add check

AlanCoding:
I'm going to be honest, `--create-db` just doesn't work for me. I'm _frequently_ just running 1 test over and over again. The migration takes a long time and dumps pages of logs.

Yes, I have been manually removing this line from my local source tree manually. Today I just got sick enough of it that I thought I would do something. @relrod expressed concern that this introduces _subtle_ bugs in tests (a column doesn't exist) when switching between different database states. Sure. But for the record, this doesn't address the need for `makemigrations`, and we live with _that_.

So, here are reproduction steps for the concern:

 - check out this branch https://github.com/ansible/django-ansible-base/compare/devel...AlanCoding:django-ansible-base:test_migration?expand=1
 - run a test `py.test test_app/tests/resource_registry/ -k test_resource_type_detail --create-db`, the --create-db put in for illustration
 - check out this branch and try to run

The idea there is that you need to add `--create-db` but it is subtle that you do. With this patch, you get:

```
E           AssertionError: Migrations not expected for app test_app, perhaps you need --create-db?
E           assert {'0003_create_system_user', '0001_initial', '0002_set_up_resources_test_data'} == {'0003_create_system_user', '0004_remove_team_encryptioner', '0001_initial', '0002_set_up_resources_test_data'}
E             Extra items in the right set:
E             '0004_remove_team_encryptioner'
E             Full diff:
E               {
E                '0001_initial',
E                '0002_set_up_resources_test_data',
E                '0003_create_system_user',
E             -  '0004_remove_team_encryptioner',
E               }

test_app/tests/conftest.py:45: AssertionError
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,7 +77,7 @@ exclude = [ 'ansible_base/authentication/migrations/*', 'test_app/migrations/*',
 
 [tool.pytest.ini_options]
 DJANGO_SETTINGS_MODULE = "test_app.settings"
-addopts = "--strict-markers --reuse-db --create-db --migrations -s -vvv"
+addopts = "--strict-markers --reuse-db --migrations -s -vvv"
 
 [tool.tox]
 legacy_tox_ini = """
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ legacy_tox_ini = """
         -r{toxinidir}/requirements/requirements_all.txt
         -r{toxinidir}/requirements/requirements_dev.txt
     allowlist_externals = sh
-    commands = sh -c 'make postgres && pytest -n auto --cov=. --cov-report=xml:coverage.xml --cov-report=html --cov-report=json --cov-branch {env:ANSIBLE_BASE_TEST_DIRS:test_app/tests} {posargs}'
+    commands = sh -c 'make postgres && pytest -n auto --cov=. --cov-report=xml:coverage.xml --cov-report=html --cov-report=json --cov-branch {env:ANSIBLE_BASE_PYTEST_ARGS} {env:ANSIBLE_BASE_TEST_DIRS:test_app/tests} {posargs}'
 
     [testenv:check]
     deps =
--- a/test_app/tests/conftest.py
+++ b/test_app/tests/conftest.py
@@ -1,4 +1,7 @@
 import datetime
+import os
+import re
+from collections import defaultdict
 from unittest import mock
 
 import jwt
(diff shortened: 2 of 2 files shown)


relrod:
I'm not in favor of this because I do think starting fresh and running migrations each time is the right thing to do. But I'm not going to block it.

I only ask this: if you _do_ end up merging this, can you at least add an `ANSIBLE_BASE_PYTEST_ARGS` env placeholder on line 101 somewhere near the end of the `pytest` command, so that those of us who do rely on this behavior can shove a line in `.bashrc` and never have to think about it again?

AlanCoding:
> can you at least add an `ANSIBLE_BASE_PYTEST_ARGS` env placeholder on line 101 somewhere near the end of the pytest command

I can't figure out what line this references.

If what you have in mind can work, it can be _the_ solution. If _I_ can set an env var and never have it create db by default, I will welcome that. But what's so challenging about this is that I don't have an easy way to remove it. It's _very easy to add the flag_ and it's necessary for everyone doing development. You just add `--create-db` onto your py.test command. But there is no equivalent that lets you _take away_ the `--create-db` flag put in by pyproject.toml. If you know of any such solution, I'll be happy to use it, and allow it to replace some or all of this patch.

relrod:
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 30dd06d..f2c5389 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ legacy_tox_ini = """
         -r{toxinidir}/requirements/requirements_all.txt
         -r{toxinidir}/requirements/requirements_dev.txt
     allowlist_externals = sh
-    commands = sh -c 'make postgres && pytest -n auto --cov=. --cov-report=xml:coverage.xml --cov-report=html --cov-report=json --cov-branch {env:ANSIBLE_BASE_TEST_DIRS:test_app/tests} {posargs}; make stop-postgres'
+    commands = sh -c 'make postgres && pytest -n auto --cov=. --cov-report=xml:coverage.xml --cov-report=html --cov-report=json --cov-branch {env:ANSIBLE_BASE_PYTEST_ARGS} {env:ANSIBLE_BASE_TEST_DIRS:test_app/tests} {posargs}; make stop-postgres'
 
     [testenv:check]
     deps =
```

This is the change I am proposing (_on top_ of what you have here already).

This lets me do `export ANSIBLE_BASE_PYTEST_ARGS=--create-db` and retain the current behavior.

We _could_ set a default (the syntax for that is `{env:ANSIBLE_BASE_PYTEST_ARGS:--create-db}`) and then _you_ could `export ANSIBLE_BASE_PYTEST_ARGS=` (nothing).

I'm fine either way, I don't particularly care what the default is, I just want to be able to keep the existing behavior in some form.

My reasoning is really that when I run tests locally, I really want it to be as close to CI as possible and CI will always be running it with a clean db.

AlanCoding:
Thanks, will try that, but https://github.com/ansible/django-ansible-base/pull/192 is a more immediately urgent change, and I'll opt to add that while doing the rebase after that.

AlanCoding:
Added that change, I'm good with this.

If people want, I could remove the extra logic and merge it that way. Whichever is fine with me.

relrod (approved):

relrod:
Note for @john-westcott-iv and others, to retain current behavior, add:

```bash
export ANSIBLE_BASE_PYTEST_ARGS="--create-db"
```

to your `~/.bashrc` or `~/.bash_profile` and `source` it.

PR #189, opened 2024-03-01
Add VSCode launch configuration for test app

fosterseth:
Adds a launch configuration for test app

This will launch the test app server via VSCode so that develops can set debug points and step through code using VSCode.

Additional changes:

bootstrap.sh can be called with args

./bootstrap.sh startpostgres

this allows the launch configuration to start the postgres server separately from runserver.

./bootstrap.sh without args runs exactly as it did before.

image

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,13 @@
+{
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "label": "Run vscodebootstrap.sh",
+      "type": "shell",
+      "command": "sh",
+      "args": [
+        "${workspaceFolder}/.vscode/vscodebootstrap.sh"
+      ],
+    }
+  ]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -27,9 +27,6 @@ venv/
 # Mac OS X
 *.DS_Store
 
-# VSCode
-.vscode/
-
 # Editors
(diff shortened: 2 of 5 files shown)


.vscode/tasks.json
```diff
@@ -0,0 +1,10 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+      {
+        "label": "bootstrap",
+        "type": "shell",
+        "command": "${workspaceFolder}/test_app/scripts/bootstrap.sh startpostgres",
```
AlanCoding: So this is like a separate button you click in VSCode? I get why you need to give it your python interpreter and have it wrap it or whatever to do it in debug mode. But it seems a lot less clear how to manage a background container.

AlanCoding: Maybe it would be easier to just let VSCode use sqlite3. We were until like this morning.

AlanCoding:
Otherwise this looks good to me. I just wonder if maintaining a postgres database at _this_ point is counter productive to the development experience. I never wanted it to be postgres-or-the-highway.

fosterseth:
@AlanCoding made some changes

- uses sqlite3 instead of postgres
- no longer uses the bootstrap.sh script, just call a custom script made for vscode situations

AlanCoding (approved):

PR #191, opened 2024-03-02
Add Beartype type checking

AlanCoding:
https://beartype.readthedocs.io/en/latest/faq/#how-do-i-only-type-check-while-running-my-test-suite

Alternative to https://github.com/ansible/django-ansible-base/pull/177 which is typeguard

> You know [typeguard](https://github.com/agronholm/typeguard)? Then you know beartype – more or less. beartype is [typeguard](https://github.com/agronholm/typeguard)’s younger, faster, and slightly sketchier brother who routinely ingests performance-enhancing anabolic nootropics.

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/base.py
+++ b/ansible_base/authentication/authenticator_plugins/base.py
@@ -65,7 +65,7 @@ class AbstractAuthenticatorPlugin:
         else:
             self.logger = logger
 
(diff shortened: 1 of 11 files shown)


AlanCoding:
Checks are actually failing but are incorrectly passing.

Status of typeguard PR is much better as of now.

PR #192, opened 2024-03-02
Fix tests passing when they should be failing

AlanCoding:
Look at some recent checks:

```
---------- coverage: platform linux, python 3.9.18-final-0 -----------
Coverage HTML written to dir htmlcov
Coverage XML written to file coverage.xml
Coverage JSON written to file coverage.json

=========================== short test summary info ============================
FAILED test_app/tests/authentication/authenticator_plugins/test_ldap.py::test_ldap_backend_authenticate_valid_user
FAILED test_app/tests/authentication/authenticator_plugins/test_ldap.py::test_ldap_backend_authenticate_unbind_exception
FAILED test_app/tests/resource_registry/test_resource_types_api.py::test_resource_type_manifest
FAILED test_app/tests/lib/utils/test_hashing.py::test_hash_serializer_data_idempotency
FAILED test_app/tests/lib/utils/test_hashing.py::test_hash_serializer_data_difference
FAILED test_app/tests/lib/utils/test_hashing.py::test_hash_serializer_with_nested_field
FAILED test_app/tests/lib/utils/test_response.py::test_csv_stream_response - ...
FAILED test_app/tests/lib/utils/test_validation.py::test_validate_bad_urls[False-4-schemes0-True]
FAILED test_app/tests/lib/utils/test_validation.py::test_validate_bad_urls[False-None-schemes7-True]
= 9 failed, 656 passed, 11 xfailed, 3 xpassed, 36 warnings in 61.15s (0:01:01) =
echo "Killing dab_postgres container"
Killing dab_postgres container
docker kill dab_postgres
dab_postgres
  py39: OK (125.81=setup[55.92]+cmd[69.89] seconds)
  congratulations :) (125.93 seconds)
```

They failed. But OK. This is actually not OK. It looks like it's considered a success because we're running multiple commands and the last command (stop-postgres) was a success. So let's put a _stop_ to that.

PR #193, opened 2024-03-03
Add test to demonstrate DO_NOTHING bug

AlanCoding:
I described this in:

https://github.com/ansible/django-ansible-base/pull/160

I wrote this first without the `check_constaints` call, and it didn't give me the error I suspected, and instead gave this:

> ValueError: User object can't be deleted because its id attribute is set to None.

However, if you remove the `yield` and `.delete` stuff from the fixtures (in https://github.com/ansible/django-ansible-base/pull/161, please merge that), you get this same error. I figure that, in any case, it will be better to show the error inside of the test itself.

```
E               django.db.utils.IntegrityError: insert or update on table "test_app_organization" violates foreign key constraint "test_app_organizatio_created_by_id_c5ac985f_fk_test_app_"
E               DETAIL:  Key (created_by_id)=(15) is not present in table "test_app_user".
```

I have not written this up in Github yet. I will allow the CI to confirm the result first. You will not see that in the result _status_ at first, because of https://github.com/ansible/django-ansible-base/pull/192, which also very much needs merging.


AlanCoding:
Tests report both the intended failure and teardown error. Not actually successful.

AlanCoding:
Still can't be merge-ready until https://github.com/ansible/django-ansible-base/pull/161 merges.

Dostonbek1 (approved):

PR #194, opened 2024-03-04
Add additional capabilities required for service migration

newswangerd:
This PR adds:
- More fine grained permissions on the resources API.
- `service-index/validate-local-user/` endpoint for checking a user's existing credentials.
- `service-index/resources/<id>/additional_data/` endpoint for retrieving additional data about a resource.
- Resource processors to allow for services to customize how shared data is serialized and de serialized.
- Set defaults for pagination and filtering so the API behaves the same everywhere.


ansible_base/resource_registry/views.py
```diff
@@ -95,8 +112,34 @@ def manifest(self, request, name, *args, **kwargs):
...
+        HasResourceRegistryPermissions,
+    ]
+    action = "retrieve"
 
     def get(self, request, **kwargs):
         registry = get_registry()
         return Response({"service_id": service_id(), "service_type": registry.api_config.service_type})
+
+
+class ValidateLocalUserView(AnsibleBaseDjangoAppApiView):
```
AlanCoding, line 172: I would suggest adding a long docstring here, because that should get rendered in the Browsable API, and make its way into swagger and stuff. Because I have a hard time figuring out what the function of this endpoint is.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -56,6 +56,8 @@ def parse_jwt_token(self, request):
             },
         )
 
+        setattr(user, "resource_api_actions", validated_body["resource_api_actions"])
```
AlanCoding: I don't know where this comes from or what content it might have, but I filed AAP-19984 a while ago. Have a look at that and let me know if it's hitting somewhere in the neighborhood of your plan.

jctanner (approved):

PR #199, opened 2024-03-05
Add organization field to teams shared serializer.

newswangerd:
```

{
    "object_id": "1",
    "name": "test",
    "ansible_id": "791b8558-001b-487e-a4ea-e7efc718c57f",
    "service_id": "9f1c372d-ed8c-4561-856e-31b888942daf",
    "resource_type": "shared.team",
    "has_serializer": true,
    "resource_data": {
        "name": "test",
        "organization": "37f44a39-2045-4e9b-a1dd-34a560218254"
    },
    "detail_url": "/api/gateway/v1/teams/1/",
    "url": "/api/gateway/v1/service-index/resources/791b8558-001b-487e-a4ea-e7efc718c57f/"
}
```

The organization field takes the ansible id for an organization object and can be used to update a Team's organization.

Right now the "organization" field is nullable in order to keep hub from breaking until we can get proper teams.

### DIFF
--- a/test_app/tests/resource_registry/test_resources_api.py
+++ b/test_app/tests/resource_registry/test_resources_api.py
@@ -5,7 +5,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.urls import reverse
 
 from ansible_base.resource_registry.models import Resource
-from test_app.models import EncryptionModel, Organization, Team
+from test_app.models import EncryptionModel
 
 
 def test_service_index_root(user_api_client):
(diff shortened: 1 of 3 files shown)


test_app/tests/resource_registry/test_resources_api.py
```diff
@@ -287,3 +281,32 @@ def test_resource_summary_fields(
...
+    # Test updating the organization field
+    data = {"resource_data": {"organization": org1_id}}
+    resp = admin_api_client.patch(url, data, format="json")
+    assert resp.status_code == 200
+    assert resp.data["resource_data"]["organization"] == org1_id
+
+    team.refresh_from_db()
+    assert team.organization == organization_1
+
+    # Test invalid organization ID
```
AlanCoding, line 317: might as well be a separate test

AlanCoding (approved):
This does introduce a `Resource` dependency, if someone were to GET the resources list to slurp up the full data list. At this point, that should still be expressible in an ordering of types. As long as teams go last, the dependency is fulfilled. Immediately, that puts some constraints on that later part of implementation. In the future, if more than 1 type has a dependency, the API will need some mechanism smarter than `order_by` querystring. Just processing out loud...

PR #200, opened 2024-03-05
Only run postgres if its not running

john-westcott-iv:
If you run `tox` from command line it will start up a postgres container.
If you attempt to run it a second time it will fail because postgres is already running. 
This change will only attempt to start up the container if its not already running allowing you to run tox back to back.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,9 +51,9 @@ check_flake8:
 check_isort:
 	tox -e isort -- --check $(CHECK_SYNTAX_FILES)
 
-## Starts a postgres container in the background
+## Starts a postgres container in the background if one is not running
 postgres:
-	docker run -d --rm --name dab_postgres -p 55432:5432 -e POSTGRES_USER=dab -e POSTGRES_PASSWORD=dabing -e POSTGRES_DB=dab_db postgres:13
+	docker start dab_postgres || docker run -d --rm --name dab_postgres -p 55432:5432 -e POSTGRES_USER=dab -e POSTGRES_PASSWORD=dabing -e POSTGRES_DB=dab_db postgres:13
 
 ## Stops the postgres container started with 'make postgres'
 stop-postgres:


relrod:
Refs #192.

I still feel like - if tox starts postgres - tox should be the thing to stop postgres at the end.

I don't want "I ran these tests" to modify my system state by the end of them. That feels really wrong to me.

So: I'm okay with this change you're making so that `make postgres` never fails if you run it twice. But I _also_ would like to see a partial revert of #192.

I'd love to see something like this go along with it:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index db368113..fe6d974d 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ legacy_tox_ini = """
         -r{toxinidir}/requirements/requirements_all.txt
         -r{toxinidir}/requirements/requirements_dev.txt
     allowlist_externals = sh
-    commands = sh -c 'make postgres && pytest -n auto --cov=. --cov-report=xml:coverage.xml --cov-report=html --cov-report=json --cov-branch {env:ANSIBLE_BASE_TEST_DIRS:test_app/tests} {posargs}; make stop-postgres'
+    commands = sh -c 'make postgres && pytest -n auto --cov=. --cov-report=xml:coverage.xml --cov-report=html --cov-report=json --cov-branch {env:ANSIBLE_BASE_TEST_DIRS:test_app/tests} {posargs}; test_rc=$?; make stop-postgres; exit $test_rc'
 
     [testenv:check]
     deps =
```

Or maybe we need something smarter that tells (via return code or something) if it actually did start postgres (and if not, we don't stop it, we keep it running).

AlanCoding (approved):
Maybe I shouldn't approve, all I mean to say is that I have no preference at this point. While tox's objective is reproducibility, CI runs in a clean VM every time. For the automated checks it just doesn't matter.

For local development, yeah this is clearly better. What I don't understand is how `docker start dab_postgres` is a thing at all, because we use the `--rm` flag that should destroy it when it starts. So how can it be stopped when it's destroy-on-stop? Riddle me that.

Anecdotally, I think I've seen similar (supposedly impossible) behavior with the `bootstrap.sh` script. This leads it to go into the polling loop until timeout, because the container exists, but doesn't process requests. Again, how exactly? If I don't think too hard, this can be a solution for that, and we can copy it over to the `boostrap.sh` script as well eventually when it annoys someone enough.

john-westcott-iv:
> Or maybe we need something smarter that tells (via return code or something) if it actually did start postgres (and if not, we don't stop it, we keep it running).

This feel like Anisble to me :)

john-westcott-iv:
> So how can it be stopped when it's destroy-on-stop?

In the current behavior nothing is ever stopping it after a tox run. So the `docker start dab_postgres` is just picking up that its already running and returns true preventing the or `||` from trying to run it again. If the start fails (because the container was stopped and destroyed) that fails and the second half of the`||` kicks in to run it again.

relrod:
> > Or maybe we need something smarter that tells (via return code or something) if it actually did start postgres (and if not, we don't stop it, we keep it running).
> 
> This feel like Anisble to me :)

But if we have an ansible playbook that does: start postgres (if needed), run tests, stop posgres (if we started it)

then we lose streaming output of pytest, we lose getting any feedback until tests are completed, and all the output is jumbled together in a JSON blob. No thanks :sweat_smile:

jctanner:
@relrod we don't have to accept the loss of streaming output ...

https://github.com/ansible/galaxy_ng/blob/master/dev/playbooks/collections/ansible_collections/galaxy_ng/tools/plugins/action/local_run.py

https://github.com/ansible/galaxy_ng/blob/master/dev/playbooks/run_unit_tests.yaml#L49-L51

PR #201, opened 2024-03-06
Move .vscode to tools/vscode

fosterseth:
Developers may wish to have run their own `.vscode/` configurations. I think it is better _not_ to source control those files.

Instead, just create some good working files in `tools/vscode/`, and provide developers instructions on how to use them.

That way users can tweak the .vscode configurations to their liking without those files being stomped on by source control.


vscode/README.md
```diff
@@ -0,0 +1,32 @@
...
+### Running Tests
+
+the `vscode/settings.json` test file allows you to run a test via VSCode.
+
+Click the Testing icon and navigate to the test you wish to run. Click either Run or Debug test.
+
+
+### Resetting database
+
+Delete the `django-ansible-base/db.sqlite3` file and re-run the server to get a fresh database.
```
john-westcott-iv: Is this still true? I thought we converted test_app tests to run on postgres by default.

john-westcott-iv (commented):
I'm not in love with `/vscode`. It might also be nice to have the README.md in the `docs` folder and maybe the code bits in `tools/vscode` with a README.md in there pointing to the one in the docs folder?

vscode/README.md
fosterseth: yeah I added it back as a separate settings file, test_app/sqlite3settings.py. It's not used except for vscode stuff

fosterseth:
`.vscode` is now in `tools/vscode`

john-westcott-iv (approved):

PR #202, opened 2024-03-06
A few misc fixes to support resource migration

newswangerd:


john-westcott-iv:
I'm not sure why Sonar didn't fail because the threshold should be 80% for coverage. Can you add additional coverage?

ansible_base/resource_registry/rest_client.py
```diff
@@ -17,7 +17,7 @@ class ResourceAPIClient:
     Client for Ansible services to interact with the service-index/ api
     """
 
-    def __init__(self, service_url: str, service_path: str, requests_auth_kwargs: dict, verify_https=True):
+    def __init__(self, service_url: str, service_path: str, requests_auth_kwargs: dict, verify_https=True, raise_if_bad_request: bool = False):
```
AlanCoding, line 20: Note: This flag is not ever set in this PR code. It is planned to be used by other code using this client, but is introduced here.

ansible_base/resource_registry/serializers.py
```diff
@@ -98,9 +102,12 @@ def create(self, validated_data):
             if not validated_data["resource_type"]:
                 raise serializers.ValidationError({"resource_type": _("This field is required for resource creation.")})
 
-            r_type = ResourceType.objects.get(name=validated_data["resource_type"])
+            resource_type = ResourceType.objects.get(name=validated_data["resource_type"])
+            if not resource_type.can_be_managed:
+                raise serializers.ValidationError({"resource_type": _(f"Resource type: {resource_type.name} cannot be managed by Resources.")})
```
AlanCoding, line 107: The other change here (out of essentially 2 major changes) is to move the "managed" validation from the models to the serializer. Is that a correct summary?

I assume so that some internal thing can modify stuff.

ansible_base/resource_registry/rest_client.py
newswangerd: It is used in this pr:

```
        resp = requests.request(**kwargs)
        if self.raise_if_bad_request:
            resp.raise_for_status()
        return resp
```

ansible_base/resource_registry/serializers.py
newswangerd: Yes. We need to be able to create new resources internally from requests to the resources api

ansible_base/resource_registry/rest_client.py
AlanCoding: That's a part of the implementation, I'm saying that nothing is using this in DAB. It's an added parameter of the client class for making requests. Nothing makes a request in DAB with that. Only something external is passing that and setting to the non-default value of `True`. It would be good to have a DAB test for this, I assume this is what the live_server type of tests were created for.

ansible_base/resource_registry/serializers.py
AlanCoding: Create managed resources from requests to the resources api? But the resource serializer prohibits those requests, so I hope you can see how it doesn't completely add up.

newswangerd: The gateway needs to request resource information from the resources API on controller and then create a copy of that resource internally using the Resource model.

We actually may want to get rid of this flag entirely and just control this with a flag on the jwt.

AlanCoding (approved):
Adds up with latest explanation. You have server A and server B. Server A is the source of truth, and requests data from server B for the initial migration. (normally server B will pull from server A) and then server A will make whatever decisions it needs, and a part of that will involve updating its own resources. Looks specific to migration scenario, but certainly needed for that.

PR #203, opened 2024-03-07
Handle no authenticator scenario when looking for latest timestamps

john-westcott-iv:
If we had no authenticators (which is valid for some tests) and attempted to call the authenticate method we would get a stack when looking for the latest timestamp of all authenticators. This change will pass the timestamp or None (if there are no authenticators) as the cache parameter value.

### DIFF
--- a/ansible_base/authentication/backend.py
+++ b/ansible_base/authentication/backend.py
@@ -28,7 +28,11 @@ def get_authentication_backends(last_updated):
 class AnsibleBaseAuth(ModelBackend):
     def authenticate(self, request, *args, **kwargs):
         logger.debug("Starting AnsibleBaseAuth authentication")
-        last_modified = Authenticator.objects.values("modified_on").order_by("-modified_on").first()["modified_on"]
+
+        # Query the database for the most recently last modified timestamp.
+        # This will be used as a cache key for the cached function get_authentication_backends below
+        last_modified_item = Authenticator.objects.values("modified_on").order_by("-modified_on").first()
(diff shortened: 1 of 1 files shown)


relrod (approved):

relrod (changes requested):
nevermind, CI is failing, I take it all back

ansible_base/authentication/backend.py
```diff
@@ -28,7 +28,11 @@ def get_authentication_backends(last_updated):
 class AnsibleBaseAuth(ModelBackend):
     def authenticate(self, request, *args, **kwargs):
         logger.debug("Starting AnsibleBaseAuth authentication")
-        last_modified = Authenticator.objects.values("modified_on").order_by("-modified_on").first()["modified_on"]
+
+        # Query the database for the most recently last modified timestamp.
+        # This will be used as a cache key for the cached function get_authentication_backends below
+        last_modified_item = Authenticator.objects.values("modified_on").order_by("-modified_on").first()
+        last_modified = None if last_modified_item is None else last_modified_item.get('modified_on', None)
```
relrod:
```suggestion
        last_modified = None if last_modified_item is None else last_modified_item.get('modified_on')
```

You don't need the `None` here, it's the default.

relrod (approved):

ansible_base/authentication/backend.py
john-westcott-iv: Fair enough, fix pushed.

PR #204, opened 2024-03-07
Adding a configurable timeout session store

john-westcott-iv:
Add a custom cached session store that accepts a dynamic timeout.

### DIFF
--- a/ansible_base/lib/sessions/stores/cached_dynamic_timeout.py
+++ b/ansible_base/lib/sessions/stores/cached_dynamic_timeout.py
@@ -15,6 +15,9 @@ class SessionStore(CachedDBSessionStore):
     def get_session_cookie_age(self):
         timeout = get_setting('SESSION_COOKIE_AGE', DEFAULT_SESSION_TIMEOUT)
         if not isinstance(timeout, int):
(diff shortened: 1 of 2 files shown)


ansible_base/lib/sessions/stores/cached_dynamic_timeout.py
```diff
@@ -0,0 +1,20 @@
...
+logger = logging.getLogger('ansible_base.lib.sessions.stores.cached_dynamic_timeout')
+
+
+class SessionStore(CachedDBSessionStore):
+    cache_key_prefix = 'ansible_base.lib.sessions.stores.cached_dynamic_timeout'
+
+    def get_session_cookie_age(self):
+        timeout = get_setting('SESSION_COOKIE_AGE', DEFAULT_SESSION_TIMEOUT)
+        if not isinstance(timeout, int):
+            logger.error(_('SESSION_COOKIE_AGE was set to {} which is an invalid int, defaulting to {}'.format(timeout, DEFAULT_SESSION_TIMEOUT)))
```
relrod: Let's avoid hardcoding the variable values in the translation string.

Something like:

```suggestion
            logger.error(_('SESSION_COOKIE_AGE was set to %(timeout)s which is an invalid int, defaulting to %(default)s') % {'timeout': timeout, 'default': DEFAULT_SESSION_TIMEOUT})
```

Otherwise, the fully compiled string is what will get translated, and any time these values change, the string won't have a translation.

relrod (approved):

PR #205, opened 2024-03-08
Add rest_pagination app

john-westcott-iv:
Implement something similar to the AWX rest pagnation. 

One key difference is that the AWX pagination appears to allow a user to request more items than MAX_PAGE_SIZE. This version respects MAX_PAGE_SIZE as the absolute maximum number of object the system will return and uses a DEFAULT_PAGE_SIZE for the default number (if page_size is unspecified).


relrod:
I think it would be better to just use https://www.django-rest-framework.org/api-guide/pagination/#limitoffsetpagination instead of reinventing the wheel here.

This seems to support everything we need (default/max limits, and custom names for the query params for limit/offset).

john-westcott-iv:
Superseded by https://github.com/ansible/django-ansible-base/pull/217

PR #207, opened 2024-03-11
Add google oauth2 plugin.

jctanner:


ansible_base/authentication/authenticator_plugins/google_oauth2.py
```diff
@@ -0,0 +1,116 @@
...
+
+    SCOPE = ListField(
+        help_text=_('The authorization scope for users. Defaults to ["openid", "email", "profile"].'),
+        required=False,
+        allow_null=False,
+        ui_field_label=_('List of OAuth2 Scope(s)'),
+        default=["openid", "email", "profile"],
+    )
+
+    EXTRA_DATA = ListField(
```
john-westcott-iv: Should we leverage the EXTRA_UNVERIFIED_FIELDS for this or is it different?

john-westcott-iv (approved):

ansible_base/authentication/authenticator_plugins/google_oauth2.py
jctanner: What I find confusing is that in AWX it is called `EXTRA_ARGUMENTS` https://github.com/ansible/awx/blob/b076cb00a968f2fe5b65cbe1ad8b328fd03b030c/awx/sso/conf.py#L619-L633 but in the backend code it's called `EXTRA_DATA` https://github.com/python-social-auth/social-core/blob/master/social_core/backends/google.py#L73-L77

jctanner: The google oauth2 backend subclasses the base oauth2 backend where this comment about `EXTRA_DATA` resides ...

https://github.com/python-social-auth/social-core/blob/master/social_core/backends/oauth.py#L30-L32
```
    Settings will be inspected to get more values names that should be
    stored on extra_data field. The setting name is created following the
    pattern SOCIAL_AUTH_<uppercase current backend name>_EXTRA_DATA.
```

john-westcott-iv: There is an `AUTH_EXTRA_ARGUMENTS` option that `BaseAuth` appears to do something with https://github.com/python-social-auth/social-core/blob/master/social_core/backends/base.py#L213C41-L213C61
I'd have to dig further to confirm but I think that the way the settings work in social-core `SOCIAL_AUTH_GOOGLE_OAUTH2_AUTH_EXTRA_ARGUMENTS` would become `AUTH_EXTRA_ARGUMENTS`.

The `EXTRA_DATA` code path is BaseAuth defiantly appears to do something very different than `AUTH_EXTRA_ARGUMENTS` though. It could be a bug in AWX.

PR #208, opened 2024-03-11
Add azuread auth plugin and tests.

jctanner:
No-Issue


john-westcott-iv (approved):

djyasin (commented):
@jctanner The only change I would make here would be to revise line 34 to `"'The OIDC secret (Client Secret Value) from your IDP."`

I will also see about putting up a PR to revise the auth readme a bit.

PR #209, opened 2024-03-12
Remove 'on' from [created|modified]_on

john-westcott-iv:
This alters the created_on/modified_on fields to be inline with AWX and just called created and modified.
Additionally we have moved `created` and `modified` to be `auto_now[_add]` fields and removed the logic from `CommonModel.save()` to set those fields.


ansible_base/authentication/management/commands/authenticators.py
```diff
@@ -73,9 +73,9 @@ def initialize_authenticators(self):
                 create_objects=True,
                 configuration={},
                 created_by=admin_user,
-                created_on=now(),
```
cutwater, line 76: Why `created` and `modified` are passed explicitly here? Wouldn't it make sense to remove it?

cutwater: Irrelevant to this PR:

```
self.stderr.write("Error message")
exit(255)
```

can be replaced with the `CommandError` exception, provided by Django

```
raise CommandError("Error message")
```

test_app/migrations/0004_remove_encryptionmodel_created_on_and_more.py
```diff
@@ -0,0 +1,109 @@
...
+        migrations.AddField(
+            model_name='user',
+            name='created',
+            field=models.DateTimeField(auto_now_add=True, default=django.utils.timezone.now, help_text='The date/time this resource was created'),
+            preserve_default=False,
+        ),
+        migrations.AddField(
+            model_name='user',
+            name='modified',
+            field=models.DateTimeField(auto_now=True, help_text='The date/time this resource was created'),
```
AlanCoding: I would, personally, find this change easier to manage if just thrown in the 0001 migration. After this merges, I have to rebase the RBAC PR and do the replacements there. By having this migration, I have to tease out what fields in the 0001 migration are _supposed_ to be `created_on` to later be changed, and which are new models and need to have the new fields.

cutwater: I don't think this is a good idea, since other projects like EDA and Hub are relying on DAB already. It will make migration painful and has to be coordinated with projects, that depend on DAB.

john-westcott-iv: Product wise if AWX and EDA have not yet shipped with any models that would be affected by this it would be fine to merge into 0001. But this would also require any dev environments to roll back to zero and then roll forward or their environments would be "stuck", right? If those statements are both accurate, are we ok with that?

ansible_base/authentication/management/commands/authenticators.py
john-westcott-iv: Good catch, I've updated this. r.e. the CommandError I filed a new issue https://github.com/ansible/django-ansible-base/issues/211.

test_app/migrations/0004_remove_encryptionmodel_created_on_and_more.py
AlanCoding: This is test_app. My problems resolving conflicts on rebase are 80% in test_app.

You could repeat this discussion for the authenticator app, but that doesn't affect my own workload.

It would also strike me as poor form to have an initial release of a server product containing a dozen or more migrations, but that's also not immediately relevant to me right now.

john-westcott-iv: Thats fair for test app, its just a playground. Should we squash all of the migrations or just merge this new one into the older ones?

AlanCoding: the 0002 migration in test_app has a technical reason for its existence. The rest do not.

john-westcott-iv: Oh, thats interesting, IDK if you can "squash with skip". I'll probably have to manually merge them.

AlanCoding (approved):

PR #45, opened 2023-12-08
RBAC system with custom roles integrated with AWX

AlanCoding:
This rewrites the roles system inside of AWX and moves them here, to this library, so that ideally we can connect it to other apps. I want to be very clear that this is mainly an **object-role** system. I added functionality to handle system-wide permissions, but this is a different sub-system.

AWX patch to use this - https://github.com/ansible/awx/pull/14735

The only reason this is possible at all is because most of the AWX gnarly permission logic exists on a layer of business logic fairly well-separated from the internals of the RBAC system. It's mainly the internals that this offers (although it tacked on a simple API), for object-roles - roles to things. Those things can be in a shallow hierarchy... which is just organizations right now. Inheritance of permissions from more than 1 parent resource has some basic tests, but isn't expected to be used yet.

Right now this is fully functional in AWX, with only 2-3 known integration test failures due to sacrifices that were knowingly made.

TODO:

 - [x] Add views for this new system. Current blocker is that basic details of the API contract are not be agreed on yet. The AWX branch offers a compatibility layer with the old roles system.
 - [x] Add `content_type` to `RoleDefinition`. This will be used for validation of permissions, because listing some permissions require listing other permissions, and most permissions are only valid for a certain model (or a parent of that model).
 - [x] Fix unit tests in AWX. The compatibility layer currently doesn't do this for us.
 - [x] Reconcile the team & organization `members` relationship that already exists in Gateway compatible with this. The role system, today, handles this by putting users in a pre-created team member role. If the role is the source of truth, then we need to add some ORM shortcuts to more easily access it.
 - ~Very large item - migrate galaxy_ng to this system like the AWX migration. This would have to go along with the introduction of an organization model, without organizations this entire system is useless. Ping @cutwater~
 - [x] ~Add to the AWX migration - split the organization admin role into 2 with 1 dealing with orgs / teams / user management, and the other dealing with AWX objects alone. This role will then be migrated into Gateway itself (not ansible_base). Oh yeah, and also convert Gateway to this system... I don't think it has a permissions system right now.~ Not doing this
 - [x] Can we give this a name? DAB RBAC

PR #212, opened 2024-03-12
Set filter backends at view level for DAB views

AlanCoding:
Explanation:

The UI was trying to use endpoints added for RBAC https://github.com/ansible/django-ansible-base/pull/45 in eda-server. The common filtering via querystrings like `?content_type__model=inventory` was not working.

eda-server has not adopted the `ansible_base.rest_filters` app. It is unclear if it ever will, or if anyone is willing to do the work to make that happen.

The generic `ModelViewSet` sets `filter_backends` to the default value from settings, even if that's an empty list. That's why the parent classes have to be reversed, so the Ansible view preference will take effect.

Otherwise, the app's defaults for filter classes will take effect. In the case of eda-server, this is nothing (again, empty list). The app does set `filter_backends` specific to many views, so filtering has to be enabled item-by-item, which is more of a standard DRF pattern as opposed to the AWX / DAB filter class of allow-all-filters-unless-disallowed.

This is a very philosophical change. Is it more important that views defined here fit in with the DAB family, or should they inherit the traits from the app using them? The UI and other development would find it difficult to accommodate the later, so this institutes the former.

Internal: AAP-21654


ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -26,6 +26,21 @@
     REST_FRAMEWORK = {}
 
 
+# Explanation - these are the filters for views provided by DRF like /authenticators/
```
john-westcott-iv: Did you mean DAB instead of DRF?

ansible_base/lib/utils/views/django_app_api.py
```diff
@@ -28,5 +29,15 @@
...
 
 
+# In case an app is using features like authentication, resources, or RBAC
+# but not using rest_filters app, we need to specify the same filter backends
+dab_filter_backends = list(api_settings.DEFAULT_FILTER_BACKENDS)
+
+for backend_name in settings.ANSIBLE_BASE_CUSTOM_VIEW_FILTERS:
+    backend = import_from_string(backend_name, 'ANSIBLE_BASE_CUSTOM_VIEW_FILTERS')
+    if backend not in dab_filter_backends:
+        dab_filter_backends.append(backend)
```
john-westcott-iv: Since these are in settings and unlikely to change w/o a reboot should we cache this somehow (maybe lru) so we aren't computing this over and over for every view?

john-westcott-iv: Also, the way this is implemented, if we do have `ansible_base.rest_filters` enabled this logic would still kick in right? That seems undesirable.

AlanCoding: > Since these are in settings and unlikely to change w/o a reboot should we cache this somehow (maybe lru) so we aren't computing this over and over for every view?

This should only get ran on import, and only once, with the way python imports work. Yes, _most_ views should inherit from this class, but the type definition of this class should only happen once.

AlanCoding: I previously pushed some logic to address the concern I think you expressed, so I just now pushed a test for that as well. We have a condition `if backend not in dab_filter_backends:`, so if the "rest_filters" includes an entry via settings (or the user does this) then it will not duplicate that entry in the list. What we get is a merged list of the backends from settings and the declared backends we want for our own DAB views via `ANSIBLE_BASE_CUSTOM_VIEW_FILTERS`. Let me know if there are any other questions, I think the concerns here have all been addressed.

AlanCoding:
REBASE NOTE: as https://github.com/ansible/django-ansible-base/pull/45 merges, on rebase, we need to delete its solution in this PR.

ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -26,6 +26,21 @@
     REST_FRAMEWORK = {}
 
 
+# Explanation - these are the filters for views provided by DAB like /authenticators/
+# we want them to be enabled by default _even if_ the rest_filters app is not used
+# so that clients have consistency, but if an app wants to turn them off, they can
+# these will be combined with the actual DRF defaults in our base view
+try:
+    ANSIBLE_BASE_CUSTOM_VIEW_FILTERS
```
john-westcott-iv: How often will we "pay the price" for the loop (which is probablly insignificant to begin with). Just thinking should we merge this somehow with the if condition down on line 51 `if 'ansible_base.rest_filters' in INSTALLED_APPS:` that we set ANSIBLE_BASE_CUSTOM_VIEW_FILTERS to `None` or `()` so that we avoid the loop altogether on services that use the rest_filters (which would automatically add the class). Or we could wrap the for loop in the if. I will still approve as I don't think this is blocking.

john-westcott-iv (approved):

ansible_base/lib/dynamic_config/dynamic_settings.py
AlanCoding: Valid comment from a DRY perspective (not really a performance perspective). I'll push a commit in a bit. This will "avoid the loop" on services that use rest_filters, like you suggest. It also avoids catching `NameError`, which is generally weird. It'll set `ANSIBLE_BASE_CUSTOM_VIEW_FILTERS` in the `else:` of `if 'ansible_base.rest_filters' in INSTALLED_APPS:` which I agree is the intuitive expectation. Instead of `NameError`, this will use `getattr(settings, 'ANSIBLE_BASE_CUSTOM_VIEW_FILTERS', ())`

PR #215, opened 2024-03-13
Fix issues where admin client fixture changed user client fixture

AlanCoding:
The `client.login()` makes any other fixtures misbehave. What we had only worked if you only used 1 of these fixtures in a single test. There are many case where I need to use several of these simultaneously.


relrod (approved):
You know you're on to something when you write tests to test your tests.

AlanCoding:
I'd argue that testing `test_app/management/commands/create_demo_data.py` is actually _more_ pedantic than this, because it's not in the DAB package. @john-westcott-iv

PR #216, opened 2024-03-13
Convert to run on postgres:15

AlanCoding:


AlanCoding:
@relrod the tox config itself uses `make postgres`, so if this passes, we should be good, right?

jctanner (approved):

PR #218, opened 2024-03-14
ImmutableModel mixin

relrod:
`ImmutableModel` will:

- Prevent a model from being save()'d if it already has a pk.
- Also remove modified_on/modified_by fields by setting them to None in case we are inheriting from CommonModel.

This is because we have some models (such as activity stream entries and some RBAC models) that should never change because they are used as a form of audit log. This prevents them from changing preventing save() from working after there's a PK on the object.

It also insures that `ImmutableModel` is the first base class so MRO is satisfied.

cc @AlanCoding because maybe it's worth squeezing this into #45


AlanCoding:
Cross-referencing with https://github.com/ansible/django-ansible-base/pull/45, I want to volunteer the fact that there is _one_ model definition I could apply this to. That is the abstract base model for user & team role assignments. Your class could be used there, because this model inherits from `CommonModel`. The two other cases where I have an immutable model, ObjectRole and RoleEvaluation, the model is not a common model to begin with. By that, I mean it doesn't have a `created_by` field and they are very performance-sensitive. Assignments are created by the user and should be expected to have all the standard API fields.

That's all fine and all, I just want to say that the overlap exists, but is fairly small between these two things.

AlanCoding:
Oh, or I guess this could be used as the first class for a model using the basic Django `Model`? That maybe I could use it in those other cases too?

AlanCoding:
^ assuming that works as expected (I don't think it is covered by tests here), then I think it would work well with the RBAC models. But right now I really want to get that branch merged ASAP and am looking for reviews on it. At this point I would prefer to merge RBAC and then rebase & apply this to those models.

john-westcott-iv (commented):
It would be nice if this was in its own file and the `super().save()` will append `modified_on/by` to the model as instance variables so we may to clear those when coming back from the save.

ansible_base/lib/abstract_models/common.py
```diff
@@ -205,3 +206,30 @@ def summary_fields(self):
...
+        if self.pk:
+            raise ValueError(f"{self.__class__.__name__} is immutable and cannot be modified.")
+
+        return super().save(*args, **kwargs)
+
+    def __getattribute__(self, name):
+        # This is a bit of a hack, there is no good way to remove these fields, because they exist up the chain
+        # if the model inherits from CommonModel.
+        if name in ('modified', 'modified_by'):
+            raise FieldDoesNotExist(f"{self.__class__.__name__} has no attribute {name}")
```
AlanCoding: I don't follow why you need to do this in the first place, and then I don't understand why you would want this exception in the case it happens. The normal behavior is `AttributeError`

```
In [2]: JobTemplate.objects.first().foobar
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 JobTemplate.objects.first().foobar

AttributeError: 'JobTemplate' object has no attribute 'foobar'
```

And I don't know how we wouldn't get this for "free" by doing nothing.

AlanCoding:
@john-westcott-iv 's comments about the `.save` method remind me of things which are very concerning.

In my 1 model that does something similar, I do:

```python
    def __init__(self, *args, **kwargs):
        """
        Because through models are created via a bulk_create, the save method is usually not called
        to get around this, we populate the user model after initialization
        """
        if not self.id:
            user = get_current_user()
            if user:
                # Hazard: user can be a SimpleLazyObject, so use id
                self.created_by_id = user.id
        # Cache fields from the associated object_role
        if self.object_role_id and not self.object_id:
            self.object_id = self.object_role.object_id
            self.content_type_id = self.object_role.content_type_id
            self.role_definition_id = self.object_role.role_definition_id

    def save(self, *args, **kwargs):
        if self.id:
            raise RuntimeError(f'{self._meta.verbose_name.title()} model is immutable, use RoleDefinition.give_permission method')
        # skip over CommonModel save because it would error due to missing modified_by and created
        return super(CommonModel, self).save(*args, **kwargs)
```

Note the `super(CommonModel, self)`, which I've said before is, unfortunately, intended. This model bypasses all of the save logic on the `CommonModel` because it was a nightmare to work with. It's going to set `modified_by`. I don't understand what the strategy of the current PR for that is.

https://github.com/ansible/django-ansible-base/blob/4429801ed2000843ff215eba94545738f8b97f54/ansible_base/lib/abstract_models/common.py#L98

Because of this line, I expected at least some change to `CommonModel.save`, if not a complete restructuring of the class inheritance.

ansible_base/lib/abstract_models/common.py
relrod: I spent a bunch of time on this. We don't get it for "free" for a few reasons.

`CommonModel` will set those fields. In the normal case, that means those fields get updated and when you save, they'll get written to the database. But since in this case they aren't fields (they are `None`), they'll end up just getting set as regular instance variables.

Why is this a problem? Well, because it's confusing if you `dir()` an instance of `ImmutableModel, CommonModel` after you `save()` it, and see those fields. And it's even more confusing if they contain an actual value.

So, then why can't we just eliminate the fields? Well, largely because they are class variables.

There is interesting play here:

```python
In [1]: class Foo():
   ...:     x = 39
   ...:

In [2]: foo = Foo()

In [3]: foo.x
Out[3]: 39

In [4]: hasattr(foo, 'x')
Out[4]: True

In [5]: delattr(foo, 'x')
AttributeError                            Traceback (most recent call last)
Cell In[5], line 1
----> 1 delattr(foo, 'x')

AttributeError: 'Foo' object has no attribute 'x'
```

Okay, fair enough, you have to delete them from the class.

```python
In [6]: hasattr(Foo, 'x')
Out[6]: True

In [7]: delattr(Foo, 'x')

In [8]: foo.x
AttributeError                            Traceback (most recent call last)
Cell In[8], line 1
----> 1 foo.x

AttributeError: 'Foo' object has no attribute 'x'
```

That kind of works. The fields don't need to exist anymore on `ImmutableModel` classes once the classes are created enough for the Django metaclass logic to run and so on, enough to notice that `modified` and `modified_by` _aren't fields_ and _should not_ be included in migrations.

But wait - if we `delattr()` them from `ImmutableModel` and the model being defined inherits from `ImmutableModel, CommonModel`, then MRO dictates that it will just fall back to using the field from `CommonModel` when you try to access one of those fields. But now you're broken, because the migration said those fields don't exist, so now you call `modified_by` and Django tries to resolve an FK relationship that isn't actually there, so you get an error _deep down_ in the Django code:

```
self = <django.db.models.fields.related.ForeignKey: modified_by>

    def resolve_related_fields(self):
        if not self.from_fields or len(self.from_fields) != len(self.to_fields):
            raise ValueError(
                "Foreign Object from and to fields must be the same non-zero length"
            )
        if isinstance(self.remote_field.model, str):
>           raise ValueError(
                "Related model %r cannot be resolved" % self.remote_field.model
            )
E           ValueError: Related model 'test_app.User' cannot be resolved

.tox/311/lib/python3.11/site-packages/django/db/models/fields/related.py:711: ValueError
```

which is completely reasonable - the relationship just doesn't exist.


As for why `FieldDoesNotExist` vs `AttributeError`, I'm fine with `AttributeError` instead.

The `FieldDoesNotExist` exception came up when I tried to do what @john-westcott-iv originally suggested with deleting the attributes after we call the superclass `save()`. 

```python
def save(self, *args, **kwargs):
    # ...
    ret = super().save(*args, **kwargs)
    delattr(ImmutableModel, 'modified')
    return ret
```

...

```python
In [1]: l = ImmutableLogEntry(message="foo")

In [2]: l.save()

In [3]: l.modified
```

This throws a _crazy long_ stack trace, that ends with:
```
FieldDoesNotExist: ImmutableLogEntry has no field named 'modified'
```

So in this case `CommonModel` (the "other" base class) says "Sure, I have a `modified` field!" but the migration for the model definition says we don't have that field (since `ImmutableModel` nullifies it so Django's metaclass logic never turns it into a field).

`AttributeError` probably does make more sense coming from `__getattribute__()` - I'll change it to that.

I expected this to all be a pretty trivial change, doing it like @john-westcott-iv said, but it actually took me on a journey. In the end I came to the conclusion that the `__getattribute__()` override seems like the simplest way to say "hey, don't use these attributes on an `ImmutableModel` instance. If you have other ideas, I'm of course open to them.

relrod:
> It's going to set `modified_by`

But that's just the thing. It *is* going to set `modified_by`, but the field won't actually exist, so it won't get saved anywhere. Yes, there will be a class variable called `modified_by` and it will be `None`. When `CommonModel#save` does `self.modified_by = ...`, it's ends up just setting the instance variable `modified_by` to some value.

That doesn't throw an error on `save()`.

```
In [1]: x = ImmutableLogEntry(message="hey")

In [2]: x.this_is_not_a_field = 42

In [3]: x.save()

In [4]:
```

We can make `CommonModel#save` check `_meta.fields` to see if `modified_by`/`modified` exist before running that logic at all, as a micro-optimization, but I think presentation wise it wouldn't change anything.

relrod:
> Oh, or I guess this could be used as the first class for a model using the basic Django `Model`? That maybe I could use it in those other cases too?

Yes the idea is that this mixin works with `CommonModel` and without `CommonModel`. The `save()` logic still is relevant in the non-`CommonModel` case.

relrod:
> At this point I would prefer to merge RBAC and then rebase & apply this to those models.

That's fine, there's no real reason to block RBAC on something like this.

ansible_base/lib/abstract_models/common.py
relrod: I've updated it to throw AttributeError now.

AlanCoding:
```python
@pytest.mark.django_db
def test_created_by_immutable_model(user):
    from crum import impersonate

    with impersonate(user):
        instance = ImmutableLogEntry(message="Oh no! An important message!")

    assert instance.created_by == user
```

gives

```
__________________________________________________________________________ test_created_by_immutable_model __________________________________________________________________________

user = <User: user>

    @pytest.mark.django_db
    def test_created_by_immutable_model(user):
        from crum import impersonate
    
        with impersonate(user):
            instance = ImmutableLogEntry(message="Oh no! An important message!")
    
>       assert instance.created_by == user
E       assert None == <User: user>
E        +  where None = <ImmutableLogEntry: ImmutableLogEntry object (None)>.created_by

test_app/tests/lib/abstract_models/test_immutable.py:61: AssertionError
```

My 2 cents here is that if it's possible to avoid doing "fancy" things like `__getattribute__` or `__init_subclass__`, then it would be better to use the simpler approach. It seems like you are not modifying `CommonModel.save`. This problem, philosophically, _can not be solved_ without making changes there. If `self.modified_by` is going to give `AttributeError`, then you have to avoid referencing it in `CommonModel` but still run ordinary logic on `self.created_by`. I don't know if a mixin is a good approach for this. The 2 approaches I would first consider are:

 - a subclass of `CommonModel` that _undoes_ some of what it does for modified/modified_by. Say it doesn't set `self.modified_by` on some `if` check, I don't know what, but it's possible, and likely tolerable.
 - break up `CommonModel` into the created and modified functionality. Those two classes would combine into `CommonModel` but only the "created" class would be used in the immutable models.

relrod:
> E       assert None == <User: user>
> E        +  where None = <ImmutableLogEntry: ImmutableLogEntry object (None)>.created_by

But this happens even before this change.

On devel:

```python
from test_app.models import RelatedFieldsTestModel

@pytest.mark.django_db
def test_created_by_immutable_model(user):
    from crum import impersonate

    with impersonate(user):
        instance = RelatedFieldsTestModel()

    assert instance.created_by == user
```

fails with the same error, because `created_by` only gets set after `.save()` is called.

Your test passes if you `instance.save()` before the assert.

AlanCoding:
Oh, I should have done `objects.create`, yeah

relrod:
> a subclass of CommonModel that undoes some of what it does for modified/modified_by. Say it doesn't set self.modified_by on some if check, I don't know what, but it's possible, and likely tolerable.

If you know of a way to prevent the field from getting created early enough (before Django's metaclass logic runs on the model class), then I'm open to that. I was unable to figure out a way to do that.

> break up CommonModel into the created and modified functionality. Those two classes would combine into CommonModel but only the "created" class would be used in the immutable models.

This could work if we did some refactoring of CommonModel where we abstract out the summary/related field stuff, too. We'd probably need an `AbstractCommonModel` that has that logic, and `CommonModel` inherits that, too.

relrod:
> This could work if we did some refactoring of CommonModel where we abstract out the summary/related field stuff, too. We'd probably need an AbstractCommonModel that has that logic, and CommonModel inherits that, too.

I put up #225 for that, if you want to go that route instead.

AlanCoding:
I think it would be good to add that test that `created_by` gets set correctly after a save. That's pretty central to what this accomplishes. I don't necessarily want to request changes here, I just don't understand it. If setting `modified_by` raises an exception... then how is this working? How does `created_by` get set, when it's set by basically the same mechanism as `modified_by`?

relrod:
I almost prefer the approach in #225 now. Maybe @john-westcott-iv has an opinion.

AlanCoding:
Clarified in call - in this implementation `self.modified_by` _does_ get set. The `AttributeError` only happens on a _getattr_ and not on a _setattr_.

That answers my question above. I sill ask that this adds a test to assert `created_by` is set before merging this or the other PR.

That said, right now I have no particular preference between the two. For this one, I'm not sure if it's _worth it_ to override `__getattribute__` to get the exception. But that's a mild preference.

relrod:
Closing in favor of #225, I think it's better in the long run.

PR #225, opened 2024-03-17
ImmutableCommonModel

relrod:
Alternative to #218 that restructures the `CommonModel` hierarchy to abstract out a `ModifiableModel`, `CreatableModel`, `AbstractCommonModel`.

- `ModifiableModel` adds the modified/modified_by fields and save logic for those
- `CreatableModel` does the same for created/created_by
- `AbstractCommonModel` simply has the logic for dealing with encrypted fields, related fields, summary fields

And then:
- `CommonModel` is a simple class that just inherits from those 3 classes
- `ImmutableCommonModel` inherits from `AbstractCommonModel` and `CreatableModel`


test_app/tests/lib/abstract_models/test_immutable.py
```diff
@@ -0,0 +1,38 @@
+import pytest
+
+from test_app.models import ImmutableLogEntry, ImmutableLogEntryNotCommon
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('model', [ImmutableLogEntry, ImmutableLogEntryNotCommon])
```
AlanCoding, line 8: Nice, I like this `parametrize`

AlanCoding (changes requested):
I want to see a test added that does `impersonate` and asserts the correct value of `created_by`. The new test_app models should work nicely for this test.

AlanCoding:
:+1: I'm good with this approach, either option is fine with me.

ansible_base/lib/abstract_models/common.py
```diff
@@ -36,54 +36,69 @@ def get_url_for_object(obj, request=None):
...
+    def save(self, *args, **kwargs):
+        '''
+        This save function will provide the following features automatically.
+          * It will automatically add the modified fields for changing items
+        '''
+        update_fields = list(kwargs.get('update_fields', []))
+
+        user = get_current_user()
+        if user is None or user.is_anonymous:
+            user = get_system_user()
```
john-westcott-iv: Its almost annoying because its only 3 lines but given the history we had with getting the user I wonder if this should be a function shared between created and modified?

ansible_base/lib/abstract_models/immutable.py
```diff
@@ -0,0 +1,18 @@
+from ansible_base.lib.abstract_models.common import AbstractCommonModel, CreatableModel
+
+
+class ImmutableCommonModel(AbstractCommonModel, CreatableModel):
```
john-westcott-iv: I was the one who asked you to make this its own file but we now have 3 models in common.py and this one on its own. Maybe we should merge this back or split out common more (which feels painful).

john-westcott-iv: The `CommonModel` appended to this throws me off because it makes me think that this is just an Immutable version of CommonModel but it does not inherit from CommonModel at all. But now that I type it out I see where you are going with this that its the `CommonModel` that is Immutable. I'm not sure if there is anything here or just a learning curve for my brain.

test_app/models.py
```diff
@@ -47,3 +46,19 @@ class RelatedFieldsTestModel(CommonModel):
     more_teams = models.ManyToManyField(Team, related_name='related_fields_test_model_more_teams')
 
     ignore_relations = ['teams_with_no_view']
+
+
+class ImmutableLogEntry(ImmutableCommonModel):
+    """
+    Testing ImmutableModel with CommonModel
```
john-westcott-iv: Are these comments out of date? These are both the same thing now, are they not?

ansible_base/lib/abstract_models/common.py
```diff
@@ -93,10 +108,23 @@ def save(self, *args, **kwargs):
                 self.created_by = user
                 update_fields.append('created_by')
 
-        if 'modified_by' not in update_fields:
-            self.modified_by = user
-            update_fields.append('modified_by')
+        return super().save(*args, **kwargs)
+
 
+class AbstractCommonModel(models.Model):
```
john-westcott-iv, line 106: Should we follow through the pattern and put the save into `EncryptedModel`? And then just have the `get_summary_fields` in a `SummaryModel` or something like that?

john-westcott-iv (commented):
I think this method is cleaner and I like it better. Should we consider `ModifiableModel` and `CreatableModel` mixins? If so, maybe all of the mixins go into a `minin.py` and then `CommonModel`, `NamedCommonModel` and `ImmutableCommonModel` go into common.py?

test_app/models.py
relrod: They are and this led me to break out ImmutableCommonModel into an ImmutableModel (because we might want immutable models that don't inherit AbstractCommonModel, etc.)

So I've done that in the latest commit and I think this comment is back to making sense :rofl:

ansible_base/lib/abstract_models/immutable.py
relrod: Yeah the idea is "This is a CommonModel and functions like a CommonModel _except_ for the fact that it's immutable and doesn't have the modified* fields."

We can rename it, but that's what I was going for.

relrod: I think `ImmutableModel` (which exists as of the latest commit) can live in `immutable.py`, and `ImmutableCommonModel` can move back into `common.py`.

ansible_base/lib/abstract_models/common.py
relrod: We could - my thought with that was that we want the encrypted/summary/related stuff to "just work" in nearly every case, so that's why I left it here.

Basically, `AbstractCommonModel` exists because my thought process was that we still do have "common" logic that we want in nearly every model, so the idea was to just split out the stuff that that isn't true for.

I'm not sure how much it's worth taking the abstraction to the extreme, and also, we can pretty easily do that later on (since `CommonModel` will inherit from those new classes - nothing in the end-user code has to change to pick that up).

relrod: I also wondered about stuff like `{Unique,}NamedCommonModel` now, though. Is there any reason to not have that extend `AbstractCommonModel` now? Because then you can do _crazy_ stuff like:

```python
class ImmutableUniqueNamedCommonModel(ImmutableModel, UniqueNamedCommonModel):
    pass
```

and you get the fancy immutable stuff _and_ the fancy unique name stuff. :scream:

relrod: Well, we'd lose created/modified. We'd need an intermediary like we did with CommonModel, I guess.

relrod: I think I'm inclined to leave it as is. If we want to split this stuff up more in the future, it's easy enough to do without breaking existing contracts.

relrod:
> I want to see a test added that does impersonate and asserts the correct value of created_by. The new test_app models should work nicely for this test.

Done.

ansible_base/lib/abstract_models/common.py
AlanCoding: My thoughts are:
 - `get_system_user` should cache. I even think of _a_ valid reason to invalidate the cache for this. Maybe if migrations ran this... which it shouldn't.
 - `get_system_user` shouldn't run its DB query if the setting is not used. AWX doesn't use it, and the task system will give None for the current user, and I don't want DAB to be running queries in this circumstances (even if use of this CommonModel in AWX is rare as of now, that will change)

AlanCoding (approved):
Final comment - just _triple check_ that makemigrations isn't producing extra migrations for... like... the authenticator app

ansible_base/lib/abstract_models/common.py
relrod: done

relrod:
@AlanCoding I pushed up one more change to make the rbac models use this (and use the "current user" function I introduced, so it can account for the system user if we need to fall back to that)

Can you give one more once-over and make sure my change there in [a1b09d2](https://github.com/ansible/django-ansible-base/pull/225/commits/a1b09d2c5b3aad3794803161f27f78b0542739ac) is reasonable?

relrod:
> Final comment - just triple check that makemigrations isn't producing extra migrations for... like... the authenticator app

I've been running makemigrations a bunch as I've worked on this, I believe/hope we are in the clear.

ansible_base/rbac/models.py
```diff
@@ -333,7 +330,7 @@ def __init__(self, *args, **kwargs):
         """
         super().__init__(*args, **kwargs)
         if not self.id:
-            user = get_current_user()
+            user = current_user_or_system_user()
             if user:
                 # Hazard: user can be a SimpleLazyObject, so use id
                 self.created_by_id = user.id
```
AlanCoding: You shouldn't need this. This should be handled by the parent class save methods now. Can't you remove the entire block starting with `if not self.id:`?

relrod: I thought that, but the doc comment of that method says:

```
        """
        Because through models are created via a bulk_create, the save method is usually not called
        to get around this, we populate the user model after initialization
        """
```

so I figured you knew of a case where we wouldn't end up setting `created_by`.

AlanCoding: Ah ha. I am pretty sure I changed that.

https://github.com/ansible/django-ansible-base/blob/d45d5a18228d68ca01e7410f207500eb140ef8cd/ansible_base/rbac/models.py#L213

That covers the main API use case. Custom through_fields don't even really work in migrations anyway, so I am not really worried about getting rid of it. I want this to be a direct use of the base immutable model.

relrod: Ok, I dropped that block from `__init__()`.

john-westcott-iv (approved):

PR #228, opened 2024-03-18
Update all deps.

jctanner:
replaced by https://github.com/ansible/django-ansible-base/pull/230

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/ldap.py
+++ b/ansible_base/authentication/authenticator_plugins/ldap.py
@@ -113,7 +113,7 @@ def validate_ldap_filter(value: Any, with_user: bool = False) -> bool:
     dn_value = value
     if with_user:
         if user_search_string not in value:
-            raise ValidationError(_(f'DN must include "{user_search_string}" placeholder for username: "{value}"'))
+            raise ValidationError(_('DN must include "{}" placeholder for username: {}').format(user_search_string, value))
         dn_value = value.replace(user_search_string, 'USER')
 
     if re.match(r'^\([A-Za-z0-9-]+?=[^()]+?\)$', dn_value):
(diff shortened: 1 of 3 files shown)


AlanCoding:
Looks relevant because it landed in DRF `3.15.0`

https://github.com/encode/django-rest-framework/commit/4abfa28e0879e2df45937ac8c7a9ffa161561955

ansible_base/authentication/authenticator_plugins/ldap.py
```diff
@@ -123,7 +123,7 @@ def validate_ldap_filter(value: Any, with_user: bool = False) -> bool:
             # We only need to check with_user at the top of the recursion stack
             validate_ldap_filter(f'({sub_filter})', with_user=False)
         return
-    raise ValidationError(_('Invalid filter: %s') % value)
+    raise ValidationError(_('Invalid filter: %s' % value))
```
relrod, line 126: This doesn't feel correct to me. If I understand right, this will do the string interpolation _before_ `_()` meaning that the fully interpolated string is what will land in the translation files.

relrod:
I think the correct way to deal with the `%(user)s` thing is to make it `%%(user)s`:

```python
In [1]: 'foo %(bar)s' % {'bar': "aa"}
Out[1]: 'foo aa'

In [2]: 'foo %%(bar)s' % {'bar': "aa"}
Out[2]: 'foo %(bar)s'
```

AlanCoding:
^ The problem is that this is _test_ input mocking _user_ input. The solution is that you don't ever format a string from a user. Before this DRF upgrade they _didn't_ format the string passed into ValidationError, thus the problem. Due to the DRF change, DAB code can't safely pass in a string with user data like it is doing now.

Link https://github.com/encode/django-rest-framework/issues/9295

PR #229, opened 2024-03-19
Fix Authenticator Maps

slemrmartin:
Fixes URL and validations for AuthenticatorMap.option

`order` validation doesn't need to be present if there is default value (that is not compatible with the way how ansible collection can work)
`triggers` validation also doesn't need to be present in PATCH request

### DIFF
--- a/ansible_base/authentication/serializers/authenticator_map.py
+++ b/ansible_base/authentication/serializers/authenticator_map.py
@@ -12,10 +12,7 @@ class AuthenticatorMapSerializer(NamedCommonModelSerializer):
 
     def validate(self, data) -> dict:
         errors = {}
-        if 'triggers' not in data or not data['triggers']:
-            errors["triggers"] = "Triggers must be a valid dict"
-        else:
-            errors.update(self.validate_trigger_data(data['triggers'], TRIGGER_DEFINITION, 'triggers'))
+        errors.update(self.validate_trigger_data(data))
 
         map_type = data.get('map_type', None)
         team = data.get('team', None)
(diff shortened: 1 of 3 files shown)


ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -27,14 +24,30 @@ def validate(self, data) -> dict:
...
-
         if errors:
             raise ValidationError(errors)
         return data
 
-    def validate_trigger_data(self, triggers: dict, definition, error_prefix: str) -> dict:
+    def validate_trigger_data(self, data):
+        errors = {}
+        request = self.context.get('request', None)
+        if not request or (request.method != 'PATCH'):
```
AlanCoding: It's difficult to imagine that `_validate_trigger_data` should be called differently when changed in PATCH vs PUT.

AlanCoding: Like, if it is a PATCH and "triggers" is in data, then you should call `_validate_trigger_data`.

slemrmartin: nice catch, yeah.

slemrmartin: switched conditions, now it should be ok

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -27,14 +24,30 @@ def validate(self, data) -> dict:
...
             raise ValidationError(errors)
         return data
 
-    def validate_trigger_data(self, triggers: dict, definition, error_prefix: str) -> dict:
+    def validate_trigger_data(self, data):
+        errors = {}
+        request = self.context.get('request', None)
+        if 'triggers' not in data or not data['triggers']:
+            if not request or (request.method != 'PATCH'):
+                errors["triggers"] = "Triggers must be a valid dict"
```
AlanCoding, line 36: Final drive-by comment about this

https://github.com/ansible/django-ansible-base/blob/5babe8742e57d45ea703986398c2777fb370f991/ansible_base/authentication/models/authenticator_map.py#L64-L69

If this shouldn't be left out of create data, or set to `{}` (the default) by omitting from PUT data, then what you're really asking for starts to sound like `blank=False`.

slemrmartin: that's true, the question is, if it was the original intention that empty dict is not possible. 

@john-westcott-iv do you know more about this?

slemrmartin: @AlanCoding ... the intention is not to allow `{}`. I tried `blank=False`, which doesn't work, it's not equal to this validation. So it has to stay here

AlanCoding (approved):

PR #231, opened 2024-03-20
Fix type of common fields which are dict-type (related / summary_fields)

AlanCoding:
~AAP-21919~

AAP-21920

### Explaining recent changes

The docs for `read_only_fields` are here:

https://www.django-rest-framework.org/api-guide/serializers/#specifying-read-only-fields

The original idea of this PR was mistaken, where we thought that putting these read-only fields in `read_only_fields` would change something in the OpenAPI spec, and that's not true. Because these `SerialzierMethodFields` are already set to `read_only=True` by default... because it's a "method" field, it's just presentation.

But the same general task of getting the OpenAPI spec (done in a fixture here) still applied to a different issue where the displayed type of "summary_fields" and "related" was wrong. You can predict why, because they are not strings and there's no way for the introspector to be able to figure that out. So this adds the fixture, adds a test to discover that, and fixes it by adding type hints to the serializer methods. That went well and we should merge it.

### DIFF
--- a/test_app/views.py
+++ b/test_app/views.py
@@ -4,6 +4,7 @@ from rest_framework.reverse import reverse
 from rest_framework.viewsets import ModelViewSet
 
 from ansible_base.lib.utils.views.ansible_base import AnsibleBaseView
+from ansible_base.rbac import permission_registry
 from ansible_base.rbac.api.permissions import AnsibleBaseObjectPermissions
 from test_app import serializers
 from test_app.models import RelatedFieldsTestModel, User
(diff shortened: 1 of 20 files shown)


relrod:
This seems reasonable to me, after CI is fixed. I would have ended up doing something similar in #155.

test_app/views.py
```diff
@@ -15,7 +16,11 @@ class TestAppViewSet(ModelViewSet, AnsibleBaseView):
     select_related = ()
 
     def get_queryset(self):
-        qs = self.serializer_class.Meta.model.access_qs(self.request.user)
+        cls = self.serializer_class.Meta.model
+        if permission_registry.is_registered(cls):
+            qs = cls.access_qs(self.request.user)
+        else:
+            qs = cls.objects.all()
```
AlanCoding, line 23: yeah, I'm pretty sure this was server erroring before this.

john-westcott-iv (approved):

relrod (approved):

PR #233, opened 2024-03-21
Cut an automatic create-only serializer for eda-server consistency

AlanCoding:
Ping for review on the expected usability side of things:
 - @ttuffin
 - @dhaustein
 - Doston (in reviewers)

### DIFF
--- a/ansible_base/lib/dynamic_config/dynamic_settings.py
+++ b/ansible_base/lib/dynamic_config/dynamic_settings.py
@@ -26,6 +26,14 @@ except NameError:
     REST_FRAMEWORK = {}
 
 
+# Tells the base serializer to auto-create a creation-only serializer
+# with the read-only fields excluded, for OpenAPI / openapi-generator / generated client
+try:
+    ANSIBLE_BASE_AUTO_CREATE_SERIALIZER
+except NameError:
+    ANSIBLE_BASE_AUTO_CREATE_SERIALIZER = False
+
+
 if 'ansible_base.api_documentation' in INSTALLED_APPS:
     if 'drf_spectacular' not in INSTALLED_APPS:
         INSTALLED_APPS.append('drf_spectacular')
--- a/ansible_base/lib/dynamic_config/dynamic_settings.py
+++ b/ansible_base/lib/dynamic_config/dynamic_settings.py
@@ -26,8 +26,7 @@ except NameError:
     REST_FRAMEWORK = {}
 
 
-# Tells the base serializer to auto-create a creation-only serializer
-# with the read-only fields excluded, for OpenAPI / openapi-generator / generated client
+# See docs/lib/views.md
 try:
     ANSIBLE_BASE_AUTO_CREATE_SERIALIZER
 except NameError:
--- a/ansible_base/lib/utils/views/ansible_base.py
+++ b/ansible_base/lib/utils/views/ansible_base.py
@@ -1,6 +1,7 @@
 import logging
 import time
 
+from django.conf import settings
 from django.utils.translation import gettext_lazy as _
 from rest_framework.views import APIView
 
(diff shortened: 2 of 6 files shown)


AlanCoding:
On the current failure:

```
    @pytest.mark.django_db
    def test_make_global_user_assignment(admin_api_client, rando, inventory):
        rd = RoleDefinition.objects.create_from_permissions(
            permissions=['change_inventory', 'view_inventory'],
            name='global-change-inv',
            content_type=None,
        )
        url = reverse('roleuserassignment-list')
        data = dict(role_definition=rd.id, user=rando.id, object_id=None)
        response = admin_api_client.post(url, data=data, format="json")
>       assert response.data['created_by']
E       KeyError: 'created_by'

test_app/tests/rbac/api/test_rbac_views.py:100: KeyError
```

Oof, it's painful because it's correct. When making a POST, there is an established expectation that you get the serialized object in the same way as when it is shown in the list or detail view. For AWX we used to have a jsonschema that verified _exactly this_. You can't do that when it differs from the create serializer. Removing this data in the POST response is exactly what is asked for here.

AlanCoding:
Proposed resolution to the above test failure:

AWX does not set the `ANSIBLE_BASE_AUTO_CREATE_SERIALIZER` setting to True, thus its responses will not be affected by this. The POST responses will continue to give `summary_fields`.

In the case of eda-server, presumably, all their responses are already omitting read-only fields like `created_at`. For the most part they don't have `summary_fields`, but if they did, then the behavior this offers is consistent with established precedent specific to that app.

DAB test_app will use the setting. Because it's test_app.

test_app/tests/rbac/api/test_rbac_views.py
```diff
@@ -74,7 +74,11 @@ def test_make_user_assignment(admin_api_client, inv_rd, rando, inventory):
     data = dict(role_definition=inv_rd.id, user=rando.id, object_id=inventory.id)
     response = admin_api_client.post(url, data=data, format="json")
     assert response.status_code == 201, response.data
-    assert response.data['created_by']
```
AlanCoding, line 77: With this change, read-only fields like `created_by` will not be returned in the POST response data. This is why we want to be selective about enabling this.

ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -26,6 +26,14 @@
     REST_FRAMEWORK = {}
 
 
+# Tells the base serializer to auto-create a creation-only serializer
+# with the read-only fields excluded, for OpenAPI / openapi-generator / generated client
+try:
+    ANSIBLE_BASE_AUTO_CREATE_SERIALIZER
```
john-westcott-iv, line 31: Total side note but I am starting to wonder with all of our settings if we should do a dict instead of individual settings like DRF does.

ansible_base/lib/utils/views/ansible_base.py
```diff
@@ -9,7 +10,34 @@
...
+        fields = create_field_list
+
+    create_serializer_name = cls.__name__.replace("Serializer", "") + "CreateSerializer"
+    return type(create_serializer_name, (cls,), {"Meta": Meta})
+
+
 class AnsibleBaseView(APIView):
+    def get_serializer_class(self):
+        serializer_cls = super().get_serializer_class()
+        if settings.ANSIBLE_BASE_AUTO_CREATE_SERIALIZER and self.action == "create":
```
john-westcott-iv, line 37: Do we ever anticipate this setting to change dynamically? If not, and this only affects fields granted from DAB, I'm wondering if we can just do something in CommonModelSerializer instead of in the view itself?

john-westcott-iv (commented):
Can we also add a doc somewhere about the setting and what its going to do if enabled?

ansible_base/lib/dynamic_config/dynamic_settings.py
AlanCoding: https://github.com/encode/django-rest-framework/blob/master/rest_framework/settings.py

Just copy DRF style for that, no problem. I think this would be a fairly large body of work, and we could leverage releases to do this in a major version bump.

ansible_base/lib/utils/views/ansible_base.py
AlanCoding: I had that thought at first. It doesn't work in practice because

 - the context is often not passed to the serializer, which I admit is incorrect, but it's not my code, and I'm not here to fix that. This is required to get the request and the method.
 - translation of http verb GET/POST/etc. to action is non-trivial as viewset customization increases. Like, a POST to the launch endpoint. Ideally this won't matter, but again that's one of those cases where it might work in theory but not in practice
 - it actually doesn't work at all, because the entire point here is the OpenAPI spec, which by drf-spectacular will inspect your views & serializers and map the view to the serializer. So the point here is to actually return a "SomethingCreateSerializer" which is a named entity in the spec.

john-westcott-iv (approved):

PR #234, opened 2024-03-21
Adding custom Redis client

john-westcott-iv:


rcarrillocruz:
I know it's early and this is wip, but just so it is in your radar it would be great if you could have a namespace/keyspace concept into the client.
Since we want to have a central redis, we want to have the various component using Redis to namespace keys.
Not sure if there is upstream support for it, it seems people in the community build a wrapper around redis-py  client class for that:

https://github.com/redis/redis-py/issues/12

john-westcott-iv:
After doing some tedious stuff in the RedisClient I found out that django supports this itself directly https://docs.djangoproject.com/en/5.0/topics/cache/#cache-key-prefixing

ansible_base/lib/redis/client.py
```diff
@@ -0,0 +1,105 @@
...
+
+# We are going to build our own cluster class to override the mget function
+# In a redis cluster, keys might not be in the same slot and this will throw off mget.
+# Instead, we are going to try and use mget and then, if we get the slot error, we will try the mget_nonatomic to make it work
+class DABRedisCluster(RedisCluster):
+    def mget(self, *args, **kwargs):
+        try:
+            return super().mget(*args, **kwargs)
+        except RedisClusterException as e:
+            if 'MGET - all keys must map to the same key slot' in str(e):
```
relrod, line 24: This scares me a little because it's "just" a string and could change between versions of the redis lib, in theory.

I don't have a better solution, but it's something to be aware of.

ansible_base/lib/redis/client.py
```diff
@@ -0,0 +1,105 @@
...
+                return super().mget_nonatomic(*args, **kwargs)
+            raise e
```
relrod: You don't need the argument here, just:
```suggestion
            raise
```
works here.

ansible_base/lib/redis/client.py
```diff
@@ -0,0 +1,105 @@
...
+                kwargs[arg_name] = attribute
+
+        # Add the DB from the URL (if passed)
+        db = None
+        try:
+            db = int(parsed_url.path.split('/')[1])
+        except (IndexError, ValueError):
+            pass
+        if db:
+            kwargs['db'] = db
```
relrod: I think we can shorten this a bit

```suggestion
        try:
            kwargs['db'] = int(parsed_url.path.split('/')[1])
        except (IndexError, ValueError):
            pass
```

ansible_base/lib/redis/client.py
```diff
@@ -0,0 +1,105 @@
...
+                    logger.error(f"Specified clustered_hosts is not a string, got: {self.clustered_hosts}")
+                    raise translated_generic_exception
+
+                host_ports = self.clustered_hosts.split(',')
+                for host_port in host_ports:
+                    try:
+                        node, port_string = host_port.split(':')
+                    except ValueError:
+                        logger.error(f"Specified cluster_host {host_port} is not valid; it needs to be in the format <host>:<port>")
+                        raise translated_generic_exception
```
relrod, line 87: Would it make sense to just use the default redis port in this case instead?

test_app/tests/lib/redis/test_client.py
```diff
@@ -0,0 +1,176 @@
...
+
+@pytest.mark.parametrize(
+    "url,host,port,username,password,db",
+    [
+        ('localhost', 'localhost', 6379, None, None, 0),
+        ('redis://my_mom:her_pass@home.com:1234/1', 'home.com', 1234, 'my_mom', 'her_pass', 1),
+        ('redis://my_mom:her_pass@home.com/1', 'home.com', 6379, 'my_mom', 'her_pass', 1),
+        ('redis://my_mom@home.com/', 'home.com', 6379, 'my_mom', None, 0),
+        ('redis://:her_pass@home.com/', 'home.com', 6379, None, 'her_pass', 0),
+        ('redis://home.com:1234/a', 'home.com', 1234, None, None, 0),
```
relrod: Nitpick, but I try to avoid using "real-ish" URLs in tests, just to avoid any future concerns. I tend to use `example.com` (which is IANA-reserved for this kind of stuff).

relrod (approved):
A few minor inlines, but nothing blocking

ansible_base/lib/redis/client.py
john-westcott-iv: Yea, I wish there was a more specific exception we could catch.

ansible_base/lib/redis/client.py
john-westcott-iv: I'd rather not make the assumption to use the default port because an unable to connect error which may ensue might be misleading vs just raising a more detailed exception here. The user had to specify this anyway so its on them if they put junk in there.

test_app/tests/lib/redis/test_client.py
john-westcott-iv: I'll change this to `example.com`, I should know better :)

PR #239, opened 2024-03-26
Supress test standard out unless failure

AlanCoding:
https://docs.pytest.org/en/7.4.x/how-to/capture-stdout-stderr.html#setting-capturing-methods-or-disabling-capturing

Looking out the docs, honestly the default behavior looks a ton better than what we're getting now which is overwhelmingly loud with DEBUG statements in test runs.

When a test fails it does print the standard out, but if it is successful, nothing gets printed. As the total test count grows... I think we need this.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,7 @@ exclude = [ 'ansible_base/authentication/migrations/*', 'test_app/migrations/*',
 
 [tool.pytest.ini_options]
 DJANGO_SETTINGS_MODULE = "test_app.settings"
-addopts = "--strict-markers --reuse-db --migrations --typeguard-packages=ansible_base -s -vvv"
+addopts = "--strict-markers --reuse-db --migrations --typeguard-packages=ansible_base -vvv"
 
 [tool.tox]
 legacy_tox_ini = """


AlanCoding:
Test output lines: 2350

Other PRs: 6936

relrod:
I'd rather see this solved by disabling logging in test_app. (I already do this in the activitystream branch by setting the console handler to use `logging.NullHandler` instead of `logging.StreamHandler`, works fine.)

Our tests should ensure that the log functions are called, we don't need to test the output of the logging system - that's testing the logger library, not our code.

AlanCoding:
My preference - I do want to see log output from failed tests.

These tend to be from mock server requests. Especially if we did this in test_app, that would shield DAB development from seeing the log output we're producing in projects downstream (for test server _as well as tests_). This output is already too much, it's a thing that needs to be fixed IMO. We just haven't had people use it enough to complain yet.

relrod:
My main concern is sometimes I want to be able to write print(foo) inside a test and see the result. Maybe losing that isn't the end of the world, I can set a breakpoint and see what I want inside pdb, or manually add `-s` back in that case. But that's what I was concerned about, because I use that often for debugging things.

AlanCoding:
> My main concern is sometimes I want to be able to write print(foo) inside a test and see the result.

This doesn't prevent that. If the test _fails_, "foo" will be printed, but if the test _passes_ "foo" will not be printed.

relrod (approved):
I'll concede, because the output is annoying and I just want it to go away.

PR #242, opened 2024-03-27
Fix project license

cutwater:
Use LICENSE file according to the official guidelines
    
Replace the markdown formatted text of the license with the
original text file from https://www.apache.org/licenses/LICENSE-2.0.txt
according to the Apache Foundation guidelines https://www.apache.org/legal/apply-license.html
    
> To apply the ALv2 to a new software distribution, include one copy of the license
> text by copying the contents of LICENSE-2.0.txt into a file called LICENSE
> in the top directory of your distribution.
    
In addition to the license text this file includes APPENDIX,
which describes how apply the Apache license.


relrod:
If we're going to do this, I'd like some kind of CI check so that they don't get forgotten in the future.

Something like https://github.com/savoirfairelinux/flake8-copyright perhaps (or any similar kind of plugin). This would fit in nicely to already existing CI workflows, since we already run `tox -e flake8` both in CI and (most of us I think) run it locally before we commit.

relrod:
(also if we do that, I would suggest we check for some substring like "Licensed under the Apache License, Version 2.0" rather than a specific copyright date + name)

LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
john-westcott-iv, line 1: Is there an advantage of using `LICENSE` over `LICENSE.md`?

cutwater: 1. The `LICENSE` and `LICENSE.txt` are standard names for the license files recognized by different tools by default.
2. The Apache license provides explicit recommendations on how to apply the license text.
3. Because it's a well-known license (unlike custom EULA), there is no need to include the license text directly to the package distribution to reduce distribution metadata bloat. Because it's specified in the package metadata as `license = {file = "LICENSE.md"}` the whole text license is included into the distribution package metadata file. Changing it to `license = {text = "Apache-2.0"}` still includes the `LICENSE` file into the distribution, but package metadata doesn't have license text duplicated.

john-westcott-iv:
Do we need to copyright migration files since they are mostly generated from another piece of work?

maxamillion:
Other than conflicts, this looks good. :+1:

relrod (approved):

PR #243, opened 2024-03-27
Change RBAC usage to filter_queryset

AlanCoding:
AAP-22130

Before, everything used `get_queryset`. It was always designed so it didn't have to do it this way, but that we could take an existing queryset and filter that. There was just one bug which this fixes in the method that gets attached.

This converts test_app and docs to suggest `filter_queryset`.

### DIFF
--- a/ansible_base/rbac/evaluations.py
+++ b/ansible_base/rbac/evaluations.py
@@ -1,4 +1,7 @@
+from typing import Optional
+
 from django.conf import settings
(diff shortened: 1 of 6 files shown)


AlanCoding:
Also, @relrod IIRC did already request this in review of the feature branch, so I think this is philosophically aligned.

Dostonbek1 (approved):

PR #248, opened 2024-03-29
Improve resource API stability

newswangerd:
This is a grab bag of improvements on the resources API:

- The post update signal now creates a resource if one didn't exist previously for some reason.
- The post migration signal now adds any missing resources, rather than just doing a one time migration for each resource type.
- Removed the resource inspector and detail url on the resources serializer. The resource inspector was designed to play a role in RBAC that is now obsolete, so it's role got downgraded to just generating a detail url in the resources serializer. This didn't work on half the AAP services, so I've gone ahead and removed it.
- Renamed service_id module to service_identifier. We were exporting function called `service_id` in `models.__init__`, and this had some potential to conflict with the `models.service_id` module.


ansible_base/resource_registry/migrations/0004_remove_resourcetype_migrated.py
```diff
@@ -0,0 +1,17 @@
...
+    dependencies = [
+        ('dab_resource_registry', '0003_alter_resource_object_id'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='resourcetype',
+            name='migrated',
+        ),
+    ]
```
AlanCoding, line 17: @john-westcott-iv and I went over this in prior PRs, but we are on a policy to squash test_app migrations. Not a blocker, I just you to be on noticed this can get squashed.

AlanCoding (approved):

ansible_base/resource_registry/migrations/0004_remove_resourcetype_migrated.py
newswangerd: This isn't a test app migration, it's in the resource registry. I'm reluctant to squash dab migrations because we've already deployed it to galaxy.ansible.com

AlanCoding: Thanks for the info, this is good to know.

PR #249, opened 2024-04-01
Revert "Cut an automatic create-only serializer for eda-server consistency"

AlanCoding:
Reverts ansible/django-ansible-base#233

Because of requirements beyond the anticipated, this solution has become non-viable to do in DAB to solve the full problem.

The plan is articulated in https://github.com/ansible/eda-server/pull/775, which is that `ANSIBLE_BASE_CUSTOM_VIEW_PARENT` will be used to make an eda-server-standard base, which will copy this code, and combine with other existing patterns in the code base.

EDIT:

that plan is not implemented in https://github.com/ansible/eda-server/pull/801


AlanCoding:
This is now ready to merge since the EDA alternative is merged there.

john-westcott-iv (approved):
The origional referenced PR had a 6th file test_app/tests/rbac/api/test_rbac_views.py that we did not revert but in looking at that it looks like it might have needed to be changed.

AlanCoding:
True, this isn't a clear-cut revert, it's a hand crafted one.

PR #254, opened 2024-04-02
[translations] common/auth update

thedoubl3j:
First PR of many to update translations. 

Had to do things a little differently here since there was a lot of use around a function that both logged an error and raised an exception all in the same call. Rather than rip that out and just do individual calls each time they were needed and translate there, I worked with @chrismeyersfsu to come up with a helper class that has been added to lib/utils/ called translations.py that can be used to not translate messages immediately (so that they can be logged) and then translate them for exceptions/other potential user facing errors. 

During testing/messing around with ways to not rip out the existing code, some research showed that the command to go look for strings marked for translation (django-admin makemessages) explicitly looks for `_("` so importing this class as `_` follows the same pattern that most all projects that needs translations use. 

Testing steps:
`pip freeze | grep Django` `pip install Django`
4. make a locale directory (makemessages looks for one and we don't have one configured yet so just create your own) `mkdir locale` 
5. run the makemessages command check the output to see if strings were extracted correctly. 
`django-admin makemessages -l es` 
`vim locale/es/LC_MESSAGES/django.po`

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -8,6 +8,7 @@ from rest_framework.authentication import BaseAuthentication
 from rest_framework.exceptions import AuthenticationFailed
 
 from ansible_base.lib.utils.settings import get_setting
+from ansible_base.lib.utils.translations import translatableConditionally as _
 
 logger = logging.getLogger("ansible_base.jwt_consumer.common.auth")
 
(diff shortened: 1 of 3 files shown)


john-westcott-iv (approved):

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -4,6 +4,7 @@
 import jwt
 import requests
 from django.contrib.auth import get_user_model
+from ansible_base.lib.utils.translation import translatable_conditionally as _
```
AlanCoding: translation -> translations

djyasin (approved):

PR #256, opened 2024-04-03
[WIP] Add view to get_summary_fields

slemrmartin:
`get_summary_fields` needs a view argument to be able to compute `user_capabilities` (permissions) in the model

### DIFF
--- a/ansible_base/lib/abstract_models/common.py
+++ b/ansible_base/lib/abstract_models/common.py
@@ -136,7 +136,7 @@ class AbstractCommonModel(models.Model):
 
         return instance
 
-    def get_summary_fields(self):
+    def get_summary_fields(self, view=None):
         response = {}
         for field in self._meta.fields:
             if isinstance(field, models.ForeignObject) and getattr(self, field.name):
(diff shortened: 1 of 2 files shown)


ansible_base/lib/abstract_models/common.py
```diff
@@ -136,7 +136,7 @@ def from_db(self, db, field_names, values):
 
         return instance
 
-    def get_summary_fields(self):
+    def get_summary_fields(self, view=None):
```
john-westcott-iv, line 139: Can you give more details here as to why view is needed? It's never referenced in this function or passed anywhere else unless I am missing something.

slemrmartin: It's required for `user_capabitilies`. In AWX, the `get_summary_fields` are placed in serializer, which has `self.context.get('view', None)`, but DAB's code calls `obj.get_summary_fields` which targets model, so the view has to be propagated there. 

Anyway this feature was postponed, so I'm converting it to draft

slemrmartin:
user capabilities are not planned now, and the concept may change, closing for now

PR #260, opened 2024-04-03
Fix enablement problem with typeguard

AlanCoding:
I'll put more details in the upstream issue https://github.com/agronholm/typeguard/issues/260

At some point, the typeguard python plugin was working, but it broke and who knows when.

I expect there will be more failures as the checks run. This validates that I, personally, needed this. Since I've added globs of code with incorrect type hints since I first tried to add enforcement.

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/utils.py
+++ b/ansible_base/authentication/authenticator_plugins/utils.py
@@ -1,8 +1,8 @@
 import logging
-from types import ModuleType
 from functools import lru_cache
 from glob import glob
 from os.path import basename, isfile, join
+from types import ModuleType
 from typing import Optional
 
 from django.conf import settings
--- a/ansible_base/authentication/authenticator_plugins/utils.py
+++ b/ansible_base/authentication/authenticator_plugins/utils.py
@@ -49,7 +49,7 @@ def get_authenticator_urls(authenticator_type: str) -> list:
         return getattr(urls_module, 'urls', [])
     except Exception as e:
         logger.error(f"Failed to load urls from {authenticator_type} {e}")
-    return None
+    return []
 
 
 def generate_authenticator_slug(type: str, name: str) -> str:
--- a/ansible_base/authentication/authenticator_plugins/utils.py
+++ b/ansible_base/authentication/authenticator_plugins/utils.py
@@ -1,4 +1,5 @@
 import logging
+from types import ModuleType
 from functools import lru_cache
 from glob import glob
 from os.path import basename, isfile, join
@@ -44,7 +45,7 @@ def get_authenticator_plugin(authenticator_type: str):
     return AuthClass()
 
 
-def get_authenticator_urls(authenticator_type: str) -> Optional[list]:
+def get_authenticator_urls(authenticator_type: str) -> Optional[ModuleType]:
     try:
         urls = __import__(authenticator_type, globals(), locals(), ['urls'], 0)
         return urls
--- a/ansible_base/authentication/authenticator_plugins/utils.py
+++ b/ansible_base/authentication/authenticator_plugins/utils.py
@@ -2,6 +2,7 @@ import logging
 from functools import lru_cache
 from glob import glob
 from os.path import basename, isfile, join
+from typing import Optional
 
 from django.conf import settings
 from django.utils.text import slugify
@@ -43,7 +44,7 @@ def get_authenticator_plugin(authenticator_type: str):
     return AuthClass()
 
 
-def get_authenticator_urls(authenticator_type: str) -> list:
+def get_authenticator_urls(authenticator_type: str) -> Optional[list]:
     try:
         urls = __import__(authenticator_type, globals(), locals(), ['urls'], 0)
         return urls
--- a/ansible_base/authentication/authenticator_plugins/utils.py
+++ b/ansible_base/authentication/authenticator_plugins/utils.py
@@ -2,8 +2,6 @@ import logging
 from functools import lru_cache
 from glob import glob
 from os.path import basename, isfile, join
-from types import ModuleType
-from typing import Optional
 
 from django.conf import settings
 from django.utils.text import slugify
@@ -45,10 +43,10 @@ def get_authenticator_plugin(authenticator_type: str):
     return AuthClass()
 
 
-def get_authenticator_urls(authenticator_type: str) -> Optional[ModuleType]:
+def get_authenticator_urls(authenticator_type: str) -> list:
     try:
-        urls = __import__(authenticator_type, globals(), locals(), ['urls'], 0)
-        return urls
+        urls_module = __import__(authenticator_type, globals(), locals(), ['urls'], 0)
+        return getattr(urls_module, 'urls', [])
     except Exception as e:
         logger.error(f"Failed to load urls from {authenticator_type} {e}")
     return None
--- a/ansible_base/lib/utils/encryption.py
+++ b/ansible_base/lib/utils/encryption.py
@@ -65,9 +65,8 @@ class Fernet256(Fernet):
 
         # Finally decode the value
         encrypted = base64.b64decode(b64data)
-        value = self.decrypt(encrypted)
 
-        return smart_str(value)
+        return smart_str(self.decrypt(encrypted))
 
 
 ansible_encryption = SimpleLazyObject(func=lambda: Fernet256())
--- a/ansible_base/rbac/models.py
+++ b/ansible_base/rbac/models.py
@@ -625,10 +625,8 @@ class RoleEvaluationFields(models.Model):
     @classmethod
     def accessible_objects(cls, model_cls, user, codename, queryset: Optional[QuerySet] = None) -> QuerySet:
         if queryset is None:
-            base_queryset = model_cls.objects
-        else:
-            base_queryset = queryset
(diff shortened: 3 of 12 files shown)


ansible_base/lib/utils/encryption.py
```diff
@@ -65,9 +65,8 @@ def decrypt_string(self, value: str) -> str:
 
         # Finally decode the value
         encrypted = base64.b64decode(b64data)
-        value = self.decrypt(encrypted)
```
AlanCoding, line 68: It doesn't like it when you re-define an input argument with a type that doesn't match its annotation. Fair, I say.

john-westcott-iv: Thats interesting, good nugget.

ansible_base/authentication/authenticator_plugins/utils.py
```diff
@@ -43,7 +45,7 @@ def get_authenticator_plugin(authenticator_type: str):
     return AuthClass()
 
 
-def get_authenticator_urls(authenticator_type: str) -> list:
+def get_authenticator_urls(authenticator_type: str) -> Optional[ModuleType]:
```
john-westcott-iv: I think something is actually out of whack here....

If you look at `ansible_base.authentication.authenticator_plugins.saml` it defines urls as:
```
urls = [
    # SAML Metadata
    re_path(r'authenticators/(?P<pk>[0-9]+)/metadata/$', SAMLMetadataView.as_view(), name='authenticator-metadata'),
]
```

Based on the name of the function (and old type hint) I think the intention of this function is to return a list[URLPattern] but its returning the module and then, in the only place we call it (django-ansible-base/ansible_base/authentication/urls.py) we are doing:
```
for plugin_name in get_authenticator_plugins():
    plugin_urls = getattr(get_authenticator_urls(plugin_name), 'urls', None)
```

It seems like this function should be extracting the 'urls' filed from the module it gets (as `urls` in this function) and then returning that or `[]`.

ansible_base/rbac/models.py
```diff
@@ -605,8 +605,10 @@ def accessible_ids(cls, model_cls, actor, codename: str, content_types: Optional
     @classmethod
     def accessible_objects(cls, model_cls, user, codename, queryset: Optional[QuerySet] = None) -> QuerySet:
         if queryset is None:
-            queryset = model_cls.objects
-        return queryset.filter(pk__in=cls.accessible_ids(model_cls, user, codename))
+            base_queryset = model_cls.objects
+        else:
+            base_queryset = queryset
+        return base_queryset.filter(pk__in=cls.accessible_ids(model_cls, user, codename))
```
john-westcott-iv: The else here feels "weird" to me. Out of curiosity, would the type checker be "happy" if instead of `model_cls.objects` we set `queryset = model_cls.objects.all()` (which would be a QuerySet). But I think that would be less efficient than just appending the filter to a Manager right?

ansible_base/rbac/permission_registry.py
```diff
@@ -46,7 +47,7 @@ def register(self, *args, parent_field_name='organization'):
     def track_relationship(self, cls, relationship, role_name):
         self._tracked_relationships.add((cls, relationship, role_name))
 
-    def get_parent_model(self, model) -> Optional[Model]:
+    def get_parent_model(self, model) -> Union[type, None]:
```
john-westcott-iv: This is the same as `Optional` which we've used in other places. Should we try to standardize on either `Optional(type)` or `Union[type, None]` in the code?

ansible_base/rbac/triggers.py
```diff
@@ -131,7 +131,7 @@ def rbac_post_init_set_original_parent(sender, instance, **kwargs):
     instance.__rbac_original_parent_id = getattr(instance, parent_id_name)
 
 
-def get_parent_ids(instance) -> list[tuple[int, int]]:
+def get_parent_ids(instance) -> list[tuple[Model, int]]:
```
john-westcott-iv, line 134: In general, should we ever try and get more specific than `Model`? i.e. if its guaranteed to be a specific model should that be added to the type hint?

ansible_base/rbac/validators.py
```diff
@@ -21,7 +20,7 @@ def codenames_for_cls(cls) -> set[str]:
     return {t[0] for t in cls._meta.permissions} | {f'{act}_{cls._meta.model_name}' for act in cls._meta.default_permissions}
 
 
-def permissions_allowed_for_system_role() -> dict[Model, set[str]]:
+def permissions_allowed_for_system_role() -> dict[type, set[str]]:
```
john-westcott-iv, line 23: I don't see `type` imported, is this literally a reference to the `type()` function? It looks like from a quick google this just means some kind of complex type, is that accurate?

ansible_base/authentication/authenticator_plugins/utils.py
AlanCoding: The name of the function seems to be wrong, yes. If you're sure there's not going to be a lot of unexpected uses of this method, then I can try to fix by changing the return value as opposed to changing the type annotation of the return value.

ansible_base/rbac/models.py
AlanCoding: Interesting theory. I will give it a try. I'm not familiar enough with this to predict what would happen. No, I don't think that would be any less efficient. Every operation you make on a QuerySet produces a new queryset, but this is only python/in-memory considerations, which I'm committed to not caring about.

ansible_base/rbac/permission_registry.py
AlanCoding: What you're proposing is `Optional[type]` and there be dragons there, because of the difference between `NoneType` and `None`. Looking for a solution, the first thing I stumbled on was `"Model"` as a string, which would specify the type, not an object of the type, also `typing.Type` was an option. There's also a _lot_ of differences in recent python versions, and although AWX is on 3.11, I don't want to assume this version if I don't have to. I'd rather make it less picky that force a certain python version. My issue with a lot of these approaches it that union/optional typing seemed to want both options to be a class, resulting in an expectation of `NoneType`.

If we were going to standardize on something, I would probably go for `Union[type, None]`, because this is brutal and reliable. This also isn't tremendously common. You will tend to hit this with _registries_ which do work crawling a collection of _classes_.

ansible_base/rbac/triggers.py
AlanCoding: I'm going to say "no" because these are models from the registry, and the app using this libraries puts its own models in there. The criteria for what goes into the registry fits `Model` like a glove. It is almost a perfect description of the expected types.

ansible_base/rbac/validators.py
AlanCoding: Yeah, it's a python built-in, and `type()` is a callable version of `class` - for my catch-phrase description. For the RBAC model registry this comes up a lot, as I intend to return a class, similar to Django `apps.get_model()`

ansible_base/rbac/models.py
AlanCoding: implemented

ansible_base/authentication/authenticator_plugins/utils.py
AlanCoding: implemented

ansible_base/rbac/permission_registry.py
relrod: I'm really struggling to grasp this. `Union[type, None]` and `Optional[type]` should absolutely be the same thing, and I have a strong preference to `Optional` for clarity/readability.

Even the python `typing` [docs](https://docs.python.org/3/library/typing.html#typing.Optional) say:

>  `typing.Optional`
>
>    `Optional[X]` is equivalent to `X | None` (or `Union[X, None]`).

Is typeguard breaking spec here?

AlanCoding: I'll put that in, but I'm starting to think it was a version-specific behavior now.

AlanCoding: alright, I guess it's working...

ansible_base/resource_registry/models/resource.py
```diff
@@ -95,7 +96,9 @@ def delete_resource(self):
             self.delete()
 
     @classmethod
-    def create_resource(cls, resource_type: ResourceType, resource_data: dict, ansible_id: str = None, service_id: str = None):
+    def create_resource(
+        cls, resource_type: ResourceType, resource_data: dict, ansible_id: Union[str, uuid.UUID, None] = None, service_id: Union[str, uuid.UUID, None] = None
```
relrod, line 100: If we were targetting Python 3.10+, then we could rewrite the unions as `Optional[str | uuid.UUID]` which IMO is more readable.

But we're not, so we can't. :cry:

relrod (approved):

ansible_base/resource_registry/models/resource.py
AlanCoding: I'm completely sure this will get better with time. If we have this check enabled, then the developer writing this is likely to re-consider passing in a `uuid.UUID` value simply for the point of simplifying the type hint. And that would be a _good_ thing! But my concern here is very strictly that type hints are _trustworthy_ in DAB to start with. I have wanted to roll out much more type hints in RBAC, as a part of better overall internal documentation, but hesitated because we didn't have this, and lately because it wasn't working.

PR #261, opened 2024-04-03
[activitystream] Make DRF render more nicely

relrod:
- Make it so the URL insertion works (by providing "activitystream-list")
- Give a nicer page title

### DIFF
--- a/ansible_base/activitystream/urls.py
+++ b/ansible_base/activitystream/urls.py
@@ -1,12 +1,13 @@
 from django.urls import include, path
+from rest_framework import routers
 
 from ansible_base.activitystream import views
-from ansible_base.lib.routers import AssociationResourceRouter
 
-router = AssociationResourceRouter()
+router = routers.SimpleRouter()
 router.register(
     'activitystream',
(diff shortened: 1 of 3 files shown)


ansible_base/activitystream/urls.py
```diff
@@ -1,12 +1,13 @@
 from django.urls import include, path
+from rest_framework import routers
 
 from ansible_base.activitystream import views
-from ansible_base.lib.routers import AssociationResourceRouter
 
-router = AssociationResourceRouter()
+router = routers.SimpleRouter()
```
john-westcott-iv, line 6: Out of curosity, why the change here? Is there something we need to/can do better in `AssociationResourceRouter`?

john-westcott-iv (approved):

ansible_base/activitystream/urls.py
relrod: It's more that I don't think we need the functionality of ARR here. I'm not actually telling it to associate anything, in the router config. So it doesn't buy us anything in this case.

PR #262, opened 2024-04-03
Be able to shut up activity stream

relrod:
This approach is ~loosely~ inspired by what AWX does.

### DIFF
--- a/ansible_base/activitystream/signals.py
+++ b/ansible_base/activitystream/signals.py
@@ -5,21 +5,30 @@ from contextlib import contextmanager
 from ansible_base.lib.utils.models import current_user_or_system_user, diff
 
 logger = logging.getLogger('ansible_base.activitystream.signals')
-thread_local = threading.local()
-thread_local.activitystream_enabled = True
+
+
+class ActivityStreamEnabled(threading.local):
+    def __init__(self):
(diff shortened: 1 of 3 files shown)


relrod:
~well...~

~That didn't go so well.~

Okay, just going to copy the AWX approach here entirely.

ansible_base/activitystream/signals.py
```diff
@@ -1,11 +1,36 @@
...
+    def __bool__(self):
+        return self.enabled
+
+
+activitystream_enabled = ActivityStreamEnabled()
+
+
+@contextmanager
+def no_activity_stream():
+    activitystream_enabled.enabled = False
```
AlanCoding, line 24: I do kind of like that AWX maintained a `prior_value` store here. It's not that you'd ever switch True-False-True-False, but you would logically want a True-False-False nesting

```python
with no_activity_stream():
  with no_activity_stream():
    pass
  obj.save()
```

In this case, I would still want `obj.save()` to leave no activity stream entry. This is perfectly testable and all.

AlanCoding (approved):
My comment is more of a suggestion since I don't think it's _actually_ production-relevant.

relrod:
Good suggestion, I'll add that back in.

PR #264, opened 2024-04-04
Remove object_role from RBAC API, cleanup logging

AlanCoding:
This is a grab-bag of a couple of things, which I'd like to address in DAB before feature branches merge.

### Permission Logging

Steps to reproduce are:
 - give user view / member permission to a team
 - log in as that user
 - navigate to the team detail page

Doing this, DAB RBAC permission class gave a whole lot of WARNING log messages. This is because DRF re-calls the permission class for other methods like PUT/PATCH/etc. even though the request is obviously just a GET request. This is distracting, and would make users lose trust in the log messages that tell _the exact permission_ that the user lacks, and thus the reason for 403 errors.

With this change, I believe logs from the permission class should be actually telling the details of a denied request, which is what I want.

### API Stabilization

Here is user_role_assignments

```json
{
    "count": 3,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 1,
            "related": {
                "role_definition": "/api/v1/role_definitions/1/",
                "user": "/api/v1/users/5/",
            },
            "summary_fields": {
                "created_by": {
                    "id": 1,
                    "username": "_system",
                    "first_name": "",
                    "last_name": ""
                },
                "role_definition": {
                    "id": 1,
                    "name": "Organization admin permissions",
                    "description": "",
                    "managed": true
                },
                "user": {
                    "id": 5,
                    "username": "org_admin",
                    "first_name": "",
                    "last_name": ""
                },
                "content_object": {
                    "id": 1,
                    "name": "AWX_community"
                }
            },
            "created": "2024-04-04T19:00:15.143023Z",
            "created_by": 1,
            "object_id": "1",
            "role_definition": 1,
            "user": 5
        },
```

This removes "object_role", both from the fields _and_ from the `summary_fields` which is an issue of the shared common serializer, ping @john-westcott-iv @relrod

We don't want to surface this, because it is internal, and we want to claim full flexibility to change the details of these models (and some talks of this have already happened).

### Unordered Queryset

I want to address the `UnorderedObjectListWarning` warnings by changing the Meta properties. This doesn't actually result in a schema change, so I'm not worried about downstream effect from putting it in the 0001 migration, but neither AWX or eda-server branches have merged yet so it wouldn't matter anyway.

### Related Names

Using test_app, I wanted to be able to link from external models (like Team and User) to role assignments. If you dig in, you find out that you can't do that without a `related_name`, and this requires a migration, so that's another thing I want to slip in before the deadline.

### DIFF
--- a/ansible_base/lib/abstract_models/common.py
+++ b/ansible_base/lib/abstract_models/common.py
@@ -142,7 +142,7 @@ class AbstractCommonModel(models.Model):
         for field in self._meta.fields:
             if isinstance(field, models.ForeignObject) and getattr(self, field.name):
                 # ignore relations on inherited django models
-                if field.name.endswith("_ptr"):
+                if field.name.endswith("_ptr") or (field.name in self.ignore_relations):
                     continue
                 if hasattr(getattr(self, field.name), 'summary_fields'):
                     response[field.name] = getattr(self, field.name).summary_fields()
--- a/ansible_base/lib/abstract_models/common.py
+++ b/ansible_base/lib/abstract_models/common.py
@@ -142,7 +142,7 @@ class AbstractCommonModel(models.Model):
         for field in self._meta.fields:
             if isinstance(field, models.ForeignObject) and getattr(self, field.name):
                 # ignore relations on inherited django models
-                if field.name.endswith("_ptr") or (field.name in self.ignore_relations):
+                if field.name.endswith("_ptr"):
                     continue
                 if hasattr(getattr(self, field.name), 'summary_fields'):
                     response[field.name] = getattr(self, field.name).summary_fields()
--- a/ansible_base/rbac/api/serializers.py
+++ b/ansible_base/rbac/api/serializers.py
@@ -11,7 +11,7 @@ from rest_framework.fields import flatten_choices_dict, to_choices_dict
 from rest_framework.serializers import ValidationError
 
 from ansible_base.lib.abstract_models.common import get_url_for_object
-from ansible_base.lib.serializers.common import CommonModelSerializer
+from ansible_base.lib.serializers.common import CommonModelSerializer, ImmutableCommonModelSerializer
 from ansible_base.rbac.models import RoleDefinition, RoleTeamAssignment, RoleUserAssignment
 from ansible_base.rbac.permission_registry import permission_registry  # careful for circular imports
 from ansible_base.rbac.validators import check_content_obj_permission, validate_permissions_for_model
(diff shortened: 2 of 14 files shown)


AlanCoding:
@vidyanambiar this makes some API changes. They are all pretty minor but I want you to be aware.

ansible_base/lib/abstract_models/common.py
```diff
@@ -141,7 +141,7 @@ def get_summary_fields(self):
         for field in self._meta.fields:
             if isinstance(field, models.ForeignObject) and getattr(self, field.name):
                 # ignore relations on inherited django models
-                if field.name.endswith("_ptr"):
+                if field.name.endswith("_ptr") or (field.name in self.ignore_relations):
```
AlanCoding: @relrod @john-westcott-iv this is a generic DAB base model change, so I am pinging you for a review here. This was discovered troubleshooting the logs & stuff for the RBAC views. The `ignore_relations` will avoid listing things in "related", but it will still show them in "summary_fields".

ansible_base/rbac/migrations/0001_initial.py
```diff
@@ -67,6 +67,7 @@ class Migration(migrations.Migration):
                 ('modified', models.DateTimeField(auto_now=True, help_text='The date/time this resource was created')),
             ],
```
john-westcott-iv, line 68: Did you do a squash migration on this? If so do we know that nothing has used this smigration yet? i.e. an AWX or EDA release? For dev environments we can just tell users to drop their DB and let it rebuild.

ansible_base/lib/abstract_models/common.py
john-westcott-iv: This makes me a little nervious incase the API is leveraging anything that might suddenly disappeared. Would you be ok with splitting this out into its own PR/commit so if we need to revert it  we can do it easily?

ansible_base/rbac/migrations/0001_initial.py
AlanCoding: Part of my reason for prioritizing this work now is because the AWX & EDA branches are not merged yet. So it's only used in feature branches... as of today.

Also, this isn't a DB schema change so we're find to make this change without disruption.

ansible_base/lib/abstract_models/common.py
AlanCoding: https://github.com/ansible/django-ansible-base/pull/303 I will push a commit to remove it here.

ansible_base/rbac/api/serializers.py
```diff
@@ -11,7 +11,7 @@
 from rest_framework.serializers import ValidationError
 
 from ansible_base.lib.abstract_models.common import get_url_for_object
-from ansible_base.lib.serializers.common import CommonModelSerializer
+from ansible_base.lib.serializers.common import CommonModelSerializer, ImmutableCommonModelSerializer
```
relrod, line 14: image

relrod (approved):

Issue #111, opened 2024-01-30
Can't log into test_app as a non-staff user


AlanCoding:
runserver

GET http://127.0.0.1:8000/api/social/login/ansible_base-authentication-authenticator_plugins-local__local-database-authenticator/

```
Internal Server Error: /api/social/login/ansible_base-authentication-authenticator_plugins-local__local-database-authenticator/
Traceback (most recent call last):
  File "/home/alancoding/repos/awx/env/lib64/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/repos/awx/env/lib64/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/repos/awx/env/lib64/python3.11/site-packages/django/views/decorators/cache.py", line 62, in _wrapper_view_func
    response = view_func(request, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/repos/awx/env/lib64/python3.11/site-packages/social_django/utils.py", line 63, in wrapper
    return func(request, backend, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/repos/awx/env/lib64/python3.11/site-packages/social_django/utils.py", line 44, in wrapper
    request.backend = load_backend(
                      ^^^^^^^^^^^^^
  File "/home/alancoding/repos/awx/env/lib64/python3.11/site-packages/social_django/utils.py", line 27, in load_backend
    return strategy.get_backend(name, redirect_uri=redirect_uri)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/repos/awx/testing/django-ansible-base/ansible_base/authentication/social_auth.py", line 96, in get_backend
    return Backend(
           ^^^^^^^^
TypeError: AuthenticatorPlugin.__init__() got multiple values for argument 'database_instance'
```

AlanCoding:
[...]

PR #265, opened 2024-04-04
Add a compose file and entrypoint for the test_app

jctanner:
I would like to have test_app in a container and I would like to be able to test the authenticators with the test_app.

default page...
image

logged in via github ...
image

### DIFF
--- a/test_app/scripts/entrypoint.sh
+++ b/test_app/scripts/entrypoint.sh
@@ -3,19 +3,8 @@
 set -e
 set -x
 
-sed -i '/^\[crb\]$/,/^enabled=0$/ s/enabled=0/enabled=1/' /etc/yum.repos.d/centos.repo
-dnf -y install \
-    python3.11 python3.11-pip python3.11-devel gcc openldap-devel \
-    xmlsec1 xmlsec1-openssl xmlsec1-devel libtool-ltdl-devel libpq-devel libpq postgresql
-if [[ ! -d /venv ]]; then
-    python3.11 -m venv /venv
-fi
-
 PIP=/venv/bin/pip
 PYTHON=/venv/bin/python3
-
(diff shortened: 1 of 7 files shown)


test_app/scripts/entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+    xmlsec1 xmlsec1-openssl xmlsec1-devel libtool-ltdl-devel libpq-devel libpq postgresql
+if [[ ! -d /venv ]]; then
+    python3.11 -m venv /venv
+fi
+
+PIP=/venv/bin/pip
+PYTHON=/venv/bin/python3
+
+$PIP install -r requirements/requirements_all.txt
+$PIP install -r requirements/requirements_dev.txt
```
relrod: For consistency with other apps, it might be best to have a real `Dockerfile` and do this there.

test_app/scripts/entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+
+echo "settings.DATABASE ..."
+$PYTHON manage.py shell -c 'from django.conf import settings; print(settings.DATABASES)'
+
+MAX_ATTEMPTS=10
+for i in $(seq 1 $MAX_ATTEMPTS); do
```
relrod, line 13: This is doing a lot of what the bootstrap script does. Would it be best to just run the bootstrap script inside the container? So when you spin the container up, it pre-loads some data already to play with?

jctanner: The bootstrap script wants to do docker commands and munge a postgres container via the make target.

cutwater (approved):

AlanCoding:
This looks like it has some tie-ins to https://github.com/ansible/django-ansible-base/issues/111 possibly? As this was mostly concerned with ability to test authentication?

PR #266, opened 2024-04-05
[activitystream] EntryReadOnlyViewSet permissions

relrod:
Allow for settings-changeable permissions for EntryReadOnlyViewSet.

This is necessary because some apps might not be using RBAC and in that case might need to be able to limit access to the activity stream to superusers (or some other permission class).

### DIFF
--- a/ansible_base/activitystream/views.py
+++ b/ansible_base/activitystream/views.py
@@ -1,10 +1,22 @@
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+from django.utils.module_loading import import_string
 from django.utils.translation import gettext_lazy as _
-from rest_framework import permissions
 from rest_framework.viewsets import ReadOnlyModelViewSet
 
 from ansible_base.activitystream.models import Entry
 from ansible_base.activitystream.serializers import EntrySerializer
 from ansible_base.lib.utils.views.django_app_api import AnsibleBaseDjangoAppApiView
+
+
+def _permission_classes():
+    entry_permission_classes = []
+    for cls in settings.ANSIBLE_BASE_ACTIVITYSTREAM_VIEW_PERMISSION_CLASSES:
+        try:
+            entry_permission_classes.append(import_string(cls))
+        except ImportError as e:
+            raise ImproperlyConfigured(f"[ANSIBLE_BASE_ACTIVITYSTREAM_VIEW_PERMISSION_CLASSES] Could not find permission class '{cls}'.") from e
+    return entry_permission_classes
 
 
 class EntryReadOnlyViewSet(ReadOnlyModelViewSet, AnsibleBaseDjangoAppApiView):
@@ -14,7 +26,7 @@ class EntryReadOnlyViewSet(ReadOnlyModelViewSet, AnsibleBaseDjangoAppApiView):
 
     queryset = Entry.objects.all()
     serializer_class = EntrySerializer
-    permission_classes = [permissions.IsAuthenticated]
+    permission_classes = _permission_classes()
 
     def get_view_name(self):
         return _('Activity Stream Entries')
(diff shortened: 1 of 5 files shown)


ansible_base/activitystream/views.py
```diff
@@ -1,20 +1,32 @@
...
+
 class EntryReadOnlyViewSet(ReadOnlyModelViewSet, AnsibleBaseDjangoAppApiView):
     """
     API endpoint that allows for read-only access to activity stream entries.
     """
 
     queryset = Entry.objects.all()
     serializer_class = EntrySerializer
-    permission_classes = [permissions.IsAuthenticated]
+    permission_classes = _permission_classes()
```
AlanCoding: Something rubs me the wrong way about this settings-based solutions. My 2 cents is that DAB can and _should_ be more prescriptive than this. Also, it's disjointed, because RBAC will define the queryset. If this is about querysets, why are you messing with the permission class? I think you should mess with both, but that puts solutions in different places hard to find.

I'd rather dynamically generate the permission class.

https://github.com/encode/django-rest-framework/blob/63063da0820e23ef0edbf92a3031103a6c2ce254/rest_framework/views.py#L274-L278

They even give an example in docs

https://www.django-rest-framework.org/api-guide/viewsets/#introspecting-viewset-actions

```python
def get_permissions(self):
    """
    Instantiates and returns the list of permissions that this view requires.
    """
    if self.action == 'list':
        permission_classes = [IsAuthenticated]
    else:
        permission_classes = [IsAdminUser]
    return [permission() for permission in permission_classes]
```

But that has an `if`, what's our `if` here? I would argue checking `"ansible_base.rbac" in settings.INSTALLED_APP` is the right thing to do.

So I'm saying, at least in this case, write an app to be compatible with using another DAB app and without that other app. For activity stream this is fairly simple and can be accomplished easily. At some point that might not be true, I realize. This is a big philosophical pill to swallow.

ansible_base/activitystream/views.py
```diff
@@ -14,7 +16,17 @@ class EntryReadOnlyViewSet(ReadOnlyModelViewSet, AnsibleBaseDjangoAppApiView):
...
     serializer_class = EntrySerializer
-    permission_classes = [permissions.IsAuthenticated]
+
+    def get_permissions(self):
+        """
+        If the RBAC app is enabled, then we delegate to it to manage permissions.
+        Otherwise, we require superuser status.
+        """
+        if 'ansible_base.rbac' in settings.INSTALLED_APPS:
+            permission_classes = [permissions.IsAuthenticated]
```
AlanCoding, line 26: Let's also get an issue filed to make the required changes to queryset (or more likely, `filter_queryset`) to support this.

test_app/tests/activitystream/test_api.py
```diff
@@ -33,3 +34,39 @@ def test_activitystream_api_read_only(admin_api_client, user):
...
+    whether or not RBAC is enabled.
+
+    If RBAC is enabled, then it locks down permissions on its own, so we allow IsAuthenticated.
+    If RBAC is not enabled, then we require IsSuperuser.
+    """
+    url = reverse("activitystream-list")
+
+    if 'ansible_base.rbac' in settings.INSTALLED_APPS:
+        if not has_rbac_app:
+            settings.INSTALLED_APPS.remove('ansible_base.rbac')
```
AlanCoding, line 58: Could you use `override_settings` to remove it less... destructively?

AlanCoding (approved):

test_app/tests/activitystream/test_api.py
relrod: This is using the pytest [settings](https://pytest-django.readthedocs.io/en/latest/helpers.html#settings) fixture, so it reverts automatically

ansible_base/activitystream/views.py
relrod: #273

PR #267, opened 2024-04-05
[activitystream] Never store encrypted values

### LINKED ISSUE #252: [activity stream] don't store encrypted fields in activity stream entries
### Bug Summary

As title.

relrod:
- Skip AbstractUser.password fields and CommonModel.encrypted_fields fields. Replace their values with ENCRYPTED_STRING instead.
Fixes #252


ansible_base/lib/utils/models.py
```diff
@@ -102,6 +103,15 @@ def current_user_or_system_user():
     return user
 
 
+def is_encrypted_field(model, field_name):
+    from django.contrib.auth.models import AbstractUser
+
+    if isinstance(model, AbstractUser) and field_name == 'password':
+        return True
+
+    return field_name in getattr(model, 'encrypted_fields', [])
```
AlanCoding, line 120: What if a field was decorated with `prevent_search`?

https://github.com/ansible/django-ansible-base/blob/392ea7713b4b7b61191982127ed17dd31968f6e8/ansible_base/lib/utils/models.py#L54-L68

relrod: Sure, added a check for that.

AlanCoding (approved):

PR #269, opened 2024-04-05
Make {created,modified}_by SET_NULL on FK deletion

### LINKED ISSUE #198: The `created_by` and `modified_by` field DO_NOTHING behaviors violate postgres constraints
### Bug Summary

Field definition:

https://github.com/ansible/django-ansible-base/blob/774644429eb63303e8e30391c326afe7562c8524/ansible_base/lib/abstract_models/common.py#L56-L64

Test corresponding to this bug:

https://github.com/ansible/django-ansible-base/pull/193

Steps:
 - user creates something
 - delete that user
 - get integrity error

If you're within a transaction, you may not see this error until it ends - thus some hacks in the proposed test.

This is a far-reaching issue, because this is a bug in the abstract base model itself, and now many models carry these fields. There's also a time premium on this, because if we lock in current migration history, that makes it much more difficult to solve later.

internal - AAP-21209

relrod:
Fixes #198 

maybe


ansible_base/rbac/migrations/0002_alter_roledefinition_created_by_and_more.py
```diff
@@ -0,0 +1,36 @@
...
+            model_name='roleteamassignment',
+            name='created_by',
+            field=models.ForeignKey(default=None, editable=False, help_text='The user who created this resource', null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='%(app_label)s_%(class)s_created+', to=settings.AUTH_USER_MODEL),
+        ),
+        migrations.AlterField(
+            model_name='roleuserassignment',
+            name='created_by',
+            field=models.ForeignKey(default=None, editable=False, help_text='The user who created this resource', null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='%(app_label)s_%(class)s_created+', to=settings.AUTH_USER_MODEL),
+        ),
+    ]
```
AlanCoding, line 36: At least for RBAC until the first AWX community release, I would still like to squash this into 0001, and AFAIK this doesn't involved a real schema change either.

If you think the migration files are helping in the other cases, I think that's over-promising, because this change will cause new `makemigration` changes in other apps as  well.

You don't have to do this now, but I'd like to consolidate before we release formally.

AlanCoding (approved):
Ping @Dostonbek1 this is going to create `makemigrations` changes in the eda-server RBAC branch due to the use of abstract models like `AbstractTeam`

Ping @john-westcott-iv this will create changes in Gateway as well, for same reason.

ansible_base/rbac/migrations/0002_alter_roledefinition_created_by_and_more.py
relrod: Yeah, I agree with squashing before release.

PR #274, opened 2024-04-05
Try a hack to resolve mapping errors with social users.

jctanner:
https://issues.redhat.com/browse/AAP-22425

The issue seems to be more of a problem with how the User and AuthenticatorUser models are auto-connected / designed.

Open to suggestion on how to do this better.


john-westcott-iv:
Maybe unnecessary with #280

PR #278, opened 2024-04-06
[dynamic_urls] Avoid masking import errors

relrod:
Also fix rest_filters whose urls.py could never get imported because
it referenced a name that wasn't defined.

### DIFF
--- a/ansible_base/lib/dynamic_config/dynamic_urls.py
+++ b/ansible_base/lib/dynamic_config/dynamic_urls.py
@@ -17,4 +17,4 @@ for app in getattr(settings, 'INSTALLED_APPS', []):
             for url_type in ['api_version_urls', 'root_urls', 'api_urls']:
                 globals()[url_type].extend(getattr(url_module, url_type, []))
         except ImportError:
-            logger.debug(f'Module {app} does not specify urls.py')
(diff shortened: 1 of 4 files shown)


AlanCoding (changes requested):
I don't like this. It confuses the contract - the expectations for what an app does.

Looking over the apps, I see that the `urls.py` module is added even in apps that add no URLs. That aligns well with your intent, but not your code here. This _is_ the de-facto standard practice that urls.py is _expected_ to exist. That being the case, you should remove the `try-except` and update the docs if it says otherwise.

With this code, it's written as if it's fine for an app to not have urls, but then does an exception log. It shouldn't, exception logging shouldn't happen on every startup, which this would imply.

relrod:
It feels weird to me to require `urls.py` even if no URLs are exposed.

What if instead we check if a urls module even exists? If it exists, we import it or fail loudly so we don't mask errors. If it doesn't exist, then we log a debug log and move on.

relrod:
> What if instead we check if a urls module even exists? If it exists, we import it or fail loudly so we don't mask errors. If it doesn't exist, then we log a debug log and move on.

I updated the PR to do this

AlanCoding (approved):

Issue #277, opened 2024-04-06
[activity stream] Be able to filter JSONField contents


relrod:
Right now, filtering doesn't seem to work for `JSONField`:

If I go to this URL in test_app: http://localhost:8000/api/v1/activitystream/?changes__added_fields__id__exact=2

I get a 400 with:

``` json
{
    "detail": "No related model for field added_fields."
}

```

This comes from our `rest_filters` source. It seems we assume that double-underscore means related model, but for `JSONField`, it doesn't mean that.

This makes it hard to do cool filtering on activity stream entries.

relrod:
Maybe @AlanCoding has some thoughts here, since most of our current filtering code came out of AWX.

AlanCoding:
I don't think this was possible in AWX. You could write a special case for this. The JSONFields in AWX were not real JSONFields @jbradberry can probably elaborate more.

relrod:
I played around a bit today with our `rest_filters` logic and I can get JSONField "working", but the issue ends up coming down to value types.

JSONFields are schemaless. The value for the filter always comes in as a string from the querystring.

So for `?changes__added_fields__id__exact=2` we end up generating `Q(changes__added_fields__id__exact="2")`

Making 2 an int there makes that a very valid query. But we have no way of knowing if it really should be an int or a string or what...

So,  with my playing around, I have it working for **string** values. You can do `?changes__added_fields__email__iexact=admin@stuff.invalid` and get what you expect. But it doesn't work for `id` or any other non-string values.

I'm not sure what the answer is here. Any solution I can think of is going to get much more complex than I'm comfortable with maintaining. But without a solution, you can't really expose a truly searchable interface to Activity Stream which is a little unfortunate.

relrod:
~We could do something custom for Activity Stream... We *know* the model the `changes` blob relates to. We can do "stuff" to see what the model expects each field to be, and try to cast (there's already machinery for that, `value_to_python_for_field`).~

~That's a very "Activity Stream-only" solution, but it could work - DRF lets us set the filter backend at the view(set) level.~

After more thought, I don't think this works at all. We *don't* know the model. I mean, we store that, but you'd have to load that in from every `Entry` and then for each `Entry`, resolve that from a content_type id into its class name and filter based on each field. That's going to be *extremely* slow.

I don't know of a good way to do this.

AlanCoding:
This works

http://127.0.0.1:8000/api/v1/activitystream/?changes__icontains=dog

[...]

PR #279, opened 2024-04-07
Rework how Activity Stream saves data

relrod:
This is mostly just me playing around with an idea and seeing what
happens. This was inspired by #277 which documents that it's very hard
to deep-filter a JSONField.

The root of the problem is that JSONField is unstructured, so we don't
have type information up front about the values inside, and that -
with Django's Q-querying for JSONFields, not having that type
information means it's impossible to coerce the user's search
parameters (always a string) into the type we need for matching to
work in all the non-string cases.

The approach taken in this commit does things differently. Instead of
storing a JSONField with the changes, I introduce a FieldChange model,
which is essentially a glorified tuple of (stream_entry, operation,
field_name, old_value, new_value).

operation is necessary because it's how we know if fields were added
or removed from the model itself, versus being set explicitly null.

When we store an activity stream entry, we also store the associated
field changes. We always store the values as strings, so that the
filtering problem in #277 is never an issue, matching can always
work. In the new serializer, we reverse the string back into its
python value using Django's Field#to_python method. This means we
render JSON that has the right types for things like numbers and
booleans, but internally they're stored as strings, and can be
filtered with the querystring filters as you'd expect.

Effectively this lets us do things like:

?changes__field_name__exact=email&changes__new_value=admin@stuff.invalid

or

?changes__field_name__exact=id&changes__new_value=2

and other cool queries like that, into the actual fields being
changed.

This is a completely different approach from the JSONField approach,
but there's really no reason that we *had* to use a JSONField here. So
this attempts to explore what happens when we don't.


relrod:
If we do take this, the AS docs need to be updated.

relrod:
Closing in favor of #282.

PR #280, opened 2024-04-08
Alter authentication

john-westcott-iv:
This change does a couple things:
* Allows a created user to be tied to an Authenticator through AuthenticatorUser on the first login. Previously, if you create user "anna" in the system and then logged in you would end up with the user "anna<hash>".
* Ensures a standard pattern where users are considered unique from all sources. The first source to authenticate a user "rick" gets that username, if another authenticator authenticates a user "rick" (same username) the second user will be "rick<hash>". 
* Better aligns the non-sso login process by adding function docs and documentation around what password authenticators should be doing.
* Adds claims process to local and tacacs authenticators.
* Ensures get_user_name 
* Splits out general authentication utils from the `utils/claims` file.

TODO:
- [ ] Test auth with TACACS (reg and collision)
- [ ] Test auth with LDAP (reg and collision)
- [ ] Test auth with social (reg and collision)
- [x] Address any unit tests and fix coverage (splitting out claims and authentication utils)
- [x] Check user hash creation from social_auth to ensure that it will be consistent the way we are using it


ansible_base/authentication/models/authenticator_user.py
```diff
@@ -21,7 +21,7 @@ class AuthenticatorUser(AbstractUserSocialAuth):
     access_allowed = models.BooleanField(default=None, null=True)
 
     @classmethod
-    def create_social_auth(cls, user, uid, slug):
+    def create_social_auth(self, user, uid, slug):
```
cutwater: Why this change? This is a class method, isn't it?

ansible_base/authentication/utils/claims.py
```diff
@@ -3,20 +3,18 @@
...
-from ansible_base.authentication.social_auth import AuthenticatorStorage, AuthenticatorStrategy
+from ansible_base.authentication.models import Authenticator, AuthenticatorMap
 
 from .trigger_definition import TRIGGER_DEFINITION
 
 logger = logging.getLogger('ansible_base.authentication.utils.claims')
 
 
-def create_claims(authenticator: Authenticator, username: str, attrs: dict, groups: list) -> (bool, bool, dict, list):
+def create_claims(authenticator: Authenticator, username: str, attrs: dict, groups: list[str]) -> (bool, bool, dict, list):
```
cutwater: The return type annotation is incorrect. Tuple of types is invalid annotation in general and this function returns dictionary on line 84.

ansible_base/authentication/utils/authentication.py
```diff
@@ -0,0 +1,120 @@
...
+        return new_username
+
+    # We didn't have an exact match but no other provider is servicing this uid so lets return that for usage
+    logger.info(f"Authenticator {authenticator.name} is is able to authenticate user {uid} as {uid}")
+    return uid
+
+
+def get_or_create_authenticator_user(
+    username: str, authenticator: Authenticator, user_details: dict = {}, extra_data: dict = {}
+) -> Tuple[AbstractUser, Optional[AuthenticatorUser], bool]:
```
cutwater: see line 96

```suggestion
) -> Tuple[Optional[AbstractUser], Optional[AuthenticatorUser], Optional[bool]]:
```

ansible_base/authentication/utils/authentication.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+        return new_username
+
+    # We didn't have an exact match but no other provider is servicing this uid so lets return that for usage
+    logger.info(f"Authenticator {authenticator.name} is is able to authenticate user {uid} as {uid}")
+    return uid
+
+
+def get_or_create_authenticator_user(
+    username: str, authenticator: Authenticator, user_details: dict = {}, extra_data: dict = {}
```
cutwater: Don't use mutable types as a default value.

Python binds default values to parameters at function definition time, which happens once when the function is created and not each time when it's executed.
Therefore it creates implicit global variable that when reused can lead to unexpected side effects and hard to debug bugs:

```
>>> def foo(a, x=[]):
...      x.append(a)
...      print(x)
... 
>>> foo(1)
[1]
>>> foo(2)
[1, 2]
```

General best practice is to avoid usage of mutable values as function defaults.

P.S. There are flake8 and ruff checks that prevent it. I think it will benefit the project if we enable them.

ansible_base/authentication/utils/authentication.py
```diff
@@ -0,0 +1,120 @@
...
+        if auth_user is not None:
+            logger.error(
+                f'Authenticator {authenticator.name} attempted to create an AuthenticatorUser for {username}'
+                f' but that id is already tied to authenticator {auth_user.provider.name}'
+            )
+            return None, None, None
+
+    # ensure the authenticator isn't trying to pass along a cheeky is_superuser in user_details
+    allowed_keys = ["first_name", "last_name", "email"]
+    details = {k: user_details.get(k, "") for k in allowed_keys if k}
```
cutwater: Why if condition in dict comprehension that iterates over list of constants?
```
 details = {k: user_details.get(k, "") for k in allowed_keys if k}
                                                             ^^^^
```

ansible_base/authentication/utils/authentication.py
```diff
@@ -0,0 +1,120 @@
...
+        auth_user = AuthenticatorUser.objects.filter(uid=username).first()
+        if auth_user is not None:
+            logger.error(
+                f'Authenticator {authenticator.name} attempted to create an AuthenticatorUser for {username}'
+                f' but that id is already tied to authenticator {auth_user.provider.name}'
+            )
+            return None, None, None
+
+    # ensure the authenticator isn't trying to pass along a cheeky is_superuser in user_details
+    allowed_keys = ["first_name", "last_name", "email"]
```
cutwater: If that's a constant, better to move it to module level and name appropriately

See https://peps.python.org/pep-0008/#constants

cutwater (changes requested):

ansible_base/authentication/utils/authentication.py
```diff
@@ -0,0 +1,120 @@
...
+logger = logging.getLogger('ansible_base.authentication.utils.authentication')
+
+
+def get_local_username(user_details: dict) -> str:
+    """
+    Converts the username provided by the backend to one that doesn't conflict with users
+    from other auth backends.
+    """
+
+    class FakeBackend:
```
cutwater: Is there a reason to use nested class, which is created on every function call for this function?

ansible_base/authentication/models/authenticator_user.py
john-westcott-iv: Thanks, reverted; I think that was an auto-pilot change.

ansible_base/authentication/utils/authentication.py
john-westcott-iv: This code is actually not new but was split out from ansible_base/authentication/utils/claims.py. I don't know why it was initially implemented this way but we can move it out since its really nothing dynamic.=

ansible_base/authentication/utils/authentication.py
john-westcott-iv: Good catch. Fixing.

ansible_base/lib/testing/fixtures.py
```diff
@@ -130,16 +130,19 @@ def expected_log(no_log_messages):
...
     def f(patch, severity, substr, assert_not_called=False):
-        with mock.patch(patch) as logger:
+        with mock.patch(f'{patch}.{severity}') as logger:
             with no_log_messages():
                 yield
-            sev_logger = getattr(logger, severity)
+
+            call_count = 0
+            for call in logger.call_args_list:
+                if substr in call.args[0]:
```
relrod: Hmm, yeah that's a good catch, that if the logger is called with other messages in the same block, we'd fail. :+1: good fix

ansible_base/lib/testing/fixtures.py
```diff
@@ -130,16 +130,19 @@ def expected_log(no_log_messages):
...
+                    call_count = call_count + 1
```
relrod:
```suggestion
                    call_count += 1
```

or just replace the whole loop with

```python
call_count = len([call for call in logger.call_args_list if substr in call.args[0]])
```

cutwater: No need to construct a temporary list that is immediately discarded.
```
call_count = sum(1 for call in logger.call_args_list if substr in call.args[0])
```

ansible_base/authentication/utils/authentication.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+        return new_username
+
+    # We didn't have an exact match but no other provider is servicing this uid so lets return that for usage
+    logger.info(f"Authenticator {authenticator.name} is is able to authenticate user {uid} as {uid}")
+    return uid
+
+
+def get_or_create_authenticator_user(
+    username: str, authenticator: Authenticator, user_details: dict = dict(), extra_data: dict = dict()
```
cutwater: Replacing dict literal `{}` with a dict call `dict()` doesn't change anything. The dictionary object is still created at the function definition time. The common pattern in Python for creating mutable function default parameter values is using an immutable default with a condition. For example:

```
def foo(data: Optional[dict] = None):
    if data is None:
        data = {}
```

ansible_base/authentication/utils/authentication.py
```diff
@@ -0,0 +1,116 @@
...
+        auth_user = AuthenticatorUser.objects.filter(uid=username).first()
+        if auth_user is not None:
+            logger.error(
+                f'Authenticator {authenticator.name} attempted to create an AuthenticatorUser for {username}'
+                f' but that id is already tied to authenticator {auth_user.provider.name}'
+            )
+            return None, None, None
+
+    # ensure the authenticator isn't trying to pass along a cheeky is_superuser in user_details
+    details = {k: user_details.get(k, "") for k in ["first_name", "last_name", "email"] if k}
```
cutwater: 1. The `if k` condition is still here, seems like a mistake. It's either meant to do something else or it simply can be deleted.
2. Minor(nit): Personally I'd move this list to a constant:

```python
USER_DETAILS_FIELDS = ["first_name", "last_name", "email"] 

# ...

def get_or_create_authenticator_user(...):
    # ...
    details = {k: user_details.get(k, "") for k in USER_DETAILS_FIELDS}
```

cutwater (changes requested):

ansible_base/lib/testing/fixtures.py
relrod: Ah, `sum()` with 1 works... I originally left out the list, but count() won't work on a generator so I changed it.

ansible_base/authentication/utils/authentication.py
john-westcott-iv: I thought about moving it to a constant but we don't use it anywhere else so I was just going to put it inline. 
The goal of this was to make sure we only accept specific fields out of the user_fields instead of users being able to put junk in. I don't remember why the `if` was there but yea, it seems useless at this point in this context.

ansible_base/authentication/utils/authentication.py
john-westcott-iv: I don't think I want to make it optional but the `()` were typos. Thanks for catching (again).

cutwater: In this case you should remove the default values. The `user_detauls: dict = dict` is incorrect. This assigns `dict` type as a default value, which I think not what was intended here.

john-westcott-iv: Default removed.

ansible_base/authentication/utils/authentication.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    username = get_username(strategy=AuthenticatorStrategy(AuthenticatorStorage()), details=user_details, backend=FakeBackend())
+
+    return username['username']
+
+
+def determine_username_from_uid_social(**kwargs: dict) -> dict:
+    uid = kwargs.get('details', {}).get('username', None)
+    if not uid:
+        raise AuthException('Unable to get associated username from: {{ kwargs.get("details", None) }}')
```
cutwater: Looks like an f-string was meant to be used here.

cutwater (approved):
Minor comment, otherwise LGTM

ansible_base/authentication/utils/authentication.py
john-westcott-iv: Good catch, this actually should have been a translated string.

relrod (approved):

PR #282, opened 2024-04-08
[activity stream] Store string-coerced values

### LINKED ISSUE #277: [activity stream] Be able to filter JSONField contents

relrod:
...and convert them back to their expected types at serialization time.

This is an alternative to #279 

In _this_ approach, unlike #279, we keep the `JSONField` around, but _like_ #279, we still only store string values.

In serialization time, we convert them back to their expected Pythonic types.

This is a smaller change than #279 and still ends with us being able to search into JSONFields (#277), although I think the changes to `rest_filters` might be backwards-incompatible.

Anyway, I just wanted to put this up for discussion as well as another alternative.

Fixes #277


relrod:
I've updated this to make `rest_filters` be backwards compatible for things like AWX because we don't want to break what users are currently expecting to work.

I added a test to test the backwards compatible behavior.

I introduced a class var on `FieldLookupBackend` that subclasses can override, and subclassed it in activitystream. This allows the JSONField traversal to work, and I added a test showing that.

ansible_base/activitystream/signals.py
```diff
@@ -39,7 +39,7 @@ def _store_activitystream_entry(old, new, operation):
 
     excluded = getattr(new, 'activity_stream_excluded_field_names', [])
     limit = getattr(new, 'activity_stream_limit_field_names', [])
-    delta = diff(old, new, exclude_fields=excluded, limit_fields=limit)
+    delta = diff(old, new, exclude_fields=excluded, limit_fields=limit, all_values_as_strings=True)
```
john-westcott-iv, line 42: Should this be a hard coded `True` here?

john-westcott-iv (approved):

ansible_base/activitystream/signals.py
relrod: That is what actually triggers storing everything as a string, so yeah it's intentionally `True` here.

relrod:
Just to note it here:

John questioned to me performance of doing the type coercion back to Pythonic types in the serializer. I did some testing with turning `DEFAULT_PAGE_SIZE` to 500 and loading the activity stream after putting a few thousand objects in the database (thanks to the bootstrap script).

I did testing against devel and against this branch, and the timing was not at all noticeably different between the two branches. On both branches it still took ~3 seconds to render 500 entries per page. (And that's slow, but this branch doesn't make it any slower.)

ansible_base/activitystream/serializers.py
```diff
@@ -20,6 +22,7 @@ class Meta:
 
     content_type_model = serializers.SerializerMethodField()
     related_content_type_model = serializers.SerializerMethodField()
+    changes = serializers.SerializerMethodField()
```
AlanCoding, line 25: I don't often see a `SerializerMethodField` that's the same name as the field, but since it works that seems like the best for this problem.

ansible_base/activitystream/serializers.py
```diff
@@ -29,6 +32,32 @@ def get_related_content_type_model(self, obj):
         if obj.related_content_type:
             return obj.related_content_type.model
 
+    def _field_value_to_python(self, entry, field_name, value):
+        model = entry.content_type.model_class()
+        field = model._meta.get_field(field_name)
+        return field.to_python(value)
```
AlanCoding, line 38: smart

AlanCoding (approved):
After reading over the code, I continue to believe that this is the best available solution. I see how you are hamstrung by how Django handles jsonfields. In the long term, jsonfield allows handling of more advanced cases with more logic, if it comes to that.

PR #288, opened 2024-04-08
Include activity stream link in root test_app view

AlanCoding:


relrod (approved):
This had been on my list, thanks.

But your branch name confuses me :rofl:

PR #292, opened 2024-04-09
Exclude last_login from test_app's User model activity stream entries

### LINKED ISSUE #285: Updates to `last_login` should not create activitystream entries
Steps:
 - log in
 - check activity stream

Expectation:

It is not the job of _activity stream_ to track my logins.

Actual result:

It shows the timestamp update of `last_login`

relrod:
Fixes #285

### DIFF
--- a/ansible_base/activitystream/signals.py
+++ b/ansible_base/activitystream/signals.py
@@ -7,20 +7,6 @@ from ansible_base.lib.utils.models import current_user_or_system_user, diff
 logger = logging.getLogger('ansible_base.activitystream.signals')
 
 
-def _always_exclude_fields():
-    """
-    Returns a list of 2-tuples of (model_or_base_class, 'field_name') that
-    should always be excluded from activity stream entries.
-    """
-
-    # The whole reason this is a function is to avoid circular imports.
-    from django.contrib.auth.models import AbstractUser
-
-    return [
-        (AbstractUser, 'last_login'),
-    ]
-
-
 class ActivityStreamEnabled(threading.local):
     def __init__(self):
         self.enabled = True
@@ -53,12 +39,6 @@ def _store_activitystream_entry(old, new, operation):
 
     excluded = getattr(new, 'activity_stream_excluded_field_names', [])
     limit = getattr(new, 'activity_stream_limit_field_names', [])
-
-    # Always exclude certain fields
-    for cls, field_name in _always_exclude_fields():
-        if isinstance(old, cls) or isinstance(new, cls):
-            excluded.append(field_name)
-
     delta = diff(old, new, exclude_fields=excluded, limit_fields=limit)
 
     if not delta:
--- a/ansible_base/activitystream/signals.py
+++ b/ansible_base/activitystream/signals.py
@@ -7,6 +7,20 @@ from ansible_base.lib.utils.models import current_user_or_system_user, diff
 logger = logging.getLogger('ansible_base.activitystream.signals')
 
 
+def _always_exclude_fields():
+    """
+    Returns a list of 2-tuples of (model_or_base_class, 'field_name') that
+    should always be excluded from activity stream entries.
+    """
+
+    # The whole reason this is a function is to avoid circular imports.
+    from django.contrib.auth.models import AbstractUser
+
+    return [
+        (AbstractUser, 'last_login'),
+    ]
+
+
 class ActivityStreamEnabled(threading.local):
     def __init__(self):
         self.enabled = True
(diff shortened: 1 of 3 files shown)


ansible_base/activitystream/signals.py
```diff
@@ -7,6 +7,20 @@
 logger = logging.getLogger('ansible_base.activitystream.signals')
 
 
+def _always_exclude_fields():
+    """
+    Returns a list of 2-tuples of (model_or_base_class, 'field_name') that
+    should always be excluded from activity stream entries.
```
AlanCoding: That's just the thing, I think we need these exceptions for more purposes than just activity stream (modified_by for example).

ansible_base/activitystream/signals.py
```diff
@@ -7,6 +7,20 @@
...
+    """
+
+    # The whole reason this is a function is to avoid circular imports.
+    from django.contrib.auth.models import AbstractUser
+
+    return [
+        (AbstractUser, 'last_login'),
```
AlanCoding: wouldn't it be easier to just add an ignore list on the `AbstractUser` model?

relrod: We could, but I wasn't sure about modifying Django base classes, it felt wrong to me.

ansible_base/activitystream/signals.py
relrod: Can you elaborate here? I don't understand the modified_by example.

ansible_base/activitystream/signals.py
AlanCoding: As long as you give it a name that won't conflict with anything else, you should be fine. Of course, there is a distinction between the DAB AbstractUser and the Django AbstactUser. But I would prefer to do this DAB-specific, because it gives apps using this library the ability to change the behavior if they wish.

Going by history, there is likely to be a lot of iteration as QE looks at the activity stream entries and goes "that shouldn't be there!" You don't want to require a DAB patch, which is also why I think an ignore list on the model is a good idea.

relrod: There's already a model-specific ignore list, models can set `activity_stream_excluded_field_names` and `activity_stream_limit_field_names` to exclude fields from showing up, or limit it to a subset of fields that can show up.

DAB doesn't have an `AbstractUser` concept. We could introduce one, but I think there was a conscious decision not to, a while back.

So I think the options are:

- Do it as-is in this PR so we can maintain a list of models -> fields to always exclude
- Do it on the Django `AbstractUser` model somewhere, by setting `AbstractUser.activity_stream_excluded_field_names = ['last_login']`
- Make every app set `activity_stream_excluded_field_names` in their own User model (but that doesn't work for something like AWX which doesn't use its own `User` model).

AlanCoding: > There's already a model-specific ignore list, models can set activity_stream_excluded_field_names and activity_stream_limit_field_names to exclude fields from showing up, or limit it to a subset of fields that can show up.

Maybe I'm just not getting it. Why wouldn't you use what you already developed and add "last_login" to `activity_stream_excluded_field_names`?

So I would say the option (3). I'm not worried about AWX. Yes, it uses `auth.User`, but it already adds _many_ things to the class in its models `__init__.py`. Take summary_fields for example. It's totally not pattern-breaking at all to glue that list onto an imported User model.

ansible_base/activitystream/signals.py
AlanCoding: Resolved elsewhere, problems were not as intertwined as I expected.

ansible_base/activitystream/signals.py
relrod: > Why wouldn't you use what you already developed and add "last_login" to `activity_stream_excluded_field_names`?

Because you opened #285 :sweat_smile: 

If we want this to be something that the apps do to their own user model, then there's nothing to do here, except maybe exclude it in test_app's User model, right?

relrod: > So I would say the option (3).

I updated this to just exclude the field in test_app's User model.

AlanCoding (approved):

ansible_base/activitystream/signals.py
AlanCoding: It was filed from the experience of using test_app. Sometimes it is hard to separate that from other issues... but still an issue.

relrod: It's all good :smile: It's nice if we don't have to maintain Yet Another List of hardcoded exceptions in DAB.

PR #294, opened 2024-04-09
Remove User FK from Authenticator in favor of AuthenticatorUser

john-westcott-iv:
We had a ForeignKey to User on Authenticator but also had a link table with AuthenticatorUser that linked User and Authenticator. This removes the User ForeignKey on Authenticator but keeps the same functionality.


jay-steurer (approved):

ansible_base/authentication/views/authenticator_users.py
```diff
@@ -0,0 +1,40 @@
...
+        logger.debug("ANSIBLE_BASE_USER_VIEWSET was not specified")
+
+    return BadViewSet()
+
+
+class AuthenticatorPluginRelatedUsersView(get_parent_view()):
+    def get_queryset(self, **kwargs):
+        authenticator_id = kwargs.get('authenticator_id', None)
+        if authenticator_id is None:
+            return get_user_model().objects.filter(pk=-1)
```
relrod: Maybe `get_user_model().objects.none()` here to avoid a round trip to the db server. It should do the same thing (return an empty queryset).

john-westcott-iv: I had initially tried `none()` first but I was getting errors but they now seem to be working. Code updated.

relrod (approved):

PR #295, opened 2024-04-09
[translations] update ansible_base/rbac 

thedoubl3j:
Not like the last PR, this one should be a little bit more straight forward. 

Testing steps:

```pip freeze | grep Django pip install Django```
3. make a locale directory (makemessages looks for one and we don't have one configured yet so just create your own) ```mkdir locale```
4. run the makemessages command check the output to see if strings were extracted correctly.
```django-admin makemessages -l es```
```vim locale/es/LC_MESSAGES/django.po```

### DIFF
--- a/ansible_base/rbac/api/serializers.py
+++ b/ansible_base/rbac/api/serializers.py
@@ -189,7 +189,7 @@ class BaseAssignmentSerializer(CommonModelSerializer):
         return fields
 
     def raise_id_fields_error(self, field1, field2):
-        msg = _(f'Provide exactly one of {self.actor_field} or {self.actor_field}_ansible_id')
+        msg = _('Provide exactly one of %(actor_field)s or %(actor_field)s_ansible_id') % {'actor_field': self.actor_field}
         raise ValidationError({self.actor_field: msg, f'{self.actor_field}_ansible_id': msg})
 
     def get_by_ansible_id(self, ansible_id, for_field):
@@ -230,7 +230,10 @@ class BaseAssignmentSerializer(CommonModelSerializer):
             obj = self.get_by_ansible_id(validated_data.get('object_ansible_id'), for_field='object_ansible_id')
             if permission_registry.content_type_model.objects.get_for_model(obj) != role_definition.content_type:
                 raise ValidationError(
-                    {'object_ansible_id': _(f'Object type of {obj._meta.model_name} does not match role type of {role_definition.content_type.model}')}
+                    {
+                        'object_ansible_id': _('Object type of %(model_name)s does not match role type of %(role_definition)s')
+                        % {'model_name': obj._meta.model_name, 'role_definition': role_definition.content_type.model}
+                    }
                 )
         return obj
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: ansible_base/rbac/api/related.py [hunk 1]
```diff
@@ -34,9 +34,9 @@ def related_permission_fields(cls):
 
 def validate_field_data(field_name: str, data: dict):
     if f'{field_name}_id' in data:
-        raise RuntimeError(f'Expected model_to_dict format of data, received {data}')
+        raise RuntimeError(_('Expected model_to_dict format of data, received %(data)s') % {'data': data})
     if field_name in data and isinstance(data[field_name], Model):
-        raise RuntimeError(f'Expected primary key for {field_name} but received {data[field_name]}')
+        raise RuntimeError(_('Expected primary key for %(field_name)s but received %(data_field_name)s') % {'field_name': field_name, 'data_field_name': data[field_name]})
```
AlanCoding: Do we really need translations for errors? This will only show up in server logs, and this in particular should only show up when doing development (like adding DAB as a library).

ansible_base/rbac/api/serializers.py
```diff
@@ -180,7 +180,7 @@ def get_fields(self):
         return fields
 
     def raise_id_fields_error(self, field1, field2):
-        msg = _(f'Provide exactly one of {self.actor_field} or {self.actor_field}_ansible_id')
+        msg = _('Provide exactly one of %(actor_field)s or %(actor_fields)s_ansible_id') % {'actor_field': self.actor_field}
```
AlanCoding: I don't know how I let such a rookie mistake in!

ansible_base/rbac/api/views.py
```diff
@@ -38,7 +39,7 @@ def get_serializer_class(self):
 
     def _error_if_managed(self, instance):
         if instance.managed is True:
-            raise ValidationError('Role is managed by the system')
+            raise ValidationError(_('Role is managed by the system'))
```
AlanCoding, line 42: Unlike my prior comment here, these errors will be surfaced by the API :+1:

ansible_base/rbac/api/related.py
thedoubl3j: @AlanCoding for that I am unsure. I always want to cast a wide net and then make it narrower if needed so if we believe these won't show up in a user facing manner, I am happy to remove. A lot of this was just "we are doing some kind of error handling that isn't logging, translate it"

ansible_base/rbac/api/serializers.py
AlanCoding: `%(actor_fields)` --> `%(actor_field)`

AlanCoding (changes requested):
plural actor_fields key error

ansible_base/rbac/api/serializers.py
thedoubl3j: good catch, thank you good sir.

djyasin (approved):

AlanCoding (approved):

PR #296, opened 2024-04-09
[HOLD] Remove ANSIBLE_BASE_ROLE_PRECREATE only used by AWX

AlanCoding:
Opening as a draft initially, the work is going on separately in AWX, along with the role rename.

The AWX change is a blocker for merging this, but I want to be sure everything is passing here.

eda-server did not use this, and I do not anticipate galaxy_ng will either, but I await confirmation of that. If you plan to use this, we can keep it.


AlanCoding:
Since https://github.com/ansible/awx/pull/15087 is merged, this can be taken out of draft... right after a rebase.

relrod (approved):

john-westcott-iv:
@newswangerd can you take a look at this and give it a thumbs up if we're good to merge?

newswangerd (approved):
Is there another mechanism for pre creating roles in place?

AlanCoding:
We might hold off on merging (at least for a bit), as we're still not totally sure if other apps might use this.

john-westcott-iv:
Checked with @AlanCoding this is still on hold.

AlanCoding:
Note: I wanted to see that the "Testing DAB Consumers" check (introduced after this was created) was passing after a rebase, and it is, so it looks like we're good there.

https://github.com/ansible/django-ansible-base/pull/435 includes the commit from this, and now also proposes a new replacement system. My focus is going to be on that, so I'm going to close this PR for now, with the note that these changes are included completely there.

PR #297, opened 2024-04-09
Implement periodic sync command and functions.

rochacbruno:
Introduces `resource_registry/tasks/sync.py` and `resource_sync` management command.

Depends on: https://github.com/ansible/django-ansible-base/pull/311  and resource_server#351

Config on `django.conf.settings` looks like

```python
RESOURCE_SERVER = {
    "URL": "https://localhost",
    "SECRET_KEY": "<VERY-SECRET-KEY-HERE>",
    "VALIDATE_HTTPS": False,
}
RESOURCE_JWT_USER_ID = "97447387-8596-404f-b0d0-6429b04c8d22"
RESOURCE_SERVICE_PATH = "/api/server/v1/service-index/"
```

```console
$ django-admin resource_sync --asyncio

----- RESOURCE SYNC STARTED -----

>>> shared.organization
Deleting 1 orphaned resources
Processing 2 resources with asyncio executor.
CREATED 3e3cc6a4-72fa-43ec-9e17-76ae5a3846ca Acme
NOOP 3e3cc6a4-72fa-43ec-9e17-76ae5a389999
Processed 3 | Created 1 | Updated 0 | Conflict 0 | Unavailable 0 | Skipped 1 | Deleted 1

>>> shared.team
Processing 1 resources with asyncio executor.

NOOP f43938cf-a618-4a73-bc90-922a6b217e4d
CREATED f43938cf-a618-4a73-bc90-922a6b28888
Processed 2 | Created 1 | Updated 0 | Conflict 0 | Unavailable 0 | Skipped 1 | Deleted 0

>>> shared.user
Processing 4 resources with asyncio executor.

UPDATED 31daab14-cb67-4c62-8dcd-39f411c82242 joe
NOOP 97447387-8596-404f-b0d0-6429b04c8d22
NOOP 4d01427e-a11e-4d0e-9408-4ef7c2b478d6
CONFLICT 53886798-29e7-426f-b9cb-f7f74b346072 mary
Processed 4 | Created 0 | Updated 1 | Conflict 1 | Unavailable 0 | Skipped 2 | Deleted 0

----- RESOURCE SYNC FINISHED -----
```

OR run in a task queue such as Celery, RQ or Pulp Task

```python
from celery import Celery
from celery.schedules import crontab
from ansible_base.resource_registry.tasks.sync import SyncExecutor

# Initialize Celery app
celery = Celery('your_app_name', broker='redis://localhost:6379/0')

# Define Celery task
@celery.task
def execute_sync_executor():
    SyncExecutor(
        asyncio=False,
        api_client=ResourceAPIClient(....),  # Omit to have a default client created
        retries=3,
        retrysleep=60,
        retain_seconds=300,  # Avoid delete resources created in the latest 5 minutes
        stdout=sys.stdout  # omit to silence it
    ).run()

# Schedule the task to run every 15 minutes
celery.conf.beat_schedule = {
    'execute-every-15-minutes': {
        'task': 'your_module.execute_sync_executor',
        'schedule': crontab(minute='*/15'),
    },
}
```


ansible_base/resource_registry/tasks/sync.py
```diff
@@ -0,0 +1,411 @@
...
+
+
+# Functions
+
+
+def create_api_client(config: APIClientConfig | None = None) -> ResourceAPIClient:
+    """Factory for pre-configured ResourceAPIClient."""
+    config = config or settings.RESOURCE_PROVIDER_CONFIG
+    config["raise_if_bad_request"] = False
+    return ResourceAPIClient(**config)
```
rochacbruno: This actually depends on #311 beign merged (also resource_server#351) so the token based rest_client can be used from that PR

rochacbruno: For now, if one wants to run locally, on the service settings.py add this

```python
RESOURCE_PROVIDER_CONFIG = {
    "service_url": "https://localhost:8000",
    "service_path": "/api/gateway/v1/service-index/",
    "verify_https": False,
    "requests_auth_kwargs": {"auth": ("admin", "admin")},
    "raise_if_bad_request": True,
}

```

ansible_base/lib/abstract_models/common.py
```diff
@@ -43,7 +43,7 @@ class Meta:
 
     modified = models.DateTimeField(
         editable=False,
-        help_text=_("The date/time this resource was created"),
+        help_text=_("The date/time this resource was modified"),
```
AlanCoding: Won't this cause `makemigrations` to make new migrations for a ton of things in test_app and other apps?

rochacbruno: AFAIK when using lazy translation django make migration will not be able to detect that there is a change on that text https://github.com/ansible/django-ansible-base/blob/92cc5e39f0dbf7b9dbc22381e433d387288a9e13/ansible_base/lib/abstract_models/common.py#L8 

I will double check, if it makes then I revert to the incorrect text and add a comment.

rochacbruno: I removed the fix, at some point someone else will probably end up fixing that when a migration is required.

docs/apps/resource_registry.md
```diff
@@ -291,32 +291,268 @@ This displays information about the service itself:
...
+
+```
+
+### Syncing local services with RESOURCE_SERVER Resources
+
+Each service connected to the RESOURCE_SERVER can schedule a sync process, this process can
+be done via the management command `resource_sync` or by wrapping the `tasks.sync.SyncExecutor().run()`
+in a main task and spawning any tasking system that the service uses.
+
+The sync will be performed only for resources managed by APP, those which `service_id`
```
AlanCoding: replace "APP"

ansible_base/resource_registry/models/resource.py
```diff
@@ -55,6 +58,9 @@ class Resource(models.Model):
     # human readable name for the resource
     name = models.CharField(max_length=512, null=True)
 
+    created = models.DateTimeField(auto_now_add=True)
+    modified = models.DateTimeField(auto_now=True)
```
AlanCoding: Similar to my prior PR comments, I very much don't like that this duplicates created/modified that _almost_ every model already has.

Can you give more details about _why_ you need these fields? It looks like it is used for cleanup up orphan objects, but that only accounts for 1 of these 2 fields. It's also awfully niche.

And if you look at docs, you see this example for creating a resource via `service-index/resources/`

```
POST

{
    "ansble_id": "4c4ef945:57289235-e68e-4abf-8e50-f868f9e5ff04",
    "resource_type": "shared.organization",
    "resource_data": {"name": "my org"}
}
```

Could we add `created` and `modified` here? So that the timestamps of the local resource match the timestamps in the remote resource?

rochacbruno: I added the fields only to safely delete local orphan resources and avoid deleting resources that are in transient state.

In fact only one field is needed,the `created`, but I added both to match the existing fields on other models.

I can remove those 2 fields and instead of filtering in the query set do a check using the related model, will be less efficient but then will not require the additional fields.

AlanCoding: Let's talk that out.

```python
    return Resource.objects.filter(
        created__lt=timezone.now() - timedelta(seconds=retain_seconds),
        service_id=manifest_list[0].service_id,
        content_type__resource_type__name=resource_type_name,
    ).exclude(ansible_id__in=[item.ansible_id for item in manifest_list])
```

This returns items that are identified as coming from the service `manifest_list[0].service_id`, but were not in the `manifest_list`. So why? When is this happening? An object is deleted in the provider service, and normally, we would not expect the _periodic sync_ to be cleaning this up, normally we would expect the provider service to make a DELETE request to the `/resources/:id/` endpoint to delete the item in the other service as a part of the transaction. If that works, I'm skeptical that this condition will _ever_ be hit. If it doesn't work, then sure, this is last resort, I get it.

I'm not compelled to worried about efficiency of this corner case.

The efficiency I _am_ worried about is that a periodic command grabs `manifest_list`, which self-apparently from even the code above, must contain _all_ objects of the given type. If we could use a `modified` time stamp to cut down on the number of items the periodic sync must obtain from the provider, that would be very worthwhile.

rochacbruno: This whole PR exists only as a rescue, it is to cover the last resort corner case, ideally this command will run and everything will be always `NOOP` because the server already propagated changes to each service.

The only way to cut the size of items returned by the manifest is if we persist the timestamp of the last sync and then take items modified after that time, is it worth adding another model to persist that data?

Next week I will do the work to remove the new fields and then check using an if condition on the created_lt of the related model.

rochacbruno: https://github.com/ansible/django-ansible-base/pull/297/commits/e5ee6d3f590ed69ba21262946b59fb2db62814f4

ansible_base/lib/testing/util.py
```diff
@@ -36,3 +43,29 @@ def delete_authenticator(authenticator):
...
+        response = Response()
+        response.encoding = "utf-8"
+
+        if path in self.router:
+            response.status_code = self.router[path]["status_code"]
+            response._content = self.router[path]["content"]
+            return response
+
+        content_file_path = Path(self.base_url) / path / "response"
+        response._content = content_file_path.read_bytes()
```
AlanCoding: Could you explain what's happening here? Its parent class does `resp = requests.request(**kwargs)` which is intuitively logically, this is a client, it makes a request to a live(?) server. But this just reads from... a file path? It reads like a URL turned into a path.

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+ResourcePK = AnsibleID = ResourceHash = ServiceID = str
+
+
+class ResourceTypeName(str, Enum):
+    # keep the order
+    ORGANIZATION = "shared.organization"
+    TEAM = "shared.team"
+    USER = "shared.user"
```
AlanCoding: Hard-coding this certainly seems like a bad practice. There is a strong shared need to make use of logic inside of `ansible_base.resource_registry.apps`, like

```python
from ansible_base.resource_registry.registry import get_registry
[f'shared.{rt.model._meta.model_name}' for rt in registry.get_resources().values() if rt.managed_serializer]
```

This would give the same result as `[item.value for item in ResourceTypeName]` you use elsewhere... except maybe for maintaining the order. But if the order is necessary, existing data like `parent_resources` on the `ResourceConfig` model could be used.

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -0,0 +1,383 @@
...
+    retain_seconds: int = 120
+    stdout: TextIOBase | None = None
+    unavailable: set = field(default_factory=set)
+    attempts: int = 0
+    deleted_count: int = 0
+    asyncio: bool = False
+
+    def write(self, text: str = ""):
+        """Write to assigned IO or simply ignores the text."""
+        self.stdout and self.stdout.write(text)
```
AlanCoding: more clear to write out `if self.stdout:`

ansible_base/resource_registry/management/commands/resource_sync.py
```diff
@@ -0,0 +1,78 @@
...
+            help="Seconds to retain orphan resources from deletion.",
+            required=False,
+        )
+        parser.add_argument("--asyncio", action="store_true", default=False, help="Enable asyncio executor")
+
+    def handle(self, *args, **options):
+        """Handle RESOURCE_PROVIDER sync"""
+        arguments = ["resource_type_names", "retries", "retrysleep", "retain_seconds", "asyncio"]
+        options = {k: v for k, v in options.items() if k in arguments}
+        SyncExecutor(**options, stdout=self.stdout).run()
```
AlanCoding: Are exceptions going to be thrown here in the event of errors? I'm thinking through how AWX, for example, will implement this in its task system. It needs to write the output to logs. We don't get log levels, but ok, we'll capture it and write it when we're done. But how do we know what the outcome was? try and catch this call?

ansible_base/lib/testing/util.py
rochacbruno: It reads mocking/fixtures from static files for testing, as we cannot have a RESOURCE_SERVER running on CI

rochacbruno: The path `https://.../api/resource-types/shared.user/manifest/` is translated to `file://.../api/resource-types/shared.user/manifest/response`

ansible_base/resource_registry/tasks/sync.py
rochacbruno: I as not aware about tihs way of retrieving the resource types, I can change the code to use it and sort it based on a hardcoded key.

ansible_base/resource_registry/tasks/sync.py
rochacbruno: ok, I think the outcome is the same, I can do this change.

ansible_base/resource_registry/management/commands/resource_sync.py
rochacbruno: The known exceptions are defined in https://github.com/ansible/django-ansible-base/pull/297/files#diff-0bc1b25451c2df54563123daaa7c9d08158c54371d231d8feccec189a0fc7194R23-R32 

Everything else can happen and then the caller must handle it, on a tasking system this will for example just log the full traceback

rochacbruno: We can potentially keep state in `SyncExecutor` object so the information is there for reporting, is that what you need?

```
executor = SyncExecutor(...)
try:
   executor.run()
except ..:
   ...

print(executor.report.created|deleted|skipped|updated|failed)


```

AlanCoding: The thing we'll probably need the most is to get the output which this class writes to `stdout` object that is passed to it. That should work fine, as the caller will initialize a `StringIO` or something before the `try:`, so it'll safely be able to get its value after the run, even if it is interrupted. I think we'll have what we need with just this.

ansible_base/resource_registry/tasks/sync.py
rochacbruno: I implemented this way, however now it is processing `shared.authenticator` and I am not sure we want to sync that kind of resource, we can execute the command passing a specific list of resources to filter

```
django-admin resource_sync organization,team,user
```

Also when instantiating the SyncExecutor
```
SyncExecutor(resource_type_names=["shared.organization", "shared.team", "shared.user"]
```

c/c @newswangerd

ansible_base/resource_registry/models/resource.py
AlanCoding: Thanks, recapping, this no longer introduces timestamp fields on the `Resource` model, and uses the content object timestamp for reaping purposes.

I will see about filing an issue to open up planning design for the bigger issue of tracking timestamps on that model for making decisions about objects from the resource_server.

AlanCoding (approved):
Thanks for responding to comments.

PR #299, opened 2024-04-09
[ModifiableModel] Pass update_fields back to super

### LINKED ISSUE #286: Updates to `last_login` should not update `modified_by` to the system user
from activity stream

```json
{
    "count": 11,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 11,
            "url": "",
            "related": {
                "created_by": "/api/v1/users/2/"
            },
            "summary_fields": {
                "created_by": {
                    "id": 2,
                    "username": "admin",
                    "first_name": "",
                    "last_name": ""
                },
                "content_object": {
                    "id": 2,
                    "username": "admin",
                    "first_name": "",
                    "last_name": ""
                }
            },
            "created": "2024-04-08T14:48:22.190325Z",
            "created_by": 2,
            "operation": "update",
            "changes": {
                "added_fields": {},
                "changed_fields": {
                    "last_login": [
                        null,
                        "2024-04-08 14:48:22.178879+00:00"
                    ],
                    "modified_by": [
                        "_system",
                        "admin"
                    ],
                    "modified_by_id": [
                        1,
                        2
                    ]
                },
                "removed_fields": {}
            },
            "content_type": 17,
            "object_id": "2",
            "related_content_type": null,
            "related_object_id": null,
            "content_type_model": "user",
        },
```

relrod:
This came up when debugging #286 which was supposed to be an easy fix
but led me down quite a rabbit hole.

For as long as modified_by has existed (from what I can tell in the
git history), we never obeyed update_fields in save(). When we split
it out into ModifiableModel, the bug was kept in tact.

What ends up happening here is that on login, we'd act as if we're
going to update the modified_by field, and we'd get that in
update_fields, but then we'd just eat update_fields and never pass it
back into the supermethod.

This led to weird things like activity stream thinking that the value
was getting updated when a user logged in (because contractually, it
was *supposed* to be getting updated), but since we forgot
update_fields, it didn't *actually* get updated.

So you'd see an activity stream entry with the change, but check in
the database and see that modified_by never actually updated on login.

This ensures we pass back in update_fields if we were given it in the
first place, and that we only update it when it makes sense (if it's
actually an update, i.e. self.pk exists).

I also add a test showing that if you update modified_by and set
update_fields to include it, the forced value you set it to gets saved
as expected.

Finally, this does *not* fix #286. So I added an XFAIL test for
that. This was found while *working* on #286, but it's a different
bug.

### DIFF
--- a/ansible_base/lib/abstract_models/common.py
+++ b/ansible_base/lib/abstract_models/common.py
@@ -62,7 +62,7 @@ class ModifiableModel(models.Model):
         '''
         update_fields = list(kwargs.get('update_fields', []))
 
-        if self.pk and 'modified_by' not in update_fields:
+        if 'modified_by' not in update_fields:
             self.modified_by = current_user_or_system_user()
             update_fields.append('modified_by')
 
--- a/ansible_base/lib/abstract_models/common.py
+++ b/ansible_base/lib/abstract_models/common.py
@@ -62,9 +62,12 @@ class ModifiableModel(models.Model):
         '''
         update_fields = list(kwargs.get('update_fields', []))
 
-        if 'modified_by' not in update_fields:
+        if self.pk and 'modified_by' not in update_fields:
             self.modified_by = current_user_or_system_user()
             update_fields.append('modified_by')
+
+        if kwargs.get('update_fields') is not None:
+            kwargs['update_fields'] = update_fields
 
         return super().save(*args, **kwargs)
 
@@ -93,12 +96,10 @@ class CreatableModel(models.Model):
         This save function will provide the following features automatically.
           * It will automatically add a created_by fields for new items
         '''
-        update_fields = list(kwargs.get('update_fields', []))
 
         if not self.pk:
             if self.created_by is None:
                 self.created_by = current_user_or_system_user()
-                update_fields.append('created_by')
 
         return super().save(*args, **kwargs)
 
--- a/test_app/tests/lib/abstract_models/test_common.py
+++ b/test_app/tests/lib/abstract_models/test_common.py
@@ -164,3 +164,14 @@ def test_modified_by_respects_given_value(system_user, random_user, user, animal
         animal.save(update_fields=['modified_by'])
     animal.refresh_from_db()
     assert animal.modified_by == random_user
+
+
+def test_modified_by_gets_saved_even_if_not_in_update_fields(system_user, random_user, user, animal):
+    animal.save()
+    assert animal.modified_by == system_user
+    animal.name = 'Bob The Fish'
+    animal.kind = 'fish'
+    with impersonate(user):
+        animal.save(update_fields=['name', 'kind'])
+    animal.refresh_from_db()
+    assert animal.modified_by == user
--- a/test_app/tests/lib/abstract_models/test_common.py
+++ b/test_app/tests/lib/abstract_models/test_common.py
@@ -166,7 +166,7 @@ def test_modified_by_respects_given_value(system_user, random_user, user, animal
     assert animal.modified_by == random_user
 
 
-def test_modified_by_gets_saved_even_if_not_in_update_fields(system_user, random_user, user, animal):
+def test_modified_by_gets_saved_even_if_not_in_update_fields(system_user, user, animal):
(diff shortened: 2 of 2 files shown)


relrod:
I know CI is still broken here, but still putting some reviewers on this. This was quite a rabbit hole and a pretty not-so-nice bug.

ansible_base/lib/abstract_models/common.py
```diff
@@ -62,10 +62,13 @@ def save(self, *args, **kwargs):
         '''
         update_fields = list(kwargs.get('update_fields', []))
 
-        if 'modified_by' not in update_fields:
+        if self.pk and 'modified_by' not in update_fields:
             self.modified_by = current_user_or_system_user()
             update_fields.append('modified_by')
 
+        if kwargs.get('update_fields') is not None:
+            kwargs['update_fields'] = update_fields
```
AlanCoding, line 70: There was a time when the prior definition of `update_fields =` did not do `list()`, in which case, this step is not needed. It's very subtle, but works. It's not surprising that this regressed.

ansible_base/lib/abstract_models/common.py
```diff
@@ -62,10 +62,13 @@ def save(self, *args, **kwargs):
         '''
         update_fields = list(kwargs.get('update_fields', []))
 
-        if 'modified_by' not in update_fields:
+        if self.pk and 'modified_by' not in update_fields:
```
AlanCoding: I don't agree with adding `self.pk and ` here, and that might be the cause of some test failure. If creating a new object, that new object should have `modified_by` set. Contractually, this is hard to argue against.

ansible_base/lib/abstract_models/common.py
```diff
@@ -93,12 +96,10 @@ def save(self, *args, **kwargs):
         This save function will provide the following features automatically.
           * It will automatically add a created_by fields for new items
         '''
-        update_fields = list(kwargs.get('update_fields', []))
 
         if not self.pk:
             if self.created_by is None:
                 self.created_by = current_user_or_system_user()
-                update_fields.append('created_by')
```
AlanCoding, line 101: :+1: :wastebasket:

ansible_base/lib/abstract_models/common.py
relrod: We can set it if there's no pk but we can't set update_fields in that case. If we set that Django treats it as an update and when there's no pk, there's nothing to update and we get an error.

AlanCoding: True. I remember how it worked before, at some point. By doing `update_fields = kwargs.get('update_fields', [])`, it made `update_fields` reference the list that existed inside of `kwargs`, if it existed there. If there was no `update_fields` in `kwargs`, then adding to the list would only add to the throw-away data structure which came from `[]`. The `list()` forces the rest of the code to be more explicit.

relrod: Ok I've removed the self.pk check.

I don't want to remove the `list()`, because it's too magical/subtle what is going on there, relying on `update_fields` sometimes being an existing reference and sometimes being a new list.

test_app/tests/lib/abstract_models/test_common.py
```diff
@@ -144,3 +144,34 @@ def test_cascade_behavior_for_created_by(user, user_api_client):
...
+
+def test_modified_by_gets_saved_even_if_not_in_update_fields(system_user, random_user, user, animal):
+    animal.save()
+    assert animal.modified_by == system_user
+    animal.name = 'Bob The Fish'
+    animal.kind = 'fish'
+    with impersonate(user):
+        animal.save(update_fields=['name', 'kind'])
+    animal.refresh_from_db()
+    assert animal.modified_by == user
```
AlanCoding, line 177: Contractually, I think this is _not_ what AWX did. IIRC, if you used update_fields but left out modified or modified_by, it wouldn't update those fields. I don't fully think I agree with that. I say we should match what the Django `auto_now` behavior is. Otherwise, no one will ever be able to predict how this will behave.

AlanCoding (approved):

PR #301, opened 2024-04-10
Add User to unique constraint on AuthenticatorUser

john-westcott-iv:
Will having migration conflict with #294


relrod (approved):
A unique :snowflake:

PR #302, opened 2024-04-10
Remove users_unique from Authenticators

john-westcott-iv:


ansible_base/authentication/models/authenticator.py
```diff
@@ -11,7 +11,6 @@ class Authenticator(UniqueNamedCommonModel):
     enabled = fields.BooleanField(default=False, help_text="Should this authenticator be enabled")
     create_objects = fields.BooleanField(default=True, help_text="Allow authenticator to create objects (users, teams, organizations)")
     # TODO: Implement unique users, remove user, etc with team and org mapping feature.
```
relrod: Should we nix this comment too?

relrod (approved):

ansible_base/authentication/models/authenticator.py
john-westcott-iv: Good call, nixed.

PR #303, opened 2024-04-11
Avoid putting ignored fields in summary_fields

AlanCoding:
Checks will fail until https://github.com/ansible/django-ansible-base/pull/264 is merged, this will just be sequentially after that.

### DIFF
--- a/test_app/tests/rbac/api/test_rbac_views.py
+++ b/test_app/tests/rbac/api/test_rbac_views.py
@@ -57,6 +57,8 @@ def test_get_user_assignment(system_user, admin_api_client, inv_rd, rando, inven
     summary_fields = response.data['summary_fields']
     assert 'content_object' in summary_fields
     assert summary_fields['content_object'] == {'id': inventory.id, 'name': inventory.name}
+    # object_role is an internal objects, and is hidden to avoid API commitments
+    assert 'object_role' not in summary_fields
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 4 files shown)


test_app/tests/rbac/api/test_rbac_views.py
```diff
@@ -57,6 +57,8 @@ def test_get_user_assignment(system_user, admin_api_client, inv_rd, rando, inven
     summary_fields = response.data['summary_fields']
     assert 'content_object' in summary_fields
     assert summary_fields['content_object'] == {'id': inventory.id, 'name': inventory.name}
+    # object_role is an internal objects, and is hidden to avoid API commitments
+    assert 'object_role' not in summary_fields
```
relrod, line 61: Can we also have an explicit test for this in `test_common.py`, just in case the rbac test is refactored and loses this case some day?

relrod (approved):
Approving, but one inline I'd really like to see fixed before merging.

test_app/tests/rbac/api/test_rbac_views.py
AlanCoding: That mainly requires finding another model in test_app, unrelated to rbac, that should arguably have a related field omitted from serializer & summary fields. Or adding such a thing.

AlanCoding:
Ok, with the test I added we should be good here now.

test_app/tests/rbac/api/test_rbac_views.py
AlanCoding: done.

PR #305, opened 2024-04-11
Add util which can tell if a user can change their password

john-westcott-iv:


ansible_base/authentication/utils/user.py
```diff
@@ -0,0 +1,33 @@
...
+from ansible_base.authentication.models import AuthenticatorUser
+from ansible_base.lib.utils.models import is_system_user
+
+
+def can_user_change_password(user: AbstractUser) -> bool:
+    """
+    See if the given user is allowed to change their password.
+    True if they are authenticated from the `local` authenticator
+    False otherwise.
+    SYSTEM_USER can never change their password
```
relrod, line 15: I think it might be worth starting to standardize docstring comments for functions and methods. We don't have to be religious about it, but if we start trying to follow the Sphinx style, then we can generate DAB-internal Python API docs and they'll look really nice. So perhaps:

```suggestion
    Determine if the given user is allowed to change their password.
    
    :param user: The user to check
    :return: `True` if the user is authenticated from the ``local`` authenticator,
        `False` otherwise. Always returns `False` if ``user`` is the system user.
```

ansible_base/authentication/utils/user.py
```diff
@@ -0,0 +1,33 @@
...
+    if auth_users.count() == 0:
+        # If the user has no associations we can set a password for them so they can login through the local authenticator
+        return True
+
+    can_change_password = False
+    for auth_user in auth_users:
+        try:
+            plugin = get_authenticator_plugin(auth_user.provider.type)
+            if plugin.type == 'local':
+                can_change_password = True
```
relrod: Can't we just return `True` in this case rather than looping the rest of the authenticator users?

ansible_base/lib/utils/models.py
```diff
@@ -76,6 +76,21 @@ def user_summary_fields(user):
...
+    """
+    setting_name = 'SYSTEM_USERNAME'
+    system_username = get_setting(setting_name)
+    if system_username is None:
+        return False
+    if user is None:
+        return False
+    if user.username == system_username:
+        return True
+    return False
```
relrod:
```suggestion
    return user.username == system_username
```

:sunglasses:

ansible_base/lib/utils/models.py
```diff
@@ -76,6 +76,21 @@ def user_summary_fields(user):
...
+def is_system_user(user):
+    """
+    Takes a user objects and returns a boolean if that user's username is the same as the SYSTEM_USERNAME
+    """
+    setting_name = 'SYSTEM_USERNAME'
+    system_username = get_setting(setting_name)
+    if system_username is None:
+        return False
+    if user is None:
+        return False
```
relrod: Maybe combine these, but I'm just being nitpicky now.

test_app/tests/authentication/utils/test_user.py
```diff
@@ -0,0 +1,48 @@
+from unittest import mock
+
+import pytest
+
+from ansible_base.authentication.models import AuthenticatorUser
+from ansible_base.authentication.utils.user import can_user_change_password
+
+
+# Typeguard prevents None so we don't need to test that scenario
```
relrod: I'm not fully to the point of trusting typeguard yet. Particularly because I don't understand how it resolves types of values coming from external code.

Like, will typeguard prevent this?

```python
# assume this library does NOT have type annotations
from cool_thirdparty_app.reallycool import a_cool_user

can_user_change_password(a_cool_user)
```

How can typeguard know what `a_cool_user` is in that case? What if `a_cool_user` is defined like this:

```python
a_cool_user = User.objects.first() if randint(1, 10) < 5 else None
```

I don't see how typeguard helps here, but maybe I'm just missing something.

test_app/tests/lib/utils/test_models.py
```diff
@@ -309,3 +309,43 @@ def test_diff_sanitizes_encrypted_fields_removed(disable_activity_stream):
...
+        ("random", False),
+    ],
+)
+def test_is_system_user_system_user_setting_set(username, expected_value, system_user, random_user):
+    if username is None:
+        user = None
+    elif username == 'system':
+        user = system_user
+    else:
+        user = random_user
```
relrod, line 330: I find myself doing this kind of thing often in parametrized tests where I end up usaing a magic string to pick between fixtures, and I hate it. I wish there were a cooler syntax for this kind of thing.

test_app/tests/lib/utils/test_models.py
```diff
@@ -309,3 +309,43 @@ def test_diff_sanitizes_encrypted_fields_removed(disable_activity_stream):
...
+
+    assert models.is_system_user(user) == expected_value
+
+
+@pytest.mark.parametrize(
+    "username",
+    [
+        (None),
+        ("system"),
+        ("random"),
```
relrod:
```suggestion
        None,
        "system",
        "random",
```

test_app/tests/lib/utils/test_models.py
john-westcott-iv: Yea, it would be nice if you could put a fixture in the parametrized data itself or somehow have a builtin way to refer to a fixture.

test_app/tests/authentication/utils/test_user.py
john-westcott-iv: Because typeguard is enabled in tests we get this if we try to run with `(None, True)`:
```
FAILED test_app/tests/authentication/utils/test_user.py::test_can_user_change_password[None-True] - TypeError: argument of type 'NoneType' is not iterable
```

I changed the function to take `Optional(AbstractUser)` and the test is now valid.

ansible_base/authentication/utils/user.py
john-westcott-iv: I 100% agree with this. Maybe we should make an issue/story to do this for all docstrings.

relrod (approved):
I don't see the docstring change (and that's non blocking), but otherwise lgtm.

PR #308, opened 2024-04-12
Only update modified_by when an editable field has changed

### LINKED ISSUE #286: Updates to `last_login` should not update `modified_by` to the system user

relrod:
Fixes #286


AlanCoding (approved):
Saying the obvious for the record... this isn't a generalized solution. And it's also not that we don't need a generalized solution, we very much do, with the strong example being `JobTemplate` (has roles, tracked by activity stream) with fields of `current_job`, `last_job`, among with a litany of others. If this was used for that, every time someone launched a job, it would set the `modified_by` back to system user, which is very obviously a bug.

The need for a general solution, though, is separate from this because the sensible thing is for non-user-editable fields to have `editable=False`, and in this case too, it _should_ be non-editable. But it's not, and that cannot be changed. That is the argument for this being a special case.

relrod:
I'm fine with doing the more generalized solution in this PR and making sure at least one field in update_fields is editable, _if_ we have an `update_fields`. But what happens if we aren't given update_fields? What if someone just does

```python
animal.name = "Fredrick"
animal.save()
```

In that case, we _want_ to update modified_by, but we wouldn't have any way of knowing that we need to.

~I think to solve this we'd have to store the values of each field and manually compare that on save if `update_fields` is `None`.~

That's not easy to do either, there's really no good place to hook in to store the current values before a .save() happens. Doing it in the constructor after a super call causes recursion issues...

relrod:
Okay, I spent a bunch of time on this, and I have to be missing something stupid.

I even tried using a signals-based approach, but I don't think that was any better, and everything you read online says "don't use pre_save to update field values in the same model, override save() instead." (It doesn't really help anyway - updated_fields is converted to a frozenset before it's passed to pre_save, so pre_save can't actually add modified_by to it...)

But the issue is we have no good way to know which fields changed from save(). This problem is trivial if we assume `update_fields` is always set. But in the case where it's not set, how do we know if the changed fields are, in fact, editable fields?

I feel like I absolutely have to be missing something here.

AlanCoding:
I approved, and I just want to leave a comment to set reasonable expectations. Forget the first part of your snippet. What if you load `animal` and then making no changes just do `animal.save()`?

I expect that in that case `modified_by` does get updated. It is not reasonable to detect what diff happened. Django itself doesn't do it, and in other cases where I've seen the app do it, it leads to extra queries, extremely unexpected behaviors, and conflicts with multiple processes working on the same objects.

But return to my example of `JobTemplate`. It is _extremely_ reasonable to expect (demand!) that the task system always uses `update_fields`. If this is the _system_ doing something, it should follow best practices. And for fields like `last_job` in particular, your action should go to great lengths to isolate the change you are making.

relrod:
This has been updated to use `editable` when `update_fields` is given.

relrod:
@AlanCoding are you good with this new approach?

ansible_base/lib/abstract_models/common.py
```diff
@@ -59,14 +60,36 @@ def save(self, *args, **kwargs):
...
+            pass
+        elif has_update_fields and not has_editable_field:
+            # No editable fields are changing, don't update modified_by
+            pass
+        else:
             self.modified_by = current_user_or_system_user()
             update_fields.append('modified_by')
 
-        if kwargs.get('update_fields') is not None:
+        if has_update_fields:
```
AlanCoding: Probably non-functional suggestion:

If `modified_by` _was not set_ then I would prefer not to make edits to `kwargs` because that would get us involved in other error paths we don't want to touch.

This may be accomplished by simply increasing the indent of this `if` by one tab.

john-westcott-iv (approved):

PR #309, opened 2024-04-12
Handle missing makemigrations change

AlanCoding:
Got warnings that makemigrations has unmigrated changes, trying to take care of that as a one-off here.


relrod (approved):
oops

PR #311, opened 2024-04-12
Add resource server config

newswangerd:
- Update the resource api client to use the service JWT token
- Add resource server config

### DIFF
--- a/ansible_base/resource_registry/resource_server.py
+++ b/ansible_base/resource_registry/resource_server.py
@@ -16,7 +16,8 @@ class ResourceServerConfig(TypedDict):
 
 def get_resource_server_config() -> ResourceServerConfig:
     defaults = {"JWT_ALGORITHM": "HS256", "VALIDATE_HTTPS": True}
-    return ResourceServerConfig(**{**defaults, **settings.RESOURCE_SERVER})
+    defaults.update(settings.RESOURCE_SERVER)
+    return defaults
 
 
 def get_service_token(user_id=None, expiration=60, **kwargs):
--- a/ansible_base/resource_registry/resource_server.py
+++ b/ansible_base/resource_registry/resource_server.py
@@ -1,0 +1,31 @@
+from django.conf import settings
+import jwt
+from typing import TypedDict
+from ansible_base.resource_registry.models import service_id
+from datetime import datetime, timedelta
+
+
+class ResourceServerConfig(TypedDict):
+    URL: str
+    SECRET_KEY: str
+    VALIDATE_HTTPS: bool
+    JWT_ALGORITHM: str
+
+
+def get_resource_server_config() -> ResourceServerConfig:
+    defaults = {"JWT_ALGORITHM": "HS256", "VALIDATE_HTTPS": True}
+    return ResourceServerConfig(**{**defaults, **settings.RESOURCE_SERVER})
+
+
+def get_service_token(user_id, expiration=60):
+    config = get_resource_server_config()
+    print(config)
+    payload = {
+        "service_id": str(service_id()),
+        "user_id": str(user_id),
+    }
+
+    if expiration is not None:
+        payload["exp"] = datetime.now() + timedelta(seconds=expiration)
+
+    return jwt.encode(payload, config["SECRET_KEY"], config["JWT_ALGORITHM"])
--- a/ansible_base/resource_registry/resource_server.py
+++ b/ansible_base/resource_registry/resource_server.py
@@ -19,7 +19,6 @@ def get_resource_server_config() -> ResourceServerConfig:
 
 def get_service_token(user_id, expiration=60):
     config = get_resource_server_config()
-    print(config)
     payload = {
         "service_id": str(service_id()),
         "user_id": str(user_id),
--- a/ansible_base/resource_registry/resource_server.py
+++ b/ansible_base/resource_registry/resource_server.py
@@ -1,8 +1,10 @@
-from django.conf import settings
-import jwt
-from typing import TypedDict
-from ansible_base.resource_registry.models import service_id
 from datetime import datetime, timedelta
+from typing import TypedDict
+
+import jwt
+from django.conf import settings
+
+from ansible_base.resource_registry.models import service_id
 
 
 class ResourceServerConfig(TypedDict):
@@ -17,11 +19,12 @@ def get_resource_server_config() -> ResourceServerConfig:
     return ResourceServerConfig(**{**defaults, **settings.RESOURCE_SERVER})
 
 
-def get_service_token(user_id, expiration=60):
+def get_service_token(user_id, expiration=60, **kwargs):
     config = get_resource_server_config()
     payload = {
-        "service_id": str(service_id()),
-        "user_id": str(user_id),
+        "iss": str(service_id()),
+        "sub": str(user_id),
+        **kwargs,
     }
 
     if expiration is not None:
--- a/ansible_base/resource_registry/resource_server.py
+++ b/ansible_base/resource_registry/resource_server.py
@@ -19,13 +19,15 @@ def get_resource_server_config() -> ResourceServerConfig:
     return ResourceServerConfig(**{**defaults, **settings.RESOURCE_SERVER})
 
 
-def get_service_token(user_id, expiration=60, **kwargs):
+def get_service_token(user_id=None, expiration=60, **kwargs):
     config = get_resource_server_config()
     payload = {
         "iss": str(service_id()),
-        "sub": str(user_id),
         **kwargs,
     }
+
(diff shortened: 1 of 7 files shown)


ansible_base/resource_registry/resource_server.py
```diff
@@ -0,0 +1,33 @@
+from datetime import datetime, timedelta
+from typing import TypedDict
+
+import jwt
+from django.conf import settings
+
+from ansible_base.resource_registry.models import service_id
+
+
+class ResourceServerConfig(TypedDict):
```
rochacbruno, line 10: We talked (on the decision channel/bot) about naming it ResourceProvider, but I am ok with ResourceServer

rochacbruno (approved):
I need it and it still needs testing

newswangerd:
@john-westcott-iv I'd like to move the JWT consumer `ANSIBLE_BASE_JWT_VALIDATE_CERT` and `ANSIBLE_BASE_JWT_KEY` (for loading the jwt key from a url) settings to the new resource config object, once we get all of the other services updated.

I also plan on deprecating disabling validate certs and requiring the user to provide a CA cert if the resource server is using a self signed certificate in order to mitigate MitM attacks. We also need this for gRPC clients in the future.

rochacbruno (approved):

ansible_base/resource_registry/resource_server.py
```diff
@@ -0,0 +1,35 @@
...
+    URL: str
+    SECRET_KEY: str
+    VALIDATE_HTTPS: bool
+    JWT_ALGORITHM: str
+
+
+def get_resource_server_config() -> ResourceServerConfig:
+    defaults = {"JWT_ALGORITHM": "HS256", "VALIDATE_HTTPS": True}
+    return ResourceServerConfig(**{**defaults, **settings.RESOURCE_SERVER})
```
john-westcott-iv: Thats a lot of stars, could we just do something like `defaults.update(settings,RESOURCE_SERVER)` and pass that in as a `**` for readability?

ansible_base/resource_registry/resource_server.py
```diff
@@ -0,0 +1,35 @@
...
+
+class ResourceServerConfig(TypedDict):
+    URL: str
+    SECRET_KEY: str
+    VALIDATE_HTTPS: bool
+    JWT_ALGORITHM: str
+
+
+def get_resource_server_config() -> ResourceServerConfig:
+    defaults = {"JWT_ALGORITHM": "HS256", "VALIDATE_HTTPS": True}
```
john-westcott-iv, line 18: I've been thinking about things like the algorithms and wondering if we should variablize that incase someone wanted to run with HS512 or whatever else. Just a thought, not asking this PR to change.

ansible_base/resource_registry/resource_server.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class ResourceServerConfig(TypedDict):
+    URL: str
+    SECRET_KEY: str
+    VALIDATE_HTTPS: bool
+    JWT_ALGORITHM: str
+
+
+def get_resource_server_config() -> ResourceServerConfig:
```
john-westcott-iv, line 17: This function takes no parameters and is just merging a couple defaults with settings. Why not just add the defaults to the settings themselves and requiring it to be there?

ansible_base/resource_registry/rest_client.py
```diff
@@ -13,25 +16,73 @@
...
             is connecting to (http://www.example.com:123).
-        service_path: path on the service where the service-index/ api is found
+        service_path (str): path on the service where the service-index/ api is found
             (/api/v1/service-index/).
-        requests_auth_kwargs: dictionary of additional args to pass to requests.request()
-            ({auth=("username", "password")})
+        verify_https (bool): check the server's SSL certificates
+        raise_if_bad_request (bool): raise an exception if the API call returns a non
+            successful status code.
+        user_id (UUID): ansible ID of the user to make the request as.
```
john-westcott-iv: Missing the `jwt_` prefix

ansible_base/resource_registry/rest_client.py
```diff
@@ -13,25 +16,73 @@
...
+        self._jwt_timeout = time.time() + (self.jwt_expiration - 2)
+        self._jwt = "Token " + get_service_token(self.jwt_user_id, expiration=self.jwt_expiration)
+
+    @property
+    def jwt(self):
+        if self._jwt is None or self._jwt_timeout is None:
+            self.refresh_jwt()
+
+        if time.time() >= self._jwt_timeout:
+            self.refresh_jwt()
```
john-westcott-iv: Add another `or` statement to line 75?

test_app/authentication/service_token_auth.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class ServiceTokenAuthentication(BaseAuthentication):
+    keyword = "Token"
+
+    def authenticate(self, request):
+        auth = get_authorization_header(request).split()
+
+        if not auth or auth[0].lower() != self.keyword.lower().encode():
+            return None
```
john-westcott-iv, line 18: If I am reading this right, the `or`, is just checking for non-UTF8 characters in the authorization header with the JWT is that correct? If so, can we add a comment as to why we are doing that? Also, if this is for security (like an injection) should we should we have a function that wraps `get_authorization_header` and does this so we are protected everywhere?

test_app/authentication/service_token_auth.py
```diff
@@ -0,0 +1,39 @@
...
+
+        token = auth[1]
```
john-westcott-iv, line 20: Do we need to worry about a malformed authoriation header? i.e. could this only every be length of 2 or is length 1 possible?

test_app/authentication/service_token_auth.py
```diff
@@ -0,0 +1,39 @@
...
+
+        cfg = get_resource_server_config()
```
john-westcott-iv, line 22: I'm not sure if I'm in love with this config. But maybe I am seeing why you are using it. Its like another layer on top of settings. I guess it makes it so that we don't have to do `settings.get('JWT_ALGORYTHM', <default value>)` everywhere but I fell like we could just do something like this in the dynamic settings or somewhere "more traditional":
```
if 'JWT_ALGORYTHM' not in <whatever settings data structure>:
    <whatever settings data structure>['JWT_ALGORYTHM'] = 'RSA256'
```

My concern with the config is people might not know about it and then end up just referencing the setting directly anyway. If we made these configs a larger pattern I suppose that would work as well but it kind of feels like a one off here.

ansible_base/resource_registry/resource_server.py
newswangerd: It is variablized. You can pick HS256 -> HS512 on the server and then configure that here.

ansible_base/resource_registry/resource_server.py
newswangerd: This is mostly for IDE convenience. Python knows that the function returns a `ResourceServerConfig` dictionary, so most editors can auto complete the key values in the dict.

ansible_base/resource_registry/rest_client.py
newswangerd: @john-westcott-iv I'm not sure what you mean here. What else should I be or'ing?

test_app/authentication/service_token_auth.py
newswangerd: I think it makes sense because it makes it a lot easier to set defaults, you can take advantage of typing and you don't run as much risk of settings colliding. I do see your point about accessing it directly though. Because this is a dict in settings.py, I can't just add it to settings as a default because redefining it would remove the default keys.

test_app/authentication/service_token_auth.py
newswangerd: Honestly, I just copied this from the DRF token auth class. I can double check that my implementation is correct.

test_app/authentication/service_token_auth.py
newswangerd: I think I would prefer to keep this pattern. It seems more sane than guessing the default value everywhere that `settings.RESOURCE_SERVER` is referenced.

ansible_base/resource_registry/rest_client.py
john-westcott-iv: The two if conditions do the same thing (unless I am missing something) so we should be able to do:
```
        if self._jwt is None or self._jwt_timeout is None or time.time() >= self._jwt_timeout:
            self.refresh_jwt()
```
Unless there is a posibility that `refresh_jwt` could return an expired session.

newswangerd: oh, duh

Issue #251, opened 2024-04-02
[activity stream] add more info about changed related objects in the detail serializer for the UI


relrod:
### Bug Summary

As title.

AlanCoding:
Steps to reproduce:

 - Create an object registered with the activity stream that has a ForeignKey with a non-null value for that field.
 - View the creation entry for that object.
 - Look for information on the related object via the ForeignKey

Example:

 - Create an `Animal` model in test_app with a non-null `owner` field

### Proposed results

These are additions this issue is seeking, may still be subject to negotiation:

 - [ ] Link to the owner in `related` to `/api/v1/users/42/`, or whatever the owner id is.
   - Corner case: if `created` timestamp of the user is less than `created` of the activity stream entry, discard this, do not include the link. Log that a deleted item was referenced.
 - [ ] Entry in `summary_fields` under "owner" with the standard `User` model fields in the sub-dictionary.
 - [ ] A link to the activity stream for the owner in `related`, which might be like `/api/v1/activity_stream/?content_type__model=user&object_id=42`

It is expected that these would only be provided for the **detail view** of entries, because there could be a large number (upwards of 5 in cases) and they are all generic relations. This would be an expensive and unoptimizable thing to put in the list view.

PR #313, opened 2024-04-13
[activitystream] Make related and summary_fields a bit more useful

relrod:
Fixes some of what #251 is asking for.

We _do_ show them on the list view too - because I was able to do it without a higher query count than what we have currently in `devel`.

It does _not_ currently include:

> A link to the activity stream for the owner in related, which might be like /api/v1/activity_stream/?content_type__model=user&object_id=42

This also adds a default ordering for `Entry`.


ansible_base/activitystream/models/entry.py
```diff
@@ -42,6 +45,44 @@ class Meta:
...
+            the pk is the new pk value, for changed fields.
+        """
+        changed_fks = {}
+
+        if self.changes is None:
+            return changed_fks
+
+        for op in ('added_fields', 'changed_fields', 'removed_fields'):
+            for field_name, value in self.changes.get(op, {}).items():
+                field = self.content_type.model_class()._meta.get_field(field_name)
```
john-westcott-iv, line 61: Since this is activity stream and there could, maybe me models that were delete during a migration but still exist in activity stream do we need a try catach around this incase something "goes missing"?

ansible_base/activitystream/models/entry.py
```diff
@@ -42,6 +45,44 @@ class Meta:
...
+        Get the content object with any changed ForeignKey fields prefetched.
+        This is useful for related and summary_fields in serializers.
+
+        :return: The content object with any changed ForeignKey fields prefetched.
+        """
+        if self.changes is None:
+            return self.content_object
+
+        changed_fks = self.changed_fk_fields.keys()
+        obj = self.content_type.model_class().objects.select_related(*changed_fks).get(pk=self.object_id)
```
john-westcott-iv, line 83: Same thing here, what if the object behind the PK gets deleted? This would give you an exception.

ansible_base/activitystream/models/entry.py
relrod: This is a good call, but this is already a broken corner case in devel.

ansible_base/lib/serializers/common.py
```diff
@@ -36,6 +36,11 @@ def __init__(self, instance=None, data=empty, **kwargs):
     def get_url(self, obj) -> str:
         return get_url_for_object(obj)
 
+    @property
+    def is_list_view(self) -> bool:
+        view = self.context.get('view')
+        return view and view.action == 'list'
```
john-westcott-iv: If you don't get a view for some reason should an exception be raised or we are good with returning `False` in that condition?

test_app/tests/activitystream/test_api.py
```diff
@@ -97,3 +100,164 @@ def test_activitystream_api_filtering(admin_api_client, user):
     response = admin_api_client.get(url + '?' + urlencode(query_params))
     assert response.status_code == 200
     assert response.data['count'] == 1
+
+
+@pytest.mark.xfail(reason="TODO: Decide if we really want to exclude these from list view.")
```
john-westcott-iv: Should we make this decision now or is it ok to do that in the future?

john-westcott-iv (approved):
Approving with minor questions.

relrod:
Hang on, hang on, hang on :rofl: this isn't ready-to-merge yet, this is waiting on #316, so I can rebase, and then I still need to address your initial comments because there are more places where we'd re-introduce what #316 is fixing.

test_app/tests/activitystream/test_api.py
relrod: I'd like to hear @AlanCoding's thoughts. I think the concern was adding a bunch of queries. But with how I've done this, I don't see any more queries (according to django-debug-toolbar) than we currently get in devel when I go to /api/v1/activitystream.

So my inclination is that's fine. But I trust @AlanCoding's instinct about this, so if he still thinks it's an issue then I'd say we shouldn't do it on the list view.

If we do it on the list view, we're kinda stuck with that, because we can't pull fields from the API response without risking breaking people. That said, I kind of like having it on the list view and keeping the list view and the singular entry response consistent.

ansible_base/activitystream/models/entry.py
```diff
@@ -42,6 +45,44 @@ class Meta:
...
+                    if op == 'changed_fields':
+                        pk = value[1]
+                    else:
+                        pk = value
+                    changed_fks[field_name] = pk
+
+        return changed_fks
+
+    @functools.cached_property
+    def content_object_with_prefetched_changed_fields(self):
```
AlanCoding: Says "prefetch" in property name, but method does `select_related`, not `prefetch_related`. Maybe saying "cached" might be more clear?

test_app/tests/activitystream/test_api.py
AlanCoding: Your use of caching is much more advanced than I anticipated. I tip my hat to @relrod 

If it feels like the methods you used in this PR are maintainable, then it would be fine to put them in the list view.

ansible_base/activitystream/serializers.py
```diff
@@ -89,4 +93,70 @@ def _get_summary_fields(self, obj) -> dict[str, dict]:
...
+            # The model was deleted
+            changed_fk_fields = {}
+
+        for field_name, pk in changed_fk_fields.items():
+            related_object = getattr(content_obj, field_name, None)
+            if related_object is None:
+                continue
+
+            if hasattr(related_object, 'summary_fields'):
+                summary_fields[f"changes.{field_name}"] = related_object.summary_fields()
```
AlanCoding, line 111: If a PATCH changes a related field (we'll say `owner` of the animal model) from Alice to Bob... then what does `related["changes.owner"]` show? Alice or Bob?

relrod: It should show Bob in that case. I added a test just now to show that.

ansible_base/activitystream/models/entry.py
relrod: Done

ansible_base/lib/serializers/common.py
relrod: Ended up not needing this, so nixing this code.

ansible_base/activitystream/serializers.py
AlanCoding: After the prior steps, `owner` is then changed to Charlie. Now, the first entry when changed Alice->Bob now what does it show? Showing Charlie would be a bit weird. Just about any single choice would be a bit weird.

relrod: Ah, okay, I see what you're saying. Good find.

relrod: #338

PR #314, opened 2024-04-13
[activitystream] Add back lost filter_backends

relrod:
These are needed so things like order_by works.

### DIFF
--- a/ansible_base/activitystream/views.py
+++ b/ansible_base/activitystream/views.py
@@ -10,6 +10,29 @@ from ansible_base.lib.utils.views.django_app_api import AnsibleBaseDjangoAppApiV
 from ansible_base.lib.utils.views.permissions import IsSuperuser
 
 
+def calculate_filter_backends():
+    """
+    Calculate the filter backends for the Activity Stream.
+
+    We want to use the ActivityStreamFilterBackend for the FieldLookupBackend, but we want to keep the
+    default filter backends (in order) for everything else.
+    """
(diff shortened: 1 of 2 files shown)


ansible_base/activitystream/views.py
```diff
@@ -10,14 +10,37 @@
...
+    added_activity_stream_backend = False
+    for backend in default_filter_backends:
+        if backend.__name__ == 'FieldLookupBackend':
+            filter_backends.append(ActivityStreamFilterBackend)
+            added_activity_stream_backend = True
+        else:
+            filter_backends.append(backend)
+
+    if not added_activity_stream_backend:
+        filter_backends.append(ActivityStreamFilterBackend)
```
john-westcott-iv, line 31: If I'm reading this right you are basically subbing `FieldLookupBackend` for `ActivityStreamFilterBackend` right?

Can we shorten this to just:
```python
if FieldLookupBackend in default_filter_backends:
    default_filter_backends.remove(FieldLookupBackend)
if ActivityStreamFilterBackend not in default_filter_backends:
    default_filter_backends.append(ActivityStreamFilterBackend)
```

Or something along those lines?

john-westcott-iv (approved):
Code reduction question but fine to merge as is.

ansible_base/activitystream/views.py
relrod: I wanted to retain order, just in case. I think DRF will apply the filtering backends consecutively and narrow the queryset further each time, so the order could matter in theory I think in some cases.

PR #318, opened 2024-04-15
Fill in user permission logic for test_app

AlanCoding:
This is concerned with the permissions of a user to view or modify something about another user.

This carries forward the rules from AWX ultimately. I would like to make some _slight_ adjustments to these rules or their implementation, but another time.

TODO:

 - [x] Re-implement rules for user CRUD actions from AWX
 - [x] Test and fix access controls for giving org/team admin/member roles via the DAB RBAC assignments API
   - [x] Just remembered, should add a test for team role assignment
 - [x] Re-implement AWX association rules in the tracked_relationships corresponding to those roles (in test_app, like `/api/v1/teams/<id>/users/`), and add tests
 - [x] Org admins can not create superusers


ansible_base/rbac/api/permissions.py
```diff
@@ -80,22 +88,28 @@ def get_required_object_permissions(self, method, model_cls, view=None):
...
+        if not self.model_is_valid(model_cls):
             if request.user.is_superuser:
                 return True
             logger.warning(f'User {request.user.pk} denied {request.method} to {obj._meta.model_name}, not in DAB RBAC permission registry')
             raise Http404
 
         perms = self.get_required_object_permissions(request.method, model_cls, view=view)
 
-        if not all(request.user.has_obj_perm(obj, perm) for perm in perms):
+        if not self.has_object_permission_by_codename(request, obj, perms):
```
AlanCoding, line 112: I don't love this, it's tortured. I am inclined to adjust `has_obj_perm` to do the special-cases for users. But I'm okay enough with this for now.

ansible_base/lib/routers/association_resource_router.py
```diff
@@ -1,29 +1,85 @@
...
 
+def filtered_association_serializer_factory(cls, qs):
+
+    class AssociationSerializer(serializers.Serializer):
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            request = self.context['request']
+            self.fields['instances'] = serializers.PrimaryKeyRelatedField(
+                queryset=cls.access_qs(request.user, queryset=qs),
+                many=True,
```
AlanCoding, line 66: @slemrmartin you can here here I wound up using the method we discussed (for associating items you don't have view for) for the association endpoints generically.

slemrmartin: awesome

slemrmartin (approved):

PR #319, opened 2024-04-16
Continuing internationalization of strings

djyasin:
Tested that the correct strings were pulled in to the .po file.

Testing steps

Clone dir and active python virt env.
install django in virtenv if it is not present via pip
make a locale directory in the project
run django-admin makemessages -l es
navigate to the generated .po file, it will be in the locale directory under the specified language (here it is spanish so es) and then under LC_MESSAGES or something like that, you are looking for the django.po file.
confirm changed strings are present
(you can also just run the command then grep for the strings if preferred)
** These steps are from @thedoubl3j. I referred to his PR https://github.com/ansible/aap-gateway/pull/255.

Reviewed the following files:
ansible_base/authentication/management
ansible_base/authentication/management/__pycache__
ansible_base/authentication/management/commands
ansible_base/authentication/management/commands/__pycache__
ansible_base/authentication/management/commands/authenticators.py
ansible_base/authentication/migrations
ansible_base/authentication/models
ansible_base/authentication/serializers/authenticator_map.py
ansible_base/authentication/serializers/authenticator.py
ansible_base/authentication/utils/claims.py
ansible_base/authentication/views/ui_auth.py

### DIFF
--- a/ansible_base/authentication/management/commands/authenticators.py
+++ b/ansible_base/authentication/management/commands/authenticators.py
@@ -9,6 +9,7 @@ except ImportError:
 
 from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand, CommandError
+
 from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
 
--- a/ansible_base/authentication/management/commands/authenticators.py
+++ b/ansible_base/authentication/management/commands/authenticators.py
@@ -9,8 +9,6 @@ except ImportError:
 
 from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand, CommandError
-
-from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
 
 from ansible_base.authentication.models import Authenticator, AuthenticatorUser
--- a/ansible_base/authentication/management/commands/authenticators.py
+++ b/ansible_base/authentication/management/commands/authenticators.py
@@ -9,6 +9,8 @@ except ImportError:
 
 from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand, CommandError
+from django.utils.timezone import now
+from django.utils.translation import gettext_lazy as _
 
 from ansible_base.authentication.models import Authenticator, AuthenticatorUser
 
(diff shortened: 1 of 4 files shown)


ansible_base/authentication/management/commands/authenticators.py
```diff
@@ -34,7 +35,7 @@ def handle(self, *args, **options):
                 try:
                     authenticator = Authenticator.objects.get(id=id)
                 except Authenticator.DoesNotExist:
-                    raise CommandError(f"Authenticator {id} does not exist")
+                    raise CommandError(_("Authenticator %(id) does not exist") % {"id": id})
```
thedoubl3j: `(_("Authenticator %(id) does not exist") % {"id": id})` -> `(_("Authenticator %(id)s does not exist") % {"id": id})`

ansible_base/authentication/serializers/authenticator.py
```diff
@@ -95,10 +96,10 @@ def validate(self, data) -> dict:
             if configuration or configuration == {}:
                 for key in authenticator.configuration_encrypted_fields:
                     if not self.instance and configuration.get(key, None) == ENCRYPTED_STRING:
-                        invalid_encrypted_keys[key] = f"Can not be set to {ENCRYPTED_STRING}"
+                        invalid_encrypted_keys[key] = _("Can not be set to %(ENCRYPTED_STRING)") % {"ENCRYPTED_STRING": ENCRYPTED_STRING}
```
thedoubl3j: same as above, need the `s` following `%(some_key)s`

ansible_base/authentication/serializers/authenticator.py
```diff
@@ -95,10 +96,10 @@ def validate(self, data) -> dict:
...
                 if invalid_encrypted_keys:
                     raise ValidationError(invalid_encrypted_keys)
                 data['configuration'] = authenticator.validate_configuration(configuration, self.instance)
             return data
         except ImportError as e:
-            raise ValidationError({'type': f'Failed to import {e}'})
+            raise ValidationError({'type': _('Failed to import %(e)') % {'e': e}})
```
thedoubl3j: same as the others.

thedoubl3j (changes requested):

ansible_base/authentication/serializers/authenticator.py
djyasin: @thedoubl3j For some reason GitHub won't let me resolve this conversation. Everything is green now! I didn't understand that the `s` is part of the pattern. I've updated my notes accordingly. How does this look now?

thedoubl3j (approved):

PR #320, opened 2024-04-16
Make USER_DN_TEMPLATE not required

john-westcott-iv:
This was not a required field in AWX and should not be in Gateway either.


relrod (approved):
LGTM, the upstream lib should handle it being null/None, since None is the default:

https://github.com/django-auth-ldap/django-auth-ldap/blob/aa04ca5ac80ce3b3e7abbbcf860d684460907937/django_auth_ldap/config.py#L77

PR #322, opened 2024-04-16
Change required logic on plugin configuration

john-westcott-iv:
Before we would simply check "if not allow_null" but this falls short if a field has a default value but does not allow null. In that case the field would be required. This change will now explicity accept the `required` field and then fallback to `not allow_null`.

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/base.py
+++ b/ansible_base/authentication/authenticator_plugins/base.py
@@ -21,6 +21,13 @@ class BaseAuthenticatorConfiguration(serializers.Serializer):
         ui_field_label=_('Additional Authenticator Fields'),
     )
 
+    def _field_required(self, field):
+        if hasattr(field, 'required'):
+            return getattr(field, 'required')
+        elif hasattr(field, 'allow_null'):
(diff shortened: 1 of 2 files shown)


ansible_base/authentication/authenticator_plugins/base.py
```diff
@@ -21,6 +21,13 @@ class BaseAuthenticatorConfiguration(serializers.Serializer):
         ui_field_label=_('Additional Authenticator Fields'),
     )
 
+    def _field_required(self, field):
+        if hasattr(field, 'required'):
+            return getattr(field, 'required')
```
relrod:
```suggestion
            return field.required
```

ansible_base/authentication/authenticator_plugins/base.py
```diff
@@ -21,6 +21,13 @@ class BaseAuthenticatorConfiguration(serializers.Serializer):
...
+        elif hasattr(field, 'allow_null'):
+            return not getattr(field, 'allow_null')
```
relrod:
```suggestion
            return not field.allow_null
```

relrod (approved):

PR #325, opened 2024-04-17
Free IPA issues

john-westcott-iv:
Fixes issues associated with the test of Free IPA LDAP which uses a different configuration than we do.

This has been rebased onto #280 for testing. Do not merge this until #280 mergers.

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/ldap.py
+++ b/ansible_base/authentication/authenticator_plugins/ldap.py
@@ -373,20 +373,27 @@ class AuthenticatorPlugin(LDAPBackend, AbstractAuthenticatorPlugin):
             data = getattr(self.settings, field, None)
             if data is None:
                 setattr(self.settings, field, None)
-            else:
+            elif not isinstance(data, config.LDAPSearch):
                 try:
                     # Search fields should be LDAPSearch objects, so we need to convert them from [] to these objects
                     search_object = config.LDAPSearch(data[0], getattr(ldap, data[1]), data[2])
                     setattr(self.settings, field, search_object)
                 except Exception as e:
-                    logger.error(f'Failed to instantiate LDAPSearch object: {e}')
+                    logger.error(f'Failed to instantiate {field} LDAPSearch object: {e}')
                     return None
 
         try:
             user_from_ldap = super().authenticate(request, username, password)
 
+            self.process_login_messages(user_from_ldap, username)
+
+            # If we didn't get a user we can return None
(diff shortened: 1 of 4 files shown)


ansible_base/authentication/models/social_auth.py
```diff
@@ -1,10 +1,25 @@
...
 from ansible_base.authentication.models import Authenticator
 
 
+def b64_encode_binary_data_in_dict(obj: Any) -> Any:
+    if isinstance(obj, list):
+        for index in range(0, len(obj)):
+            obj[index] = b64_encode_binary_data_in_dict(obj[index])
+    elif isinstance(obj, dict):
+        for key, value in obj.items():
+            obj[key] = b64_encode_binary_data_in_dict(key)
```
relrod: value?

john-westcott-iv: Thank you.

ansible_base/authentication/models/social_auth.py
```diff
@@ -1,10 +1,25 @@
...
+from typing import Any
+
 from django.conf import settings
 from django.db import models
 from social_django.models import AbstractUserSocialAuth
 
 from ansible_base.authentication.models import Authenticator
 
 
+def b64_encode_binary_data_in_dict(obj: Any) -> Any:
```
relrod: This concerns me a little in that we're not tracking which fields are getting base64-encoded. So we have no way to ever revert the process, because we don't know which fields to decode.

I was originally going to ask down below where you call this in `save()` if we need a way to revert the process on object load - but there's no way to do that with this implementation. Could that come back to bite us?

john-westcott-iv: This field is the extra data that we get from an authentication. For example, it might look like:
```json
{
    "displayname": ["User Name"],
    "uid": ["username"],
    "phone_number": "123-456-7890",
    "country": "Germany",
    ....
    "auth_time": "2024-04-17T15:45:56.066473Z"
}
```

We reference this data when processing `attribute` authenticator_maps. So if there is a binary blob in the extra data, like maybe a GPG key, whoever is setting up the authenticator_map would just need to know to use the b64 version of the key. To me that seems like a pretty rare use case so I don't think there is really a need for us to be able to un-convert the field.

But I'm open to other opinions. We can't really change the field type because its probvide by social_auth but we could add a second text field with a copy but again, kind of an edge case to begin with IMHO.

relrod (approved):

PR #328, opened 2024-04-18
Fix authenticator/:id/users endpoint

john-westcott-iv:
This adds the URL back into the related dict on /authenticators/:id and corrects a bug where it always returned 0 users.

### DIFF
--- a/ansible_base/authentication/serializers/authenticator.py
+++ b/ansible_base/authentication/serializers/authenticator.py
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 
 from django.utils.translation import gettext_lazy as _
+from rest_framework.reverse import reverse
 from rest_framework.serializers import ChoiceField, ValidationError
 
 from ansible_base.authentication.authenticator_plugins.utils import get_authenticator_plugin, get_authenticator_plugins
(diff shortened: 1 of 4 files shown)


ansible_base/authentication/serializers/authenticator.py
```diff
@@ -103,3 +103,14 @@ def validate(self, data) -> dict:
             return data
         except ImportError as e:
             raise ValidationError({'type': _('Failed to import %(e)s') % {'e': e}})
+
+    def _get_related(self, obj) -> dict[str, str]:
+        related = super()._get_related(obj)
+        from ansible_base.authentication.views.authenticator_users import get_authenticator_user_view
```
relrod, line 110: Is this import recursive if we put it up top?

ansible_base/authentication/serializers/authenticator.py
```diff
@@ -103,3 +103,14 @@ def validate(self, data) -> dict:
...
+
+        if get_authenticator_user_view():
+            from rest_framework.reverse import reverse
```
relrod: Let's move this up too

ansible_base/authentication/serializers/authenticator.py
john-westcott-iv: Yea

ansible_base/authentication/serializers/authenticator.py
john-westcott-iv: Moved

relrod (approved):

PR #329, opened 2024-04-18
Update EDA auth class to not use old Roles

Dostonbek1:
The old Roles model is removed in favor of DAB RBAC implementation in EDA so this PR aims to remove its use in EDA Auth class and authorize the user as is.

We need this merged before https://github.com/ansible/eda-server/pull/710 is merged.


AlanCoding (approved):
This looks mostly like a change for git history management sanity.

PR #330, opened 2024-04-18
Internationalization for lib & resource_registry dirs

djyasin:
Testing steps

Clone dir and active python virt env.
install django in virtenv if it is not present via pip
make a locale directory in the project
run django-admin makemessages -l es
navigate to the generated .po file, it will be in the locale directory under the specified language (here it is spanish so es) and then under LC_MESSAGES or something like that, you are looking for the django.po file.
confirm changed strings are present
(you can also just run the command then grep for the strings if preferred)
** These steps are from @thedoubl3j. I referred to his PR https://github.com/ansible/aap-gateway/pull/255.


thedoubl3j (approved):
nice and easy help text translations. lgtm

PR #331, opened 2024-04-18
[translations] /rest_filters updates

thedoubl3j:
Updates to rest_filters dir for translation. thanks to @AlanCoding for calling out during the work on this PR for making sure we don't translate when necessary (and also syntax checking me to make sure we are actually translating all that is needed). 

**Testing Steps**
`pip freeze | grep Django`
`pip install Django`
3. make a locale directory (makemessages looks for one and we don't have one configured yet so just create your own) `mkdir locale`
4. run the makemessages command check the output to see if strings were extracted correctly.
`django-admin makemessages -l es`
`vim locale/es/LC_MESSAGES/django.po`

### DIFF
--- a/ansible_base/rest_filters/rest_framework/field_lookup_backend.py
+++ b/ansible_base/rest_filters/rest_framework/field_lookup_backend.py
@@ -90,7 +90,8 @@ class FieldLookupBackend(BaseFilterBackend):
             try:
                 return self.to_python_related(value)
             except ValueError:
-                raise ParseError(_('Invalid {field_name} id: {field_id}').format(field_name=getattr(field, 'name', 'related field'), field_id=value))
+                field_name = getattr(field, 'name', 'related field')
+                raise ParseError(_("Invalid %(field_name)s id: %(field_id)s") % {"field_name": field_name, "field_id": value})
         else:
             return field.to_python(value)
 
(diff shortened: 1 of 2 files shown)


thedoubl3j:
Technically a WIP, spoke more @AlanCoding and need to change a few more things (but everything here is good for review)

djyasin (approved):

ansible_base/rest_filters/rest_framework/field_lookup_backend.py
```diff
@@ -90,15 +90,16 @@ def value_to_python_for_field(self, field, value):
...
+                raise ParseError(_("Invalid %(field_name)s id: %(field_id)s") % {"field_name": field_name, "field_id": value})
         else:
             return field.to_python(value)
 
     def value_to_python(self, model, lookup, value):
         try:
             lookup.encode("ascii")
         except UnicodeEncodeError:
-            raise ValueError("%r is not an allowed field name. Must be ascii encodable." % lookup)
+            raise ValueError(_("%(r)s is not an allowed field name. Must be ascii encodable.") % {"r": lookup})
```
AlanCoding: This might be a little surprising, since `%r` has special meaning as using `repr()` in the old-style python format strings.

thedoubl3j: so are you thinking I should just replace `%(r)s` with a new var or do you think we think we were using the repr functionality here and see if there is another route to go for getting the string translated?

AlanCoding: I would just replace it with a new var

john-westcott-iv (approved):

AlanCoding (approved):

AlanCoding:
Ignore the DAB Consumers failure, I've confirmed other PRs are seeing the same errors.

PR #335, opened 2024-04-21
[AS] Add related AS link to all AuditableModels

relrod:
This takes #251 a little further and adds an activity_stream link to
all AuditableModels' "related" fields.

Kind of playing around here, but this might be useful.
Checked queries with DjDT and didn't see any added queries in list views to do this.


ansible_base/activitystream/models/entry.py
```diff
@@ -65,3 +67,14 @@ def activity_stream_entries(self):
...
         return Entry.objects.filter(content_type=ContentType.objects.get_for_model(self), object_id=self.pk).order_by('created')
+
+    def extra_related_fields(self, request):
+        content_type = ContentType.objects.get_for_model(self)
+        query_kwargs = {
+            'content_type': content_type.pk,
+            'object_id': self.pk,
+        }
+        activity_stream_url = reverse('activitystream-list') + '?' + urlencode(query_kwargs)
+        return {
```
AlanCoding, line 111: ... `if "ansible_base.activity_stream" in settings.INSTALLED_APPS and isinstance(self, AuditableModel)`?

AlanCoding (changes requested):
Predicting this will break everything for apps not using the activity stream right now, like eda-server and AWX

ansible_base/activitystream/models/entry.py
relrod: This is on `AuditableModel`, not `CommonModel`, so I think we're fine here. The only way something would subclass `AuditableModel` is by having activity stream installed.

AlanCoding (approved):
I stand corrected that this is only added to `AuditableModel` in the first place. This looks in good shape.

PR #337, opened 2024-04-22
Remove permission_classes from un-opinionated views to allow customization

AlanCoding:


ansible_base/lib/utils/views/ansible_base.py
```diff
@@ -10,6 +11,8 @@
 
 
 class AnsibleBaseView(APIView):
+    permission_classes = [permissions.IsAuthenticated]
```
AlanCoding: @john-westcott-iv @relrod @slemrmartin Maybe the better solution here would be to _not_ set this at all, with the intent that it'll just use the default permission class from DRF? In practice, we would do this for test_app.

All the real apps using DAB will have something concrete set for this. By not putting it here, apps can customize the permission class (to its own default) without changing the base view.

slemrmartin: I agree, it sounds more generic

john-westcott-iv (approved):

ansible_base/lib/utils/views/ansible_base.py
john-westcott-iv: I agree with removing this from ansible-base but added an approval on this.

AlanCoding: Thanks. I will go ahead and make this change.

AlanCoding:
Updated title of PR to reflect the latest change.

Confirmed that /api/v1/authenticators/ is still working in the browser.

slemrmartin (approved):

PR #338, opened 2024-04-22
[AS] Address incorrect FK refs in summary_fields

relrod:
Followup to #313. Don't use the current FK, use the one that it was
set to when the AS entry was logged.

This ended up triggering a lot more queries because there's really no
way to cache this effectively. So that led to reverting back to not
showing the extra summary_fields and related on list views.

Additionally, clean up a lot of redundant, verbose code, and make
things easier to read.

### DIFF
--- a/ansible_base/activitystream/models/entry.py
+++ b/ansible_base/activitystream/models/entry.py
@@ -49,7 +49,9 @@ class Entry(ImmutableCommonModel):
     def changed_fk_fields(self):
         """
         :return: A dictionary of {field_name: pk} for any ForeignKey fields that have changed.
-            the pk is the new pk value, for changed fields.
+            the pk is the new pk value, for changed fields. In the case of added/removed fields,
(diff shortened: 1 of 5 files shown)


ansible_base/activitystream/models/entry.py
```diff
@@ -58,31 +58,19 @@ def changed_fk_fields(self):
...
-                    if op == 'changed_fields':
-                        pk = value[1]
-                    else:
-                        pk = value
-                    changed_fks[field_name] = pk
+                    try:
+                        fk_model = field.related_model
+                    except AttributeError:  # Likely the model was deleted
+                        continue
+                    pk = value[1] if op == 'changed_fields' else value
```
AlanCoding, line 73: Might be good to have a comment here...

So for _updates_ this implements a rule to show the _latter_ of the two values involved.

AlanCoding (approved):
Looks pretty solid.

Issue #159, opened 2024-02-23
[RBAC] Remove the "relationship tracking" or "role tracking" feature


AlanCoding:
We don't need this anymore. JWT data exchange details were unclear at the time.

Tests at `test_app/tests/rbac/features/test_role_tracking.py`

AlanCoding:
@fosterseth I want you to know that I've got some stuff like this lying around. I'm just not prioritizing it at the moment.

AlanCoding:
Blocked because we might use this temporarily to solve an API compatibility problem.

PR #339, opened 2024-04-22
Allow use of role-tracking stuff in reverse relationship

AlanCoding:
I kind of remember intentionally not doing this before. At the time of development, I was probably thinking I would return to it, and then I thought we would remove this entire functionality, see https://github.com/ansible/django-ansible-base/issues/159

But now we're using it, and things were broken. Looked at it, like hmmm, is this because `reverse=True` in the signals? And yep, it was.

Clarifying: the test added here is the _first_ test to test this synchronization (role to relationship) for a reverse relationship.

### DIFF
--- a/ansible_base/rbac/triggers.py
+++ b/ansible_base/rbac/triggers.py
@@ -1,5 +1,6 @@
 import logging
 from contextlib import contextmanager
+from typing import Optional
 
 from django.apps import apps
 from django.conf import settings
@@ -325,7 +326,7 @@ class TrackedRelationship:
         else:
             manager.remove(actor)
 
-    def _sync_actor_to_role(self, actor_model, instance, action, pk_set, reverse):
+    def _sync_actor_to_role(self, actor_model: type, instance: Model, action: str, pk_set: Optional[set[int]]):
         if self._active_sync_flag:
             return
         if action.startswith('pre_'):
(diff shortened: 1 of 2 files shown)


ansible_base/rbac/triggers.py
```diff
@@ -343,14 +344,22 @@ def _sync_actor_to_role(self, actor_model, instance, action, pk_set, reverse):
...
+                self._sync_actor_to_role(permission_registry.team_model, model(pk=pk), action, {instance.pk})
 
-    def sync_user_to_role(self, instance, action, model, pk_set, reverse, **kwargs):
-        self._sync_actor_to_role(permission_registry.user_model, instance, action, pk_set, reverse)
+    def sync_user_to_role(self, instance: Model, action: str, model: type, pk_set: Optional[set[int]], reverse: bool, **kwargs):
+        if not reverse:
+            self._sync_actor_to_role(permission_registry.user_model, instance, action, pk_set)
+        else:
+            for pk in pk_set:
+                self._sync_actor_to_role(permission_registry.user_model, model(pk=pk), action, {instance.pk})
```
john-westcott-iv, line 362: Nit, these are the same less the `permission_registry.<whatever>`. Should we make this a function and just pass the params through?

john-westcott-iv (approved):

ansible_base/rbac/triggers.py
AlanCoding: Not 100% with you, `_sync_actor_to_role` is a single function, called for user or team. We need to pass in the `actor_model` (I know because I tried to remove it) because we register different methods for user vs. team model. The user/team model might be the same as `model`, that could be an interesting simplification, just not sure if it would work.

PR #341, opened 2024-04-23
Related mixin parent query

slemrmartin:
The RelatedListMixin is using query_params for parent queryset, which generates error


AlanCoding:
Test: https://github.com/slemrmartin/django-ansible-base/pull/1

AlanCoding:
Confirmed locally that the linked test passes with this fix. So we should be good to merge that here, and then merge the combined patch.

AlanCoding (approved):

PR #343, opened 2024-04-23
Allow m2m relations in related_fields

john-westcott-iv:
The associative router knows how to deal with reverse m2m fields.

This allows things like organization/team to show up in users related fields.

### DIFF
--- a/ansible_base/lib/abstract_models/common.py
+++ b/ansible_base/lib/abstract_models/common.py
@@ -3,7 +3,6 @@ from collections import OrderedDict
 
 from django.conf import settings
 from django.db import models
-from django.db.models.fields.reverse_related import ManyToManyRel
 from django.urls.exceptions import NoReverseMatch
 from django.utils.translation import gettext_lazy as _
 from inflection import underscore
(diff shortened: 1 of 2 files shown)


AlanCoding:
I agree we should show reverse m2m links in "related" of the serializers. But for all our sanity, we have got to do something about this log

```python
logger.error(f"Wanted to add {reverse_view} for {self.__class__} but view was missing")
```

This will attempt the reverse (by itself somewhat concerning performance-wise) and then issue a log for about as many reverse relations as there are forward relations. The forward relations are, in major part, filled in, but we categorically don't want to bother with the reverse in _most_ cases. This will open a floodgate of logs. Maybe we could introduce a management command or something to diagnose missing links, but it's time to cut that log off.

Also, no test.

john-westcott-iv:
How would you feel about making it a debug instead of an error? I've found this message helpful for when you are adding a new view and its not showing up in the API as expected.

AlanCoding:
I have to be a little extreme here. Even a `logger.debug` would be unacceptable for the context in which this is issued. I would negotiate that down to a debug log only when `settings.DEBUG` is True. For your expressed use case that should still work.

AlanCoding:
You might even want to consolidate those logs, adding the missing fields to a list, and then after the loop, if DEBUG is true, then do a debug log telling the model, and listing the fields that can't be reversed. This would translate better into dropping them straight into the model's ignore relations list.

This is also necessary because those issues are, so far, unfixable. In particular, the un-named reverse relations that end in `_set`, those can't be skipped coherently.

john-westcott-iv:
I left it at error but wrapped it in a `if settings.DEBUG` and also grabbed all bad view names and put them into a single message.

ansible_base/lib/abstract_models/common.py
```diff
@@ -167,17 +166,21 @@ def related_fields(self, request):
...
                 continue
             reverse_view = f"{basename}-{field_name}-list"
             try:
                 response[field_name] = reverse(reverse_view, kwargs={'pk': self.pk})
             except NoReverseMatch:
-                logger.error(f"Wanted to add {reverse_view} for {self.__class__} but view was missing")
+                missing_relations.append(reverse_view)
+
+        if missing_relations:
+            if settings.DEBUG:
```
AlanCoding: For ultimate pedantry, I remember that SonarCloud likes for double-conditions to go in the same `if`

AlanCoding (approved):
Looks great, thanks.

ansible_base/lib/abstract_models/common.py
john-westcott-iv: 100%, had to fix it for SQ!

PR #345, opened 2024-04-23
Add conglomeration of multiple fields for reverse m2m views

john-westcott-iv:


AlanCoding:
I would rather not do this if we don't have to.

relrod:
I'm going to summarize some thoughts here from elsewhere.

I agree this isn't the right approach. The example in question (and mentioned in the docs) is an `Organization` model which has two m2m fields, `users` and `admins`. And the goal is that we are considering admins to be users/members of the organization.

This seems like it stems from a design error. If admins are members of the organization, then we should make them be members of the organization.

Whether or not they are an admin is a property of their membership. So the best way to solve this is to get rid of 'admins' as a field on the `Organization` model, make `users` use a custom through table, and make the through table have an `is_admin` flag. Semantically, this describes the situation better.

But this comes with its own set of issues. Namely - our association serializer in the router is very simplistic. So at association time, there's no real way to specify that added "is_admin" attribute.

I don't have an easy solution to that problem yet, but this seems like a more general problem that is going to come up one way or another in the future. Basically right now, our association router is unusable for any m2m with a custom through table that has properties that need to be set.

I'm not sure what fixing that looks like, but I think that's the better problem to solve here. Perhaps as a stop-gap. We could just use a fully custom serializer for these cases, but that seems less than ideal too because then it's easy to create accidental inconsistencies across them.

Maybe we allow the `instances` field of the existing router to take optional JSON objects, so if you pass `instances: [1,2,3]` it works as-is, but if you pass `instances: [{id: 1, is_admin: false}, {id: 2, is_admin: true}, {id: 3, is_admin: false}]` then it fills in those fields dynamically on the membership row.

Or maybe there's a simpler solution.

PR #346, opened 2024-04-23
Allow configuring permissions needed to related objects

AlanCoding:
There was a sentiment that requiring "change" permission to related objects might be too much. I expected this, but was trying to finish development fast earlier.

Normally I would expect an app to rely on "use" permission, like AWX for `use_role` of credentials and other things. But that requires re-configuring permission structure and another permission for each model. I'm not ready to take that step now, and without that your choices are "change" or "view". If required is downgraded to "view", that's perfectly coherent. It's a policy question that I have no specific opinion on.

In short, this will allow apps using the DAB RBAC API utilities to configure what level of related permission they want to check. This comes in a precedence list defined in setting. If nothing in the preference list applies to the target model, then we won't check permissions for that item.

Ping @Dostonbek1

### DIFF
--- a/ansible_base/lib/dynamic_config/dynamic_settings.py
+++ b/ansible_base/lib/dynamic_config/dynamic_settings.py
@@ -133,6 +133,9 @@ if 'ansible_base.rbac' in INSTALLED_APPS:
 
     # Permissions a user will get when creating a new item
     ANSIBLE_BASE_CREATOR_DEFAULTS = ['add', 'change', 'delete', 'view']
+    # Permissions API will check for related items, think PATCH/PUT
+    # This is a precedence order, so first action related model has will be used
+    ANSIBLE_BASE_CHECK_RELATED_PERMISSIONS = ['use', 'change', 'view']
     # If a role does not already exist that can give those object permissions
     # then the system must create one, this is used for naming the auto-created role
     ANSIBLE_BASE_ROLE_CREATOR_NAME = '{obj._meta.model_name}-creator-permission'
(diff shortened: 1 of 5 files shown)


ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -133,6 +133,9 @@
 
     # Permissions a user will get when creating a new item
     ANSIBLE_BASE_CREATOR_DEFAULTS = ['add', 'change', 'delete', 'view']
+    # Permissions API will check for related items, think PATCH/PUT
+    # This is a precedence order, so first action related model has will be used
+    ANSIBLE_BASE_CHECK_RELATED_PERMISSIONS = ['use', 'change', 'view']
```
Dostonbek1, line 138: Is this an AND or OR combination, meaning does the user need at least one of these permissions or all of them to be able to use the object as related?

AlanCoding: Formally I think this is neither AND or OR. It is first-found. Because there are 2 conditions here:
 - Is there a permission for this action, for this model
 - Given the prior criteria is satisfied, does the user have that permission

That's why it is a precedence order. First it determines if the model has a "use" permission, and if it does, the answer is a True/False the user has "use" permission. But for _models_ that don't have that first option, it will go on to "change" and if the model has the permission, return True/False user has "change" permission, and so-on.

Dostonbek1 (approved):

PR #350, opened 2024-04-24
Attempt to run AWX tests against changes

relrod:
In my testing this takes around 6 minutes to run, so it adds about 2 minutes to our current test time.

I argue that the benefit is worth it.

I tested this on my fork by reverting #349 and ensuring this check failed.

### DIFF
--- a/.github/workflows/dab-consumers.yml
+++ b/.github/workflows/dab-consumers.yml
@@ -1,0 +1,42 @@
+---
+name: Testing DAB Consumers
+on:
+  push:
+  pull_request:
+jobs:
+  awx:
+    name: AWX
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          show-progress: false
+
+      - name: Pull AWX devel image
+        run: docker pull ghcr.io/ansible/awx_devel:devel
+
+      # Generate a docker image that includes current DAB
+      - name: Create image for AWX devel that includes our DAB
+        run: |
+          mkdir awx-dockerfile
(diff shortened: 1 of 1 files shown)


.github/workflows/dab-consumers.yml
```diff
@@ -0,0 +1,42 @@
...
+        run: docker pull ghcr.io/ansible/awx_devel:devel
+
+      # Generate a docker image that includes current DAB
+      - name: Create image for AWX devel that includes our DAB
+        run: |
+          mkdir awx-dockerfile
+          cat <<EOF > awx-dockerfile/Dockerfile
+          FROM ghcr.io/ansible/awx_devel:devel
+          COPY . /opt/dab/
+          RUN pip install -e /opt/dab/
```
AlanCoding, line 30: I'm a little so-so about the overall concept, but I have to hand it to you, this is :fire:

relrod: @AlanCoding are you less so-so about this if we keep it non-voting? I think we can't make it voting anyway because there might be times where we _intentionally_ break AWX, and in that case we still need to be able to merge.

I really just want this to run as a visual indicator and some extra re-assurance in the normal case.

john-westcott-iv (approved):

AlanCoding (approved):

relrod:
Merging, but also noting there's a potential race condition here: If we pull the image after an AWX commit lands, but before the new devel image gets pushed, then we might test on an AWX commit that doesn't correspond to the AWX devel image we have.

I think this is pretty easily solved by adding a `git_commit=` label to the image metadata when the image is built in AWX CI. Then we can pull :latest, check which commit it's built against, and check out that exact commit SHA.

I don't think this is enough of an issue to warrant doing that immediately, but it's a corner case worth noting here.

PR #351, opened 2024-04-24
Fix bug where org admins could modify superusers

AlanCoding:
Sorry I've kind of slow-rolled the work on this a little bit. Let me show the criteria this is borrowing from AWX in `UserAccess.can_admin`.

```python
        if obj.is_superuser or obj.is_system_auditor:
            # must be superuser to admin users with system roles
            return False
```

That's pretty clearly the same as the criteria I am adding here. I missed this because I was trying to _avoid_ this other criteria, which was:

```python
        if data is not None and ('is_superuser' in data or 'is_system_auditor' in data):
            if to_python_boolean(data.get('is_superuser', 'false'), allow_none=True) and not self.user.is_superuser:
                return False
            if to_python_boolean(data.get('is_system_auditor', 'false'), allow_none=True) and not (self.user.is_superuser or self.user == obj):
                return False
```

My reason for avoiding that was because it is re-validation and should be avoided (like `to_python_boolean` can't be fully trusted b/c it is not the proper source-of-truth). It should instead happen in the serializers (which has largely already been done). But that caused me to miss this other criteria. This should take care of it.

### DIFF
--- a/ansible_base/rbac/policies.py
+++ b/ansible_base/rbac/policies.py
@@ -34,6 +34,8 @@ def visible_users(request_user, queryset=None) -> QuerySet:
 def can_change_user(request_user, target_user) -> bool:
     if request_user.is_superuser:
         return True
+    elif target_user.is_superuser:
+        return False  # target is a superuser and request user is not
 
     if not get_setting('MANAGE_ORGANIZATION_AUTH', False):
         return False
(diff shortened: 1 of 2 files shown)


Dostonbek1 (approved):
confirmed on eda-server, this resolves the issue of normal user being able to edit superuser.

PR #352, opened 2024-04-25
Superuser: Visible Users for Related ViewSets

slemrmartin:
Superuser and SystemAuditor should see only the team/org members/admins where they GET i.e. `/organizations/<pk>/admins/`

### DIFF
--- a/ansible_base/rbac/policies.py
+++ b/ansible_base/rbac/policies.py
@@ -18,7 +18,10 @@ def visible_users(request_user, queryset=None) -> QuerySet:
     if has_super_permission(request_user, 'view') or (
         get_setting('ORG_ADMINS_CAN_SEE_ALL_USERS', False) and org_cls.access_ids_qs(request_user, 'change').exists()
     ):
-        return user_cls.objects.all()
+        if queryset is not None:
+            return queryset
+        else:
+            return user_cls.objects.all()
 
     object_id_fd = ObjectRole._meta.get_field('object_id')
     members_of_visble_orgs = ObjectRole.objects.filter(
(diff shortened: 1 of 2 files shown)


ansible_base/rbac/policies.py
```diff
@@ -18,7 +18,10 @@ def visible_users(request_user, queryset=None) -> QuerySet:
     if has_super_permission(request_user, 'view') or (
         get_setting('ORG_ADMINS_CAN_SEE_ALL_USERS', False) and org_cls.access_ids_qs(request_user, 'change').exists()
     ):
-        return user_cls.objects.all()
+        if queryset is not None:
+            return queryset
+        else:
+            return user_cls.objects.all()
```
AlanCoding, line 24: We have an identical condition later on in this method, we might want to combine those into 1 conditional at the start that may redefine `queryset`. But this should be perfectly functional

test_app/tests/rbac/api/test_user_permissions.py
```diff
@@ -197,6 +197,57 @@ def test_team_admins_can_add_children(self, user, user_api_client, organization,
...
+        self._initial_check(url, user_api_client, 1)
+        self._assign_users(org_member_rd, organization)
+
+        response = user_api_client.get(url)
+        assert response.status_code == 200
+        assert response.data['count'] == 3
+
+    def test_superadmin_list_org_members(self, user, user_api_client, organization, org_member_rd):
+        user.is_superuser = True
+        user.save()
```
AlanCoding, line 240: Would be more direct to use `admin_api_client` instead of doing this. But again, still perfectly functional, just a suggestion for code cleanliness.

AlanCoding (approved):

Issue #324, opened 2024-04-17
Support OPTIONS data show/hide of POST / PATCH / PUT capabilities


AlanCoding:
This might already work, but there are not tests for it. Examples:

 - If user doesn't have `change_inventory` permission, OPTIONS to `/api/v1/inventories/42/` should not show PATCH or PUT actions. Otherwise it should show these.
 - If user doesn't have `add_organization` permission, either by being a superuser, or having a global role that gives this permission, OPTIONS to `/api/v1/organizations/` should not show POST. Otherwise it should show it.
 - If user doesn't have `add_inventory` permission to any organizations, OPTIONS to `/api/v1/inventories/` should not show POST. Otherwise it should show it.

This issue asks to add tests for these 3 cases to show that it works for test_app.

PR #354, opened 2024-04-26
Test and fix hide/show of OPTIONS based on requesting user's permissions

AlanCoding:
Addresses the test cases given in https://github.com/ansible/django-ansible-base/issues/324

This was working correctly for the detail view already, and this only adds a condition to the list view, for the most part.


ansible_base/rbac/api/permissions.py
```diff
@@ -69,13 +69,21 @@ def has_permission(self, request, view):
...
             if parent_field_name is None:
                 result = self.has_create_permission(request, model_cls)
                 if (not result) and (not is_cloned_request(request)):
                     logger.warning(f'User {request.user.pk} lacks global add_{model_cls._meta.model_name} permission to create {model_cls._meta.model_name}')
                 return result
+        elif request.method == 'POST' and is_cloned_request(request):
+            # If this is OPTIONS purposes
+            # return a speculative answer about whether user might be generally able to create
+            model_cls = self._queryset(view).model
+            parent_model = permission_registry.get_parent_model(model_cls)
```
slemrmartin: this is crashing when model_cls is User

AlanCoding: ok good point, I'll add a new test case.

slemrmartin (approved):

PR #355, opened 2024-04-26
Use DRF clone_request in favor of copy which is unstable

AlanCoding:


relrod (approved):
I think flake8 will whine about the unused `copy` import, but lgtm after that.

AlanCoding:
Well this didn't work as intended, and then I watered it down significantly and now the tests work as intended, but I'm unsure if it still fixes the problem. We could test it and see...

ansible_base/lib/routers/association_resource_router.py
```diff
@@ -35,7 +36,7 @@ def get_parent_object(self):
         with the specified pk.
         """
         parent_view = self.parent_viewset()
-        parent_view.request = copy.copy(self.request)
+        parent_view.request = clone_request(self.request, 'GET')
         parent_view.request._request = copy.copy(self.request._request)
```
relrod, line 40: will this still trigger the recursion issue on old DRF?

slemrmartin (approved):

ansible_base/lib/routers/association_resource_router.py
AlanCoding: Let's try to answer by looking at the traceback in https://github.com/encode/django-rest-framework/discussions/8682

That saw

```
  File "~/python_playground/.venv/lib/python3.10/site-packages/rest_framework/request.py", line 426, in __getattr__
    return getattr(self._request, attr)
```

In the updated version of this, `self._request` is still copied. However, I think this does avoid the blanket `getattr` which is the problem. And if you look at the fix, it implies that the root cause is `__getattr__` deferring to the same-name property on `self._request` when `getattr` is called with `"_request"`.

Yeah, I kind of wish this fix was cleaner, but I also think maybe I should just trust the data and merge this knowing pretty well it will avoid the version-specific error.

relrod: It would probably be good to have a test that we know hits this bug in old DRF and expand the CI matrix to also test against old DRF, if we're intending to support it (which we seem to be with this patch).

Thoughts on that? (Let's assume it wouldn't make CI take any longer)

AlanCoding:
```
pip install djangorestframework==3.14.0
py.test test_app/tests/lib/
```

On `devel` branch this gives `21 failed, 239 passed, 2 xfailed, 10 warnings in 29.45s`

Now test fix

```
git checkout the_clone_wars
py.test test_app/tests/lib/ --create-db
```

That gives `260 passed, 2 xfailed, 10 warnings in 24.73s`

All that is with no addition test, or test change.

I guess this is my gripe about CI matrix with python versions. Sometimes it finds an issue, but only with relatively newer python stuff, which is more like asyncio or type hints. For the most part it doesn't do any good, while _python dependencies_ will aggressively break if anyone bothered to cross-version test. Those should theoretically go 1 of 2 paths, adding a version pin, or patching the code (like what this is doing). I'm sure somewhere there is tooling that will do a good job at this, but tox / github CI checks seems woefully insufficient.

PR #356, opened 2024-04-26
[Low Prio] add debug_timer decorator and other logs to auth.py

kdelee:
@john-westcott-iv  lmk if this is harmless enough to include

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -82,7 +82,7 @@ class JWTCommonAuth:
 
     @debug_timer
     def log_and_raise(self, conditional_translate_object, expand_values={}):
-        logger.debug(conditional_translate_object.not_translated() % expand_values)
+        logger.error(conditional_translate_object.not_translated() % expand_values)
         raise AuthenticationFailed(conditional_translate_object.translated() % expand_values)
 
     @debug_timer
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -1,4 +1,5 @@
 import logging
+import time
 from urllib.parse import urljoin, urlparse
 
 import jwt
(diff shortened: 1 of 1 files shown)


ansible_base/jwt_consumer/common/auth.py
```diff
@@ -64,10 +80,12 @@ def parse_jwt_token(self, request):
 
         return user, validated_body
 
+    @debug_timer
     def log_and_raise(self, conditional_translate_object, expand_values={}):
-        logger.error(conditional_translate_object.not_translated() % expand_values)
+        logger.debug(conditional_translate_object.not_translated() % expand_values)
```
kdelee:
```suggestion
        logger.error(conditional_translate_object.not_translated() % expand_values)
```

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -56,6 +70,8 @@ def parse_jwt_token(self, request):
                 "is_superuser": validated_body["is_superuser"],
             },
         )
+        delta_ms = (time.time() - start) * 1000
+        logger.debug(f'Getting/saving user took {delta_ms:.0f} (ms)')
```
AlanCoding, line 74: It would be better to refactor this code into a method and them apply the decorator the same as elsewhere.

AlanCoding:
My quick 2 cents, following the DAB philosophy.

The promise of DAB is to standardize things. Having a one-off utility method like `debug_timer` is fine in other libraries, and is fine if there's a valid case that the utility will not be encountered again soon. But here is a method in AWX which does what that does with bells & whistles:

https://github.com/ansible/awx/blob/4754819a0916cbd9352550fb0f8fbc65273854b0/awx/main/utils/common.py#L1197-L1225

So I want to make the case here that we already have a widely shared need established.

That established, the (1)st change I'd seek is putting this into proper `ansible_base/lib`. But I would go further, the (2)nd change I'd like to see is setting us up to delete `log_excess_runtime` in AWX.

I don't love everything about that method. I'd suggest that long-term it would be better to use a local logger (starting with `ansible_base.lib`) as opposed to passing in a logger in my prior implementation. This would allow us to have special policies for performance logs from ansible_base in the logger configuration. Anything else you want to change about its feature set, please have a shot at it.

john-westcott-iv:
I agree with Alan on this. Personally I can't decide if it should be `ansible_base.lib.decerators` or `ansible_base.lib.performance`. I kind of like the idea of `performace` because then you would end up with a logger like `ansible_base.lib.performance` and anything else you put in there would use that same logger. @kdelee if you don't have the time to dedicate to this now please cut a JIRA and we can do this as part of 2.5-next.

john-westcott-iv:
@kdelee do you have capacity to make the requested changes?

kdelee:
Closing with https://issues.redhat.com/browse/AAP-31678 filed as follow up

PR #360, opened 2024-04-30
Partially migrate another user editing condition from AWX

AlanCoding:
Let me link the legacy code here:

https://github.com/ansible/awx/blob/918d5b3565e5997bdbd57117b92852ba13bb55d6/awx/main/access.py#L713-L726

When you look at that, first note that there is a condition for "orphan" users --> this means users who are not in any organizations.

The need for this is clear from the test. If a user is not in _any_ organization, then we can't allow _any other user_ to manage that user. That's why we need an extra condition in this case. That's not up for debate.

Here, I'm using `is_superuser`, because I'm reluctant to copy over all the rules from AWX to DAB. This one, in particular, seems dubious. The reason why the AWX criteria had extra stuff was so that organization admins could POST to `/api/v2/users/` and then after that add the user as a member of their organization. I would like to punt this over to require an API solution, so that a user can be given an organization on creation @john-westcott-iv @slemrmartin

### DIFF
--- a/ansible_base/rbac/policies.py
+++ b/ansible_base/rbac/policies.py
@@ -44,6 +44,10 @@ def can_change_user(request_user, target_user) -> bool:
     if not get_setting('MANAGE_ORGANIZATION_AUTH', False):
         return False
 
+    # All users can chang their own password and other details
+    if request_user.pk == target_user.pk:
+        return True
+
     # If the user is not in any organizations, answer can not consider organization permissions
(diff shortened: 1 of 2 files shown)


slemrmartin (approved):
Approving, but (similarly to the PR's description) in the next version I would like to force User to be an Org Member when he/she becomes Team Member/Team Admin. Otherwise the change permissions are quite confusing

AlanCoding:
^ Ping @chadmf because there is already a sentiment that we need to take holistic view of platform-wide permissions and address the pain points. This ties into the larger UX flow, docs, and user recommendations.

PR #362, opened 2024-04-30
Using cache in jwt authentication

john-westcott-iv:
This leverages Djangos cache to store:
* The JWT decryption key
* References to the users that have authenticated 

It will default to reloading the JWT if a decryption fails or doing a user create if the user was cached but unable to be laoded from the DB.

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -1,9 +1,12 @@
 import logging
+from typing import Optional, Tuple
 from urllib.parse import urljoin, urlparse
 
 import jwt
 import requests
+from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.core.cache import caches
 from rest_framework.authentication import BaseAuthentication
 from rest_framework.exceptions import AuthenticationFailed
 
@@ -21,11 +24,21 @@ default_mapped_user_fields = [
     "is_superuser",
     "is_system_auditor",
 ]
+# This setting allows a service to override which django cache we want to use
+jwt_cache_name = getattr(settings, 'ANSIBLE_BASE_JWT_CACHE_NAME', 'default')
+cache = caches[jwt_cache_name]
+# This is the cache name we will use for the JWT key
+cache_key = 'ansible_base_jwt_public_key'
 
 
 class JWTCommonAuth:
     def __init__(self, user_fields=default_mapped_user_fields) -> None:
         self.mapped_user_fields = user_fields
+
+    def get_cache_timeout(self):
+        # If unspecified the cache will expire in 7 days
+        cache_timeout = get_setting('ANSIBLE_BASE_JWT_CACHE_TIMEOUT_SECONDS', 604800)
+        return cache_timeout
 
     def parse_jwt_token(self, request):
         logger.debug("Starting JWT Authentication")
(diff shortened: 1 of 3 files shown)


ansible_base/jwt_consumer/common/auth.py
```diff
@@ -21,6 +24,11 @@
     "is_superuser",
     "is_system_auditor",
 ]
+# This setting allows a service to override which django cache we want to use
+cache_setting = getattr(settings, 'ANSIBLE_BASE_JWT_CACHE_NAME', 'default')
```
relrod: This isn't used anywhere

john-westcott-iv: Good catch, the next line was supposed to use it.

john-westcott-iv:
@AlanCoding @relrod Since the AWX tests failed, should I notify someone?
I don't think this change should have affected the general AWX login but maybe it could have 🤷 ?

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -40,35 +48,96 @@ def parse_jwt_token(self, request):
...
-    def get_decryption_key_from_url(self, url, timeout, validate_certs):
+    def check_user_in_cache(self, validated_body: dict) -> Tuple[bool, dict]:
+        # These are the defaults which will get passed to the user creation and what we expect in the cache
+        expected_cache_value = {
+            "first_name": validated_body["first_name"],
+            "last_name": validated_body["last_name"],
+            "email": validated_body["email"],
+            "is_superuser": validated_body["is_superuser"],
+        }
+        cached_user = cache.get(validated_body["sub"], None)
```
newswangerd, line 122: It might be better to use the JWT signature as the cache key here with the user ID as the value. That way you can validate the incoming token by just taking the signature and checking if it is in the cache. If it is, you can assume that the token has already been validated and return the user that corresponds to the cached ID.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -40,35 +48,96 @@ def parse_jwt_token(self, request):
...
+        # If the user was in the cache and the values of the cache match the expected values we had it in cache
+        if cached_user is not None and cached_user == expected_cache_value:
+            return True, expected_cache_value
+        cache.set(validated_body["sub"], expected_cache_value)
```
kdelee:
```suggestion
        # The user was not previously in the cache, set the user in the cache so it is found on future requests
        cache.set(validated_body["sub"], expected_cache_value)
```

kdelee (approved):
LGTM left one optional suggestion for a comment

ansible_base/jwt_consumer/common/auth.py
john-westcott-iv: I thought about this but a JWT will be reissued if it expires which, I think, will change the signature. Going by the fields gets us around that while still causing processing if they subject of the key actually changes.

ansible_base/jwt_consumer/common/auth.py
newswangerd: we should put an expiration on this, even if it is a long one.

ansible_base/jwt_consumer/common/auth.py
newswangerd: ah, so this is just to determine if the user details need to be updated. Got it. I had in mind that we'd be caching the whole token and redoing the check every time the token expires.

ansible_base/jwt_consumer/common/auth.py
john-westcott-iv: Added a configurable 7 day timeout.

newswangerd (approved):

PR #364, opened 2024-05-01
Fix bug detecting running container that needs migration

AlanCoding:
I've been hitting this a lot lately, and the prior solution just wasn't working --> it went to error state when given a perfectly good running (but un-migrated) database.

```
python manage.py migrate --check; migrate_needed=$?
```

This was supposed to capture the exit code, but it didn't. In the first command, it just went straight to the trap. This seems to fix it... as devious and ugly as it is.

### DIFF
--- a/test_app/scripts/bootstrap.sh
+++ b/test_app/scripts/bootstrap.sh
@@ -25,7 +25,7 @@ then
     make postgres
 else
     echo "dab_postgres container is already running, will use that container"
-    python manage.py migrate --check; migrate_needed=$?
+    migrate_needed=$(echo $(python manage.py migrate --check > /dev/null 2> /dev/null; echo $?))
 fi
 
 MAX_ATTEMPTS=10
@@ -43,6 +43,7 @@ done
 
 if [ "${migrate_needed}" -ne 0 ]
 then
+    echo "Migrating database"
     python3 manage.py migrate
     DJANGO_SUPERUSER_PASSWORD=password DJANGO_SUPERUSER_USERNAME=admin DJANGO_SUPERUSER_EMAIL=admin@stuff.invalid python3 manage.py createsuperuser --noinput
     python3 manage.py authenticators --initialize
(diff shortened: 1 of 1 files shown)


test_app/scripts/bootstrap.sh
```diff
@@ -25,7 +25,7 @@ then
     make postgres
 else
     echo "dab_postgres container is already running, will use that container"
-    python manage.py migrate --check; migrate_needed=$?
+    migrate_needed=$(echo $(python manage.py migrate --check > /dev/null 2> /dev/null; echo $?))
```
relrod: This might be a cleaner fix.

Setting `-e` at the top is the reason we're bailing when the first command fails.

```suggestion
    set +e
    python manage.py migrate --check; migrate_needed=$?
    set -e
```

relrod: Or we can just get rid of the `-e` at the top if we wanted.

AlanCoding: This does result in a fairly long traceback that ends in:

```
django.db.utils.OperationalError: connection failed: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

But verbosity isn't really my concern, it still comes up. So I'll commit this.

relrod: well we don't need the output from the command here, so we can pipe it to >/dev/null (or 2>/dev/null if it's going to stderr, I'm not sure).

That would go before the semicolon, so:

```suggestion
    set +e
    python manage.py migrate --check 2> /dev/null; migrate_needed=$?
    set -e
```

I didn't test this - if it's still verbose then change the `2>` to just `>`

AlanCoding: Yeah, but we don't do that in any other cases. I'd worry about suppressing other errors. I've put in your first suggestion. That looks good to me.

relrod (approved):

PR #366, opened 2024-05-02
REMOTE_ADDR changes

john-westcott-iv:
Add function to `utils/encryption` for generating a validating hmac sha256 hashes based on a pre-shared secret.

Add `utils/requests` with methods for `get_remote_host` and `get_remote_hosts`.
`get_remote_hosts` will look at any headers in the setting `REMOTE_HOST_HEADERS` which defaults to `['REMOTE_ADDR', 'REMOTE_HOST']` and, if from a trusted proxy also `['HTTP_X_ENVOY_EXTERNAL_ADDRESS', 'HTTP_X_FORWARDED_FOR']`. 
`get_remote_host` will return the first address `get_remote_hosts` finds.

Alters TACACS authenticator to use `get_remote_host` instead of hard coding to `REMOTE_ADDR`.


ansible_base/lib/utils/encryption.py
```diff
@@ -70,3 +79,26 @@ def decrypt_string(self, value: str) -> str:
 
 
 ansible_encryption = SimpleLazyObject(func=lambda: Fernet256())
+
+
+def generate_hmac_sha256_shared_secret(nonce: Optional[str] = None) -> str:
+    shared_secret = getattr(settings, 'ANSIBLE_BASE_SHARED_SECRET', None)
+    if shared_secret is None:
+        logger.error("The setting ANSIBLE_BASE_SHARED_SECRET was not set, insecurely using default")
+        shared_secret = 'ansible_base_shared_secret_insecure'
```
relrod: I really don't like defaulting to a well-known PSK. That seems like it is going to get us a CVE.

test_app/tests/lib/utils/test_encryption.py
```diff
@@ -1,6 +1,10 @@
 import pytest
+from django.test import override_settings
```
relrod, line 2: This works, you can also just use the `settings` fixture and do `settings.FOO = 'bar'` in the test, it'll get reverted after the test finishes.

john-westcott-iv: Is there an advantage of one over the other? I've always just used `override_settings`.

ansible_base/lib/utils/encryption.py
john-westcott-iv: I made it raise an exception and will just not validate/generate the headers if this exception is thrown. For other exceptions I am doing to do a `logger.exectpion`.

ansible_base/lib/utils/encryption.py
```diff
@@ -70,3 +79,30 @@ def decrypt_string(self, value: str) -> str:
...
+
+
+class SharedSecretNotFound(Exception):
+    pass
+
+
+def generate_hmac_sha256_shared_secret(nonce: Optional[str] = None) -> str:
+    shared_secret = getattr(settings, 'ANSIBLE_BASE_SHARED_SECRET', None)
+    if shared_secret is None or not shared_secret:
+        logger.error("The setting ANSIBLE_BASE_SHARED_SECRET was not set, insecurely using default")
```
relrod: So I think this is outdated now :sweat_smile: 

Maybe something like:

```suggestion
        logger.error("The setting ANSIBLE_BASE_SHARED_SECRET was not set, some functionality may be disabled.")
```

ansible_base/lib/utils/requests.py
```diff
@@ -0,0 +1,46 @@
...
+        return remote_hosts
+
+    headers = get_setting('REMOTE_HOST_HEADERS', ['REMOTE_ADDR', 'REMOTE_HOST'])
+
+    # If we are connected to from a trusted proxy then we can add some additional headers
+    try:
+        if 'HTTP_X_TRUSTED_PROXY' in request.META and validate_hmac_sha256_shared_secret(request.META['HTTP_X_TRUSTED_PROXY']):
+            headers.insert(0, 'HTTP_X_FORWARDED_FOR')
+            headers.insert(0, 'HTTP_X_ENVOY_EXTERNAL_ADDRESS')
+    except SharedSecretNotFound:
```
relrod, line 37: Should we add another logger here? Otherwise it might not be obvious why the header isn't getting used.

docs/lib/requests.md
```diff
@@ -0,0 +1,12 @@
+## Remote Headers
+
+If you want to know what machine is making the request to your service you might be tempted to look at the `REMOTE_ADDR` header passed by Nginx.
+
+If you are running behind a proxy or load balancer the value in this header could be the IP of said device thus not indicating the actual host making the request.
+
+DAB offers two functions called `get_remote_host` and `get_remote_hosts` to help deal with this.
+
+`get_remote_hosts` will attempt to look at the setting `REMOTE_HOST_HEADERS` (defaulted to `['REMOTE_ADDR', 'REMOTE_HOST']`). For any named header in that array the code will split the header (if present) on `,` and then return an array of all the addresses found (in order with duplicates).
+Additionally if the header `HTTP_X_TRUSTED_PROXY` and is present and is properly signed with a pre-shared key, the method will automatically prepend the following to `REMOTE_HOST_HEADERS`: `['HTTP_X_FORWARDED_FOR', 'HTTP_X_ENVOY_EXTERNAL_ADDRESS']`.
```
relrod: Should we note the setting that needs to be set for that to work?

test_app/tests/lib/utils/test_encryption.py
relrod: Not really, just an alternative. I like to use it because then I don't have to remember where to import the other thing from, etc. You just add `settings` as a param and go to work.

relrod (approved):
A few more minor inlines, but approving because I think they are easy fixes.

ansible_base/lib/utils/requests.py
john-westcott-iv: Yea,I split this out into a nested if/else.

docs/lib/requests.md
john-westcott-iv: Done.

PR #369, opened 2024-05-07
Remove unwanted URL registrations from related endpoints

### LINKED ISSUE #275: Server error accessing URL that shouldn't exist (got 500, expected 404)
Reproduce by
 - `./test_app/scripts/bootstrap.sh`
 - GET http://127.0.0.1:8000/api/v1/organizations/1/teams/2/

### Expected result

404 Not Found

No values are expected at the end of this URL. If we were writing our the URL patterns, we would have written a regex like:

```python
r'^/api/v1/organizations/(?P<pk>[0-9]+)/teams/$'
```

Thus, Django should simply pass over this pattern in the URL search as not matched. The view shouldn't even be involved here, as the view should never be routed such a request.

### Actual result

Server error

```
[05/Apr/2024 20:52:28] "GET /api/v1/organizations/1/teams/2 HTTP/1.1" 301 0
Internal Server Error: /api/v1/organizations/1/teams/2/
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/views/decorators/csrf.py", line 56, in wrapper_view
    return view_func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/rest_framework/viewsets.py", line 124, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/rest_framework/mixins.py", line 56, in retrieve
    instance = self.get_object()
               ^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/rest_framework/generics.py", line 100, in get_object
    obj = get_object_or_404(queryset, **filter_kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/rest_framework/generics.py", line 19, in get_object_or_404
    return _get_object_or_404(queryset, *filter_args, **filter_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/shortcuts.py", line 85, in get_object_or_404
    return queryset.get(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/db/models/query.py", line 623, in get
    clone = self._chain() if self.query.combinator else self.filter(*args, **kwargs)
                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/db/models/query.py", line 1436, in filter
    return self._filter_or_exclude(False, args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/db/models/query.py", line 1454, in _filter_or_exclude
    clone._filter_or_exclude_inplace(negate, args, kwargs)
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/db/models/query.py", line 1461, in _filter_or_exclude_inplace
    self._query.add_q(Q(*args, **kwargs))
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/db/models/sql/query.py", line 1546, in add_q
    clause, _ = self._add_q(q_object, self.used_aliases)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/db/models/sql/query.py", line 1577, in _add_q
    child_clause, needed_inner = self.build_filter(
                                 ^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/db/models/sql/query.py", line 1427, in build_filter
    lookups, parts, reffed_expression = self.solve_lookup_type(arg, summarize)
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/db/models/sql/query.py", line 1237, in solve_lookup_type
    _, field, _, lookup_parts = self.names_to_path(lookup_splitted, self.get_meta())
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.11/site-packages/django/db/models/sql/query.py", line 1725, in names_to_path
    raise FieldError(
django.core.exceptions.FieldError: Cannot resolve keyword 'teams' into field. Choices are: admins, created, created_by, created_by_id, description, encryptioner, encryptioner_id, has_roles, id, member_roles, modified, modified_by, modified_by_id, name, organization, organization_id, related_fields_test_model_more_teams, related_fields_test_model_teams_with_no_view, resource, roleteamassignment, team_children, team_parents, tracked_users, users
[05/Apr/2024 20:52:28] "GET /api/v1/organizations/1/teams/2/ HTTP/1.1" 500 172747
```

### Commentary

Relevant registration is at:

https://github.com/ansible/django-ansible-base/blob/84d78132a4b92e699e4114254ff9b6aa455a8cac/test_app/router.py#L14

of course this triggers dynamically generated viewset & serializer for the purpose.

The root problem seems to be that `TeamViewSet` has all the viewset methods required for vendoring `/api/v1/teams/`, which includes its internal router automatically adding tails to the URL for detail view and other stuff.

AlanCoding:
Fixes https://github.com/ansible/django-ansible-base/issues/275

I finally hit a wall with this trying to add related `resource` fields to the user/team association endpoints. I went into http://127.0.0.1:8000/api/v1/docs/#/role_team_assignments/role_team_assignments_create and it had _duplicate view names_. This was caused by the same problem as 275. Random views were unintentionally getting registered.

So I had to bite the bullet and fix this. This bug was... kind of bad already. Those custom actions look concerning. It is true that the lookup field probably makes it unusable, but this isn't a good thing to rely on.

### DIFF
--- a/ansible_base/lib/routers/association_resource_router.py
+++ b/ansible_base/lib/routers/association_resource_router.py
@@ -1,5 +1,7 @@
 import copy
 import logging
+from itertools import chain
+from typing import Type
 
 from django.conf import settings
 from django.db.models.fields import IntegerField
@@ -11,6 +13,7 @@ from rest_framework.decorators import action
 from rest_framework.permissions import SAFE_METHODS
 from rest_framework.request import clone_request
 from rest_framework.response import Response
+from rest_framework.viewsets import ViewSetMixin
 
 from ansible_base.rbac.permission_registry import permission_registry
 
@@ -159,6 +162,11 @@ class AssociateMixin(RelatedListMixin):
         return super().get_serializer_class()
 
 
+@property
+def attribute_raiser(cls):
+    raise AttributeError
+
+
 class AssociationResourceRouter(routers.SimpleRouter):
     def get_method_map(self, viewset, method_map):
         is_associate_viewset = issubclass(viewset, AssociateMixin)
@@ -170,6 +178,26 @@ class AssociationResourceRouter(routers.SimpleRouter):
                     continue
                 bound_methods[method] = action_str
         return bound_methods
+
+    def associated_viewset_cls_factory(self, viewset: Type[ViewSetMixin]) -> Type[ViewSetMixin]:
+        """Given viewset (as a class) return a subclass containing all its actions except for list"""
+
+        class AssociatedViewSetType(type(viewset)):
+            """Metaclass that turns off viewset methods other than list"""
+
+            pass
+
+        custom_methods = chain.from_iterable(action.mapping.values() for action in viewset.get_extra_actions())
+
+        for method in ('retrieve', 'update', 'partial_update', 'destroy') + tuple(custom_methods):
+            setattr(AssociatedViewSetType, method, attribute_raiser)
+
+        class AssociatedViewSet(viewset, metaclass=AssociatedViewSetType):
+            """Adjusted version of given viewset for related endpoint with only list views"""
+
+            pass
+
+        return AssociatedViewSet
 
     def register(self, prefix, viewset, related_views={}, basename=None):
         if basename is None:
@@ -186,10 +214,13 @@ class AssociationResourceRouter(routers.SimpleRouter):
                 is_reverse_view = True
                 mixin_class = RelatedListMixin
 
+            # Start with a viewset that only has list action enabled
+            associated_viewset = self.associated_viewset_cls_factory(related_view)
+
             # Generate the related viewset
             modified_related_viewset = type(
                 f'Related{related_view.__name__}',
-                (mixin_class, related_view),
+                (mixin_class, associated_viewset),
                 {
                     'association_fk': fk,
                     'parent_viewset': viewset,
--- a/test_app/tests/lib/routers/test_association_resoure_router.py
+++ b/test_app/tests/lib/routers/test_association_resoure_router.py
@@ -3,7 +3,7 @@ from django.urls import reverse
 
 from ansible_base.lib.routers import AssociationResourceRouter
 from test_app import views
-from test_app.models import Inventory, User
+from test_app.models import Cow, Inventory, User
 
 
 def validate_expected_url_pattern_names(router, expected_url_pattern_names):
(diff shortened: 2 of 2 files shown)


test_app/tests/lib/routers/test_association_resoure_router.py
```diff
@@ -205,3 +205,30 @@ def test_sublist_filtering(inventory, organization, admin_api_client):
...
+def test_related_custom_actions_get_scrubed(organization, method, admin_api_client):
+    cow = Cow.objects.create(organization=organization)
+    url = f'/api/v1/organizations/{organization.pk}/cows/{cow.pk}/cowsay/'
+    # Can not use the reverse function like this, because post-fix, the view does not exist
+    # url = reverse('organization-cows-cowsay', kwargs={'pk': organization.pk, 'cows': cow.pk})
+    if method in ('PUT', 'POST'):
+        response = admin_api_client.get(url, data={})
+    else:
+        response = admin_api_client.get(url)
+    assert response.status_code == 404
```
AlanCoding, line 243: All of these tests were failing before this patch.

test_app/tests/lib/routers/test_association_resoure_router.py
```diff
@@ -186,12 +174,33 @@ def test_association_router_related_viewset_all_mapings(db):
     expected_urls = [
         'my_test_basename-detail',
         'my_test_basename-list',
-        'my_test_basename-teams-detail',
```
AlanCoding, line 189: I'm not going to lie, this kind of test assertion is a little bit upsetting. This URL should not be registered, but someone probably added it in just to make the test pass.

ansible_base/lib/routers/association_resource_router.py
```diff
@@ -159,6 +162,11 @@ def get_serializer_class(self):
         return super().get_serializer_class()
 
 
+@property
+def attribute_raiser(cls):
+    raise AttributeError
```
john-westcott-iv, line 167: Is an Attribute Error correct for these? Shouldn't this return a MethodNotAllowed response?

AlanCoding: Not a 405, a 404. See title, unwanted _URL registrations_, meaning endpoints that shouldn't exist were giving 500 errors. These are endpoints that _no_ method is valid for. If you have the wrong method (GET/POST) but another method would have worked, that's a 405, but that's not what this is fixing. This is _how_ we get that behavior. This is guided missile to programmed to a specific `hasattr` in `SimpleRouter.get_method_map`:

```python
        bound_methods = {}
        for method, action in method_map.items():
            if hasattr(viewset, action):
                bound_methods[method] = action
        return bound_methods
```

I skipped ahead without explaining this in great detail. But this is the way viewsets work, if a method is defined on a viewset, then that's a URL. That method can have multiple valid methods. I need to remove the `retrieve` method and others. If left in, these would create "phantom" endpoints, as in, a detail endpoint nested in the association endpoint, because it got left in the association viewset. So we need to leave it in the original (parent) viewset (because it is used for other things) and remove it as we subclass it.

The unorthodox thing is that `viewset` is a _class_, and we need a _subclass_ that gives `False` for `hasattr(viewset, 'retrieve')`. That is what `AttributeError` does. It's a method-scrubber, isolated to the new subclass.

AlanCoding:
Other PR is so good I think I want that instead. https://github.com/ansible/django-ansible-base/pull/370

PR #370, opened 2024-05-08
Fix 500 on unintended endpoints and further refactor association router

AlanCoding:
Something drove me crazy after testing with my last patch - which is that certain serializers (with the same name) were created twice. This is logical if you think about what's being done. This PR adds a registry to fix the problem.

It also trades a few things in code organization. It gets rid of the "factory" method, and adds a new utility method for getting the parent class, which has to follow a certain format. Then the registry management becomes the main job of `get_serializer_class`

Recapping of major changes being done here:

- Serializers for association & disassociation are refactored so that they don't do "rbac" filtering by their own logic. This change gets rid of the user-specialized association serializer using `visible_users` and the general `access_qs` call in favor of calls to `filter_queryset`.
- As a subpoint of the above point, `get_serializer_class` no longer builds querysets (for `instances`). This changes it to use methods from the view _instance_ inside of the serializer context. This has to happen on the serializer `__init__`.
- Disassociation validation that the object is already attached is deleted, replaced by queryset logic, which now uses the same `get_queryset` as the listing does, which only includes attached objects to begin with.
- New parent class for related model viewsets is added which is inserted and _lower_ precedence than the passed-in class. This means that methods defined here can be overwritten in order to customize behavior. These are listed in docs too.
- When synthesizing the new association viewset class, all methods _other than list_ are disabled by a metaclass. This avoids the duplicate registration I complained above above.
- A serializer registry is added so that we don't create the same-name class multiple times, which confuses OpenAPI.

### DIFF
--- a/ansible_base/lib/routers/association_resource_router.py
+++ b/ansible_base/lib/routers/association_resource_router.py
@@ -22,6 +22,54 @@ from ansible_base.rbac.permission_registry import permission_registry
 logger = logging.getLogger('ansible_base.lib.routers.association_resource_router')
 
 
+# Registry contains subclasses of AssociationSerializerBase indexed by name
+# this prevents duplicate names which would cause schema to not render correctly
+serializer_registry = {}
+
+
+class AssociationSerializerBase(serializers.Serializer):
+    """It is expected that final subclasses will set the instances field"""
+
+    instances = None
+
+    def get_queryset_on_init(self, original_qs: QuerySet) -> QuerySet:
+        raise NotImplementedError
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        new_qs = self.get_queryset_on_init(self.fields['instances'].child_relation.queryset)
+        self.fields['instances'].child_relation.queryset = new_qs
+
+
+class DisassociationSerializerBase(AssociationSerializerBase):
+    """Serializer used for removing objects that are currently associated via a many-to-many relationship"""
+
+    def get_queryset_on_init(self, original_qs: QuerySet) -> QuerySet:
+        if 'view' in self.context:
+            view = self.context['view']
+            if 'pk' in view.kwargs:
+                return view.get_queryset()
+        return original_qs
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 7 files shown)


AlanCoding:
I've created https://github.com/ansible/django-ansible-base/issues/393 to discuss 2 remaining things I'm still not happy with. But I don't want those in scope here.

ansible_base/lib/routers/association_resource_router.py
```diff
@@ -120,45 +148,82 @@ def disassociate(self, request, **kwargs):
...
+        super().__init__(*args, **kwargs)
+        self.fields['instances'] = serializers.PrimaryKeyRelatedField(queryset=self.get_queryset_on_init(), many=True)
+
+
+class FilteredAssociationSerializer(AssociationSerializerBase):
+    viewset_instance = None
+
+    def get_queryset_on_init(self) -> QuerySet:
+        qs = super().get_queryset_on_init()
+        if self.viewset_instance and hasattr(self.viewset_instance, 'filter_associate_queryset'):
```
AlanCoding: This callback could still be an issue... and this also leaves a viewset object referenced by objects in the registry which creates garbage collection problems. So it might be better, if possible, to refer to the view inside of the context. As I previously deleted `request = self.context['request']`. We might want to resurrect that to get the view instead of the request. Like this:

https://github.com/ansible/awx/blob/d4bdaad4d83a912d05dfbc833b5aeef4e915403c/awx/conf/serializers.py#L64

john-westcott-iv:
@AlanCoding to make a merge meeting for this PR sometime next week.

docs/lib/routers.md
```diff
@@ -45,6 +51,26 @@ Related views are expressed in the following format:
...
 
 
+## Customization
+
+Several methods defined in the `<ViewSet for relation>` will have an effect on constructed related endpoints.
+Those are:
+
+ - `get_sublist_queryset` - items shown in the listing _before_ filtering, OR candidate items for disassociation
+ - `filter_queryset` - filter applied to items shown in sublist, which works the same as the viewset by itself
+ - `get_queryset` - candidate items to associate _before_ filtering
```
AlanCoding: delete

relrod (approved):

Issue #363, opened 2024-04-30
AssociationResourceRouter limitations


relrod:
### Bug Summary

This ticket exists just to track known limitations with AssociationResourceRouter.

This will some day motivate me to write a better router.

The current `AssociationResourceRouter`...
* Cannot be used with custom through-tables for m2m relationships. If you want to store any metadata about a relationship besides "is related or is not related" then the current implementation fails at this.
* Is not composable. Like, at all. If I register a "users" route with some related views in App A, and then I write the Animal Ownership Tracker app in DAB, I cannot, from DAB, inject something in the related views of the route App A registered, so that /users/N/animals/ works. The only workaround right now is to document "if you want this endpoint to work, add this line to your route registration in your app" which isn't great.
* #275 
* Has no way to filter down the queryset. If you want to expose routes for `/users/N/authorized_tokens/` and `/users/N/personal_tokens/` you can't do that easily.

AlanCoding:
I tried to stay away from these concerns, but with https://github.com/ansible/django-ansible-base/pull/370 I have bumped up against some.

> If you want to store any metadata about a relationship besides "is related or is not related" then the current implementation fails at this.

In that case you couldn't use the `PrimaryKeyRelatedField` field. It would hardly be legitimate to call this an "association". You're creating the through object.

> I cannot, from DAB, inject something in the related views of the route App A registered, so that /users/N/animals/ works. 

Ok, but don't. If you want a `/users/N/animals/` endpoint, you should define this in the routes for App A. You _should_ not doing that "from DAB". I have this problem with AWX, as users should probably have a sublist of role assignments... but AWX shares none of this. So just write it however AWX writes stuff. If DAB made this endpoint, that would offer an inconsistent API with different patterns for association than the rest of the API, for example.

> Has no way to filter down the queryset. If you want to expose routes for `/users/N/authorized_tokens/` and `/users/N/personal_tokens/` you can't do that easily.

Good example, but yes you can do that. The solution I put in so-far was to allow `filtered_queryset` from the related model viewset that you pass in to work on the list. So if you don't want entries to be filtered, then introduce a `UnfilteredTokenViewSet` as opposed to another `TokenViewSet`, the latter of which is filtered to what you can see.

[...]

PR #374, opened 2024-05-08
[Low Priority] View documentation

### LINKED ISSUE #371: Document view-level `extra_related_fields()`
... in `docs/lib/views.md`

Refs: https://github.com/ansible/django-ansible-base/pull/133#discussion_r1592715843

john-westcott-iv:
Fixes https://github.com/ansible/django-ansible-base/issues/371
Requires https://github.com/ansible/django-ansible-base/pull/133 (needs rebase after this has merged)

### DIFF
--- a/docs/lib/views.md
+++ b/docs/lib/views.md
@@ -48,3 +48,32 @@ class DefaultAPIView(AnsibleBaseView):
     all my good stuff
     ...
 ```
+
+
+## Adding extra related fields
+
+There may be cases where the associative router does not provide a way to add specific related views to your view. As a way to help combat this the `AnsibleBaseView` class has a method defined called `extra_related_fields` which returns an empty dict (`{}`) by default. The entries expected in this dictionary are in the format:
+```
+{
+    "string": "url",
+}
+```
+Where string is the name of the field under the related section in the API and URL is the URL to that object.
+
+For example, if `extra_related_fields` returned:
+```
+{
+    'authenticators': reverse('user-authenticators-list', kwargs={'pk': obj.pk}),
+}
+```
+
+We would expect to see the related fields for an object to be something along the lines of:
+```
+    "related": {
+        "authenticators": "/api/gateway/v1/users/1/authenticators/"
+    }
+```
+
+The URL for that entry, in our example, will be generated from the reverse lookup user-authenticators-list url and should end up as something like `/api/users/:id/authenticators/` where :id comes from the `obj.pk`. 
+
+This can be particularly useful if a feature in DAB wants to extend a view that is created by a service. In this case you can create a MixIn like `ansible_base.oauth2_provider.views.user_mixin.DABOAuth2UserViewsetMixin` which can set `extra_related_fields` and then have the view in the service extend the Mixin. 


docs/lib/views.md
```diff
@@ -48,3 +48,22 @@ class DefaultAPIView(AnsibleBaseView):
...
+There may be cases where the associative router does not provide a way to add specific related views to your view. As a way to help combat this the `AnsibleBaseView` class has a method defined called `extra_related_fields` which returns an empty dict (`{}`) by default. The entries expected in this dictionary are in the format:
+```
+{
+    "string": "url",
+}
+```
+Where string is the name of the field under the related section in the API and URL is the URL to that object. For example, a return like:
+```
+{
+    'authenticators': reverse('user-authenticators-list', kwargs={'pk': obj.pk}),
```
AlanCoding, line 66: Where does `obj` come from?

AlanCoding: ...a return like:

...from `UserViewSet.extra_related_fields`...

docs/lib/views.md
```diff
@@ -48,3 +48,22 @@ class DefaultAPIView(AnsibleBaseView):
...
+}
+```
+Would cause `authenticators` to be listed under the related field and the related URL will be generated from the user-authenticators-list url and should end up as something like `/api/users/:id/authenticators/`.
+
+This can be particularly useful if a feature in DAB wants to extend a view that is created by a service. In this case you can create a MixIn like `ansible_base.oauth2_provider.views/user_mixin.DABOAuth2UserViewsetMixin` which can set `extra_related_fields` and then have the view in the service extend the MixIn.
```
AlanCoding: > This can be particularly useful if a feature in DAB wants to extend a view that is created by a service.

As I said in https://github.com/ansible/django-ansible-base/issues/363, I don't think that's actually a good idea.

> `ansible_base.oauth2_provider.views/user_mixin.DABOAuth2UserViewsetMixin`

`/` -> `.`?

> and then have the view in the service extend the MixIn.

I'd prefer capitalization "Mixin"

And I also don't get it. Wouldn't the app just import a `AuthenticatorViewSet` from DAB and declare that in related views when registering `UserViewSet`? This thing here is really complicated and I don't see what it does that's useful.

docs/lib/views.md
```diff
@@ -48,3 +48,22 @@ class DefaultAPIView(AnsibleBaseView):
...
+    "string": "url",
+}
+```
+Where string is the name of the field under the related section in the API and URL is the URL to that object. For example, a return like:
+```
+{
+    'authenticators': reverse('user-authenticators-list', kwargs={'pk': obj.pk}),
+}
+```
+Would cause `authenticators` to be listed under the related field and the related URL will be generated from the user-authenticators-list url and should end up as something like `/api/users/:id/authenticators/`.
```
AlanCoding: > cause `authenticators` to be listed under the related field and the related URL

The language isn't super clear. I would suggest "under the related field" --> "inside of the "related" dictionary in the User model's serializer"

docs/lib/views.md
AlanCoding: There's too much ground to cover here. Start with the basics. You have a model in the end-user app, and a related model in DAB. You want to define related endpoints of the end-user's model (User) in the DAB app. You make it sound like you're doing this in DAB, but that's not 100% true even in what you're describing here. You define a mixin _and use that in the end-user's app_. That needs to be spelled out in greater detail. It _is_ defined in the end-user app, what you're describing is just a method for vendoring helper code that the end-user app _can_ use, from DAB, when there is references between those models.

jctanner (approved):
Given these are internal project docs, it doesn't have to be perfect on the first commit.

AlanCoding (approved):

PR #375, opened 2024-05-08
Disallow editing of system user

bhavenst:
Adds new base serializer for users that disallows editing of the system user (settings.SYSTEM_USERNAME).

### DIFF
--- a/ansible_base/lib/serializers/common.py
+++ b/ansible_base/lib/serializers/common.py
@@ -3,6 +3,7 @@ import logging
 from django.db.models.fields import NOT_PROVIDED
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
+from rest_framework.exceptions import PermissionDenied
 from rest_framework.fields import empty
 from rest_framework.serializers import ValidationError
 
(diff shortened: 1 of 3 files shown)


ansible_base/lib/serializers/common.py
```diff
@@ -92,3 +95,14 @@ class Meta(CommonModelSerializer.Meta):
...
+
+
+class UneditableSystemUserSerializer(CommonModelSerializer):
+    """
+    Disallows editing of system user (settings.SYSTEM_USERNAME).
+    Make sure to call super if you override validate(..).
+    """
+
+    def validate(self, data):
+        if data["username"] == settings.SYSTEM_USERNAME:
```
AlanCoding: This would not prevent, say, a PATCH to the system user to change their username to something other than the system username. You also need to check if `self.instance.username` is the system username.

bhavenst: Thanks!  I'd already noticed it wasn't working in all cases after testing some things.  You clued me in to self.instance though, so I think comparing the IDs there will be more robust.  Please let me know if you see issues w/ the new method..

ansible_base/lib/serializers/common.py
```diff
@@ -98,3 +101,15 @@ class Meta(CommonModelSerializer.Meta):
 class ImmutableCommonModelSerializer(AbstractCommonModelSerializer):
     class Meta(AbstractCommonModelSerializer.Meta):
         fields = AbstractCommonModelSerializer.Meta.fields + ['created', 'created_by']
+
+
+class UneditableSystemUserSerializer(CommonModelSerializer):
```
AlanCoding: I would be inclined to move the `validate_is_superuser` method here as well, and have it be a generic DAB user serializer.

ansible_base/lib/serializers/common.py
```diff
@@ -98,3 +101,15 @@ class Meta(CommonModelSerializer.Meta):
...
+    """
+    Disallows editing of system user (settings.SYSTEM_USERNAME).
+    Make sure to call super if you override validate(..).
+    """
+
+    def validate(self, data):
+        if hasattr(self.instance, "id") and self.instance.id == models.get_system_user().id:
```
AlanCoding: It should always have an id (or at least a pk), but `hasattr(self, 'instance')` might be a necessary check.

bhavenst: I found that id isn't present for creates (since this runs before it gets populated by the ORM).  But I didn't see a case where instance wasn't populated.  No problem to add that though.

ansible_base/lib/serializers/common.py
bhavenst: Sounds reasonable to me.

bhavenst: OK, did that, please take a look.

AlanCoding (approved):

PR #376, opened 2024-05-08
Add more test_app api v1 links

AlanCoding:
This adds links to more things, like authenticator plugins IIRC, because that app adds a lot of URLs which are not via a router.

### DIFF
--- a/test_app/views.py
+++ b/test_app/views.py
@@ -1,6 +1,8 @@
 from itertools import chain
 
 from django.shortcuts import render
+from django.urls.exceptions import NoReverseMatch
+from django.urls.resolvers import URLPattern
 from rest_framework.decorators import action, api_view
 from rest_framework.response import Response
 from rest_framework.reverse import reverse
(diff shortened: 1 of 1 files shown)


test_app/views.py
```diff
@@ -138,9 +140,14 @@ def api_root(request, format=None):
...
-    for url in docs_urls:
-        list_endpoints[url.name] = reverse(url.name, request=request, format=format)
+    from ansible_base.authentication.urls import api_version_urls as authentication_urls
+
+    for url in docs_urls + authentication_urls[1:]:
+        if isinstance(url, URLPattern):
+            try:
+                list_endpoints[url.name] = reverse(url.name, request=request, format=format)
+            except NoReverseMatch:
+                pass
```
john-westcott-iv, line 168: We have a sorted API view page in gateway, I wonder if we should put that in DAB for others to use? Not relavent to this PR though.

john-westcott-iv (approved):

test_app/views.py
AlanCoding: This is rather hard to maintain. @fosterseth added it a while ago... and yeah it was needed, but it's a challenge. If there is a code pattern that does this more generally (for top-level links) then I'd love to have that in the main DAB code base for general use. Maybe some sub-selection by app would be nice too.

PR #377, opened 2024-05-08
More fixes from warnings in /docs/ page

AlanCoding:
This is addressing things you see in the console when you navigate to the /api/v1/docs/ page in test_app.

Also ping @dhaustein

### DIFF
--- a/ansible_base/activitystream/serializers.py
+++ b/ansible_base/activitystream/serializers.py
@@ -1,5 +1,6 @@
 import logging
 from copy import deepcopy
+from typing import Optional
 
 from rest_framework import serializers
 
@@ -32,7 +33,7 @@ class EntrySerializer(ImmutableCommonModelSerializer):
         if obj.content_type:
             return obj.content_type.model
 
-    def get_related_content_type_model(self, obj) -> str:
+    def get_related_content_type_model(self, obj) -> Optional[str]:
         if obj.related_content_type:
             return obj.related_content_type.model
 
--- a/ansible_base/activitystream/serializers.py
+++ b/ansible_base/activitystream/serializers.py
@@ -28,11 +28,11 @@ class EntrySerializer(ImmutableCommonModelSerializer):
     related_content_type_model = serializers.SerializerMethodField()
     changes = serializers.SerializerMethodField()
 
-    def get_content_type_model(self, obj):
+    def get_content_type_model(self, obj) -> str:
         if obj.content_type:
             return obj.content_type.model
 
-    def get_related_content_type_model(self, obj):
+    def get_related_content_type_model(self, obj) -> str:
         if obj.related_content_type:
             return obj.related_content_type.model
 
@@ -47,7 +47,7 @@ class EntrySerializer(ImmutableCommonModelSerializer):
         field = model._meta.get_field(field_name)
         return field.to_python(value)
 
-    def get_changes(self, obj):
+    def get_changes(self, obj) -> dict[str, dict]:
         """
         We store strings, we have to convert them back to the correct type.
         """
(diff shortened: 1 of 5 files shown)


ansible_base/api_documentation/customizations.py
```diff
@@ -0,0 +1,9 @@
+def apply_authentication_customizations() -> None:
+    """Make the schema generator of the type of DAB authentication classes"""
+    from drf_spectacular.authentication import SessionScheme
+
+    from ansible_base.authentication.session import SessionAuthentication
+
+    class MyAuthenticationScheme(SessionScheme):
+        target_class = SessionAuthentication  # full import path OR class ref
+        name = 'SessionAuthentication'
```
john-westcott-iv: So if authentication application is enabled we create this class but we don't seem to be doing anything with it? Does this get "magically" consumed by drf_spectacular or something? If so can we add a comment around that?

AlanCoding: In general the drf_spectacular library works like this - you have certain classes in imported modules, and it crawls them to build the schema.

But what you might be confused about here is that I only want this imported if _both_ the api_documentation and authentication apps are enabled, which is why this is called from the api_documentation ready method, gated on authentication being installed.

We could just define this below the `SessionAuthentication` class, and I think other places do that. But I think that's wrong, because `requirements_authentication.in` does not list drf_spectacular as a dependency, so it should be an import error if someone _only_ used authentication.

I would leave more of this in a comment, but I'm not sure if any of this is what your question is.

AlanCoding: added more docs

ansible_base/activitystream/serializers.py
```diff
@@ -28,11 +29,11 @@ class Meta:
     related_content_type_model = serializers.SerializerMethodField()
     changes = serializers.SerializerMethodField()
 
-    def get_content_type_model(self, obj):
+    def get_content_type_model(self, obj) -> str:
         if obj.content_type:
             return obj.content_type.model
 
-    def get_related_content_type_model(self, obj):
+    def get_related_content_type_model(self, obj) -> Optional[str]:
```
AlanCoding, line 36: There are some subtleties in the type hints here, @relrod

ansible_base/activitystream/serializers.py
```diff
@@ -28,11 +29,11 @@ class Meta:
     related_content_type_model = serializers.SerializerMethodField()
     changes = serializers.SerializerMethodField()
 
-    def get_content_type_model(self, obj):
+    def get_content_type_model(self, obj) -> str:
```
relrod: I disagree with this type-hint. What if `not obj.content_type`?

relrod: ~also, wait, `str`? That can't be right~

or yes it is, I'm dumb. Still though, this can return `None`

AlanCoding: You have 1 test that returned `None` by artificially creating an content type. That's fine to test but that's not realistic.

relrod: What if the model class goes away entirely in a future migration? That's what the test is really testing, iirc. And that seems like a realistic situation, even if rare.

AlanCoding: ContentTypes are not cleaned up when removing a model, intended by design.

These type hints produce documentation for _API clients_. If a model class goes away, and the app removes its content type, yet for some reason doesn't remove the activity stream entries, then that's a bug of the app developer.

I believe it's a bad decision, but I will make this change.

relrod: I mean I'm also fine with just removing the conditional if we're _that_ confident. But I think the type signature should match the code (and really that should be enforced by the type checker, but alas)

relrod (approved):

PR #378, opened 2024-05-08
Encrypt `extra_data` to not expose things that might be used by authenticators

thedoubl3j:
When a user logs in through an authenticator we capture information about the login into a field called extra_data. The authenticator (ldap/keycloak/github/any others) could add sensitive information which is then plain text in the database. 

This PR encrypts that data before it goes in to the database and includes test coverage for those functions.

I also included an update to the requirements file for the dev environment as they are needed for testing/migration making for test app. I can split those out into a separate PR if needed. Django is pinned to the same versions that are required for regular install. 

# Testing Steps
1. Add any authenticator you wish to the API. Easiest is just enable ldap via the dev stack in the container start up file and un comment the `ldap_enabled: True` line. 
2. login in with a user via the authenticator. 
3. open a dbshell `docker exec -it aap_gw_1 aap-gateway-manage $* dbshell`
4. select the extra_data column from authenticatoruser (this will return any authenticatorUser in the db, change the query if you wish to select a specific one but _everything_ should be encrypted for `extra_data`) `select extra_data from dab_authentication_authenticatoruser;`
5. confirm the field starts with `$(encrypted)`


BrennanPaciorek:
Can confirm that the extra_data fields start with `$encrypted` when following the `Testing Steps` for both the ldap authenticator and the local authenticator

requirements/requirements_dev.txt
```diff
@@ -1,7 +1,10 @@
 ansible  # Used in build process to generate some configs
 build
-django-extensions
+Django>=4.2.5,<4.3.0
```
BrennanPaciorek, line 3: what change introduces this version constraint?

BrennanPaciorek (approved):
lgtm. Manual testing works on my end. only have minor questions about the PR that should not block merging.

requirements/requirements_dev.txt
thedoubl3j: this what the project pins to as whole, we just didn't have django in the dev environment so its not picked up in the bootstrapping of a dev environment so I added it to the dev reqs since in some cases (needing any kind of django management utility for example), django is needed in the virtenv.

ansible_base/authentication/models/authenticator_user.py
```diff
@@ -50,3 +55,6 @@ def save(self, *args, **kwargs):
         self.extra_data = b64_encode_binary_data_in_dict(self.extra_data)
 
         super().save(*args, **kwargs)
+
+        # After super().save extra_data will be encrypted but we need this to field to not be encrypted during the login process
+        self.extra_data = json.loads(ansible_encryption.decrypt_string(self.extra_data))
```
john-westcott-iv: We shouldn't need json.loads here anymore if we are going to do it inside of the decrypt_string function.

ansible_base/lib/abstract_models/common.py
```diff
@@ -161,8 +161,11 @@ def from_db(self, db, field_names, values):
 
         for field in self.encrypted_fields:
             field_value = getattr(instance, field, None)
-            if field_value:
-                setattr(instance, field, ansible_encryption.decrypt_string(field_value))
+            if field_value and type(field_value) is str and field_value.startswith("$encrypted$"):
+                decrypted_field_value = ansible_encryption.decrypt_string(field_value)
+                if isinstance(self._meta.get_field(field), models.JSONField):
+                    decrypted_field_value = json.loads(decrypted_field_value)
+                setattr(instance, field, decrypted_field_value)
```
john-westcott-iv: Same thing with json.loads here. Also note, that decrypt_string will not auto check the value and return it directly so I don't think we need any changes in here anymore.

ansible_base/lib/utils/encryption.py
```diff
@@ -84,27 +85,18 @@ def is_encrypted_string(self, value: Any) -> Tuple[bool, Optional[str], Optional
         if not isinstance(value, str):
             value = str(value)
 
+    def encrypt_string(self, value: Any) -> str:
+        # Its possible for a serializer to accept a number for a CharField (like 5). In the serializer its "5" but when we get here it might be 5
+        if type(value) is not str:
+            value = json.dumps(value)
+
```
john-westcott-iv: This got merged into the middle of the `is_encrypted_function` it needs to be moved below the `return False, None, None`.

requirements/requirements_dev.txt
john-westcott-iv, line 3: Can you expand on your comment around this? We don't typically run django management commands in the dev env directly. Usually we do it through a running gateway container.

ansible_base/lib/utils/encryption.py
```diff
@@ -83,18 +84,18 @@ def is_encrypted_string(self, value: Any) -> Tuple[bool, Optional[str], Optional
...
 
-    def encrypt_string(self, value: str) -> str:
+    def encrypt_string(self, value: Any) -> str:
         # Its possible for a serializer to accept a number for a CharField (like 5). In the serializer its "5" but when we get here it might be 5
-        if not isinstance(value, str):
-            value = str(value)
-
+        if type(value) is not str:
+            value = json.dumps(value)
+
```
john-westcott-iv:
```suggestion
        value = json.dumps(value)
```

ansible_base/lib/utils/encryption.py
```diff
@@ -120,7 +121,14 @@ def decrypt_string(self, value: str) -> str:
...
 
-        return smart_str(self.decrypt(encrypted))
+        decrypted_value = smart_str(self.decrypt(encrypted))
+        try:
+            return json.loads(decrypted_value)
+        except json.JSONDecodeError:
+            # If we had a string like 'test' JSON will go crazy if its not wrapped in quotes and file the decryption
+            # So here we are going to assume that the decode failed because of this and try decoding with the string wrapped in ""'s
+            # If this fails, we will just raise up the error because decoding is not working
+            return json.loads(f'"{decrypted_value}"')
```
john-westcott-iv:
```suggestion
        decrypted_value = smart_str(self.decrypt(encrypted))
        try:
            return json.loads(decrypted_value)
        except json.JSONDecodeError as e:
             logger.exception("Failed to decode encrytped value as json from database")
             raise e
```

jctanner (approved):

trucdg (approved):
great work! :hand_with_index_finger_and_thumb_crossed:

PR #380, opened 2024-05-08
create orgs&teams from auth maps

jctanner:


ansible_base/authentication/utils/claims.py
```diff
@@ -268,8 +273,96 @@ def update_user_claims(user: Optional[AbstractUser], database_authenticator: Aut
...
+
+    # Load existing teams
+    existing_teams = load_existing_teams(all_teams)
+
+    # Create missing teams
+    create_missing_teams(all_teams, team_map, existing_orgs, existing_teams)
+
+
+def load_existing_orgs(org_names):
+    Organization = get_organization_model()
```
cutwater: No need to  call that every time, you can call `get_organization_model()` and `get_team_model()` once at the module level.

ansible_base/authentication/utils/claims.py
```diff
@@ -268,8 +273,96 @@ def update_user_claims(user: Optional[AbstractUser], database_authenticator: Aut
...
+            org_name = team_map[team_name]
+            if org_name in existing_orgs:
+                logger.info(f"creating team {team_name} in org {org_name}")
+                Team.objects.get_or_create(name=team_name, organization_id=existing_orgs[org_name])
+            else:
+                logger.error(f"attempting to create team {team_name} but its organization does not exist")
+
+
 class ReconcileUser:
     def reconcile_user_claims(user, authenticator_user):
```
cutwater: Please add `@staticmethod` here. This method definition is currently incorrect.

ansible_base/lib/utils/auth.py
```diff
@@ -0,0 +1,24 @@
+from django.apps import apps as django_apps
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+
+def get_model_from_settings(setting_name):
+    """
+    Return the User model that is active in this project.
+    """
+    setting = getattr(settings, setting_name)
```
cutwater:
```suggestion
    try:
        setting = getattr(settings, setting_name)
    except AttributeError:
        raise ImproperlyConfigured(f"{setting_name} is not defined in settings.")
```

ansible_base/lib/utils/auth.py
```diff
@@ -0,0 +1,24 @@
+from django.apps import apps as django_apps
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+
+def get_model_from_settings(setting_name):
```
cutwater: I'd encourage using type hints when writing new code:
```suggestion
def get_model_from_settings(setting_name: str) -> Any:
```

ansible_base/lib/utils/auth.py
```diff
@@ -0,0 +1,24 @@
...
+    setting = getattr(settings, setting_name)
+    try:
+        return django_apps.get_model(setting, require_ready=False)
+    except ValueError:
+        raise ImproperlyConfigured(f"{setting_name} must be of the form 'app_label.model_name'")
+    except LookupError:
+        raise ImproperlyConfigured(f"{setting_name} refers to model '{setting}' that has not been installed")
+
+
+def get_team_model():
```
cutwater:
```suggestion
def get_team_model() -> Type[AbstractTeam]:
```

ansible_base/lib/utils/auth.py
```diff
@@ -0,0 +1,24 @@
...
+    return get_model_from_settings('ANSIBLE_BASE_TEAM_MODEL')
+
+
+def get_organization_model():
```
cutwater:
```suggestion
def get_organization_model() -> Type[AbstractOrganization]:
```

ansible_base/authentication/utils/claims.py
```diff
@@ -268,8 +273,96 @@ def update_user_claims(user: Optional[AbstractUser], database_authenticator: Aut
...
-        logger.error("TODO: Fix reconciliation of user claims")
+
+        if not authenticator_user.provider.create_objects:
+            logger.info("Skip reconciling user claims")
+            return
```
cutwater: If I understand requirements correctly when `create_objects = False`, the user reconciliation logic should assign user to existing teams and organizations but not to create them.
In fact this is something I'm currently working and if this logic is supposed to land here, it would be nice to add some comment that this code is not fully implemented and that skipping reconciling user claims when `create_objects = False` is not final desired behavior.

ansible_base/authentication/utils/claims.py
```diff
@@ -268,8 +273,96 @@ def update_user_claims(user: Optional[AbstractUser], database_authenticator: Aut
     return user
 
 
+def process_organization_and_team_memberships(results):
+    # Extract organizations where the user is a member
+    org_list = [org_name for org_name, is_member in results['claims']['organization_membership'].items() if is_member]
+
+    # Build a mapping of teams to their respective organizations, filtering out non-members
+    team_map = {team_name: org_name for org_name, teams in results['claims']['team_membership'].items() for team_name, is_member in teams.items() if is_member}
```
cutwater: In my opinion this wide one-liner dict comprehension is hard to read. Partially this is due line length code style accepted on the project. I'd prefer a regular loop here instead of one-liner.

cutwater (changes requested):

ansible_base/authentication/utils/claims.py
```diff
@@ -268,8 +273,96 @@ def update_user_claims(user: Optional[AbstractUser], database_authenticator: Aut
     return user
 
 
+def process_organization_and_team_memberships(results):
+    # Extract organizations where the user is a member
+    org_list = [org_name for org_name, is_member in results['claims']['organization_membership'].items() if is_member]
```
cutwater: The only purpose of `org_list` from this code is to be converted to set on line 289. Why not create set from the beginning?

cutwater: Shouldn't we create organizations where user is not currently a member as well?

cutwater (changes requested):

ansible_base/authentication/utils/claims.py
john-westcott-iv: Correct, we should still do the user reconciliation even if create_objects is false.

ansible_base/authentication/utils/claims.py
jctanner: @john-westcott-iv is that what we want? and if so, presumably we want to make all teams too?

cutwater (approved):
I suggest we merge this PR, so that I can rebase on top of it and iterate further.

ansible_base/authentication/utils/claims.py
john-westcott-iv: Creating orgs/teams takes time during the login process and we have some customers who have hundreds or thousands of these. So if we limit this to just what the user is a member of it will minimize the impact on any one given user.

Ideally, we would deal with this in a better way. Like maybe there should be a button on an authenticator that says "create assets" so if you have a ton of them an administrator can "click the button" after they configure all of the authenticator maps. Maybe that can be a 2.5-next thing. Even if we did that I think this code will still remain in place because if a user attempts to login in between the admin completing the authenticator maps and clicking the button we would still need to create the org/team for the user.

ansible_base/authentication/utils/claims.py
```diff
@@ -268,8 +277,104 @@ def update_user_claims(user: Optional[AbstractUser], database_authenticator: Aut
...
+
+
+def load_existing_teams(team_names: Iterable[str]) -> set[str]:
+    existing_teams = set(Team.objects.filter(name__in=team_names).values_list('name', flat=True))
+    return existing_teams
+
+
+def create_missing_teams(team_names: Iterable[str], team_map: dict[str, str], existing_orgs: dict[str, int], existing_teams: set[str]) -> None:
+    for team_name in team_names:
+        if team_name not in existing_teams:
```
john-westcott-iv: Team names are not unique. What happens if I have a team named "My team" in "Org A" but the login wants a "My Team" in "Org B". This would detect "My Team" from "Org A" but not create the team in "Org B" or am I missing something?

jctanner: That is currently unaccounted for. I'm fixing it ..

jctanner: resolved with latest commit

PR #384, opened 2024-05-10
Move the rest_filters reserved names to a setting

AlanCoding:
My 2 cents here - this code was really just copied over from AWX. There is a process that code needs to go through in order to generalize it so that it's just abstract, for use via a library. This never happened for at least this part of rest_filters. Since it came up, I wanted to go ahead and get it done.

### DIFF
--- a/ansible_base/rest_filters/rest_framework/field_lookup_backend.py
+++ b/ansible_base/rest_filters/rest_framework/field_lookup_backend.py
@@ -22,8 +22,6 @@ class FieldLookupBackend(BaseFilterBackend):
     Filter using field lookups provided via query string parameters.
     """
 
-    RESERVED_NAMES = ('page', 'page_size', 'format', 'order', 'order_by', 'search', 'type', 'host_filter', 'count_disabled', 'no_truncate', 'limit', 'validate')
-
     SUPPORTED_LOOKUPS = (
         'exact',
         'iexact',
@@ -143,6 +141,17 @@ class FieldLookupBackend(BaseFilterBackend):
             value = self.value_to_python_for_field(field, value)
         return value, new_lookup, needs_distinct
 
+    def reserved_names(self, view):
+        """The names in query_params to ignore given the current settings and current view"""
+        from django.conf import settings
+
+        reserved_set = set(settings.ANSIBLE_BASE_REST_FILTERS_RESERVED_NAMES)
+
+        if hasattr(view, 'rest_filters_reserved_names'):
+            reserved_set |= set(view.rest_filters_reserved_names)
+
+        return reserved_set
+
     def filter_queryset(self, request, queryset, view):
         try:
             # Apply filters specified via query_params. Each entry in the lists
@@ -158,7 +167,7 @@ class FieldLookupBackend(BaseFilterBackend):
             # If 'OR' is used, an item just needs to satisfy one condition to appear in results.
             search_filter_relation = 'OR'
             for key, values in request.query_params.lists():
-                if key in self.RESERVED_NAMES:
+                if key in self.reserved_names(view):
                     continue
 
                 # HACK: make `created` available via API for the Django User ORM model
--- a/test_app/tests/rest_filters/rest_framework/test_field_lookup_backend.py
+++ b/test_app/tests/rest_filters/rest_framework/test_field_lookup_backend.py
@@ -1,7 +1,9 @@
 from unittest.mock import MagicMock, Mock
 
 import pytest
+from django.conf import settings
 from django.core.exceptions import FieldDoesNotExist, FieldError, ValidationError
+from django.test.utils import override_settings
 from django.urls import reverse
 from django.utils.http import urlencode
 from rest_framework.exceptions import ParseError, PermissionDenied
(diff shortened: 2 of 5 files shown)


relrod (approved):
I might suggest renaming the settings variable to indicate that those are always added even if a view overrides them, if that's purely additive.

But overall LGTM.

ansible_base/rest_filters/rest_framework/field_lookup_backend.py
```diff
@@ -143,6 +141,17 @@ def value_to_python(self, model, lookup, value):
...
+    def reserved_names(self, view):
+        """The names in query_params to ignore given the current settings and current view"""
+        from django.conf import settings
+
+        reserved_set = set(settings.ANSIBLE_BASE_REST_FILTERS_RESERVED_NAMES)
+
+        if hasattr(view, 'rest_filters_reserved_names'):
+            reserved_set |= set(view.rest_filters_reserved_names)
+
+        return reserved_set
```
john-westcott-iv, line 153: I always find the operators hard to read for some reason and prefer using the `.` functions on them i.e. 
```
settings_reserved_names = set(getattr(settings, 'ANSIBLE_BASE_REST_FILTERS_RESERVED_NAMES', []))
view_override_reserved_naames = set(getattr(view, 'rest_filters_reserved_names', []))
return settings_reserved_names.union(view_override_reserved_naames)
```

test_app/tests/rest_filters/rest_framework/test_field_lookup_backend.py
```diff
@@ -173,3 +175,28 @@ def test_filter_jsonfield_as_text(admin_api_client):
...
+    response = admin_api_client.get(url, data={'foofield': 'bar'})
+    assert response.status_code == 400, response.data
+
+
+def test_app_ignore_field(admin_api_client):
+    base_list = tuple(settings.ANSIBLE_BASE_REST_FILTERS_RESERVED_NAMES)
+    with override_settings(ANSIBLE_BASE_REST_FILTERS_RESERVED_NAMES=base_list + ('foofield',)):
+        url = reverse('organization-list')
+        response = admin_api_client.get(url, data={'foofield': 'bar'})
+        assert response.status_code == 200, response.data
```
john-westcott-iv, line 191: Do you want to parameterize this function to ensure that you can use `foofield`, `cud` and also something from `ANSIBLE_BASE_REST_FILTERS_RESERVED_NAMES`? To ensure that they are properly unioned?

test_app/tests/rest_filters/rest_framework/test_field_lookup_backend.py
```diff
@@ -173,3 +175,28 @@ def test_filter_jsonfield_as_text(admin_api_client):
     response = admin_api_client.get(url + '?' + urlencode(query_params))
     assert response.status_code == 200
     assert response.data['count'] == 1
+
+
+def test_filter_unexpected_field(admin_api_client):
+    url = reverse('organization-list')
+    response = admin_api_client.get(url, data={'foofield': 'bar'})
+    assert response.status_code == 400, response.data
```
john-westcott-iv, line 183: Isn't this technically covered by test_view_level_ignore_field?

ansible_base/rest_filters/rest_framework/field_lookup_backend.py
```diff
@@ -143,6 +141,17 @@ def value_to_python(self, model, lookup, value):
             value = self.value_to_python_for_field(field, value)
         return value, new_lookup, needs_distinct
 
+    def reserved_names(self, view):
+        """The names in query_params to ignore given the current settings and current view"""
+        from django.conf import settings
+
+        reserved_set = set(settings.ANSIBLE_BASE_REST_FILTERS_RESERVED_NAMES)
```
john-westcott-iv, line 148: Do you want to add any handeling for if someone completely borks `ANSIBLE_BASE_REST_FILTERS_RESERVED_NAMES`? i.e. makes it a dict or an int. Or are we just ok with the function blowing up/potentially doing something "weird"?

john-westcott-iv (approved):
Mostly minor comments. If it went in as built it wouldn't be the end of my world.

ansible_base/rest_filters/rest_framework/field_lookup_backend.py
AlanCoding: Personally I struggle with the _named methods_ because it is hard to know if they return the value or operate in-place.

https://docs.python.org/2/library/sets.html#set-objects

```python
>>> s = set(range(5))
>>> s.union(set(range(3, 7)))
{0, 1, 2, 3, 4, 5, 6}
>>> s.update(set(range(3, 7)))
```

The first prints and leaves `s` unchanged, but the second doesn't, it just updates `s` in memory.

It doesn't even feel like there's a formal "name" for the in-place group of methods vs the returning methods. But from the docs, they seem to pretty consistently end in `update`... provided it is a set-to-set operation. IMO that's a lot of qualifiers to expect someone to generally remember. But this is me griping about python in general.

Here, since `reserved_set` is artificially created, I think `reserved_set.update(view.rest_filters_reserved_names)` (works for iterables, as per docs) is would be the most consistent thing. Otherwise, if you want the returning methods, we should be using `frozenset`s for everything! To make these objects disposable.

ansible_base/rest_filters/rest_framework/field_lookup_backend.py
```diff
@@ -22,8 +22,6 @@ class FieldLookupBackend(BaseFilterBackend):
     Filter using field lookups provided via query string parameters.
     """
 
-    RESERVED_NAMES = ('page', 'page_size', 'format', 'order', 'order_by', 'search', 'type', 'host_filter', 'count_disabled', 'no_truncate', 'limit', 'validate')
```
mkanoor, line 25: @AlanCoding Isn't this list also preventing users from using these column names in their database model. If they have a "type" column in their DB then they cannot filter on it.

mkanoor: Should there be a warning if the column name also shows up in the reserved list.

AlanCoding: I don't think a warning with the request would be the right tool for this. The Django checks framework might be a better fit

https://docs.djangoproject.com/en/5.0/topics/checks/

ansible_base/rest_filters/rest_framework/field_lookup_backend.py
AlanCoding: I think this is pretty defensibly standard. Lots of settings basically work given the thing is an iterable, if not it'll error.

PR #386, opened 2024-05-13
[WIP] Fix missing check in sync triggers

cutwater:
Check if `pk_set` is a True-like (e.g. non-empty set) object.

The `pk_set` parameter may be `None` in cases when user teams list is cleared (e.g. calling `user.teams.clear()`).

This code will fail with the following exception:

```
TypeError: 'NoneType' object is not iterable
```

due to attempt to iterate over `None` value.

### DIFF
--- a/ansible_base/rbac/triggers.py
+++ b/ansible_base/rbac/triggers.py
@@ -350,14 +350,14 @@ class TrackedRelationship:
     def sync_team_to_role(self, instance: Model, action: str, model: type, pk_set: Optional[set[int]], reverse: bool, **kwargs):
         if not reverse:
             self._sync_actor_to_role(permission_registry.team_model, instance, action, pk_set)
-        else:
+        elif pk_set:
             for pk in pk_set:
                 self._sync_actor_to_role(permission_registry.team_model, model(pk=pk), action, {instance.pk})
 
     def sync_user_to_role(self, instance: Model, action: str, model: type, pk_set: Optional[set[int]], reverse: bool, **kwargs):
         if not reverse:
             self._sync_actor_to_role(permission_registry.user_model, instance, action, pk_set)
-        else:
+        elif pk_set:
             for pk in pk_set:
                 self._sync_actor_to_role(permission_registry.user_model, model(pk=pk), action, {instance.pk})
 
--- a/test_app/tests/rbac/conftest.py
+++ b/test_app/tests/rbac/conftest.py
@@ -7,6 +7,7 @@ from ansible_base.rbac.validators import combine_values, permissions_allowed_for
 from test_app.models import Inventory, Organization
 
 
+# REVIEW(cutwater): Fixture name is obscure and non-descriptive.
 @pytest.fixture
 def rando():
     return get_user_model().objects.create(username='rando')
--- a/test_app/tests/rbac/features/test_role_tracking.py
+++ b/test_app/tests/rbac/features/test_role_tracking.py
@@ -71,3 +71,13 @@ def test_add_organization_member_to_relationship(rando, organization, org_member
     object_role = org_member_rd.object_roles.first()
     assert rando in object_role.users.all()
     assert rando.has_obj_perm(organization, 'member')
+
+
+@pytest.mark.django_db
+def test_user_teams_clear(rando):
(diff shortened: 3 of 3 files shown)


AlanCoding (changes requested):
`user.teams.clear()` is expected to remove that user from the corresponding team's member role. This change would do nothing in that event, causing the relationship's membership to differ from the role's membership.

cutwater:
@AlanCoding I added tests that demonstrate this behavior.

Unless it's fixed the `user.teams.clear()` call will raise an exception:

```
rando = <User: rando>

    @pytest.mark.django_db
    def test_user_teams_clear(rando):
>       rando.teams.clear()

test_app/tests/rbac/features/test_role_tracking.py:78: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.tox/py311/lib/python3.11/site-packages/django/db/models/fields/related_descriptors.py:1176: in clear
    signals.m2m_changed.send(
.tox/py311/lib/python3.11/site-packages/django/dispatch/dispatcher.py:176: in send
    return [
.tox/py311/lib/python3.11/site-packages/django/dispatch/dispatcher.py:177: in <listcomp>
    (receiver, receiver(signal=self, sender=sender, **named))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <ansible_base.rbac.triggers.TrackedRelationship object at 0x7fc28b07a110>, instance = <User: rando>, action = 'pre_clear', model = <class 'test_app.models.Team'>, pk_set = None, reverse = True, kwargs = {'sender': <class 'test_app.models.Team_users'>, 'signal': <django.db.models.signals.ModelSignal object at 0x7fc28da59790>, 'using': 'default'}
memo = <typeguard.TypeCheckMemo object at 0x7fc28a46ddc0>

    def sync_user_to_role(self, instance: Model, action: str, model: type, pk_set: Optional[set[int]], reverse: bool, **kwargs):
        if not reverse:
            self._sync_actor_to_role(permission_registry.user_model, instance, action, pk_set)
        else:
>           for pk in pk_set:
E           TypeError: 'NoneType' object is not iterable
```

In addition to that the `pg_set` parameter in function is annotated as `Optional`, which means it's nullable. So iterating over it without additional checks is a mistake.

cutwater:
I analyzed the code a bit deeper and while my root cause analysis is accurate, it seems that the fix is not sufficient. The code doesn't handle `pre_clear` event correctly, which is what should be really fixed.

cutwater:
This PR is obsolete, since the TrackedRelationship feature is about to be deprecated and will be removed in future.

PR #388, opened 2024-05-14
fix podman-compose on selinux enforcing systems

BrennanPaciorek:
using podman-compose to run the test application fails on selinux enforcing systems since podman does not perform file relabeling by default, resulting in access denial when attempting to use files mounted from the host machine, so this change enables the option to relabel mounted files in docker_compose.yml

The change has been tested on a machine using docker-ce, and does not appear to affect setups using it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
     working_dir: /src
     command: 'test_app/scripts/container_startup.sh'
     volumes:
-      - '.:/src'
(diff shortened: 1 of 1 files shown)


BrennanPaciorek:
Note: there are still some issues, such as podman-compose not respecting the `depends_on` option (at least on my system), but this should clear up permissions issues when using podman-compose.

jctanner (approved):

PR #389, opened 2024-05-15
Replace X-TRUSTED-HEADER with pub/priv encrypted message

john-westcott-iv:
In simple testing this method appears to be fast but significantly slower than the shared secret method.

Fixes: [AAP-24036](https://issues.redhat.com/browse/AAP-24036)

### DIFF
--- a/ansible_base/jwt_consumer/common/util.py
+++ b/ansible_base/jwt_consumer/common/util.py
@@ -1,0 +1,52 @@
+import logging
+import time
+
+from cryptography.exceptions import InvalidSignature
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import padding
+
+from ansible_base.jwt_consumer.common.cert import JWTCert, JWTCertException
+
+logger = logging.getLogger('ansible_base.jwt_consumer.common.util')
+
+_SHARED_SECRET = 'trusted_proxy'
+
+
+def generate_x_trusted_proxy_header(key: str) -> str:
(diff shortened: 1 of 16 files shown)


BrennanPaciorek:
@john-westcott-iv Do i need to fast-forward this barnch to get the CI tests to run here?

BrennanPaciorek:
There is one remaining line to be covered, though it is covered, since return statements from both code branches are covered. Should we `# NOSONAR` the line?

ansible_base/jwt_consumer/common/util.py
```diff
@@ -0,0 +1,52 @@
...
+    try:
+        public_key.verify(
+            bytes.fromhex(signature),
+            bytes(f'{_SHARED_SECRET}-{timestamp}', 'utf-8'),
+            padding.PSS(mgf=padding.MGF1(hashes.SHA256()), salt_length=padding.PSS.MAX_LENGTH),
+            hashes.SHA256(),
+        )
+        return True
+    except InvalidSignature:
+        if not cert.cached:
```
BrennanPaciorek:
```suggestion
        if not cert.cached: # NOSONAR
```

BrennanPaciorek:
I'll add one more test for a bad cached jwt pubkey, but good not-cached key in `test_util.py`, since that code path is not directly tested. If that does not fix the last line of new code coverage, I will likely `# NOSONAR` the uncovered line.

djyasin (approved):
Pulled this down and confirmed it works as expected

PR #390, opened 2024-05-15
 Reconcile user teams and organizations 

cutwater:


ansible_base/authentication/utils/claims.py
```diff
@@ -263,21 +266,34 @@ def update_user_claims(user: Optional[AbstractUser], database_authenticator: Aut
...
+        module = importlib.import_module(module_path)
+    except ImportError:
+        logger.warning("Failed to load module '%s'.", module_path)
+        return None
+
+    try:
+        return getattr(module, 'ReconcileUser')
+    except AttributeError:
+        logger.warning("Failed to load ReconcileUser class in module '%s'.", module_path)
+        return None
```
john-westcott-iv, line 293: This might actually be nice to make generic in a "lib/utils" file at some point. I think there are other places where we try and load things dynamically just like this. Non-blocking on this PR.

ansible_base/authentication/utils/claims.py
```diff
@@ -330,39 +347,61 @@ def process_organization_and_team_memberships(results) -> None:
...
-        claims = getattr(user, 'claims', getattr(authenticator_user, 'claims'))
+
+        claims = getattr(user, 'claims', authenticator_user.claims)
+        org_member_rd = RoleDefinition.objects.get(name=constants.ORGANIZATION_MEMBER_ROLE_NAME)
+        team_member_rd = RoleDefinition.objects.get(name=constants.TEAM_MEMBER_ROLE_NAME)
+
+        remove_users = authenticator_user.provider.remove_users
+        if remove_users:
+            cls._remove_role_from_user(user, org_member_rd)
+            cls._remove_role_from_user(user, team_member_rd)
```
john-westcott-iv, line 377: @AlanCoding Is there a way to just drop all of a users RBAC in one liner?

ansible_base/authentication/constants.py
```diff
@@ -0,0 +1,4 @@
+TEAM_MEMBER_ROLE_NAME = "Team Member"
+TEAM_ADMIN_ROLE_NAME = "Team Admin"
+ORGANIZATION_MEMBER_ROLE_NAME = "Organization Member"
+ORGANIZATION_ADMIN_ROLE_NAME = "Organization Admin"
```
john-westcott-iv: We might want to make these properties in the `ReconcileUser` class so that if a specific service had a different role name for some reason they could extend their `ReconcileUser` class to change the property.

ansible_base/authentication/utils/claims.py
```diff
@@ -330,39 +347,61 @@ def process_organization_and_team_memberships(results) -> None:
...
+
+        org_names = claims['organization_membership'].keys() | claims['team_membership'].keys()
+        orgs_by_name = {org.name: org for org in Organization.objects.filter(name__in=org_names)}
 
         for org_name, is_member in claims['organization_membership'].items():
+            org = orgs_by_name.get(org_name)
+            if org is None:
+                logger.info("Skipping organization '%s'", org_name)
```
john-westcott-iv: I might change this to add "because the org does not exist" or something to indicate why we skipped.

ansible_base/authentication/utils/claims.py
```diff
@@ -330,39 +347,61 @@ def process_organization_and_team_memberships(results) -> None:
...
-                except Organization.DoesNotExist:
-                    continue
-                if not org.users.filter(pk=user.pk).exists():
-                    logger.info(f"Adding {user.username} to {org_name} organization")
-                    org.users.add(user)
-                else:
-                    logger.info(f"Skip adding {user.username} to {org_name} organization")
+                logger.info("Adding user '%s' to organization '%s'", user.username, org_name)
+                org_member_rd.give_permission(user, org)
+            else:
```
john-westcott-iv: We could make this an `elif not remove_users` right? Because the permission will already be gone if `remove_users` is True.

ansible_base/authentication/utils/claims.py
```diff
@@ -330,39 +347,61 @@ def process_organization_and_team_memberships(results) -> None:
...
+                logger.info("Removing '%s' from organization '%s'", user.username, org_name)
+                org_member_rd.remove_permission(user, org)
 
         for org_name, team_info in claims['team_membership'].items():
-            try:
-                org = Organization.objects.get(name=org_name)
-            except Organization.DoesNotExist:
+            org = orgs_by_name.get(org_name)
+            if org is None:
+                logger.info("Skipping organization '%s'", org_name)
```
john-westcott-iv: Make do `Skipping team addition because related organization is missing` with the team and org name showing up in there.

ansible_base/authentication/utils/claims.py
```diff
@@ -330,39 +347,61 @@ def process_organization_and_team_memberships(results) -> None:
...
                 continue
+
+            teams_by_name = {team.name: team for team in Team.objects.filter(organization=org, name__in=team_info.keys())}
```
john-westcott-iv, line 402: We probablly want to pre-load all of the team names for efficency. If we are going to do that in the next version of this PR where we cache the created that is fine with me.

ansible_base/authentication/utils/claims.py
```diff
@@ -330,39 +347,61 @@ def process_organization_and_team_memberships(results) -> None:
...
-                    except Team.DoesNotExist:
-                        continue
-                    if not team.users.filter(pk=user.pk).exists():
-                        logger.info(f"Adding {user.username} to {team_name} team")
-                        team.users.add(user)
-                    else:
-                        logger.info(f"Skip adding {user.username} to {team_name} team")
+                    logger.info("Adding user '%s' to team '%s'", user.username, team_name)
+                    team_member_rd.give_permission(user, team)
+                else:
```
john-westcott-iv: Sme thing on the else here w/ regards to being an elif.

ansible_base/authentication/utils/claims.py
cutwater: Yeah, makes sense.

ansible_base/authentication/utils/claims.py
john-westcott-iv: Non blocking, closing for now.

john-westcott-iv (approved):
Merge when tests are green.

PR #391, opened 2024-05-15
Redis client will now check read access to file settings

john-westcott-iv:


ansible_base/lib/redis/client.py
```diff
@@ -63,6 +64,12 @@ def connect(self, index: int = 0) -> Union[Redis, RedisCluster]:
         except (IndexError, ValueError):
             pass
 
+        for file_setting in ['ssl_certfile', 'ssl_keyfile', 'ssl_ca_certs']:
+            file = kwargs.get(file_setting, None)
+            if file:
+                if not os.access(file, os.R_OK):
```
cutwater: Minor nit: Can be combined in a single if expression:
```suggestion
            if file and not os.access(file, os.R_OK):
```

ansible_base/lib/redis/client.py
```diff
@@ -63,6 +64,12 @@ def connect(self, index: int = 0) -> Union[Redis, RedisCluster]:
         except (IndexError, ValueError):
             pass
 
+        for file_setting in ['ssl_certfile', 'ssl_keyfile', 'ssl_ca_certs']:
+            file = kwargs.get(file_setting, None)
```
cutwater, line 68: Minor nit: `dict.get` returns `None` by default, no need to explicitly set it.

john-westcott-iv: Good to know! I feel like one of the functions like this will raise an exception if the value is not in the collection so I always just explicitly returned None.

cutwater (approved):

PR #392, opened 2024-05-16
RBAC: add checks for when request is from Anonymous User

Dostonbek1:
This fixes opeanapi schema issues in EDA when the `/v1/docs/` endpoint is called from unauthenticated requests. With these changes in place the schema seems to get generated mostly correctly and we are down to the following warnings from DAB
```
eda-api-1                      | /app/venv/lib64/python3.11/site-packages/ansible_base/lib/routers/association_resource_router.py: Warning [RelatedRoleTeamAssignmentViewSet]: could not derive type of path parameter "team_assignments" because it is untyped and obtaining queryset from the viewset failed. Consider adding a type to the path (e.g. <int:team_assignments>) or annotating the parameter type with @extend_schema. Defaulting to "string".
eda-api-1                      | /app/venv/lib64/python3.11/site-packages/ansible_base/lib/routers/association_resource_router.py: Warning [RelatedRoleUserAssignmentViewSet]: could not derive type of path parameter "user_assignments" because it is untyped and obtaining queryset from the viewset failed. Consider adding a type to the path (e.g. <int:user_assignments>) or annotating the parameter type with @extend_schema. Defaulting to "string".
eda-api-1                      | /app/venv/lib64/python3.11/site-packages/ansible_base/rbac/api/views.py: Warning [RoleTeamAssignmentViewSet]: could not derive type of path parameter "id" because it is untyped and obtaining queryset from the viewset failed. Consider adding a type to the path (e.g. <int:id>) or annotating the parameter type with @extend_schema. Defaulting to "string".
eda-api-1                      | /app/venv/lib64/python3.11/site-packages/ansible_base/rbac/api/views.py: Warning [RoleUserAssignmentViewSet]: could not derive type of path parameter "id" because it is untyped and obtaining queryset from the viewset failed. Consider adding a type to the path (e.g. <int:id>) or annotating the parameter type with @extend_schema. Defaulting to "string".
```
There is still some schema data that doesn't follow the expected behavior which probably results from these warnings
```
        "/role_definitions/{id}/user_assignments/": {
            "get": {
                "operationId": "role_definitions_user_assignments_retrieve",

VS

        "/role_definitions/{id}/user_assignments/{user_assignments}/": {
            "get": {
                "operationId": "role_definitions_user_assignments_retrieve_2",
```

### DIFF
--- a/ansible_base/rbac/evaluations.py
+++ b/ansible_base/rbac/evaluations.py
@@ -1,6 +1,7 @@
 from typing import Optional
 
 from django.conf import settings
+from django.contrib.auth.models import AnonymousUser
 from django.db.models.functions import Cast
 from django.db.models.query import QuerySet
 
@@ -23,6 +24,9 @@ models in an app using these RBAC internals.
 
 def has_super_permission(user, full_codename=None) -> bool:
     "Analog to has_obj_perm but only evaluates to True if user has this permission system-wide"
+    if isinstance(user, AnonymousUser):
+        return False
+
     if user._meta.model_name == permission_registry.user_model._meta.model_name:
         # Super permission flags only exist for users, teams can use global roles
         for super_flag in settings.ANSIBLE_BASE_BYPASS_SUPERUSER_FLAGS:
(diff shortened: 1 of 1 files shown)


AlanCoding:
My only concern is that there is a known issue that a user passed in may be a `SimpleLazyObject` from Django... in which case `isinstance` checks won't work correctly. This is generally why I've favored the `_meta` attribute. But if we can't count on the meta... and we can't count on type checking... and we don't know what the user model will be. I just don't know what options they're giving us.

AlanCoding (approved):
Good fix, thanks.

It may hypothetically be possible to add tests for this, but it's not a huge deal.

PR #397, opened 2024-05-17
Use different method name from AWX for get_parent_object

AlanCoding:


AlanCoding:
I'm thinking there probably isn't any overlap here, because AWX doesn't use any of this in its sublist. Only DAB stuff does, and I think we can validly say they are analogs of each other that do the same thing.

PR #398, opened 2024-05-17
Move rest_filter default excludes to unconditional settings include

AlanCoding:


Dostonbek1 (approved):
Tested on eda-server, resolves the AttributeError we encountered.

PR #403, opened 2024-05-20
Only check redis file settings if tls is enabled

john-westcott-iv:
When we merged #391 we broke installations where `tls` was disabled but the related file settings for specifying cert/key/ca were still present.  These installations started failing with exceptions that the files were unreadable (which is technically accurate but not needed because tls isn't enabled).

### DIFF
--- a/docs/lib/redis.md
+++ b/docs/lib/redis.md
@@ -8,19 +8,19 @@ To leverage this client you can setup your Django cache like:
         "default": {
             "BACKEND": "django_redis.cache.RedisCache",
             # Note, the location is not really used but we parse it in the client to get settings like host/port/username/etc.
-            "LOCATION": f"redis://{os.environ.get('REDIS_USER')}:{os.environ.get('REDIS_PASS')}@{os.environ.get('REDIS_HOST')}:{os.environ.get('REDIS_PORT')}",
+            "LOCATION": f"<redis, rediss, file or unix schemed URL>",
             "KEY_PREFIX": '<service name>',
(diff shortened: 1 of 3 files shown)


rcarrillocruz (approved):

docs/lib/redis.md
```diff
@@ -8,19 +8,19 @@ To leverage this client you can setup your Django cache like:
         "default": {
             "BACKEND": "django_redis.cache.RedisCache",
             # Note, the location is not really used but we parse it in the client to get settings like host/port/username/etc.
-            "LOCATION": f"redis://{os.environ.get('REDIS_USER')}:{os.environ.get('REDIS_PASS')}@{os.environ.get('REDIS_HOST')}:{os.environ.get('REDIS_PORT')}",
+            "LOCATION": f"<redis, rediss, file or unix schemed URL>",
```
pavankrao, line 11: nit: typo `rediss`

john-westcott-iv: I know this looks weird but `rediss` is actually a legitimate scheme; Its like `http` vs `https` but it looks weird because redis ends with `s` already. So its just specify a tls enabled connection to redis.

pavankrao: oh! sounds good, sorry, just knew about this.

pavankrao (approved):

jamesmarshall24 (approved):

AlanCoding (approved):

rooftopcellist (approved):

PR #404, opened 2024-05-20
[oauth2] Prevent filtering on client_secret

relrod:

### DIFF
--- a/ansible_base/oauth2_provider/models/application.py
+++ b/ansible_base/oauth2_provider/models/application.py
@@ -3,8 +3,10 @@ from django.conf import settings
 from django.db import models
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
+from oauth2_provider.generators import generate_client_secret
 
 from ansible_base.lib.abstract_models.common import NamedCommonModel
+from ansible_base.lib.utils.models import prevent_search
 
 activitystream = object
 if 'ansible_base.activitystream' in settings.INSTALLED_APPS:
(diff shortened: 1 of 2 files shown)


ansible_base/oauth2_provider/models/application.py
```diff
@@ -54,14 +56,25 @@ class Meta(oauth2_models.AbstractAccessToken.Meta):
...
+
     # It would be nice to just use our usual encrypted_fields flow here
     # until DOT makes a release with https://github.com/jazzband/django-oauth-toolkit/pull/1311
     # there is no way to disable its expectation of using its own hashing
     # (which is Django's make_password/check_password).
     # So we use their field here.
     # Previous versions of DOT didn't hash the field at all and AWX pins
     # to <2.0.0 so AWX used the AWX encryption with no issue.
+    # We still override it here so that we can prevent_search() on it.
+    client_secret = prevent_search(
```
chrismeyersfsu, line 68: I'm surprised this doesn't require/complain about a migration.

From 0001 migration.
`('client_secret', models.CharField(blank=True, db_index=True, default=oauth2_provider.generators.generate_client_secret, help_text='Used for more stringent verification of access to an application when creating a token.', max_length=1024)),`

relrod: 0002 modifies it some and restores it to the upstream defaults. And this change keeps all those parameters the same as upstream, just wrapping it in `prevent_search()` which IIRC sets a field property but not one that Django knows about.

chrismeyersfsu (approved):

PR #407, opened 2024-05-21
AuthenticationMap role

slemrmartin:
This PR adds a `role` field to AuthenticationMap and `map_type == 'role'`

Role field is relation to RBAC's RoleDefinition through name (it's a CharField), so it's not a real foreign key

Conditions:

---

- If map_type is `organization`, `team`, or `role`
  - `role` field is mandatory

---

- if map_type is `organization`
  - role's content type must be 'organization'
- if map_type is `team`
  -  role's content type must be 'team'
- if map_type is `role`
  -  role's content type can be any or blank

---

- If role's content type is `organization`:
  - fields `organization` can't be blank
- If role's content type is `team`:
  - fields `team` and `organization` can't be blank


ansible_base/authentication/models/authenticator_map.py
```diff
@@ -35,38 +36,49 @@ class Meta:
         default=False,
         help_text=(_("If a user does not meet this rule should we revoke the permission")),
     )
+    map_type_choices = [
+        ('allow', 'allow'),
+        ('is_superuser', 'is_superuser'),
+        ('organization', 'organization'),
+        ('team', 'team'),
+    ]
+    if 'ansible_base.rbac' in settings.INSTALLED_APPS:
```
slemrmartin: TODO: If rbac dependency is not optional, let's remove the condition

ansible_base/authentication/models/authenticator_map.py
```diff
@@ -35,38 +36,49 @@ class Meta:
...
+        from ansible_base.rbac.models import RoleDefinition
+
+        role = models.ForeignKey(
+            RoleDefinition, null=True, on_delete=models.SET_NULL, related_name="authenticator_maps", help_text=(_("The role this mapping belongs to"))
```
john-westcott-iv: I think I might the help text to something like: `The role this map will grant the authenticating user to the targeted object`

ansible_base/authentication/models/authenticator_map.py
john-westcott-iv: I don't think our implementation of authentication_maps can exist without RBAC.

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -17,17 +18,42 @@ def validate(self, data) -> dict:
...
             errors["team"] = "You must specify a team with the selected map type"
-        if map_type == 'team' and (not org or org == ''):
-            errors["organization"] = "You must specify an organization with the selected map type"
-        if map_type == 'organization' and (not org or org == ''):
+        if map_type in ['team', 'organization'] and (not org or org == ''):
             errors["organization"] = "You must specify an organization with the selected map type"
+        if map_type in ['team', 'organization', 'role'] and role is None:
+            errors["role"] = "You must specify a role with the selected map type"
+        if map_type in ['allow', 'is_superuser'] and role is not None:
+            errors["role"] = "You cannot specify role with the selected map type"
```
john-westcott-iv: These string should probablly be translated as well.

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -17,17 +18,42 @@ def validate(self, data) -> dict:
...
 
+    def validate_role_data(self, role, org, team):
+        errors = {}
+
+        if not role.content_type:
+            return errors
+
+        model_class = role.content_type.model_class()
+        is_org_role = model_class == get_organization_model()
+        is_team_role = model_class == get_team_model()
```
john-westcott-iv: How does `==` compare to `isinstance`?

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -17,17 +18,42 @@ def validate(self, data) -> dict:
...
+
+        if (is_org_role or is_team_role) and (not org or org == ''):
+            errors["organization"] = "You must specify an organization with the selected role"
+
+        if is_team_role and (not team or team == ''):
+            errors["team"] = "You must specify a team with the selected role"
```
john-westcott-iv: These strings should probably be translated as well.

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -17,17 +18,42 @@ def validate(self, data) -> dict:
...
-        if map_type == 'organization' and (not org or org == ''):
+        if map_type in ['team', 'organization'] and (not org or org == ''):
             errors["organization"] = "You must specify an organization with the selected map type"
+        if map_type in ['team', 'organization', 'role'] and role is None:
+            errors["role"] = "You must specify a role with the selected map type"
+        if map_type in ['allow', 'is_superuser'] and role is not None:
+            errors["role"] = "You cannot specify role with the selected map type"
+
+        if role is not None:
+            errors.update(self.validate_role_data(role, org, team))
```
john-westcott-iv: This is an interesting problem because above we are validating the map types and here we are validating the roles. This function is role centric, looking at the role and then validating the inputs.
But I think we are missing checks for "does this role make sense for the map type". If I am reading things right and I selected an organization map type but passed in a Team and a Team type role I think I would get through this validation logic.

So we likely need an additional check somewhere to say "if you selected org map type and your role is not an org map type raise 'For an organization map type you must select an organization based role'".

ansible_base/authentication/models/authenticator_map.py
AlanCoding: This condition would conflict with the migration. Allowing using with/without the other app would be inflexible, at best.

AlanCoding:
My very-fast gut reaction to this is:

```python
    organization = models.CharField(
        max_length=512,
        null=True,
        default=None,
        blank=True,
```

A related `RoleDefinition` is similar in nature to a related `Organization`. This model doesn't have a ForeignKey link to the organization model due to the massive complexities that would introduce. It just keeps the name of it. I would favor a solution like that for roles as well.

However, we need to talk more about the proposal for a secondary identifier for role definitions, before we commit to using names. Upstream discussion I started is in https://github.com/ansible/django-ansible-base/issues/408, which doesn't get into custom roles, but that needs to be a part of this too.

ansible_base/authentication/serializers/authenticator_map.py
slemrmartin: this works well, because the class and `get_*` are derived from settings. 
moreover they are classes, not instances, so `isinstance` won't work. But I can use `issubclass` for more clean solution

slemrmartin:
@AlanCoding if I'm correct, we need to make this compatible with hub, which uses pulp rbac atm. So both RoleDefinition and ManagedRoleDefinition introduces the same issue, which is RBAC dependency. Thus using the role as a name is not a question for this version.

Anyway I need to clarify the 2nd paragraph of your previous comment, what does it mean

ansible_base/authentication/models/authenticator_map.py
```diff
@@ -35,38 +35,48 @@ class Meta:
...
-            ('is_superuser', 'is_superuser'),
-            ('is_system_auditor', 'is_system_auditor'),
             ('allow', 'allow'),
+            ('is_superuser', 'is_superuser'),
+            ('role', 'role'),
             ('organization', 'organization'),
+            ('team', 'team'),
         ],
-        help_text=(_('What does the map work on, a team, a user flag or is this an allow rule')),
+        help_text=(_('What does the map work on, a team, organization, a user flag or is this an allow rule')),
```
AlanCoding: The extra `()` looks unnecessary.

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -17,17 +25,69 @@ def validate(self, data) -> dict:
...
+
+        from ansible_base.rbac.models import RoleDefinition
+
+        try:
+            rbac_role = RoleDefinition.objects.get(name=role)
+            is_system_role = rbac_role.content_type is None
+
+            # system role is allowed for map type == role without further conditions
+            if is_system_role and map_type == 'role':
+                return errors
```
AlanCoding, line 67: Don't you have to add something to `errors` before returning it? It's `{}` right here, as far as I can tell.

AlanCoding:
Left some minor comments, but :+1: to the overall thing being done here.

ansible_base/authentication/models/authenticator_map.py
john-westcott-iv:
```suggestion
        help_text=(_('What will the map grant the user? System access (allow) a team or organization membership, the superuser flag or a role in the system')),
```

ansible_base/authentication/models/authenticator_map.py
```diff
@@ -35,38 +35,48 @@ class Meta:
...
+    )
+
+    role = models.CharField(
```
john-westcott-iv, line 53: Was it not possible to do a ForeignKey here? We made the others char fields because we want the system to be able to create teams/organizations on the fly but we won't be able to build roles on the fly.

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -17,17 +25,69 @@ def validate(self, data) -> dict:
...
         org = data.get('organization', None)
-        if map_type == 'team' and (not team or team == ''):
-            errors["team"] = "You must specify a team with the selected map type"
-        if map_type == 'team' and (not org or org == ''):
-            errors["organization"] = "You must specify an organization with the selected map type"
-        if map_type == 'organization' and (not org or org == ''):
-            errors["organization"] = "You must specify an organization with the selected map type"
+        role = data.get('role', None)
+
+        if map_type == 'team' and not team:
```
john-westcott-iv: Should we change these to a re to require a character that is not a white space? i.e. I probably shouldn't be able to create a team whose name is just a tab character. Maybe this is a constraint we actually want to put on NamedCommonModel?

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -17,17 +25,69 @@ def validate(self, data) -> dict:
...
             raise ValidationError(errors)
         return data
 
+    def validate_role_data(self, map_type, role, org, team):
+        errors = {}
+
+        # Validation is possible only if RBAC is installed
+        if not apps.is_installed('ansible_base.rbac'):
+            logger.warning(_("You specified a role without RBAC installed "))
+            return errors
```
john-westcott-iv: I kind of like what you are doing here but there are likely going to be a lot of cases where one feature in DAB requires another. For the jwt_consumer I was going to do something like this in `lib/dynamic_config/dynamic_settings.py`:

```
if 'ansible_base.jwt_consumer' in INSTALLED_APPS:
    if 'ansible_base.rbac' not in INSTALLED_APPS:
        INSTALLED_APPS.append('ansible_base.rbac')
```

I'm open to either way of doing this but we should likely pick something consistant. @AlanCoding @relrod any thoughts on which approach we take?

ansible_base/authentication/serializers/authenticator_map.py
john-westcott-iv: This is ok. On line 40 we are taking the return from this function and updating the parents errors condition which would then get displayed (if there are any).

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -17,17 +25,69 @@ def validate(self, data) -> dict:
...
+            # system role is allowed for map type == role without further conditions
+            if is_system_role and map_type == 'role':
+                return errors
+
+            if is_system_role:
+                is_org_role, is_team_role = False, False
+            else:
+                model_class = rbac_role.content_type.model_class()
+                is_org_role = issubclass(model_class, get_organization_model())
+                is_team_role = issubclass(model_class, get_team_model())
```
john-westcott-iv, line 74: We should be able to get rid of this if/else?

```suggestion
            is_org_role, is_team_role = False, False
            model_class = rbac_role.content_type.model_class()
            is_org_role = issubclass(model_class, get_organization_model())
            is_team_role = issubclass(model_class, get_team_model())
```

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -17,17 +25,69 @@ def validate(self, data) -> dict:
...
+
+            # role type and map type must correspond
+            if map_type == 'organization' and (is_system_role or is_team_role):
```
john-westcott-iv: Can we simplify this if?

```suggestion
            if map_type == 'organization' and not is_org_role:
```

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -17,17 +25,69 @@ def validate(self, data) -> dict:
...
+                errors['role'] = _("For an organization map type you must specify an organization based role")
+
+            if map_type == 'team' and (is_system_role or is_org_role):
```
john-westcott-iv:
```suggestion
            if map_type == 'team' and not is_team_role:
```

ansible_base/authentication/serializers/authenticator_map.py
```diff
@@ -17,17 +25,69 @@ def validate(self, data) -> dict:
...
+            # org/team role needs organization field
+            if (is_org_role or is_team_role) and (not org or org == ''):
+                errors["organization"] = _("You must specify an organization with the selected role")
+
+            # team role needs team field
+            if is_team_role and (not team or team == ''):
+                errors["team"] = _("You must specify a team with the selected role")
+
+        except ObjectDoesNotExist:
+            errors['role'] = _("RoleDefinition {role} doesn't exist").format(role=role)
```
john-westcott-iv, line 92: If we move to a ForeignKey we don't need this catch.

test_app/tests/authentication/serializers/test_authententicator_map.py
```diff
@@ -0,0 +1,153 @@
...
+    )
+
+
+@pytest.fixture
+def organization_role():
+    return RoleDefinition.objects.create(
+        name=ORG_ROLE_NAME,
+        content_type=permission_registry.content_type_model.objects.get_for_model(get_organization_model()),
+        managed=True,
+    )
```
john-westcott-iv: These seem like they might be useful in a larger context should we move them either into `conftest` or into `lib/testing/fixtures.pm`?

docs/apps/authentication/authentication.md
```diff
@@ -11,6 +11,7 @@ Add `ansible_base.authentication` to your installed apps:
 INSTALLED_APPS = [
     ...
     'ansible_base.authentication',
+    'ansible_base.rbac'  # dependency
```
john-westcott-iv: I think we should either make this a hard dependency in dynamic_settings or specify here that its an optional dependency. If we are going to make it optional we should also add to this doc what it would mean if the dependency is either there or not.

ansible_base/authentication/serializers/authenticator_map.py
slemrmartin: There wasn't a white space check, just an empty string, which is covered by this. 
If we want to be more precise, we can check against team.strip()

ansible_base/authentication/serializers/authenticator_map.py
slemrmartin: Somewhere in the Django documentation I found a note that INSTALLED_APPS shouldn't be used directly (but not mentioned why)

ansible_base/authentication/models/authenticator_map.py
slemrmartin: ForeignKey means dependency on RBAC. If I'm not wrong, Hub is going to use this code, but isn't using DAB RBAC now, so it's not possible. Is it correct?

ansible_base/authentication/serializers/authenticator_map.py
slemrmartin: we can't, because in case of system_role the content_type is None, which means `.model_class` will be called on None

ansible_base/authentication/serializers/authenticator_map.py
slemrmartin: I believe my condition is better understandable - it's saying what we are testing, instead of what we are _not_ testing

test_app/tests/authentication/serializers/test_authententicator_map.py
slemrmartin: in case of larger context, fixtures and constants must be renamed, because they're creating member roles. 

I can move it to conftest, but rather not to lib's fixtures - rbac is creating it with permissions etc., which is not needed for these tests, but can be required somewhere else

ansible_base/authentication/serializers/authenticator_map.py
john-westcott-iv: Ah, good call.

ansible_base/authentication/serializers/authenticator_map.py
slemrmartin: so, I guess you were not talking about the format of `if` condition, right? 

To the meaning, we agreed that authentication app doesn't need RBAC to be installed (at least now). 
So the condition in `dynamic_settings.py` is not applicable, because after this condition it's using the RBAC models, which are optional

docs/apps/authentication/authentication.md
slemrmartin: Instead of this line I made a chapter describing the optional dependency

ansible_base/authentication/models/authenticator_map.py
```diff
@@ -27,46 +27,56 @@ class Meta:
...
             ('organization', 'organization'),
+            ('team', 'team'),
         ],
-        help_text=(_('What does the map work on, a team, a user flag or is this an allow rule')),
+        help_text=_('What will the map grant the user? System access (allow) a team or organization membership, the superuser flag or a role in the system'),
+    )
+
+    role = models.CharField(
+        max_length=512,
+        null=True,
```
AlanCoding, line 55: SonarCloud doesn't like this `null=True` flag because it doesn't make sense for a `CharField`. I've seen this before and I think it's very valid, although I accept that other existing fields may do the same thing.

Trying to remember the academic issue: Null isn't actually a value in the database, which is what this is about. If you want serializers to accept null or blank values, that's a separate topic, and not what _this_ means.

slemrmartin: I want to be compatible with the other char fields representing "foreign keys by name". So it makes sense to either have `null=True` for all or none of them

ansible_base/authentication/serializers/authenticator_map.py
AlanCoding: > Somewhere in the Django documentation I found a note that INSTALLED_APPS shouldn't be used directly (but not mentioned why)

Looks like https://stackoverflow.com/a/43189917. So sure, runtime checking should maybe be `apps.is_installed('ansible_base.rbac')`, but that should be rolled out generally. Also, I'm skeptical that this would work on import, making the advice potentially risky.

I like the idea that the authenticator app can be used without RBAC. I would still like RBAC to be usable without all the other apps as well. I think it's fine to have fields in some models that are _unused_ unless a related app is enabled.

I don't like automatically adding apps to INSTALLED_APPS. There's a lot of configuration for RBAC, and just enabling it is a bad idea. It would just delay the error which is the opposite of what we want here. If you're not using roles, then the claims should list roles.

ansible_base/lib/utils/string.py
```diff
@@ -6,3 +6,10 @@ def make_json_safe(value):
         return value
 
     return smart_str(value)
+
+
+def is_empty(value, check_stripped=True):
```
AlanCoding: I'd really rather remove `check_stripped=True` from this since it's never not used.

And I'd like for pure-python utility methods like this to have some type hints and doc strings.

test_app/tests/authentication/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+
+
+@pytest.fixture
+def organization_member_role():
+    return RoleDefinition.objects.create(
+        name=ORG_MEMBER_ROLE_NAME,
+        content_type=permission_registry.content_type_model.objects.get_for_model(get_organization_model()),
+        managed=True,
+    )
```
AlanCoding, line 34: Both of these fixtures already exist in the path where fixtures are available as `member_rd` and `org_member_rd`. I'd prefer to not introduce new fixtures for this purpose. But this is can be touched up later, since it's just a test change.

test_app/tests/authentication/serializers/test_authententicator_map.py
```diff
@@ -0,0 +1,126 @@
...
+        """map_type='organization' requires fields organization and role"""
+        with pytest.raises(ValidationError) as e:
+            serializer.validate(dict(name="authentication_map_1", map_type="organization"))
+        assert set(e.value.detail.keys()) == {'organization', 'role'}
+
+        with pytest.raises(ValidationError) as e:
+            serializer.validate(dict(name="authentication_map_1", map_type="organization", organization='test_org'))
+        assert list(e.value.detail.keys()) == ['role']
+
+        serializer.validate(dict(name="authentication_map_1", map_type="organization", organization='test_org', role=SYSTEM_ROLE_NAME))
```
AlanCoding: You didn't include the `system_role` fixture, so using `role=SYSTEM_ROLE_NAME` should throw a `ValidationError` because the role doesn't exist.

Am I missing something?

test_app/tests/authentication/serializers/test_authententicator_map.py
```diff
@@ -0,0 +1,126 @@
+from unittest.mock import MagicMock
```
AlanCoding: It looks like you misspelled the name of this file.

test_app/tests/authentication/serializers/test_authententicator_map.py
AlanCoding: oooooh, it mocked that.

AlanCoding (approved):
I see the final bit for functionality is planned after this to fully apply the permissions. So I'm on board with the plan finally.

I left some minor comments unresolved that are minor organizational or aesthetic suggestions.

test_app/tests/authentication/conftest.py
slemrmartin: they are in rbac's conftest, and they're expecting roles exists. So I would rather spend time on it later

AlanCoding: > they are in rbac's conftest

That is not the case https://github.com/ansible/django-ansible-base/blob/devel/test_app/tests/conftest.py#L630

PR #411, opened 2024-05-21
disable POST in options for role definitions if custom roles disabled

AlanCoding:


slemrmartin (commented):
it looks good, I'll approve when the conflict is solved

john-westcott-iv (approved):
Per Martins comment, conflict is resolved so approving.
Also one of the CI job got skipped so I kicked the jobs.

PR #413, opened 2024-05-22
RBAC: User can't delete self

slemrmartin:
User shouldn't have an ability to delete self.

### DIFF
--- a/ansible_base/rbac/api/permissions.py
+++ b/ansible_base/rbac/api/permissions.py
@@ -10,7 +10,7 @@ from ansible_base.lib.utils.models import is_add_perm
 from ansible_base.lib.utils.settings import get_setting
 from ansible_base.rbac import permission_registry
 from ansible_base.rbac.evaluations import has_super_permission
-from ansible_base.rbac.policies import can_change_user
+from ansible_base.rbac.policies import can_change_user, can_delete_user
 
 logger = logging.getLogger('ansible_base.rbac.api.permissions')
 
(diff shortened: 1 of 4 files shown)


AlanCoding:
Is this really a _permission_ check though? Doesn't this apply to superusers? Should a superuser get a _403_ when trying to delete themselves? Should they get a message that reads "You don't have permission to perform this action" in the response?

A shared `CommonUserSerializer` was just introduced. And I know, I know, you can't enforce deletion rules using the serializer. That's not going to help you, but it seems like a better place.

This is just too low in the stack for me to put this check. If you must raise a `PermissionDenied` (as opposed to another exception), then put a custom message with it.

slemrmartin:
In my opinion this is a permission check, it's a Delete permission. 
Serializer validates changes in fields, not this kind of "change".
It should be valid also for superusers as we discussed it in the RBAC doc. 

What could be the alternative (but I don't think it's worth the effort)  - that superuser can delete self, except if it's the last superuser in the system - that check should be part of serializer's validation.
But this simple check covers that condition too with much less and simpler code

AlanCoding:
Then could you raise a PermissionDenied exception with a message explaining "you can't delete yourself" (would have to move to the permission class)? This is going down the path of non-trivial undocumented policies rules, which can be extremely bad for the user experience.

slemrmartin:
changed as suggested

AlanCoding (approved):
thanks!

PR #414, opened 2024-05-22
disallow filtering by encrypted_fields

BrennanPaciorek:
Raise a PermissionDenied exception when attempting to get encrypted fields via ansible_base.rest_filters.utils::get_field_from_path, so such requests get handled in the same way that attempting to filter by a password field does.

No-Issue

### DIFF
--- a/ansible_base/rest_filters/utils.py
+++ b/ansible_base/rest_filters/utils.py
@@ -42,8 +42,8 @@ def get_fields_from_path(model, path, treat_jsonfield_as_text=True):
         else:
             new_parts.append(name)
 
-        if name in getattr(model, 'PASSWORD_FIELDS', ()):
-            raise PermissionDenied(_('Filtering on password fields is not allowed.'))
+        if name in getattr(model, 'PASSWORD_FIELDS', ()) or name in getattr(model, 'encrypted_fields', ()):
+            raise PermissionDenied(_('Filtering on field %(name)s is not allowed.') % {'name': name})
         elif name == 'pk':
             field = model._meta.pk
         else:
--- a/test_app/tests/rest_filters/test_utils.py
+++ b/test_app/tests/rest_filters/test_utils.py
@@ -1,11 +1,20 @@
 import pytest
-from rest_framework.exceptions import ParseError
+from rest_framework.exceptions import ParseError, PermissionDenied
 
 from ansible_base.authentication.models import Authenticator
 from ansible_base.rest_filters.utils import get_field_from_path
+from test_app.models import EncryptionModel
 
 
 def test_invalid_field_hop():
     with pytest.raises(ParseError) as excinfo:
         get_field_from_path(Authenticator, 'created_by__last_name__user')
     assert 'No related model for' in str(excinfo)
+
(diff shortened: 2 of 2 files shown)


BrennanPaciorek:
Personally, I think having ansible_base.rest_filter.utils accept an allow-list of fields (rather than blocking attempts to filter over marked sensitive data) would be a better long term solution, but this should address the issue with our rest filters leaking sensitive data for now, assuming everyone marks their encrypted fields properly.

ansible_base/rest_filters/utils.py
```diff
@@ -44,6 +44,8 @@ def get_fields_from_path(model, path, treat_jsonfield_as_text=True):
 
         if name in getattr(model, 'PASSWORD_FIELDS', ()):
             raise PermissionDenied(_('Filtering on password fields is not allowed.'))
+        elif name in getattr(model, 'encrypted_fields', ()):
+            raise PermissionDenied(_('Filtering on encrypted fields is not allowed.'))
```
john-westcott-iv: In thinking about this I don't like the origional message we had in here about password fields and I'm thinking we won't want to add multiple message types if we wanted to prevent filtering on some other kind of field. What if we return a generic message like this:

```suggestion
        if name in getattr(model, 'PASSWORD_FIELDS', ()) or name in getattr(model, 'encrypted_fields', ()):
            raise PermissionDenied(_('Filtering on field %(name)s is not allowed.') % {'name': name})
```

If we do this, we will need to change the test as well.

test_app/tests/rest_filters/test_utils.py
```diff
@@ -1,11 +1,25 @@
...
         get_field_from_path(Authenticator, 'created_by__last_name__user')
     assert 'No related model for' in str(excinfo)
+
+
+def test_invalid_field_filter():
+    class CommonModelTest(AbstractCommonModel):
+        encrypted_fields = ["secret"]
+
+        def __init__(self):
+            self.secret = "$encrypted$"
```
john-westcott-iv: These tests run in the context of test_app which has models which likely already have encrypted_fields so we can save these lines of code and just reuse one of those models; like `test_app.models.EncryptionModel`.

ansible_base/rest_filters/utils.py
BrennanPaciorek: Do you think we have to worry about this change affecting Ansible Connector's test results? Does not seem like it will, but I figured I would check with you as well.

test_app/tests/rest_filters/test_utils.py
BrennanPaciorek: This change sounds good to me. I will make it when I am back at my workstation.

ansible_base/rest_filters/utils.py
BrennanPaciorek: Applied the requested changes, and modified test case to match.

john-westcott-iv (approved):

BrennanPaciorek:
Rebased on devel.

AlanCoding (approved):

PR #415, opened 2024-05-23
Tidying up system user and other functions

john-westcott-iv:
Modify create_dab_permissions to return a boolean
Create a _WARNED_ABOUT_SYSTEM_USER and only show the SYSTEM_USERNAME is defined but no user matches once per thread based on the variable.
Alter `get_system_user` to create the `system_user` if its missing.
Refactor the system user functions.
Make sure the system_user functions are used globally instead of people referencing the SYSTEM_USER setting or getting the system user directly.

### DIFF
--- a/ansible_base/lib/utils/models.py
+++ b/ansible_base/lib/utils/models.py
@@ -13,6 +13,7 @@ from ansible_base.lib.utils.settings import get_setting
 from ansible_base.lib.utils.string import make_json_safe
 
 logger = logging.getLogger('ansible_base.lib.utils.models')
+_WARNED_ABOUT_SYSTEM_USER = False
 
 
 def get_all_field_names(model, concrete_only=False, include_attnames=True):
@@ -88,11 +89,13 @@ def is_system_user(user):
 
 
 def get_system_user():
+    global _WARNED_ABOUT_SYSTEM_USER
     system_user = None
     setting_name = 'SYSTEM_USERNAME'
     system_username = get_setting(setting_name)
     system_user = get_user_model().objects.filter(username=system_username).first()
-    if system_username is not None and system_user is None:
+    # We are using a global variable to try and track if this thread has already spit out the message, if so ignore
+    if system_username is not None and system_user is None and not _WARNED_ABOUT_SYSTEM_USER:
         logger.error(
             _(
                 "{setting_name} is set to {system_username} but no user with that username exists.".format(
@@ -100,6 +103,7 @@ def get_system_user():
                 )
             )
         )
+        _WARNED_ABOUT_SYSTEM_USER = True
     return system_user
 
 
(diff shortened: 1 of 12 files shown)


ansible_base/rbac/management/__init__.py
```diff
@@ -9,19 +9,19 @@
 logger = logging.getLogger(__name__)
 
 
-def create_dab_permissions(app_config, verbosity=2, interactive=True, using=DEFAULT_DB_ALIAS, apps=global_apps, **kwargs):
+def create_dab_permissions(app_config, verbosity=2, interactive=True, using=DEFAULT_DB_ALIAS, apps=global_apps, **kwargs) -> bool:
```
AlanCoding, line 12: For background, this is modeled after

https://github.com/django/django/blob/main/django/contrib/auth/management/__init__.py

To a fault, it was trying to not change any signatures. But in the weeks and months after merging this, it became pretty apparent that we won't be able to keep its scope so limited in the long run. It's still dangerous code to mess with since it'll run with any given model states and has (or will have) dependencies on post-migrate code from multiple other apps (content types, resource registry).

ansible_base/rbac/management/__init__.py
```diff
@@ -86,6 +86,11 @@ def create_dab_permissions(app_config, verbosity=2, interactive=True, using=DEFA
             permission.content_type = ct
             perms.append(permission)
 
+    if not perms:
+        return False
+
     Permission.objects.using(using).bulk_create(perms)
     for perm in perms:
         logger.debug("Adding permission '%s'" % perm)
```
AlanCoding, line 94: instead of adding `if not perms:` as a separate block, it might be cleaner to add an `else:` and `return False` there. In terms of code flow...

ansible_base/lib/utils/models.py
```diff
@@ -76,34 +78,111 @@ def user_summary_fields(user):
...
+    return system_user
+
+
+def get_system_username() -> Tuple[Optional[str], str]:
+    # Returns (system_username, setting_name)
+    setting_name = 'SYSTEM_USERNAME'
+    return get_setting(setting_name), setting_name
+
+
+def create_system_user() -> AbstractUser:
```
AlanCoding, line 118: I think it is a very bad idea to not take `apps` as an input argument here. We need to be much better disciplined about this since we're building a library with DAB and we don't want it to randomly blow up things deep in the migrations spaghetti of some project. Look at the source code for Django `get_user_model`. It is very explicitly clear that it is created with _current_ apps. That means, never call that method in migrations. I think you need to use this in migrations, no?

john-westcott-iv: Yea, I think I need to rewrite this function to make it callable from migrations.

PR #416, opened 2024-05-28
Make create_dab_permissions return a boolean

john-westcott-iv:
Make create_dab_permissions return a boolean.

### DIFF
--- a/ansible_base/rbac/management/__init__.py
+++ b/ansible_base/rbac/management/__init__.py
@@ -9,19 +9,19 @@ from ansible_base.rbac import permission_registry
 logger = logging.getLogger(__name__)
 
 
-def create_dab_permissions(app_config, verbosity=2, interactive=True, using=DEFAULT_DB_ALIAS, apps=global_apps, **kwargs):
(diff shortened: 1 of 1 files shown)


AlanCoding (approved):
I'm okay with having this return a bool. I'm iffy about how that would be used, but this is pretty targeted if considered in isolation.

AlanCoding:
To have it pasted here, this was modeled after:

https://github.com/django/django/blob/main/django/contrib/auth/management/__init__.py

Generally I wanted the method signatures and everything to remain the same as that from `auth.Permission`, but adding a return value isn't conflicting.

Django signals does have documentation on return values to some minor extent

https://docs.djangoproject.com/en/5.0/topics/signals/

> All four methods return a list of tuple pairs [(receiver, response), ...], representing the list of called receiver functions and their response values.

But that makes it dependent on the sender.

https://github.com/django/django/blob/99f23eaabd8da653f046dc1d19f5008c030a4f79/django/core/management/sql.py#L52-L59

and the sender of the post_migrate signal doesn't seem to care.

john-westcott-iv:
We no longer need this because of a different direction.

PR #418, opened 2024-05-28
Create system user as needed

### LINKED ISSUE #428: CI test "test_radius.py::test_authenticator_plugin" fails randomly
### Bug Summary

Across multiple commits and PRs, `test_radius.py::test_authenticator_plugin` appears to fail randomly in the same way:
```python
backend_cls = <MagicMock name='RADIUSBackend' id='140395987507952'>
unauthenticated_api_client = <rest_framework.test.APIClient object at 0x7fb07cf5a8e0>
radius_configuration = {'PORT': 1812, 'SECRET': 'testingsecret', 'SERVER': 'localhost'}
radius_authenticator = <Authenticator: Test RADIUS Authenticator>
user = <User: user>

    @mock.patch("rest_framework.views.APIView.authentication_classes", [SessionAuthentication])
    @mock.patch("ansible_base.authentication.authenticator_plugins.radius.RADIUSBackend")
    def test_authenticator_plugin(backend_cls, unauthenticated_api_client, radius_configuration, radius_authenticator, user):
        """
        Test RADIUS authenticator logic.
    
        Test that the RADIUS backend is created with the correct settings.
        Test that login credentials are passed to the backend's authenticate() method call.
        Test that associated AuthenticatorUser record is created.
        """
        backend = backend_cls.return_value
        backend.authenticate.return_value = RADIUSUser(
            username="user",
            groups=["users", "managers"],
            is_staff=False,
            is_superuser=False,
        )
    
        client = unauthenticated_api_client
        client.login(username="user", ***)
    
        url = reverse(authenticated_test_page)
        response = client.get(url)
    
>       backend_settings = backend_cls.call_args[0][0]
E       TypeError: 'NoneType' object is not subscriptable

test_app/tests/authentication/authenticator_plugins/test_radius.py:77: TypeError
```

This case fails in devel commit https://github.com/ansible/django-ansible-base/commit/0b4d0138f3572829ba6e21efbf16758dadba26fb and the PR to revert this commit: #426
Since applying and undoing this patch both produce this error, I do not think the commit is the root cause.

Bug Reproduction: I cannot reproduce this locally. I have attempted restricting local pytest worker counts to 4 (the amount used by GitHub actions job)
I have not tried testing running pytest in a virtualized Ubuntu instance.

Expected Resolution: One of the below
- Figure out the trigger for this failure, and fix the root cause (ideal)
- Refactor the test case `test_radius.py::test_authenticator_plugin` to be more resilient to the random failure (not as good)
-  xfail `test_radius.py::test_authenticator_plugin` (not good solution)

john-westcott-iv:
Superceeds #415 

This PR restructures the system user functions in`lib/utils/models.py`. Specifically, it also creates a create_system_user function into `lib/utils/create_system_user.py` This file is split out because we want to be able to call create_system_user from either a migration (where models are not loaded) or at runtime. Loading this function in the larger `models.py` was problematic because of the model loads. 

Additionally, I went through the code in DAB and made sure references to SYSTEM_USERNAME or the manual getting of that user was done only through the functions provided in models.py. 

We had to add special handeling logic for the created_by/modified_by fields to make sure we did not attempt to set that to the system user if we are in the middle of creating the sytem user.

Fixes #428


djyasin (approved):
I pulled this down and tested it. I can confirm I was unable to change the password for `_system`

ansible_base/lib/abstract_models/common.py
```diff
@@ -123,7 +127,7 @@ def save(self, *args, **kwargs):
           * It will automatically add a created_by fields for new items
         '''
 
-        if not self.pk:
+        if not self.pk and (not hasattr(self, 'username') or self.username != get_system_username()[0]):
```
AlanCoding: This condition is too hard to read. I suggest introducing an instance method `is_the_system_user` which returns `hasattr(self, 'username') and self.username == get_system_username()[0]`

AlanCoding: I realize this is a really base-level abstract model class. If you didn't want this method to be available on almost all instances, then add it as a static method at the top of this file which would take a `model.Model` as its argument.

ansible_base/lib/utils/create_system_user.py
```diff
@@ -0,0 +1,60 @@
...
+from ansible_base.lib.utils.settings import get_setting
+
+logger = logging.getLogger('ansible_base.lib.utils.create_system_user')
+
+"""
+These functions are in its own file because it is loaded during migrations so it has no access to models.
+"""
+
+
+def create_system_user(user_model=None) -> object:  # Note: We can't load models here so we can typecast to anything better than an object
```
AlanCoding: I'm pretty sure you could say `Model` instead of `object`, but neither are wrong. You might even be able to say the Django `AbstractUser`. Because while the principle is that our utils shouldn't import from "our" models, our utils can import for _Django_ models while preserving the intended import ordering.

ansible_base/lib/abstract_models/common.py
john-westcott-iv: How would you feel if I modified lib.utils.models.is_system_user to take `model.Model` and then add a check for "Is this model a user"? That way there is just one function in the system for this.

ansible_base/lib/utils/create_system_user.py
```diff
@@ -0,0 +1,60 @@
...
+    #
+    # Creating the system user is a little tricky because we need to reference ourselves
+    #
+    if not user_model:
+        logger.error("You must pass a user_model to create_system_user")
+        return None
```
AlanCoding, line 23: Then why make it optional??

ansible_base/lib/utils/create_system_user.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # First create a User object for the system user
+    system_user, created = user_model.objects.get_or_create(username=get_system_username()[0])
```
AlanCoding, line 26: No `defaults`? Don't have an email, or first_name?

ansible_base/lib/utils/create_system_user.py
```diff
@@ -0,0 +1,60 @@
...
+
+    if created:
+        logger.info(f"Created system user {system_user.username}")
+    else:
+        logger.debug("System user is already created")
+        return system_user
+
+    system_user.is_active = False
```
AlanCoding, line 34: You could probably move this, and maybe only this, into `defaults` for the `get_or_create`.

ansible_base/lib/utils/create_system_user.py
john-westcott-iv: I tried to make it AbstractUser but that was unable to be imported because we're still loading when calling this from a migration and, from that I can tell, you can't make a type hint variable (to load the model from an app).

john-westcott-iv: I can try model.Model, that would at least be more specific than object.

ansible_base/lib/utils/create_system_user.py
john-westcott-iv: The only required field on User is username. So I left all of the defaults off. We could do defaults for the is_active but we have to apply other things (like set_unusable_password and created_by) anyway so I just made this `username`.

test_app/models.py
```diff
@@ -45,6 +45,10 @@ def summary_fields(self):
         return user_summary_fields(self)
 
 
+class ManagedUser(User):
```
AlanCoding, line 48: :scream: this is using concrete inheritance. Are you sure you want to do this?

ansible_base/lib/utils/create_system_user.py
john-westcott-iv: Yea, I thought about doing this but since we have to call save either way I didn't think it would make a difference.

ansible_base/lib/utils/create_system_user.py
AlanCoding: Using the DAB `AbstractUser` would be categorically off the table. That's why I say the Django `AbstractUser`. But, you might want to allow custom user classes which do not subclass that, which would make that wrong. So it's likely that Model is the most specific you could get (while still being correct), given the intent.

ansible_base/lib/abstract_models/common.py
AlanCoding: Yeah, I saw there was a very similar method in that module. It would make sense to reuse that somewhere in the call stack for this. However it shakes out should be fine.

ansible_base/lib/utils/create_system_user.py
john-westcott-iv: Good call, I can make it non-optional.

ansible_base/lib/utils/create_system_user.py
john-westcott-iv: `models.Model` it is.

AlanCoding:
Given this change, wouldn't you prefer to delete `create_system_user` from test_app?

https://github.com/ansible/django-ansible-base/blob/e09740278c7d7d8ca69252c7aef417f50fe4248d/test_app/migrations/0001_initial.py#L13

Or am I missing something?

john-westcott-iv:
| Given this change, wouldn't you prefer to delete create_system_user from test_app

@AlanCoding ideally yes :)

AlanCoding:
Conflicts introduced by me in https://github.com/ansible/django-ansible-base/pull/422, because I really wanted an internal cacher in RBAC managers. I'd love to see this adopt stuff from that.

ansible_base/lib/utils/create_system_user.py
```diff
@@ -0,0 +1,61 @@
...
+    system_user.created_by = system_user
+    system_user.modified_by = system_user
+    update_fields = ['is_active', 'password', 'created_by', 'modified_by']
+
+    # If our model has a managed flag set it to true
+    if hasattr(user_model, 'managed'):
+        system_user.managed = True
+        update_fields.append('managed')
+
+    system_user.save(update_fields=update_fields)
```
AlanCoding, line 45: Should this save be done with activity stream disabled?

AlanCoding (changes requested):
I checked out your branch and ran `./test_app/scripts/bootstrap.sh`. This resulted in a "RecursionError: maximum recursion depth exceeded" error.

AlanCoding:
Here is the cycle for you:

```
  File "/home/alancoding/repos/awx/testing/django-ansible-base/ansible_base/lib/utils/models.py", line 123, in create_system_user
    system_user = get_user_model().objects.create(username=get_system_username()[0], is_active=False, modified_by=None, created_by=None)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.12/site-packages/django/db/models/manager.py", line 87, in manager_method
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.12/site-packages/django/db/models/query.py", line 658, in create
    obj.save(force_insert=True, using=self.db)
  File "/home/alancoding/venvs/awx/lib64/python3.12/site-packages/django/contrib/auth/base_user.py", line 76, in save
    super().save(*args, **kwargs)
  File "/home/alancoding/repos/awx/testing/django-ansible-base/ansible_base/lib/abstract_models/common.py", line 153, in save
    return super().save(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/repos/awx/testing/django-ansible-base/ansible_base/lib/abstract_models/common.py", line 128, in save
    self.created_by = current_user_or_system_user()
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/repos/awx/testing/django-ansible-base/ansible_base/lib/utils/models.py", line 192, in current_user_or_system_user
    user = get_system_user()
           ^^^^^^^^^^^^^^^^^
  File "/home/alancoding/repos/awx/testing/django-ansible-base/ansible_base/lib/utils/models.py", line 106, in get_system_user
    system_user = create_system_user()
                  ^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/repos/awx/testing/django-ansible-base/ansible_base/lib/utils/models.py", line 123, in create_system_user
    system_user = get_user_model().objects.create(username=get_system_username()[0], is_active=False, modified_by=None, created_by=None)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

john-westcott-iv:
@AlanCoding Can you double check your checkout? create_system_user isn't in ansible_base/lib/util/models.py anymore.

AlanCoding:
Yeah, I had the wrong branch. After running it now, I see these entries in the activity stream

```json
        {
            "id": 2,
            "url": "/api/v1/activitystream/2/",
            "related": {
                "created_by": "/api/v1/users/1/"
            },
            "summary_fields": {
                "created_by": {
                    "id": 1,
                    "username": "_system",
                    "first_name": "",
                    "last_name": ""
                }
            },
            "created": "2024-06-03T18:29:32.503350Z",
            "created_by": 1,
            "operation": "update",
            "changes": {
                "added_fields": {},
                "changed_fields": {
                    "password": [
                        "$encrypted$",
                        "$encrypted$"
                    ],
                    "is_active": [
                        true,
                        false
                    ],
                    "created_by": [
                        null,
                        1
                    ],
                    "modified_by": [
                        null,
                        1
                    ]
                },
                "removed_fields": {}
            },
            "content_type": 17,
            "object_id": "1",
            "related_content_type": null,
            "related_object_id": null,
            "content_type_model": "user",
        },
        {
            "id": 1,
            "url": "/api/v1/activitystream/1/",
            "related": {
                "created_by": "/api/v1/users/1/"
            },
            "summary_fields": {
                "created_by": {
                    "id": 1,
                    "username": "_system",
                    "first_name": "",
                    "last_name": ""
                }
            },
            "created": "2024-06-03T18:29:32.497655Z",
            "created_by": 1,
            "operation": "create",
            "changes": {
                "added_fields": {
                    "id": 1,
                    "email": "",
                    "created": "2024-06-03T18:29:32.493292Z",
                    "is_staff": false,
                    "modified": "2024-06-03T18:29:32.493287Z",
                    "password": "$encrypted$",
                    "username": "_system",
                    "is_active": true,
                    "last_name": "",
                    "created_by": null,
                    "first_name": "",
                    "date_joined": "2024-06-03T18:29:32.493139Z",
                    "modified_by": null,
                    "is_superuser": false
                },
                "changed_fields": {},
                "removed_fields": {}
            },
            "content_type": 17,
            "object_id": "1",
            "related_content_type": null,
            "related_object_id": null,
            "content_type_model": "user",
        }
    ]
```

I want to get confirmation that you're good with this.

AlanCoding (approved):
Never mind, this does seem to be working after a rebase.

I think I see it working as you intended here.

PR #420, opened 2024-05-30
Alter JWT authentication

john-westcott-iv:


ansible_base/jwt_consumer/hub/auth.py
```diff
@@ -1,15 +1,33 @@
...
-            from galaxy_ng.app.models.auth import Group
+            from galaxy_ng.app.models import Team
         except ImportError:
             raise InvalidService("automation-hub")
 
-        for team in claims.get("teams", []):
-            hub_group, _ = Group.objects.get_or_create(name=team["name"])
-            hub_group.user_set.add(user)
+        for role_name in token.get('object_roles', {}).keys():
+            if role_name.startswith('Team'):
```
bhavenst, line 24: Error log for non-Team requests?

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -265,27 +278,118 @@ def validate_token(self, token, decryption_key):
...
+    def process_user_data(self):
+        self.common_auth.map_user_fields()
 
-    def process_permissions(self, user, claims, token):
-        logger.info("process_permissions was not overridden for JWTAuthentication")
+    def process_permissions(self):
+        if self.use_rbac_permissions:
+            self.common_auth.process_rbac_permissions()
+        else:
+            logger.info("process_permissions was not overridden for JWTAuthentication")
```
bhavenst, line 302: Maybe message could say that RBAC is switched off, doing nothing or something like that.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -265,27 +278,118 @@ def validate_token(self, token, decryption_key):
...
     map_fields = default_mapped_user_fields
+    use_rbac_permissions = False
+
+    def __init__(self):
+        self.common_auth = JWTCommonAuth(self.map_fields)
 
     def authenticate(self, request):
-        common_auth = JWTCommonAuth(self.map_fields)
-        user, token = common_auth.parse_jwt_token(request)
+        logger.debug("Calling JWTAuthetnication.authenticate")
```
bhavenst: Intending to leave this in?

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -216,34 +234,29 @@ def get_decryption_key(self, url_or_string: str, ignore_cache: bool = False, val
         cache.set(cache_key, key, timeout=self.get_cache_timeout())
         return key, cached
 
-    def map_user_fields(self, user, token):
+    def map_user_fields(self):
```
bhavenst, line 121: Do you need protection from this and process(_rbac)_permissions being called before authenticate since they are exposed on the top classes?

test_app/tests/jwt_consumer/hub/test_auth.py
```diff
@@ -10,23 +11,38 @@
...
-def test_hub_jwt_teams(user, token):
-    mock_group = MagicMock()
-    mocked_Group = MagicMock(**{'Group.objects.get_or_create.return_value': (mock_group, True)})
-    sys.modules['galaxy_ng.app.models.auth'] = mocked_Group
-    mocked_authenticator = HubJWTAuth()
-    mocked_authenticator.process_permissions(user, token, {})
-    call_count = len(token.get('teams', []))
-    assert mocked_Group.Group.objects.get_or_create.call_count == call_count
+def test_hub_jwt_teams(user, token, num_roles):
+    assert True
```
bhavenst, line 27: Now that is a way to get a test to pass.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -34,32 +39,46 @@
...
+        self.user = None
+        self.token = None
+
         logger.debug("Starting JWT Authentication")
-        token = request.headers.get("X-DAB-JW-TOKEN", None)
-        if not token:
+        if request is None:
+            return
+
+        unencrypted_token = request.headers.get("X-DAB-JW-TOKEN", None)
```
bhavenst: Is this more accurately undecoded_token?  Isn't this code decrypting it?  Or maybe it was already decrypted in some way..

bhavenst (commented):
Just some minor comments..

ansible_base/jwt_consumer/hub/auth.py
john-westcott-iv: Hub does not use the RBAC system to its fullest potential so for Hub we just ignore anything that is not a Team role.

ansible_base/jwt_consumer/common/auth.py
john-westcott-iv: The initial intent is that every service would have a custom class extending this one and implement their own process_permissions to work however it needed to. Since then we've implemented RBAC and add the method that works with standard RBAC. So the patten is that services will have to make a conscious decision to use the standard RBAC processor. So this is intended really more as a "he dev you didn't do this" whether RBAC is installed or not.

test_app/tests/jwt_consumer/hub/test_auth.py
john-westcott-iv: lol, 100% test passing.... @jctanner is going to do a follow up on this PR to cover the Hub unit testing. I'm not familiar with Hub so this is more of a swag with what I am doing and the tests are just about completely dead from the changes.

ansible_base/jwt_consumer/common/auth.py
john-westcott-iv: No, good catch, that needs a better name. Its probably encoded and encrypted when its in the header.

ansible_base/jwt_consumer/common/auth.py
john-westcott-iv: process_permissions I'm not worried about because it will either call process_rbac_permissions or raise an error if its not overridden. I updated process_rbac_permissions and map_user_fields to error and return if the token or user are not defined.

ansible_base/jwt_consumer/common/auth.py
john-westcott-iv: Thank you for catching that.

bhavenst (approved):

BrennanPaciorek (commented):
lgtm. waiting for the rebase before approving though, since it seems like it will be a pretty involved process to resolve conflicts with current devel.

BrennanPaciorek (approved):

AlanCoding:
Here is my proposal to automatically create roles in the claims processing if it's in the constructor registry thing I did. https://github.com/john-westcott-iv/django-ansible-base/pull/6

ansible_base/jwt_consumer/hub/auth.py
```diff
@@ -1,15 +1,42 @@
...
+
+            self.team_content_type = ContentType.objects.get_for_model(Team)
+            self.org_content_type = ContentType.objects.get_for_model(Organization)
         except ImportError:
             raise InvalidService("automation-hub")
 
-        for team in claims.get("teams", []):
-            hub_group, _ = Group.objects.get_or_create(name=team["name"])
-            hub_group.user_set.add(user)
+        for role_name in self.common_auth.token.get('object_roles', {}).keys():
```
AlanCoding, line 23: It looks like you fixed the issue with references to `self.token` and user? You squashed commits

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -146,27 +166,122 @@ def validate_token(self, token, decryption_key):
...
+
+        # The resource was missing so we need to create its stub
+        if content_type == 'team':
+            # For a team we first have to make sure the org is there
+            org_id = data['org']
+            organization_data = self.token['objects'][self.org_content_type.model][org_id]
+
+            # Now that we have the org we can build a team
+            resource, org = self.get_or_create_resource(self.org_content_type.model, organization_data)
+            team, created = self.team_content_type.model_class().objects.get_or_create(
```
AlanCoding, line 256: Instead of calling `self.team_content_type.model_class().objects.get_or_create`, would it be cleaner to call `resource.create_resource`?

https://github.com/ansible/django-ansible-base/blob/9f0f05168dbcb18c1f542aec480b80eedb5ffc3e/ansible_base/resource_registry/models/resource.py#L99

...ideally we would remove the content-type-specific branches here.

AlanCoding (approved):
I don't see a _functional_ reason right now to block merge. I left some comments, but I could re-apply my patch against devel branch.

jctanner (approved):

ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -144,6 +144,10 @@
     REST_FRAMEWORK['DEFAULT_PAGINATION_CLASS'] = 'ansible_base.rest_pagination.DefaultPaginator'
 
 
+if 'ansible_base.jwt_consumer' in INSTALLED_APPS:
+    if 'ansible_base.rbac' not in INSTALLED_APPS:
+        INSTALLED_APPS.append('ansible_base.rbac')
```
AlanCoding, line 149: This would break galaxy_ng if they added the jwt_consumer to installed_apps.

https://github.com/ansible/galaxy_ng/blob/bc70fe3aecdeede74eb08f50fb3e5f3e8231d2f0/galaxy_ng/app/settings.py#L83

Right now I think they don't, and just use the auth class. But this does seem to undermine the point of having this switch. I would prefer to get rid of it. But non-blocking comment.

ansible_base/jwt_consumer/hub/auth.py
```diff
@@ -1,15 +1,42 @@
...
+                        except Exception as e:
+                            logger.error(f"Failed to request object {ansible_id} from gateway to grant permissions to {self.common_auth.user.username}: {e}")
+
+                    # Because RoleDefinition, Team and Group are all linked we don't have to go through individual lookups nor do we need to catch DoesNotExist
+                    team = Team.objects.get(id=resource.object_id)
+                    team.group.user_set.add(self.common_auth.user)
+
+        if "Platform Auditor" in self.common_auth.token.get('global_roles', []):
+            # Add platform operator here
+            pass
```
AlanCoding, line 42: :+1:  nice, thanks for adding that. Ping @jctanner

AlanCoding (approved):

hsong-rh:
We tested it for EDA, looks good.

ansible_base/jwt_consumer/common/auth.py
john-westcott-iv: Will follow up in a separate PR.

ansible_base/lib/dynamic_config/dynamic_settings.py
john-westcott-iv: galaxy_ng has issues for sure but that will be addressed laster.

ansible_base/jwt_consumer/common/auth.py
AlanCoding: Actually that's a classmethod, and doesn't do quite what i expected. Maybe don't bother.

PR #421, opened 2024-05-30
oauth2_provider: support adding, removing, clearing tokens

BrennanPaciorek:
Add support for the django-admin commands: create_oauth2_token, revoke_oauth2_tokens, cleanup_tokens.

Fixes: [AAP-18812](https://issues.redhat.com/browse/AAP-18812)

### DIFF
--- a/ansible_base/oauth2_provider/management/commands/cleanup_tokens.py
+++ b/ansible_base/oauth2_provider/management/commands/cleanup_tokens.py
@@ -1,0 +1,27 @@
+import logging
+
+from django.core.management import call_command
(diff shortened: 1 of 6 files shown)


BrennanPaciorek:
I don't think I've made any changes to break these tests.

BrennanPaciorek:
These tests seem to be producing inconsistent results. One of py39, 310, 311 will produce a failure for test_radius.py::test_authenticator_plugin. Are any of my tests modifying things that that particular test depends on?

djyasin (approved):
I can confirm I was able to test these commands locally and they worked with no issues!

PR #422, opened 2024-05-30
Introduce caching utility for lazy managed role references

AlanCoding:
This is a baby-step in the direction I want to go for defining managed RoleDefinitions (or at least some of them in DAB), but for now, this removes the error seen in test logs:

```
INFO     ansible_base.authentication.utils.claims:claims.py:368 Reconciling user claims
ERROR    ansible_base.authentication.utils.claims:claims.py:275 Failed to reconcile user claims: RoleDefinition matching query does not exist.
  File "/home/alancoding/repos/awx/testing/django-ansible-base/ansible_base/authentication/utils/claims.py", line 273, in update_user_claims
    reconcile_user_class.reconcile_user_claims(user, authenticator_user)
  File "/home/alancoding/repos/awx/testing/django-ansible-base/ansible_base/authentication/utils/claims.py", line 371, in reconcile_user_claims
    org_member_rd = RoleDefinition.objects.get(name=cls.ORGANIZATION_MEMBER_ROLE_NAME)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.12/site-packages/django/db/models/manager.py", line 87, in manager_method
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alancoding/venvs/awx/lib64/python3.12/site-packages/django/db/models/query.py", line 637, in get
    raise self.model.DoesNotExist(
ansible_base.rbac.models.RoleDefinition.DoesNotExist: RoleDefinition matching query does not exist.
```

I really don't want to require that each `RoleDefinition` exists to be able to use basic authentication. If you get a _claim_ for a role definition that doesn't exist, yeah, that'll be a problem. But if there are no claims to a `RoleDefinition` that doesn't exist, I really don't want to do anything about that. The tests via test_app won't have these role definitions loaded, and I don't want to change that.

### DIFF
--- a/ansible_base/rbac/models.py
+++ b/ansible_base/rbac/models.py
@@ -44,7 +44,38 @@ class DABPermission(models.Model):
         return f"<{self.__class__.__name__}: {self.codename}>"
 
 
+class ManagedRoleFromSetting:
+    def __init__(self, role_name):
+        super().__init__()
+        self.role_name = role_name
+
+    def __get__(self, obj, objtype=None):
+        if obj is None:
+            return self
+        if self.role_name not in obj._cache:
+            obj._cache[self.role_name] = RoleDefinition.objects.filter(name=self.role_name).first()
+        return obj._cache[self.role_name]
+
+
+class ManagedRoleManager:
+    def __init__(self):
+        self._cache = {}
+
+    def clear(self) -> None:
+        "Clear any managed roles already loaded into the cache"
+        self._cache = {}
+
+    org_admin = ManagedRoleFromSetting('Organization Admin')
+    org_member = ManagedRoleFromSetting('Organization Member')
+    team_admin = ManagedRoleFromSetting('Team Admin')
+    team_member = ManagedRoleFromSetting('Team Member')
+
+
 class RoleDefinitionManager(models.Manager):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.managed = ManagedRoleManager()
+
     def give_creator_permissions(self, user, obj) -> Optional['RoleUserAssignment']:
         # If the user is a superuser, no need to bother giving the creator permissions
         for super_flag in settings.ANSIBLE_BASE_BYPASS_SUPERUSER_FLAGS:
--- a/ansible_base/authentication/utils/claims.py
+++ b/ansible_base/authentication/utils/claims.py
@@ -353,13 +353,11 @@ def create_organizations_and_teams(results) -> None:
 #  removal or update of a pluggable interface.
 class ReconcileUser:
 
-    TEAM_MEMBER_ROLE_NAME = "Team Member"
-    TEAM_ADMIN_ROLE_NAME = "Team Admin"
-    ORGANIZATION_MEMBER_ROLE_NAME = "Organization Member"
-    ORGANIZATION_ADMIN_ROLE_NAME = "Organization Admin"
-
     @classmethod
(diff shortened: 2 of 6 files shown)


ansible_base/rbac/models.py
```diff
@@ -44,7 +44,38 @@ def __str__(self):
...
+        self._cache = {}
+
+    def clear(self) -> None:
+        "Clear any managed roles already loaded into the cache"
+        self._cache = {}
+
+    org_admin = ManagedRoleFromSetting('Organization Admin')
+    org_member = ManagedRoleFromSetting('Organization Member')
+    team_admin = ManagedRoleFromSetting('Team Admin')
+    team_member = ManagedRoleFromSetting('Team Member')
```
john-westcott-iv, line 71: Would we put System Auditor in here as well or, because not all services support it, would we have a service extend this class to add it? If a service wanted to extend this what would that look like since its hardcoded on line 77 below?

ansible_base/rbac/models.py
```diff
@@ -44,7 +44,38 @@ def __str__(self):
...
+    def __init__(self, role_name):
+        super().__init__()
+        self.role_name = role_name
+
+    def __get__(self, obj, objtype=None):
+        if obj is None:
+            return self
+        if self.role_name not in obj._cache:
+            obj._cache[self.role_name] = RoleDefinition.objects.filter(name=self.role_name).first()
+        return obj._cache[self.role_name]
```
john-westcott-iv, line 57: If someone deleted the role "Organization Admin" and this caches None what would happen in the code above? Would we get a NoneType has no attribute give_permission? Should we have a dummy object incase the objects.filter returns None that has the methods like give_permission/remove_permission but they are no-ops?

ansible_base/rbac/models.py
AlanCoding: "service wanted to extend this" - I'm not doing that here. My approach:

 - introduce a caching data structure for managed roles
 - introduce a registry, a part of that, including the algorithm to create the managed role

so this is doing the first part but not the second part. They will interact, but we'll have the caching in any case.

ansible_base/rbac/models.py
AlanCoding: Generally managed roles shouldn't get deleted... but if it was deleted in a migration, no functional python process should be active both before and after the migration.

But I am thinking about a different version of these problems as the registry comes into things. If you have a managed RoleDefinition and remove it from the registry, it is a problem if it still exists, but if we have a registry this shouldn't return the object if it's not in the registry. Say you have a registry, reference a managed role, but that role isn't created. Well in that case it should return a `TentativeRoleDefinition` which will create itself when it is needed.

john-westcott-iv: I am thinking about role based authenticator_maps here. If we have a map that uses a role that gets deleted (either through a migration or because it was a custom role and it was no longer needed) it could still be in the authenticator_map. So we would either need to:

1. Cascade the delete to authenticator_maps (which feels wrong)
2. Prevent the deletion if the role is tied to an authenticator_map (which might be a pain for upgrades)
3. Modify claims.py to check to see if `RoleDefinition.objects.managed.org_member` returns None so we don't try and call give_permission/remove_permission. 

But this can be tackled later.

john-westcott-iv (approved):

PR #425, opened 2024-05-31
Fix queryset of RoleDefinition-assignment list with new utils

### LINKED ISSUE #304: Role definition user/team assignment sublists should be filtered to visible assignments
We added a sublist in "related" links for the RoleDefinition model, so `/api/v1/role_definitions/` have a related user_assignments / team_assignments links.

These links show assignments that the current user doesn't have permission to view.

Props to @chrismeyersfsu for discovery.

I don't know if filtered sublists have ever been done in DAB up to this point.

AlanCoding:
Fixes https://github.com/ansible/django-ansible-base/issues/304

AAP-25895

A lot of changes were recently made to the contract for `related_views` in the router. The docs are at:

https://github.com/ansible/django-ansible-base/blob/devel/docs/lib/routers.md#customization

So by using `filter_queryset` a lot of things will "just work", but the views have not really attempted to adopt these abilities until now. This sublist view is a good example where we obviously have to filter the list.


jctanner (approved):
I've tested locally, and can confirm the user_assigments endpoints found under role_definitions are responding with empty lists where needed.

PR #426, opened 2024-05-31
Revert "disallow filtering by encrypted_fields"

BrennanPaciorek:
Reverts ansible/django-ansible-base#414


BrennanPaciorek:
Just making this PR to show that effectively undoing devel HEAD does not fix the current failures in test_radius.py

PR #427, opened 2024-05-31
Split DAB RBAC app docs into 4 parts by intended audience

AlanCoding:
Previously, the `docs/apps/rbac.md` document was organized into 3 sections, based on persona.

I installed Inkscape and started doodling. Before I knew it, I had made the 4 svg files included in this PR, so I packaged them up into a new document for another persona, so now we have 4 personas. This splits and links those.

The svg files have a formatting problem where the canvas is too big. I will try to fix that in another commit.

Ping @tvo318 @TheRealHaoLiu @fosterseth @dcdacosta


docs/apps/rbac/for_users.md
```diff
@@ -0,0 +1,126 @@
...
+![User getting an organization role](images/user_org.svg)
+
+### Resource Tree
+
+This concept is mainly intended for organizations, but it generalized
+in DAB RBAC, so that any model can have child models if they have
+a relational link between them.
+
+So if you consider the above illustration of a user gaining permission
+to an organization, the grouping of objects could have nested groupings.
```
AlanCoding, line 82: @chrismeyersfsu mentioned to me that we don't have any visualizations for nested resources. This is a good point, but there's a fairly clear place-holder here for where it would be added. I would still like to leave that as future work. Hopefully we know exactly what it would look like.

Once we support many-to-many links to organizations from resources, that also would need a diagram clarifying how the grouping mechanics would work.

docs/apps/rbac/for_users.md
```diff
@@ -0,0 +1,126 @@
+# Using DAB Role-Based Access Control
+
+Access control allows or denies requests from a user based on some permission specification.
+Roles are an implementation detail of some access control systems, such as this one.
```
AlanCoding, line 4: TODO for later, before the "Model Permissions" section, we should introduce a "Content Type" section, because this document does not assume the reader has a background with Django development.

docs/apps/rbac/for_users.md
```diff
@@ -0,0 +1,126 @@
...
+This is illustrated in the image below. 1 user gets 1 or more permissions to the object.
+
+![User getting permission to an object](images/user_obj.svg)
+
+## Role Team Assignments
+
+Teams are treated like other objects in the system by DAB RBAC,
+with the exception that its **member** permission acts as a multiplier.
+By receiving a team's member permission, you automatically receive all
+permissions that the team has.
```
AlanCoding, line 45: One more TODO I'd like to illustrate: A new diagram showing team membership _as a role assignment_ via the team member role or admin role, thus showing how that role assignment defines the grouping that the subsequent diagram shows. The point is to quickly cover this to avoid confusion and say "when we draw a line around users, this is what we mean".

docs/apps/rbac/for_users.md
```diff
@@ -0,0 +1,126 @@
...
+to multiple objects to multiple users.
+
+![Team getting an organization role](images/team_org.svg)
+
+## System-wide Roles
+
+DAB RBAC allows creating **system-wide role definitions**.
+System-wide role definitions are idenified by a **null content_type**.
+Any assignments given for this role will, correspondingly, have a **null object**.
+
```
trucdg, line 97: "DAB RBAC allows creaating **system-wide rode definitions**":
- However, I don't see any “null” option in the Content Type list selection either in the UI or the API HTML form when trying to add a new role defitinion. I can only create through the raw_data format in the /api/v1/role_definitions/.

I wonder if this is intended or it is something we should fix to not cause future confusion

docs/apps/rbac/for_clients.md
```diff
@@ -0,0 +1,114 @@
+## Using DAB RBAC as an API Client
+
+This section tells how to use the API endpoints as a client;
+what requests to make and how to build request data.
+
+You need a server running a Django project that uses this system.
+Use test_app in this repo for a demo, see `test_app/README.md` for bootstrap.
+http://127.0.0.1:8000/admin/ allows login as admin user (password is "password").
+
```
trucdg, line 10: - I have pinged you in Slack regarding the inconsistency between the test_app/README.md and the bootstrap.sh script.
- I think the password for admin user can be either "password" or "admin" depending on how we fix the bug in bootstrap.sh

docs/apps/rbac/for_clients.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+Name can be any string that's not blank. Description can be any string.
+
+### Assigning a User a Role to an Object
+
+Select a role definition from `/api/v1/role_definitions/`, use the id as `role_definition`
+and a user from `/api/v1/users/` and use the id as `user`.
+Given the type of the role definition, check the available objects of that type,
+in this case `/api/v2/inventories/` and obtain the desired id, this will become `object_id`.
```
trucdg: is it `/api/v2/inventories/` or /api/v1/inventories/`?
I can't find `/api/v2/inventories/` anywhere so I'm not sure if this is a typo

docs/apps/rbac/for_clients.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Assigning a Team a Role to an Object
+
+If you used the test_app bootstrap script, then you will find the user "angry_spud"
+is a member of the "awx_devs" team.
+
+Assuming the team id is 2, POST to
+
+http://127.0.0.1:8000/api/v1/role_team_assignments/
+
```
trucdg, line 72: I think the description for Role **User** and Role **Team** Assignment List are swapped. 
The sentence where it say "The needed data is the **user** ..." is under **Role Team** Assignment List. And it's the reverse for Role User Assignment List

docs/apps/rbac/for_clients.md
```diff
@@ -0,0 +1,114 @@
...
+With all 3 ids ready, POST to http://127.0.0.1:8000/api/v1/role_user_assignments/
+
+```json
+{
+    "role_definition": 3,
+    "object_id": 3,
+    "user": 3
+}
+```
```
trucdg, line 52: The json content template is:
{
    "object_id": "",
    "object_ansible_id": null,
    "role_definition": null,
    "user": null,
    "user_ansible_id": null
}
not sure if we should fix this

docs/apps/rbac/for_users.md
AlanCoding: Right now the API HTML form doesn't work at all, but this is due to the `permissions` field. In that form, I think the `content_type` is nullable as expected, I see an option like "-------". Manually, I can create a role definition using like:

```json
{
    "permissions": ["view_inventory"],
    "content_type": null,
    "name": "Inventory Viewer"
}
```

but yeah, this is using the raw_data, due to the issue of permissions being a multi-select.

docs/apps/rbac/for_clients.md
AlanCoding: Yeah, this is a valid issue. I think it would be best to file it in https://github.com/ansible/django-ansible-base/issues, because in that case it's unclear what to change (the demo data command, or the script?), so the issue is a good place to formalize the decision.

docs/apps/rbac/for_clients.md
AlanCoding: Good find, that's inconsistent. I'd like it to all be v1. Actionable steps in test_app might be good.

docs/apps/rbac/for_clients.md
AlanCoding: AWX had a solution for this with a generic `update_raw_data` method

https://github.com/ansible/awx/blob/devel/awx/api/generics.py#L361-L372

But it's fairly messy to implement, because a flag has to be set on the serializer in the renderer class, which we don't even do anything custom to in DAB yet. So I'm not yet ready to dive into that customization.

I'm wondering if we could, instead, change the fields to allow null, and then do nothing when getting a null value.

AlanCoding: https://github.com/ansible/django-ansible-base/pull/468

docs/apps/rbac/for_clients.md
trucdg: I'll go ahead and open a new issue.

trucdg: #471

trucdg (approved):
Thanks for making a detailed doc!

docs/apps/rbac/for_clients.md
```diff
@@ -0,0 +1,114 @@
+## Using DAB RBAC as an API Client
+
+This section tells how to use the API endpoints as a client;
+what requests to make and how to build request data.
+
+You need a server running a Django project that uses this system.
+Use test_app in this repo for a demo, see `test_app/README.md` for bootstrap.
+http://127.0.0.1:8000/admin/ allows login as admin user (password is "password").
```
trucdg: I think after your #471 , password will be "admin"

AlanCoding: Good call-out. This whole thing is out-of-date, and I'll update it.

PR #429, opened 2024-06-03
Fix issue where tests fail to teardown

john-westcott-iv:


jctanner (approved):
a more specific exception would be nice in the future

ansible_base/lib/utils/models.py
```diff
@@ -101,7 +101,12 @@ def get_system_user() -> Optional[AbstractUser]:
                 )
             )
         )
-        system_user = create_system_user(user_model=get_user_model())
+        try:
+            # There are issues in the tests where, during teardown the system user can not be created
```
AlanCoding, line 106: Because of a broken transaction?

john-westcott-iv: I haven't been able to determine why this happening. I thought pytest was run within a transaction but it for some reason the resource_registry signals are being processed like there has been a reverse migration. I tried adding code to `initialize_resources` in `ansible_base.resource_registry.apps` to detect a reverse migration but it didn't help. This method eventually does `ResourceType.objects.update_or_create` which will then try and create the system user. But things are "missing" at this point causing the exception we are catching above. But this seems like can rear its head in multiple exception types depending on the test which is why I just catch Exception.

john-westcott-iv:
>  a more specific exception would be nice in the future

It would be but depending on the test it seems like it can rear up differently. We've seen KeyError, and UserNotDefined, and I think maybe one or two more. I suppose I could run the tests and then capture all of them and add them in here instead of just Exception.

AlanCoding (approved):
The `ResourceType.DoesNotExist` exception is very narrow and self-explanatory. You have a comment in the method which is actually outdated given that change, better to delete it before merge.

PR #431, opened 2024-06-04
Reconcile User Claims with AuthMap roles

slemrmartin:
- **depends on** #407 

This PR applies AuthenticationMap RBAC roles to the Reconcile User Claims.

The AuthenticatorMap.remove_users = True now removes all role_user_assignments (it used to removed only Team Member, Organization Member)


ansible_base/lib/testing/fixtures.py
```diff
@@ -50,7 +50,7 @@ def local_authenticator(db):
         name="Test Local Authenticator",
         enabled=True,
         create_objects=True,
-        remove_users=True,
+        remove_users=False,
```
slemrmartin, line 53: when `True` this breaks some rbac tests. Because `reconcile_user_claims()` used to remove only member roles, now it removes all user roles (so it influences `org_admin` fixture)

test_app/tests/authentication/utils/test_claims_reconcile_user.py
```diff
@@ -0,0 +1,262 @@
+import random
```
slemrmartin, line 1: this file was renamed from `test_claims_orgs_and_teams.py` (here as deleted)

ansible_base/authentication/utils/claims.py
```diff
@@ -77,29 +83,73 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
             access_allowed = False
         elif auth_map.map_type == 'is_superuser':
             is_superuser = has_permission
-        elif auth_map.map_type == 'is_system_auditor':
+        elif auth_map.map_type == 'is_system_auditor':  # TODO: remove?
             is_system_auditor = has_permission
```
john-westcott-iv: Yes, we can remove this

ansible_base/authentication/utils/claims.py
```diff
@@ -77,29 +83,73 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
...
+            _add_rbac_role_mapping(has_permission, rbac_role_mapping, auth_map.role)
+
         else:
             logger.error(f"Map type {auth_map.map_type} of rule {auth_map.name} does not know how to be processed")
 
     return {
         "access_allowed": access_allowed,
         "is_superuser": is_superuser,
-        "is_system_auditor": is_system_auditor,
+        "is_system_auditor": is_system_auditor,  # TODO: remove
```
john-westcott-iv: Yes, remove.

ansible_base/authentication/utils/claims.py
```diff
@@ -77,29 +83,73 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
...
+            role_mapping['organizations'][organization] = {'roles': {}, 'teams': {}}
+        # Organization role
+        if organization and not team:
+            role_mapping['organizations'][organization]['roles'][role] = has_permission
+        # Team role
+        elif organization and team:
+            if team not in role_mapping['organizations'][organization]['teams']:
+                role_mapping['organizations'][organization]['teams'][team] = {'roles': {}}
+            role_mapping['organizations'][organization]['teams'][team]['roles'][role] = has_permission
+
```
john-westcott-iv, line 149: I get nervous not having an else. Really, it shouldn't be possible here so can we either:
Make the elif just an else or add an else with a logged error message like `Got a role mapping that is impossible to process`. 

It sounds dumb but if a dev ever added another role mapping type this would at least alert them as to where else they need to add code.

ansible_base/authentication/utils/claims.py
```diff
@@ -77,29 +83,73 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
...
+        if organization not in role_mapping['organizations']:
+            role_mapping['organizations'][organization] = {'roles': {}, 'teams': {}}
+        # Organization role
+        if organization and not team:
+            role_mapping['organizations'][organization]['roles'][role] = has_permission
+        # Team role
+        elif organization and team:
+            if team not in role_mapping['organizations'][organization]['teams']:
+                role_mapping['organizations'][organization]['teams'][team] = {'roles': {}}
+            role_mapping['organizations'][organization]['teams'][team]['roles'][role] = has_permission
```
john-westcott-iv, line 146: If a user has access to a team in an org do we also need to give them org member to the org itself?

test_app/tests/authentication/utils/test_claims.py
```diff
@@ -5,88 +5,233 @@
...
+            None,
+            {},
+            [],
+            True,
+            None,
+            None,
+            {"team_membership": {}, "organization_membership": {}, 'rbac_roles': {'system': {'roles': {}}, 'organizations': {}}},
+            [{1: "skipped"}],
+        ),
+        # This one is ambiguous, create_claims() evaluates triggers in wrong order
```
john-westcott-iv: Is this a bug or am I misunderstanding this comment? If its not a bug can we elaborate on the comment and delete the test? If it is a bug can we add something like "Once xyz is reaolved uncomment this test".

test_app/tests/authentication/utils/test_claims.py
```diff
@@ -5,88 +5,233 @@
 
 from ansible_base.authentication.models import AuthenticatorUser
 from ansible_base.authentication.utils import claims
+from test_app.tests.authentication.conftest import SYSTEM_ROLE_NAME
 
 
-@pytest.mark.xfail(connection.vendor == 'postgresql', reason='Test only reliable in sqlite3 due to pk references')
 @pytest.mark.parametrize(
```
john-westcott-iv, line 11: Should we add a label/comment to the various tests in here to give someone a better idea of what each is doing?

ansible_base/authentication/utils/claims.py
```diff
@@ -352,58 +401,279 @@ def create_organizations_and_teams(results) -> None:
...
+        return None
+
+    def add(self, role_name: str, organization: Optional[AbstractOrganization] = None, team: Optional[AbstractTeam] = None) -> None:
+        """Marks role assignment's params and (optionally) associated object in the cache.
+        If role_user_assignment (a.k.a. permission) existed before, marks it to do nothing
+        """
+        content_type_id = self._get_content_type_id(organization, team)
+        object_id = self._get_object_id(organization, team)
+
+        self._init_key(role_name, content_type_id=content_type_id)
```
john-westcott-iv:
```suggestion
        self._init_cache_key(role_name, content_type_id=content_type_id)
```

ansible_base/authentication/utils/claims.py
```diff
@@ -352,58 +401,279 @@ def create_organizations_and_teams(results) -> None:
...
+
+    def _get_object_id(self, organization, team) -> Optional[int]:
+        object_id = None
+        if organization:
+            object_id = organization.id
+        elif team:
+            object_id = team.id
+        return object_id
+
+    def _init_key(self, role_name: str, content_type_id: Optional[int]) -> None:
```
john-westcott-iv:
```suggestion
    def _init_cache_key(self, role_name: str, content_type_id: Optional[int]) -> None:
        """
        Initialize a key in the cache for later user
        """
```

ansible_base/authentication/utils/claims.py
```diff
@@ -352,58 +401,279 @@ def create_organizations_and_teams(results) -> None:
...
+
+        return content_type.id if content_type is not None else None
+
+    def _get_object_id(self, organization, team) -> Optional[int]:
+        object_id = None
+        if organization:
+            object_id = organization.id
+        elif team:
+            object_id = team.id
+        return object_id
```
john-westcott-iv:
```suggestion
    def _get_object_id(self, organization Optional[AbstractOrganization], team Optional[AbstractTeam]) -> Optional[int]:
        """
        If return an object id to work on, if both items are set organization will take priority over a team id.
        """
        
        if organization:
            return organization.id
        elif team:
            return team.id
        else:
            return None
```

ansible_base/authentication/utils/claims.py
```diff
@@ -352,58 +401,279 @@ def create_organizations_and_teams(results) -> None:
...
+
+    def _remove_permission(self, role_definition: CommonModel, obj: Union[AbstractOrganization, AbstractTeam, None] = None) -> None:
+        if obj:
+            logger.info(
+                _("Removing role '{rd}' from user '{username}' in '{object}").format(
+                    rd=role_definition.name, username=self.user.username, object=obj.__class__.__name__
+                )
+            )
+        else:
+            logger.info(_("Removing role '{rd}' to user '{username}'").format(rd=role_definition.name, username=self.user.username))
```
john-westcott-iv:
```suggestion
            logger.info(_("Removing role '{rd}' from user '{username}'").format(rd=role_definition.name, username=self.user.username))
```

ansible_base/authentication/utils/claims.py
```diff
@@ -352,58 +401,279 @@ def create_organizations_and_teams(results) -> None:
...
+            self.user.save()
+
+    def apply_permissions(self) -> None:
+        """See RoleUserAssignmentsCache for more details."""
+        for role_name, role_permissions in self.permissions_cache.items():
+            if self.permissions_cache.rd_by_name(role_name):
+
+                for content_type_id, content_type_permissions in role_permissions.items():
+                    for _object_id, object_with_status in content_type_permissions.items():
+                        self._apply_permission(object_with_status, role_name)
```
john-westcott-iv:
```suggestion
        for role_name, role_permissions in self.permissions_cache.items():
            if self.permissions_cache.rd_by_name(role_name):
                # If we failed to load this role for some reason we can't continue setting the permissions, log message has already emitted
                continue

            for content_type_id, content_type_permissions in role_permissions.items():
                for _object_id, object_with_status in content_type_permissions.items():
                    self._apply_permission(object_with_status, role_name)
```

ansible_base/authentication/utils/claims.py
```diff
@@ -352,58 +401,279 @@ def create_organizations_and_teams(results) -> None:
...
-            if org is None:
-                logger.info("Skipping all teams in organization '%s', because the organization does not exist", org_name)
+    def _compute_team_permissions(self, org: AbstractOrganization, teams_dict: dict[str, dict]) -> None:
+        teams_by_name = self._get_teams_by_name(org.id, teams_dict.keys())
+
+        for team_name, team_details in teams_dict.items():
+            if (team := teams_by_name.get(team_name)) is None:
+                logger.info(
+                    _("Skipping team '{team}' in organization '{organization}', because the team does not exist").format(team=team_name, organization=org.name)
+                )
```
john-westcott-iv: We need to build the objects if they don't exist.

ansible_base/authentication/utils/claims.py
```diff
@@ -352,58 +401,279 @@ def create_organizations_and_teams(results) -> None:
...
+            else:
+                self.permissions_cache.remove(role_name, organization=None, team=None)
+
+    def _compute_organization_permissions(self) -> Iterator[Tuple[AbstractOrganization, dict]]:
+        orgs_by_name = self._get_orgs_by_name(self.claims['rbac_roles'].get('organizations', {}).keys())
+
+        for org_name, org_details in self.claims['rbac_roles'].get('organizations', {}).items():
+            if (org := orgs_by_name.get(org_name)) is None:
+                logger.info(_("Skipping organization '{organization}', because the organization does not exist").format(organization=org_name))
+                continue
```
john-westcott-iv: We need to build the objects if they don't exist.

ansible_base/authentication/utils/claims.py
```diff
@@ -352,58 +401,279 @@ def create_organizations_and_teams(results) -> None:
...
+        self._init_key(role_name, content_type_id=content_type_id)
+
+        # If the permission haven't existed, mark it to ADD, otherwise noop
+        current_status = self.cache[role_name][content_type_id].get(object_id, {}).get('status')
+        if current_status in [self.STATUS_EXISTING, self.STATUS_NOOP]:
+            self.cache[role_name][content_type_id][object_id] = {'object': organization or team, 'status': self.STATUS_NOOP}
+        elif current_status is None:
+            self.cache[role_name][content_type_id][object_id] = {'object': organization or team, 'status': self.STATUS_ADD}
+
+    def remove(self, role_name: str, organization: Optional[AbstractOrganization] = None, team: Optional[AbstractTeam] = None) -> None:
```
john-westcott-iv: Add and remove are almost identical and we have a couple places where we are doing:
```
if has_role:
    add
else:
    remove
```

Should we combine these two functions into one and have it take "has role" as a param and do the if inside of it?

ansible_base/authentication/utils/claims.py
```diff
@@ -352,58 +401,279 @@ def create_organizations_and_teams(results) -> None:
...
+        else:
+            logger.info(_("Removing role '{rd}' to user '{username}'").format(rd=role_definition.name, username=self.user.username))
+
+        if obj:
+            role_definition.remove_permission(self.user, obj)
+        else:
+            role_definition.remove_global_permission(self.user)
+
+
+class RoleUserAssignmentsCache:
```
john-westcott-iv, line 552: This is making me think should we have an RBAC cache inside of Redis for all nodes to use. We would need a good cache-busting strategy to do that though. Not something specifically for this PR.

ansible_base/lib/testing/fixtures.py
AlanCoding: This isn't totally clear to me. There's an `org_admin_rd` fixture. That fixture by itself just gives the role definition and shouldn't have any connection to claims processing. If you do claims processing with an empty roles list, then sure, it might remove org admin permissions, but I don't know where that would be a problem.

ansible_base/authentication/utils/claims.py
```diff
@@ -352,58 +401,279 @@ def create_organizations_and_teams(results) -> None:
...
+    STATUS_NOOP = 'noop'
+    STATUS_EXISTING = 'existing'
+    STATUS_ADD = 'add'
+    STATUS_REMOVE = 'remove'
+
+    def __init__(self):
+        self.cache = {}
+        self.content_types = {
+            'organization': ContentType.objects.get_for_model(get_organization_model()),
+            'team': ContentType.objects.get_for_model(get_team_model()),
```
AlanCoding: You can call `get_for_models` instead to save 1 query. Not a blocker, but you could just leave a TODO. The data format it returns might already be what you need here.

ansible_base/authentication/utils/claims.py
slemrmartin: yeah, this needs to be discussed. But current behaviour in AWX is no, so I understood we won't do this now here too.

test_app/tests/authentication/utils/test_claims.py
slemrmartin: I was quite surprised, that the order of `always` and `never` wasn't important, it was always processed as `never`. So I'm not quite sure it is a bug or not, I'll try to look at it

ansible_base/authentication/utils/claims.py
slemrmartin: we should rather log error. Teams and organizations are created during `create_claims()`, so this shouldn't happen

ansible_base/authentication/utils/claims.py
slemrmartin: the same, we may increase log severity instead, because this line shouldn't happen

ansible_base/lib/testing/fixtures.py
slemrmartin: bad example. It's failing i.e. on 
```
FAILED tests/rbac/api/test_user_permissions.py::TestUserDetailView::test_superuser_can_delete_new_user[True] - assert 403 == 204
```

I think the issue is in the `admin_api_client` + `admin_user` fixtures. 
My guess is that the `admin_user` is created with flag `is_superuser`, then `admin_api_client` calls `login`, which call the `reconcile_user_claims`, which removes the `is_superuser` flag

test_app/tests/authentication/utils/test_claims.py
slemrmartin: it's like I wrote, 'never' is always first in `auth_map.triggers` (not sure why). 
But the order is nowhere guaranteed, so if somebody creates contradictory rules, I think it's correct that the result is unpredictable. 
We might add some check and at least log warning, but it's beoynd scope of this PR

ansible_base/authentication/utils/claims.py
slemrmartin: confirmed, marking as resolved

test_app/tests/authentication/utils/test_claims.py
slemrmartin: added

ansible_base/authentication/utils/claims.py
slemrmartin: this needs opposite condition, but get the point - added

ansible_base/authentication/utils/claims.py
slemrmartin: done

ansible_base/authentication/utils/claims.py
slemrmartin: it would need quite more effort indeed

test_app/tests/authentication/utils/test_claims.py
slemrmartin: -> test removed

ansible_base/authentication/utils/claims.py
john-westcott-iv: Ah, ok. Yea, lets make this an error message and have it say something `because the team does not exist but it should already have been created`. That will also help clarify in the code that this shouldn't happen.

test_app/tests/authentication/utils/test_claims.py
john-westcott-iv: I'm not fully understanding this. `always` and `never` are `triggers` on an `authenticator_map`. Are you saying that you are specifying both `always` and `never` inside the `trigger` of a single `authenticator_map`? I don't see that happening anywhere so I am assuming not. You can have them in multiple `authenticator_maps` but those have an `order` field. If two `authenticator_maps` have the same order we say its undefined as to which `authenticator_map` would run first.

ansible_base/lib/testing/fixtures.py
john-westcott-iv: It might be better to not user admin for some tests and user a random user instead.

john-westcott-iv:
@slemrmartin I left a couple comments on this but I think we are in pretty good shape so far. I'd like to get this merged tomorrow.

test_app/tests/authentication/utils/test_claims.py
slemrmartin: That's your test, John :) I just removed the `xfail` and extended the params in `pytest.parametrize` with map_typ=role

ansible_base/lib/testing/fixtures.py
slemrmartin: the `admin_api_client` fixture is based on `admin_user`. Moreover the `remove_users=True` doesn't make much sense as a default behaviour for tests, because this situation can happen elsewhere and it'll be tough to reveal this "magic" for other developers

ansible_base/authentication/utils/claims.py
slemrmartin: messages updated

test_app/tests/authentication/utils/test_claims.py
slemrmartin: anyway we are in agreement that the test is not useful, so this is solved

john-westcott-iv (approved):

Issue #424, opened 2024-05-31
Allow disabling object-level permissions for a given model in the RBAC registry


AlanCoding:
This is a "model registration flexibility" issue similar to https://github.com/ansible/django-ansible-base/issues/365

The ask here is to allow disabling of using object-level roles for a particular model. Spitball of the contract for this:

```python
permission_registry.register(Instance, allow_object_roles=False)
```

We would have `allow_object_roles` default to True, as this is how it is always used today. ~With this change, the `/api/v1/role_metadata/` endpoint would not show the content type for the `Instance` model~ (see later notes). _However, importantly_, the permissions related to the `Instance` model would still be shown as possible for system-wide roles. This is a valid use-case for this feature, because you may want to allow someone to manage instances in AWX without making them a superuser, but you still don't want to build out a UI for giving permissions to single instances, because you just don't want that.

The AWX `Instance` model does not have a related organization, but you could repeat for an organization-scoped model.

Our use case for this second version (org-scoped) mode would be the OAuth2 Application model. In AWX organization admins got permissions to this model. Implementation here would be that the model's permissions are listed in the organization admin roles. Optionally, someone could create custom organization-level roles that include OAuth2 Application permissions. But a user would not be able to create custom roles that give permission to _individual applications_, simply because we do not want to build a UI for this.

AlanCoding:
@relrod my only hangup for doing a fast 30-minute version of this is _where_ this data could live. Right now the `permission_registry` keeps some simple dictionaries and sets that were introduced for specific problems.

 - We could introduce another... `set` on `permission_registry` that lists model names that don't allow object roles? That would get the job done.
 - Another option to ninja this in would be that we could take the _model class itself_ and do something like `setattr(cls, '_allow_object_roles', True)`, and then use that flag on the objects themselves in the serializer.

Thinking it over, I'm okay with either approach.

AlanCoding:
This has a problem with the need for "creator" roles after someone creates a new object and does not have a role which would give implicit access to that object. Options for ways around that are:

1. For this type of model, require that the "add" permission is coupled with all other permissions. You can't have a role with "add" permission by itself.
2. Prohibit the "add" permission for models that have no object-level roles and use "change", "delete", or "view" permissions as proxy. This is complex and confusing.
3. Still allow system object-level roles but prevent custom object-level roles. But how would we communicate this to clients?

None of these options are slam dunks. If I were to choose one, I would lean towards (1). However that still leaves unanswered questions.

[...]

PR #434, opened 2024-06-04
Lock down OAuth2 views

relrod:
Doing this properly with RBAC depends on #424.

For now, we limit Application views to superusers. We limit Token
views using a custom DRF permission class based on what was specified
in AWX's access.py


ansible_base/oauth2_provider/migrations/0004_alter_oauth2accesstoken_scope.py
relrod, line 1: This is technically fixing an unrelated issue, I can split it out into another PR if needed.

ansible_base/oauth2_provider/views/token.py
```diff
@@ -51,7 +51,41 @@ def create_token_response(self, request):
             return request.build_absolute_uri(), {}, str(e), e.status_code
 
 
+class OAuth2TokenPermission(permissions.BasePermission):
```
relrod: @AlanCoding thoughts on this implementation?

ansible_base/oauth2_provider/migrations/0004_alter_oauth2accesstoken_scope.py
john-westcott-iv: Why remove the choices? Isn't read just a subset of write? i.e. doing `'read write'` is redundant? Or are they really separate things like in a unix directory where we would want to allow someone to write something but not be able to see it later? I suppose this would poise us if there was ever additional scopes possible.

ansible_base/oauth2_provider/models/access_token.py
```diff
@@ -60,7 +55,6 @@ class Meta(oauth2_models.AbstractAccessToken.Meta):
         blank=True,
         default='write',
         max_length=32,
-        choices=SCOPE_CHOICES,
```
john-westcott-iv, line 63: If we are going to pull the choices off here do we want a custom validator so that the model itself can't be saved with 'read dead redemption write'?

ansible_base/oauth2_provider/serializers/token.py
```diff
@@ -22,7 +22,7 @@
 class BaseOAuth2TokenSerializer(CommonModelSerializer):
     refresh_token = SerializerMethodField()
     token = SerializerMethodField()
-    ALLOWED_SCOPES = [x[0] for x in OAuth2AccessToken.SCOPE_CHOICES]
```
john-westcott-iv, line 25: I'm kind of hung up on this. Even if we remove the choice field it might be nice to have this defined in the model for a custom field validator and pull it in here.

ansible_base/oauth2_provider/views/token.py
```diff
@@ -51,7 +51,41 @@ def create_token_response(self, request):
...
+            if request.data.get('application'):
+                # TODO: Change this once dab#424 is fixed
+                return request.user.is_superuser
+        return request.user.is_authenticated
+
+    def has_object_permission(self, request, view, obj):
+        if request.method in ['GET', 'PATCH', 'PUT', 'DELETE']:
+            if request.user.is_superuser:
+                return True
+            if obj.application and obj.application.organization.admins.filter(id=request.user.id).exists():
```
john-westcott-iv: I think we are trying to nix the `users`/`admins` field off Organizations so we will likely need another way to get at his, probably through a RoleDefinition or something.

ansible_base/oauth2_provider/views/token.py
john-westcott-iv: How would you feel about have a "permissions" folder (or permissions.py) in the apps and then putting any custom permissions objects in those folders?

ansible_base/oauth2_provider/views/token.py
```diff
@@ -51,7 +51,41 @@ def create_token_response(self, request):
...
+    def has_permission(self, request, view):
+        # Handle PAT and app token creation separately
+        if request.method == 'POST':
+            if request.data.get('application'):
+                # TODO: Change this once dab#424 is fixed
+                return request.user.is_superuser
+        return request.user.is_authenticated
+
+    def has_object_permission(self, request, view, obj):
+        if request.method in ['GET', 'PATCH', 'PUT', 'DELETE']:
```
john-westcott-iv: Do we need to factor in a `GET` request from an auditor?

ansible_base/oauth2_provider/views/token.py
john-westcott-iv: Should we move any custom permissions for apps into a "permissions" folder or (or permissions.py)?

ansible_base/oauth2_provider/views/token.py
john-westcott-iv: Do we need to consider a `GET` from an auditor?

ansible_base/oauth2_provider/views/token.py
john-westcott-iv: I think we are working on nixing the `users/admins` fields off the Org/Team roles so this will likely need to change to a RoleDefinition or something like that.

ansible_base/oauth2_provider/models/access_token.py
relrod: Yeah, we can add a validator on the field in the model.

ansible_base/oauth2_provider/migrations/0004_alter_oauth2accesstoken_scope.py
relrod: > Why remove the choices?

It's because in the help text we say "space separated" but then we don't actually allow that. And also you might want a token scoped for multiple scopes in the future, so it seems really odd to limit it to just one. In DRF what we had before my change corresponded to just a dropdown where you could only pick one, that was a bug.

ansible_base/oauth2_provider/views/token.py
relrod: I can move it into permissions.py, sure

ansible_base/oauth2_provider/views/permissions.py
```diff
@@ -0,0 +1,35 @@
...
+    #   - I am the user of the token
+    # - A PAT can be created if:
+    #   - I am a user
+
+    def has_permission(self, request, view):
+        # Handle PAT and app token creation separately
+        if request.method == 'POST':
+            if request.data.get('application'):
+                # TODO: Change this once dab#424 is fixed
+                return request.user.is_superuser
```
AlanCoding, line 26: :+1: bugs processing `request.data` are a nagging concern of mine, but this is simple enough that it'll be perfectly fine for now.

ansible_base/oauth2_provider/views/permissions.py
```diff
@@ -0,0 +1,35 @@
...
+        return request.user.is_authenticated
+
+    def has_object_permission(self, request, view, obj):
+        if request.method in ['GET', 'PATCH', 'PUT', 'DELETE']:
+            if request.user.is_superuser:
+                return True
+            if obj.application and obj.application.organization.admins.filter(id=request.user.id).exists():
```
AlanCoding: Isn't this in-congruent with the creation permission (`has_permission`)?

If we require permission to the related application of a token (for the type of tokens that have an application) then creation would require organization admin, and change permission would require organization admin.

If we can't support that easily for creation, I think it might make more sense to leave both create/edit at superuser level (for type of tokens that have an application) and leave the full implementation as a TODO. Otherwise a half-way solution can cause even more confusion.

AlanCoding (changes requested):
Seeing the test `xfail` the near-term plan is starting to come together. I'm okay if we say that right now organization admins can't create application tokens, although that is not our intent. It looks like the things that need to happen are
 - replace the `organization.admins` reference, which I think you're aware of
 - squash the 2 migration files into 1

ansible_base/oauth2_provider/views/permissions.py
AlanCoding: This is knowingly not complete, as per test `xfail`, so resolving this for now. I will not block merge on org admins not being able to create tokens, and we can leave it this way.

ansible_base/oauth2_provider/views/token.py
john-westcott-iv: Here is what I did in my JWT test:
```
RoleDefinition.objects.managed.org_member.give_permission(user, organization)
```

ansible_base/oauth2_provider/views/permissions.py
```diff
@@ -0,0 +1,37 @@
...
+    def has_permission(self, request, view):
+        # Handle PAT and app token creation separately
+        if request.method == 'POST':
+            if request.data.get('application'):
+                # TODO: Change this once ansible/django-ansible-base#424 is fixed
+                return request.user.is_superuser
+        return request.user.is_authenticated
+
+    def has_object_permission(self, request, view, obj):
+        if request.method in SAFE_METHODS and getattr(request.user, 'is_system_auditor', False):
```
AlanCoding, line 30: I really don't want to keep a is_system_auditor check long term, but I guess that should be replaced as soon as we're able to add the model to the registry.

AlanCoding (approved):

john-westcott-iv (approved):

PR #435, opened 2024-06-05
Vendoring system for shared role definitions

### LINKED ISSUE #408: Introduce a managed RoleDefinition "registry"
This proposes to copy many patterns from the `CredentialType` model in AWX.

https://github.com/ansible/awx/blob/devel/awx/main/models/credential/__init__.py

That has entries like:

```python
ManagedCredentialType(
    namespace='registry',
    kind='registry',
    name=gettext_noop('Container Registry'),
    inputs={
        'fields': [
            {
                'id': 'host',
                'label': gettext_noop('Authentication URL'),
                'type': 'string',
                'help_text': gettext_noop('Authentication endpoint for the container registry.'),
                'default': 'quay.io',
            },
            {
                'id': 'username',
                'label': gettext_noop('Username'),
                'type': 'string',
            },
            {
                'id': 'password',
                'label': gettext_noop('Password or Token'),
                'type': 'string',
                'secret': True,
                'help_text': gettext_noop('A password or token used to authenticate with'),
            },
            {
                'id': 'verify_ssl',
                'label': gettext_noop('Verify SSL'),
                'type': 'boolean',
                'default': True,
            },
        ],
        'required': ['host'],
    },
)
```

So this proposes to create a similar in-memory `ManagedRoleDefinition` class.

### Registry

Instances of `ManagedRoleDefinition`s will be added to a registry. You will be able to get them back out of the registry using a key similar to the `namespace` key from the credential type example. TBD: it's still unclear if we want a globally unique namespace, or a namespace unique within a given content type.

### Enablement

The registry will have to be filled in by the app using DAB RBAC, because there is no way we can predict what managed roles the app using DAB RBAC will want. So this proposes that DAB hosts a "courtesy" registry. The app will then be able to copy things from the courtesy registry to its real registry.

### Function

Each `ManagedRoleDefinition` will, obviously, have the full data required to make a `RoleDefinition` which will, obviously, set `managed=True`.

It may contain a translated string for some fields, in which case the non-translated version will be saved to the DB, but (similar to credential type), the serializer may present a translated version of the field's data.

When will the `RoleDefinition` be created? I would like to give the app some autonomy to decide this. However, DAB must provide a method that will create all items in the registry to the database, and this method will take `apps` so that it can be used safely in migrations.

We may also introduce a method on the `RoleDefinition` model that gets a managed role definition given the namespace (or whatever we use for that), which may create the role definition if it doesn't exist.

AlanCoding:
Fixes https://github.com/ansible/django-ansible-base/issues/408

As of creating this, I still have 1 failure, but it's meaningful as a type of pre-existing issue so I'm happy to see it.

Opening as a draft, it's still fairly early in development of this, but I have a framework that works for me.

TODO:
 - customization in test_app
   - ~add a managed role for "Cow Admin", shortname "cow_admin" to test an app-defined thing~
   - ~duplicate some of the managed roles, because this is part of the intent of the system~
   - ~add tests for creating managed roles in migrations (will not use in AWX or test_app)~

post-merge TODO:
   - Various downstream work in AWX because this replaces code that exists there (only partially)
   - ...more I forget...

This is the replacement for removals in https://github.com/ansible/django-ansible-base/pull/296, so some kind of merge order solution will be developed for the two of these.


ansible_base/rbac/managed.py
```diff
@@ -0,0 +1,145 @@
...
+        from ansible_base.rbac.validators import combine_values, permissions_allowed_for_role
+
+        return combine_values(permissions_allowed_for_role(self.get_model()))
+
+
+class OrganizationMixin:
+    def get_model(self):
+        from ansible_base.lib.utils.auth import get_organization_model
+
+        return get_organization_model()
```
AlanCoding: @john-westcott-iv I wanted to use the shared method for doing this, but after getting further in development it's just not going to work. I want to further customize this _in settings_ so no imports or python objects. It makes the most sense to use the setting and have the logic here do `get_model`. Also we can't import from utils at top level. I'll leave a comment about this tension.

ansible_base/rbac/managed.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class OrganizationAdmin(OrganizationMixin, ManagedAdminBase):
+    name = gettext_noop("Organization Admin")
+    description = gettext_noop("Has all permissions to a single organization and all objects inside of it")
+
+
+class OrganizationMember(OrganizationMixin, ManagedActionBase):
+    name = gettext_noop("Organization Member")
+    description = gettext_noop("Has all permissions given to a single team")
```
slemrmartin: typo: `..single organization`
also, member has only view permissions, not all, correct?

ansible_base/rbac/managed.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class TeamAdmin(TeamMixin, ManagedAdminBase):
+    name = gettext_noop("Team Admin")
+    description = gettext_noop("Can manage a single team and has all permissions given the team")
+
+
+class TeamMember(TeamMixin, ManagedActionBase):
+    name = gettext_noop("Team Member")
+    description = gettext_noop("Has all permissions given to a single team")
```
slemrmartin: also here, doesn't have all permissions (can't change)

ansible_base/rbac/models.py
```diff
@@ -58,23 +58,28 @@ def __get__(self, obj, objtype=None):
...
         self._cache = {}
 
-    org_admin = ManagedRoleFromSetting('Organization Admin')
-    org_member = ManagedRoleFromSetting('Organization Member')
-    team_admin = ManagedRoleFromSetting('Team Admin')
-    team_member = ManagedRoleFromSetting('Team Member')
+    def __getattr__(self, attr):
+        if attr in self._cache:
+            return self._cache[attr]
+        code_definition = permission_registry._managed_roles.get(attr)
```
slemrmartin: it `_managed_roles` is accessed from outside, is could be "public", thus `managed_roles`

ansible_base/rbac/managed.py
john-westcott-iv: Yea, I've been thinking about `uitls` lately and wondering if we should take anything with a model import out so that utils would /always/ be safe to import.

slemrmartin (approved):
It looks good, comments on cosmetic things only => approving

ansible_base/rbac/managed.py
AlanCoding: Documented https://github.com/ansible/django-ansible-base/issues/443

ansible_base/rbac/managed.py
AlanCoding: User has all permissions assigned to the team via the role_team_assignments. I'll reword slightly.

ansible_base/rbac/models.py
AlanCoding: I made a new method to reference publicly, there's an existing pattern for `_` internal data in the permission registry. I think this should address this issue.

ansible_base/rbac/managed.py
AlanCoding: Members can view all other members of an organization, so I think that's possibly the most consequential permission that comes as a part of the "member" permission. But my revisions will leave this fairly ambiguous.

PR #437, opened 2024-06-06
Fix issues with saving secrets starting with $encrypted$

trucdg:
Description:

-  Encrypted strings are prefixed with $encrypted$ but a user can pass a value through the API like $encrypted$something. This then messes everything up.

Changes made:
- Implemented a helper function to identify encrypted strings prefixed with "$encrypted$".


ansible_base/lib/utils/encryption.py
```diff
@@ -43,13 +43,33 @@ def __init__(self):
...
         self._backend = default_backend()
 
+    def is_encrypted_string(self, value:str) -> bool:
+        # Ensure input is a string already encrypted by our algorithm
+        encrypt_sequence = value.split('$')
+
+        if len(encrypt_sequence) != 5:
+            return False
+        
+        _, flag, encoding, algo, b64data = value.split('$')
```
john-westcott-iv: This is going to assume that `ENCRYPTED_STRING` is always going to be `$encrypted$`; which it really should be but you never know when someone might be like "we need to change this". If we change this up a bit, I think we can protect ourselves incase that day ever comes. My inclination would be to move some of the starting code from `decrypt_string` into here:
```
        # This is needed because we don't enforce type hints at runtime, only in tests
        if type(value) is not str:
            raise ValueError("decrypt_string can only accept string")

        if not value.startswith(ENCRYPTED_STRING):
            return False
        
        encoding, algo, b64data = value[len(ENCRYPTED_STRING) :].split('$')
```

Then we can have `decrypt_string` do something like:
```
    def decrypt_string(self, value: str) -> str:
        if not self.is_encrypted_string(value):
              return value
        .... do the rest of the decryption ....
```

ansible_base/lib/utils/encryption.py
```diff
@@ -43,13 +43,33 @@ def __init__(self):
...
+        
+        if (flag != 'encrypted' or encoding != 'UTF8' or algo == ENCRYPTION_METHOD):
+            return False
```
john-westcott-iv: Also note that in the decrypt method the `UTF-8$` marker may or may not be in the encrypted string.

ansible_base/lib/utils/encryption.py
```diff
@@ -43,13 +43,33 @@ def __init__(self):
...
+
+        s = smart_str(self.decrypt(base64.b64decode(b64data)))
+        encrypted_data = self.encrypt(smart_bytes(s))
+        b64data = smart_str(base64.b64encode(encrypted_data))
+        encrypted_s = f'{ENCRYPTED_STRING}UTF8${ENCRYPTION_METHOD}${b64data}'
```
john-westcott-iv: What happens if any of these fail? i.e. If I gave you a string that was like `$encrypted$UTF8$something$junk` we would try and b64decode `junk` which would likely raise an exception.

ansible_base/lib/utils/encryption.py
```diff
@@ -43,13 +43,33 @@ def __init__(self):
...
+
+        return encrypted_s == value
+
     def encrypt_string(self, value: str) -> str:
         # Its possible for a serializer to accept a number for a CharField (like 5). In the serializer its "5" but when we get here it might be 5
         if type(value) is not str:
             value = str(value)
 
-        if value.startswith(ENCRYPTED_STRING):
+        if value.startswith(ENCRYPTED_STRING) and self.is_encrypted_string(value):
```
john-westcott-iv: I think here we could just say `if self.is_encrytped_string(value)`

ansible_base/lib/utils/encryption.py
trucdg: thank you for pointing this out.
At first I thought the string deconstruction part in decrypt_string() was lengthy. However, with your insight, I now understand the rationale behind each step

ansible_base/lib/utils/encryption.py
```diff
@@ -37,34 +38,62 @@ def __init__(self):
         self._encryption_key = self.key[32:]
         self._backend = default_backend()
 
+    def extract_data(self, raw_data: str) -> str:
+        # Sequential data extract for code reuse in is_encrypted_string and decrypt_string
+        raw_data = raw_data[len(ENCRYPTED_STRING) :]
```
john-westcott-iv: I noticed that we are not doing the `if raw_data.startswith(ENCRYPTED_STRING)` here. Which is ok for how you have the function setup but would not be ok if we called this directly and the string was missing it. I think we should add that here to protect someone from calling this incorrectly. What are your thoughts?

ansible_base/lib/utils/encryption.py
```diff
@@ -37,34 +38,62 @@ def __init__(self):
...
+        if not value.startswith(ENCRYPTED_STRING):
+            return False
+
+        try:
+            data = self.extract_data(value)
+        except ValueError as ve:
+            # value error is raised when algorithm is unsupported
+            if invalid_algo_is_fatal:
+                raise ve
+
```
john-westcott-iv: Don't we want to return `False` here if we got a ValueError but `invalid_algo_is_fatal` is `False`? If so, can you write a test case where we detect that not returning causes an error so we can test it before and after the change?

ansible_base/lib/utils/encryption.py
```diff
@@ -37,34 +38,62 @@ def __init__(self):
...
+        # Check if data is base64 encoded
+        # **important** Note: this can fail because some strings resemble base 64 encoded ones
+        # **important** for instance, if user's secret is $encrypted$UTF8$AESCBC$junk, the code will break
+        # use regex pattern: ^([A-Za-z0-9+/]{4})* means the string starts with 0 or more base64 groups.
+        # ([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)$ means
+        # the string ends in one of three forms: [A-Za-z0-9+/]{4}, [A-Za-z0-9+/]{3}= or [A-Za-z0-9+/]{2}==.
+        reg = r'^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$'
```
john-westcott-iv: `junk` [matches your regular expression](https://pythex.org/?regex=(%5BA-Za-z0-9%2B%2F%5D%7B4%7D)*(%5BA-Za-z0-9%2B%2F%5D%7B3%7D%3D%7C%5BA-Za-z0-9%2B%2F%5D%7B2%7D%3D%3D)%3F%24&test_string=junk&ignorecase=0&multiline=0&dotall=0&verbose=0). 

I like where you are going with this though, if we are able to introducing a regex for testing data can we just regex this whole method?
`^\$encrypted\$(UTF8\$)?AESCBC\$<your b64 check>$` (obviously we would want to put variables in there instead of hard coding.

john-westcott-iv:
@trucdg now that you have provided us with a very eloquent way to see if a string is (most likely) encrypted I would like to ask you to extend this PR. 

Per the [encryption docs](https://github.com/ansible/django-ansible-base/blob/devel/docs/lib/encryption.md), we say roughly, import ENCRYPTION_STRING and if a string.startswith(ENCRYPTION_STRING) you can call decrypt_string. However, this new code will handle getting an already decrypted string.

1. Can we fix the docs.
2. Can we find areas in the code like this particular [example](https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/authentication/models/authenticator.py#L58-L59) and clean them up?
```python
from ansible_base.lib.utils.encryption import ENCRYPTED_STRING

for field in getattr(authenticator, 'configuration_encrypted_fields', []):
    if field in instance.configuration and instance.configuration[field].startswith(ENCRYPTED_STRING):
        instance.configuration[field] = ansible_encryption.decrypt_string(instance.configuration[field])
```

trucdg:
@john-westcott-iv  I think this is all set :-?

jctanner (approved):

test_app/tests/lib/utils/test_encryption.py
```diff
@@ -43,3 +46,56 @@ def test_fernet256_unsupported_algorithm():
...
+        ('$encrypted$UTF8$AESCBC$junk', (True, 'AESCBC', 'junk')),
+        ('$encrypted$UTF8$AESCBC$junk12==', (True, 'AESCBC', 'junk12==')),
+        ('$encrypted$UTF8$AESCBC$junk123=', (True, 'AESCBC', 'junk123=')),
+        ('$encrypted$UTF8$AESCBC$junk123=asdf', (False, None, None)),
+        ('$encrypted$UTF8$AESCBC$junk123a', (True, 'AESCBC', 'junk123a')),
+        (smart_str(base64.b64encode(b"This is a test")), (False, None, None)),
+        (f'$encrypted$UTF8$AESCBC${smart_str(base64.b64encode(b"This is a test"))}', (True, 'AESCBC', smart_str(base64.b64encode(b"This is a test")))),
+        (f'$encrypted$AESCBC${smart_str(base64.b64encode(b"This is a test"))}', (True, 'AESCBC', smart_str(base64.b64encode(b"This is a test")))),
+        (
+            f'$encrypted$some_other_algo${smart_str(base64.b64encode(b"This is a test"))}',
```
bhavenst, line 73: This is an awesome set of tests.  Perhaps for the valid data versions, the string could be varied such that it produces base64 with no padding, single =, and double ==?

bhavenst (approved):
Looks great!

test_app/tests/lib/utils/test_encryption.py
trucdg: yes, we have tests for those which are `junk`, `junk123=`, and `junk12==`

bhavenst: I meant the opposite actually, "This is a test" -> VGhpcyBpcyBhIHRlc3Q= but one could be changed to "This is a test." -> VGhpcyBpcyBhIHRlc3QuCg== and "This is not a test" -> VGhpcyBpcyBub3QgYSB0ZXN0

The difference is just decrypting valid base64 that has all the possible endings.  But no big deal of course, just a small suggestion to round it out.

trucdg: ahh, "junk" is actually a valid base64 ending. we can do base64.b64decode(b'junk')

Issue #443, opened 2024-06-07
Remove model imports in any utils/ modules


AlanCoding:
To prevent circular imports, we want ansible_base/utils imports to always be "safe" to import inside of Django apps.

Right now this doesn't work due to a stray import of `AbstractUser` which needs apps to be loaded before importing (or will throw an error).

This can be done by moving these imports in-line, or more elegantly, but further code refactoring.

Kuppjaerk:
Hi!

Is this issue still awaiting resolution? If so, I'd like to take a look at it. I'm a uni student who's currently taking an open source course, which includes making contributions.

AlanCoding:
Yes, still unresolved. The import in question is:

https://github.com/ansible/django-ansible-base/blob/5f1ca896a4dc94a4ed7f1da0dc371ddc4217669f/ansible_base/lib/utils/models.py#L9

In many cases, `Model`, would be an acceptable replacement, which does not require apps to be loaded to be able to import.

A solution can be verified by using the `python3` terminal, without doing any other setup before that.

```python
>>> from ansible_base.lib.utils import *
>>> from ansible_base.lib.utils.models import *
```

The first line is successful, but the 2nd line errors. Make that not error, and this is solved.

Kuppjaerk:
Alright, I'll start working on this then, thanks for the quick reply! I'm quite new to contributing to ansible, is there a certain Matrix room I could ask questions regarding django-ansible-base?

AlanCoding:
This is not fixed. Current example as a traceback:

```
Traceback (most recent call last):
  File "/home/arominge/venvs/awx/bin/pytest", line 8, in <module>
    sys.exit(console_main())
             ^^^^^^^^^^^^^^
  File "/home/arominge/venvs/awx/lib64/python3.11/site-packages/_pytest/config/__init__.py", line 201, in console_main
    code = main()
           ^^^^^^
  File "/home/arominge/venvs/awx/lib64/python3.11/site-packages/_pytest/config/__init__.py", line 156, in main
    config = _prepareconfig(args, plugins)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arominge/venvs/awx/lib64/python3.11/site-packages/_pytest/config/__init__.py", line 342, in _prepareconfig
    config = pluginmanager.hook.pytest_cmdline_parse(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #444, opened 2024-06-07
model import fix - #443

darshannv:
Fixes - Model imports that was giving an error - #443

### DIFF
--- a/ansible_base/lib/utils/models.py
+++ b/ansible_base/lib/utils/models.py
@@ -7,7 +7,6 @@ from crum import get_current_user
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AbstractUser
-from django.db import models
 from django.utils.translation import gettext_lazy as _
 from inflection import underscore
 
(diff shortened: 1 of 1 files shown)


ansible_base/lib/utils/models.py
```diff
@@ -7,7 +7,6 @@
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AbstractUser
-from django.db import models
```
AlanCoding, line 10: Checks are failing because this is used in type hints.

And also, I don't believe this is a problem, here's how you test: just run `python` by itself (without DJANGO_SETTINGS_MODULE set) and in the interpreter try these imports. The import `from django.db import models` works, but the import `from django.contrib.auth.models import AbstractUser` throws an error.

john-westcott-iv:
@darshannv Thanks so much for your contribution! @AlanCoding left some comments, please let us know what your thoughts are and if you will be able to address them. If we don't hear back from your in a couple weeks we will close this request but it can always be reopened later.

john-westcott-iv:
Closing due to inactivity, please reopen or reach out to us if you would like to continue this work in the future.

PR #447, opened 2024-06-10
resource_registry: sync organization and team descriptions across services

BrennanPaciorek:
Fixes: AAP-23755

### DIFF
--- a/ansible_base/resource_registry/shared_types.py
+++ b/ansible_base/resource_registry/shared_types.py
@@ -59,8 +59,7 @@ class OrganizationType(SharedResourceTypeSerializer):
 
     name = serializers.CharField()
     description = serializers.CharField(
-        default="",
-        allow_blank=True,
+        required=False,
     )
 
 
@@ -69,3 +68,6 @@ class TeamType(SharedResourceTypeSerializer):
 
     name = serializers.CharField()
     organization = AnsibleResourceForeignKeyField("shared.organization", required=False, allow_null=True)
+    description = serializers.CharField(
+        required=False,
+    )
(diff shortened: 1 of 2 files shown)


ansible_base/resource_registry/shared_types.py
```diff
@@ -58,6 +58,10 @@ class OrganizationType(SharedResourceTypeSerializer):
     RESOURCE_TYPE = "organization"
 
     name = serializers.CharField()
+    description = serializers.CharField(
+        default="",
+        allow_blank=True,
+    )
```
john-westcott-iv, line 64: Do we also need this field on the TeamType?

BrennanPaciorek: Not for the purposes of addressing the issue, but I don't think it would hurt to sync description for Team models as well. What are your thoughts on this as the epic assignee @rochacbruno?

rochacbruno: Hi @BrennanPaciorek @john-westcott-iv The issue mentions only org description, but we can go ahead and add team descriptions.

BrennanPaciorek:
To make the failing test pass, it seems like we'd need to change AWX's test suite.

https://github.com/ansible/awx/blob/4738c8333a917ef09fa85ff2b50b9c9c03adaf4f/awx/main/tests/functional/api/test_immutablesharedfields.py#L33

BrennanPaciorek:
> To make the failing test pass, it seems like we'd need to change AWX's test suite.
> 
> https://github.com/ansible/awx/blob/4738c8333a917ef09fa85ff2b50b9c9c03adaf4f/awx/main/tests/functional/api/test_immutablesharedfields.py#L33

@rochacbruno @john-westcott-iv. I am not certain on the above, can anyone confirm if this is the case?

john-westcott-iv (approved):

john-westcott-iv:
We should notify the AWX team of the failure as it will likely affect them in the long run.
@TheRealHaoLiu @chrismeyersfsu can one of you all take a look at the failing AWX test and let us know if we are cleared to merge this or if you want to have us wait while you preemptively clean up the AWX test?

PR #448, opened 2024-06-10
Add handling for jwt enc OIDC user info

bhavenst:
For blocker bug related to failure to decode jwt encoded OIDC user info response.  This attempts to decode it using JWT specified algorithms and public key.

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/oidc.py
+++ b/ansible_base/authentication/authenticator_plugins/oidc.py
@@ -253,6 +253,7 @@ class AuthenticatorPlugin(SocialAuthMixin, OpenIdConnectAuth, AbstractAuthentica
         """
         user_data = self.request(self.userinfo_url(), headers={"Authorization": f"Bearer {access_token}"})
         if user_data.headers["Content-Type"] == "application/jwt":
+            # If the content type is application/jwt than we can assume that the token is encrypted. Otherwise it should be application/json
             pubkey = self.public_key()
             if not pubkey:
                 logger.error(_("OIDC client sent encrypted user info response, but no public key found."))
--- a/ansible_base/authentication/authenticator_plugins/oidc.py
+++ b/ansible_base/authentication/authenticator_plugins/oidc.py
@@ -1,9 +1,12 @@
+import http
 import logging
 
+import jwt
 from django.core.exceptions import ValidationError
 from django.core.validators import MinValueValidator
 from django.utils.translation import gettext_lazy as _
 from jwt.algorithms import get_default_algorithms
+from jwt.exceptions import PyJWTError
 from social_core.backends.open_id_connect import OpenIdConnectAuth
 
 from ansible_base.authentication.authenticator_plugins.base import AbstractAuthenticatorPlugin, BaseAuthenticatorConfiguration
@@ -221,3 +224,28 @@ class AuthenticatorPlugin(SocialAuthMixin, OpenIdConnectAuth, AbstractAuthentica
         # This is a copy of super without caching, which avoids data
         # from one OIDC based authenticator showing up in another
         return self.get_json(self.oidc_endpoint() + "/.well-known/openid-configuration")
+
+    def public_key(self):
+        return "\n".join(
+            [
+                "-----BEGIN PUBLIC KEY-----",
+                self.setting("PUBLIC_KEY"),
+                "-----END PUBLIC KEY-----",
+            ]
+        )
+
+    def user_data(self, access_token, *args, **kwargs):
+        user_data = self.request(self.userinfo_url(), headers={"Authorization": f"Bearer {access_token}"})
+        if user_data.headers["Content-Type"] == "application/jwt":
+            try:
+                data = jwt.decode(
+                    access_token,
+                    key=self.public_key(),
+                    algorithms=self.setting("JWT_ALGORITHMS"),
+                    audience=self.setting("KEY"),
+                )
+                return data
+            except PyJWTError as e:
+                logger.error(_(f"Unable to decode user info response JWT: {e}"))
+                return None
+        return user_data.json()
--- a/ansible_base/authentication/authenticator_plugins/oidc.py
+++ b/ansible_base/authentication/authenticator_plugins/oidc.py
@@ -225,17 +225,22 @@ class AuthenticatorPlugin(SocialAuthMixin, OpenIdConnectAuth, AbstractAuthentica
         return self.get_json(self.oidc_endpoint() + "/.well-known/openid-configuration")
 
     def public_key(self):
+        key = self.setting("PUBLIC_KEY")
         return "\n".join(
             [
                 "-----BEGIN PUBLIC KEY-----",
-                self.setting("PUBLIC_KEY"),
+                key,
                 "-----END PUBLIC KEY-----",
             ]
-        )
+        ) if key else None
 
     def user_data(self, access_token, *args, **kwargs):
         user_data = self.request(self.userinfo_url(), headers={"Authorization": f"Bearer {access_token}"})
         if user_data.headers["Content-Type"] == "application/jwt":
+            pk = self.public_key()
+            if not pk:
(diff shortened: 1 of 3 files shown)


ansible_base/authentication/authenticator_plugins/oidc.py
```diff
@@ -215,3 +217,37 @@ def extra_data(self, user, backend, response, *args, **kwargs):
...
+    def public_key(self):
+        key = self.setting("PUBLIC_KEY")
+        return (
+            "\n".join(
+                [
+                    "-----BEGIN PUBLIC KEY-----",
+                    key,
+                    "-----END PUBLIC KEY-----",
+                ]
+            )
```
john-westcott-iv: Is this guaranteed to not be there? If not should we add a test to see if its there before we just add it?

ansible_base/authentication/authenticator_plugins/oidc.py
```diff
@@ -215,3 +217,37 @@ def extra_data(self, user, backend, response, *args, **kwargs):
...
+            if key
+            else None
+        )
+
+    def user_data(self, access_token, *args, **kwargs):
+        user_data = self.request(self.userinfo_url(), headers={"Authorization": f"Bearer {access_token}"})
+        if user_data.headers["Content-Type"] == "application/jwt":
+            pk = self.public_key()
+            if not pk:
```
john-westcott-iv: Can we change the `pk` variable name since that is so commonly referenced as primary key?

ansible_base/authentication/authenticator_plugins/oidc.py
```diff
@@ -215,3 +217,37 @@ def extra_data(self, user, backend, response, *args, **kwargs):
...
+                logger.error(_("OIDC client sent encrypted user info response, but no public key found."))
```
john-westcott-iv, line 259: Just to confirm, there is no way for an OIDC provider to return an unencrypted token right? Knowing that would be insecure but sometimes people do unexpected things. I think the OIDC specification does state that it has to be encrypted though.

ansible_base/authentication/authenticator_plugins/oidc.py
```diff
@@ -215,3 +217,37 @@ def extra_data(self, user, backend, response, *args, **kwargs):
...
+                    "-----BEGIN PUBLIC KEY-----",
+                    key,
+                    "-----END PUBLIC KEY-----",
+                ]
+            )
+            if key
+            else None
+        )
+
+    def user_data(self, access_token, *args, **kwargs):
```
john-westcott-iv, line 245: Can we add a comment summarizing what we are doing here? (feel free to correct this if needed)
```suggestion
    def user_data(self, access_token, *args, **kwargs):
       """
       This overrides the the user_data function provided by the social auth class OpenIdConnectAuth
       The function in that class assumes the request cming back from userinfo_url is not an encrypted token.
       We want to assume that if we get back a Content-Type of application/jwt that we have an encrypted token.
       So we will decrypt the token and return that data or we will raise a PyJWTError that we can't decrypt the token
       If we don't get that Content-Type back, we will just return whatever data was given to us just like the parent method currently does. 
       """
```

ansible_base/authentication/authenticator_plugins/oidc.py
```diff
@@ -215,3 +217,37 @@ def extra_data(self, user, backend, response, *args, **kwargs):
...
+                return None
+            try:
+                data = jwt.decode(
+                    access_token,
+                    key=self.public_key(),
+                    algorithms=self.setting("JWT_ALGORITHMS"),
+                    audience=self.setting("KEY"),
+                )
+                return data
+            except PyJWTError as e:
```
john-westcott-iv, line 269: If we can't decrypt for some reason here should we just attempt to return `user_data.json()` just like the original method would?

john-westcott-iv (commented):
I'm kind of surprised this isn't already a feature in social_auth somewhere already.

ansible_base/authentication/authenticator_plugins/oidc.py
bhavenst: Yeah sure, thanks.

ansible_base/authentication/authenticator_plugins/oidc.py
bhavenst: sure!

ansible_base/authentication/authenticator_plugins/oidc.py
bhavenst: Ah, you mean that the header and footer might already be present?  Yeah sure, it just depends how it was pasted into the configuration.  I will add that.  I copied it from the keycloak version, which does even less checking. :)  https://github.com/python-social-auth/social-core/blob/d7bba223c0036581b63b01d05e53b115c606dbec/social_core/backends/keycloak.py#L115

ansible_base/authentication/authenticator_plugins/oidc.py
bhavenst: As far as I have seen, it is more common that it is not encrypted and/or signed.  Even in RH SSO I believe it comes back unencrypted.  So yeah, it can and usually does come back unencrypted for sure.

The standard is here and 5.3.2 just says that the server may encrypt/sign it.

https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

But yeah, it is strange that social auth doesn't support that, especially considering that their own keycloak implementation does.  Keycloak itself though by default does not encrypt/sign that response and you have to go into the advanced settings to enable it..

James Tanner added the PUBLIC_KEY to the dab oidc config but it wasn't doing anything yet ("The public key from your IDP. Only necessary if using keycloak for OIDC.").  It doesn't come from social auth and would be needed if they were to do this jwt unpacking.

ansible_base/authentication/authenticator_plugins/oidc.py
bhavenst: That will just throw the original json decode failed at char 1 exception, but fine with me if that is preferred.  That one will not trigger the exception handling middleware that would redirect back to login w/ a message (assuming that gets implemented) so that's why I returned the None.  But it could also be wrapped in one that would trigger the handler.  Or just raise AuthException here directly.

ansible_base/authentication/authenticator_plugins/oidc.py
john-westcott-iv: So what happens in this code if the JWT is decrypted? Or would this code just not get called if its not encrypted?

ansible_base/authentication/authenticator_plugins/oidc.py
john-westcott-iv: I would be fine with either so long as a this code can still handle a not encrypted token.

ansible_base/authentication/authenticator_plugins/oidc.py
bhavenst: Yeah, they have application/json content type rather than jwt type, so they just get returned by the last line.

ansible_base/authentication/authenticator_plugins/oidc.py
```diff
@@ -215,3 +217,49 @@ def extra_data(self, user, backend, response, *args, **kwargs):
...
+        """
+        This function overrides the one in social auth class OpenIdConnectAuth, since
+        it assumes the user info endpoint response returns plain json. Depending on
+        server config it may instead be a JWT. This function notices the JWT via the
+        content type header and if found, attempts to decode it using the configured
+        public key and algorithm(s). None is returned to signify a failed login in the
+        case of improper config or other decoding failure.
+        """
+        user_data = self.request(self.userinfo_url(), headers={"Authorization": f"Bearer {access_token}"})
+        if user_data.headers["Content-Type"] == "application/jwt":
```
john-westcott-iv, line 255:
```suggestion
        if user_data.headers["Content-Type"] == "application/jwt":
            # If the content type is application/jwt than we can assume that the token is encrypted. Otherwise it should be application/json
```

john-westcott-iv (approved):

PR #458, opened 2024-06-12
Avoid import from resource_registry at top of file

AlanCoding:
I saw an issue about this, and since I had just done a similar solution for DAB RBAC I wanted to just go in and try to make it look more standard.

For the root problem, I might need some help understanding how galaxy_ng feature flags work.

https://github.com/ansible/galaxy_ng/blob/2f19483e1a7564afece88557833119969e255857/galaxy_ng/app/dynaconf_hooks.py#L67

This linked code sure strongly suggests that, in some cases, the resource_registry app might not be installed. If that's the case, then this would likely be the right kind of fix.

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -40,14 +40,6 @@ class JWTCommonAuth:
         if hasattr(settings, 'ANSIBLE_BASE_ORGANIZATION_MODEL'):
             self.org_content_type = ContentType.objects.get_for_model(apps.get_model(settings.ANSIBLE_BASE_ORGANIZATION_MODEL))
 
-    @property
-    def resource_cls(self):
-        return apps.get_model('dab_resource_registry', 'Resource')
-
-    @property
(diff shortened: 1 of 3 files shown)


AlanCoding:
I tried to do this as a `@property` but the Hub class doesn't directly subclass, so that's not so obvious, there's not really a big benefit from code reuse. So I set it back to the original (fairly dumb) method.

john-westcott-iv (approved):

john-westcott-iv:
If we are not in a rush for this it would be nice if we can up the coverage just a bit.

AlanCoding:
I tried. I wound up adding some tests, but they're much more marginal.

Looking at it for a few minutes, I realized that there is no way to easily solve the problem of joint testing with galaxy_ng. That is a big ball of yarn that I will not be able to help with efficiently right now. I already had some similar tests, but this gets more of the new method interface directly. It's something.

AlanCoding:
This will probably become defunct with https://github.com/ansible/django-ansible-base/pull/477 changes

PR #459, opened 2024-06-12
Rename is_system_auditor to is_platform_auditor

relrod:

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/oidc.py
+++ b/ansible_base/authentication/authenticator_plugins/oidc.py
@@ -207,7 +207,7 @@ class AuthenticatorPlugin(SocialAuthMixin, OpenIdConnectAuth, AbstractAuthentica
     configuration_encrypted_fields = ['SECRET']
 
     def extra_data(self, user, backend, response, *args, **kwargs):
-        for perm in ["is_superuser", "is_system_auditor"]:
+        for perm in ["is_superuser", "is_platform_auditor"]:
             if perm in response:
                 kwargs["social"].extra_data[perm] = response[perm]
         data = super().extra_data(user, backend, response, *args, **kwargs)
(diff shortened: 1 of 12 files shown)


john-westcott-iv (commented):
My only concern with this is the authenticators. i.e. if I'm an awx customer and I already have the is_system_auditor flag defined in my saml/oidc will this cause the customer to have to change something in their provider for the functionality to work. Keycloak is a net new authenticator so I am fine with that one changing.

AlanCoding:
Can we turn it into a setting? Then moving forward @john-westcott-iv, can AWX continue to respect the current field, and apps can change for their own expectations.

ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -139,6 +139,8 @@
     SOCIAL_AUTH_STRATEGY = "ansible_base.authentication.social_auth.AuthenticatorStrategy"
     SOCIAL_AUTH_LOGIN_REDIRECT_URL = "/"
 
+    ANSIBLE_BASE_SOCIAL_AUDITOR_FLAG = "is_system_auditor"
```
john-westcott-iv, line 142: Should we change this to default to `is_platform_auditor` as a "going forward" approach?
Also, we should add this to the docs for authentication to let people know they need to set it if required?

AlanCoding: My logic here is that "is_system_auditor" isn't going away, it just gives an indicator of "local" permissions, not permissions that are shared with some other system. And maybe we don't migrate it at all, and move towards role mappings instead. So I still like this as the default. But maybe it should be added in docs.

john-westcott-iv (approved):

ansible_base/authentication/authenticator_plugins/oidc.py
```diff
@@ -207,7 +207,9 @@ class AuthenticatorPlugin(SocialAuthMixin, OpenIdConnectAuth, AbstractAuthentica
     configuration_encrypted_fields = ['SECRET']
 
     def extra_data(self, user, backend, response, *args, **kwargs):
-        for perm in ["is_superuser", "is_system_auditor"]:
+        from ansible_base.lib.utils.settings import get_setting
```
relrod: Does this have to be a local import or can it move up?

AlanCoding: This is in plugins, which is why I'm hesitant to do it top-level.

AlanCoding: ok, well I'll give it a shot, and I assume if it still passes checks that's ok.

PR #460, opened 2024-06-12
Remove OIDC backend config data caching

bhavenst:
Fixes blocker bug where subsequent OIDC based authenticators were redirecting to the wrong server (from another authenticator) after log in/log out.


bhavenst:
Yeah, there is 50% coverage on new code, but it is a one-liner copied from the super so doesn't make sense to add a unit test I don't think.

ansible_base/authentication/authenticator_plugins/oidc.py
```diff
@@ -215,3 +215,8 @@ def extra_data(self, user, backend, response, *args, **kwargs):
 
     def get_user_groups(self, extra_groups=[]):
         return extra_groups
+
+    def oidc_config(self):
+        # This is a copy of super without caching, which avoids data
+        # from one OIDC based authenticator showing up in another
+        return self.get_json(self.oidc_endpoint() + "/.well-known/openid-configuration")
```
john-westcott-iv, line 223: If you call the oidc_config from super would it still be cached? If not it would be nice to just `return super().oidc_config` or whatever.

bhavenst: Yeah that doesn't work to fix the issue, guess the caching works in that case.

john-westcott-iv (approved):

PR #463, opened 2024-06-13
Social Auth reflects Authentication maps

slemrmartin:
Social Auth (i.e. github-enterprise) calls `update_user_claims` which parses Authentication Maps in the pipeline specified by `SOCIAL_AUTH_PIPELINE` in settings.

To reflect that the auth map refuses to authenticate, the pipeline must return something other than `dict`,  `False` or `None`

### DIFF
--- a/ansible_base/authentication/backend.py
+++ b/ansible_base/authentication/backend.py
@@ -27,6 +27,7 @@ def get_authentication_backends(last_updated):
 
 class AnsibleBaseAuth(ModelBackend):
     def authenticate(self, request, *args, **kwargs):
+        from ansible_base.authentication.social_auth import SOCIAL_AUTH_PIPELINE_FAILED_STATUS
         logger.debug("Starting AnsibleBaseAuth authentication")
 
         # Query the database for the most recently last modified timestamp.
@@ -36,6 +37,11 @@ class AnsibleBaseAuth(ModelBackend):
 
         for authenticator_id, authenticator_object in get_authentication_backends(last_modified).items():
             user = authenticator_object.authenticate(request, *args, **kwargs)
+
+            # Social Auth pipeline can return False when update_user_claims fails (authentication maps deny access)
+            if user == SOCIAL_AUTH_PIPELINE_FAILED_STATUS:
+                continue
+
             if user:
                 # The local authenticator handles this but we want to check this for other authentication types
                 if not getattr(user, 'is_active', True):
@@ -46,5 +52,4 @@ class AnsibleBaseAuth(ModelBackend):
 
                 logger.info(f'User {user.username} logged in from authenticator with ID "{authenticator_id}"')
                 return user
-
         return None
--- a/ansible_base/authentication/backend.py
+++ b/ansible_base/authentication/backend.py
@@ -28,6 +28,7 @@ def get_authentication_backends(last_updated):
 class AnsibleBaseAuth(ModelBackend):
     def authenticate(self, request, *args, **kwargs):
         from ansible_base.authentication.social_auth import SOCIAL_AUTH_PIPELINE_FAILED_STATUS
+
         logger.debug("Starting AnsibleBaseAuth authentication")
 
         # Query the database for the most recently last modified timestamp.
@@ -38,7 +39,7 @@ class AnsibleBaseAuth(ModelBackend):
         for authenticator_id, authenticator_object in get_authentication_backends(last_modified).items():
             user = authenticator_object.authenticate(request, *args, **kwargs)
 
-            # Social Auth pipeline can return False when update_user_claims fails (authentication maps deny access)
+            # Social Auth pipeline can return status string when update_user_claims fails (authentication maps deny access)
             if user == SOCIAL_AUTH_PIPELINE_FAILED_STATUS:
                 continue
 
--- a/ansible_base/authentication/social_auth.py
+++ b/ansible_base/authentication/social_auth.py
@@ -15,7 +15,11 @@ from social_django.strategy import DjangoStrategy
 from ansible_base.authentication.authenticator_plugins.utils import generate_authenticator_slug, get_authenticator_class, get_authenticator_plugins
 from ansible_base.authentication.models import Authenticator, AuthenticatorUser
 
+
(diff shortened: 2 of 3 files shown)


ansible_base/authentication/social_auth.py
BrennanPaciorek, line 1: If we want to return something unique, we could also create a new class and a singleton of that class, use that for the logic. Something like:
```python
class SocialAuthPipelineFailedStatus():
    # Make objects of this type falsy
    def __bool__(self):
        return False

SOCIAL_AUTH_PIPELINE_FAILED_STATUS = SocialAuthPipelineFailedStatus()
```
The upside to this is that we would not need to do any explicit checking for the pipeline failure in `ansible_base/authentication/backend.py` while still returning a unique value, so we would not need to alter anything in that file. If we need to serialize it into a string, we could probably just give it a `__str__` method.

Doing this will also reduce the number of branches we need to test for in the new code, assuming that making the change wouldn't require a new change somewhere else.

If you think this will simplify the code, make the change, otherwise feel free to silently resolve this suggestion.

BrennanPaciorek (commented):
I'm alright with the changes made, but think that we should add test cases for the new code added, or incorporate code that will trigger the new logic into existing test cases.

ansible_base/authentication/backend.py
```diff
@@ -36,6 +38,11 @@ def authenticate(self, request, *args, **kwargs):
 
         for authenticator_id, authenticator_object in get_authentication_backends(last_modified).items():
             user = authenticator_object.authenticate(request, *args, **kwargs)
+
+            # Social Auth pipeline can return status string when update_user_claims fails (authentication maps deny access)
```
bhavenst, line 42: I guess that this would happen when the user has selected this specific authenticator, but there is an auth map that blocks the authentication (despite it maybe being a successful auth with the external system).  So this will just hop over this intended authenticator and let the rest fail as well?  I wonder if it's better to throw an AuthForbidden exception or similar (from social auth), which will kill the process and show some sort of error message (TBD, waiting on UI update).  I guess in any case the end result will be the AuthException from determine_username_from_uid_social but maybe it's better to go directly to an exception here rather than allowing the rest of the (fruitless) activity to happen?

bhavenst (commented):
Looks like it will work fine, just left a comment regarding immediately bailing rather than finishing the auth loop.

ansible_base/authentication/social_auth.py
slemrmartin: This won't work, because the problem is in the social-django's pipeline evaluation. It accepts None, False and any dict (which sounds like a bug in social-django), so the `if` condition then won't cause the desired effect.

ansible_base/authentication/backend.py
slemrmartin: The AuthException in social pipeline will cause 500 error, regarding doc, pipeline method doesn't expect any Exception.

Regarding this line, it follows the pattern which is there.  If the authenticate() is not successful, it continues to next in a loop

Other authenticators return None or False (directly from `update_user_claims`), but the social pipeline **can't** return None or False otherwise the `authenticate()` returns the user. 

So there is no change in the main logic, just fixing the weird behaviour of expected social-django pipeline return values.

slemrmartin: there is a use case like - `if not getattr(user, 'is_active', True) and user == SOCIAL_AUTH_PIPELINE_FAILED_STATUS` together, 
I'll investigate that if it's possible to handle it through social pipeline

slemrmartin: checked, it's not possible to set 'is_active' or similar user attribute through pipeline - https://github.com/python-social-auth/social-core/blob/master/social_core/backends/base.py#L85

slemrmartin:
@BrennanPaciorek tests added
@bhavenst proposed changes in the authentication chain are beyond of scope of this bug fix.

ansible_base/authentication/social_auth.py
```diff
@@ -188,4 +191,6 @@ def create_user_claims_pipeline(*args, backend, response, **kwargs):
     from ansible_base.authentication.utils.claims import update_user_claims
 
     extra_groups = response["Group"] if "Group" in response else None
-    update_user_claims(kwargs["user"], backend.database_instance, backend.get_user_groups(extra_groups))
+    user = update_user_claims(kwargs["user"], backend.database_instance, backend.get_user_groups(extra_groups))
+    if user is None:
+        return SOCIAL_AUTH_PIPELINE_FAILED_STATUS
```
BrennanPaciorek, line 196: These lines appear to be untested, is there any way we can test these (ex. `patching update_user_claims`)?

ansible_base/authentication/social_auth.py
BrennanPaciorek: In this case, why would having string constant as `SOCIAL_AUTH_PIPELINE_FAILED_STATUS` not pose the same issue?

BrennanPaciorek (approved):
As soon as SonarCloud approves, this PR lgtm.

ansible_base/authentication/backend.py
bhavenst: Yeah OK, I went and looked at the pipeline sequence and I see this is the very last element, so no use in short circuiting it anyway.  The only reason to throw an exception would be to display the new UI message from PR 467, since I guess this continue will cycle through the rest of the authenticators and invisibly fail? What happens from a user/UI perspective if no authenticators can authenticate?

bhavenst (approved):

ansible_base/authentication/social_auth.py
slemrmartin: [because this](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/base.py#L118)
it's not evaluated to False and the social backend returns what the pipeline returns

ansible_base/authentication/backend.py
slemrmartin: It returns None, so authentication failed

ansible_base/authentication/social_auth.py
slemrmartin: it's quite a test just for test coverage. For these cases of one line means 50% uncovered, the sonarcloud is not mandatory

PR #464, opened 2024-06-13
Add another variable which might house the public cert

john-westcott-iv:
If this code was used on the service which houses the jwt key we might not have the setting `ANSIBLE_BASE_JWT_KEY` but instead we could have the setting `jwt_public_key`.

### DIFF
--- a/ansible_base/jwt_consumer/common/cert.py
+++ b/ansible_base/jwt_consumer/common/cert.py
@@ -25,7 +25,9 @@ class JWTCert:
     def __init__(self):
         self.cached = None
         self.key = None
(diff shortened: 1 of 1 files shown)


AlanCoding:
Hate to say this, but I would like some references of more explanation. Like, would this mean that the `ANSIBLE_BASE_JWT_PUBLIC_KEY` setting would become valid? The conversion involving lowercase+prefix is non-obvious to me.

BrennanPaciorek (approved):

AlanCoding (approved):
I found the source of the "jwt_public_key", so that makes sense now.

PR #465, opened 2024-06-13
Add logging util

john-westcott-iv:
This creates a django logging filter which extracts a specific header.


kdelee (approved):

relrod (changes requested):
I'm gonna block this just to make sure it doesn't get merged - @john-westcott-iv and I were discussing alternatives, to avoid using `inspect` and I'd much prefer that.

relrod (approved):
Approving this as a *temporary* solution - we've agreed that we won't ship this.

PR #466, opened 2024-06-13
Avoid references to User.resource which may not be there

AlanCoding:
Explanation:

test_app adds the `resource` property to the `User` model as a handy short-hand for referencing its related resource.

https://github.com/ansible/django-ansible-base/blob/cff4a1106b58bd75587eb5e18e131939a2d8fa5d/test_app/models.py#L37-L42

But again, this is test_app, and it should be up to the discretion of the maintainers of test_app (mentally substitute some other real app there) whether or not this is added. Additionally, some apps get their user model from `auth.User`... and while [you can add a related resource](https://github.com/ansible/awx/blob/d94f766fcbe5cf68370830430d980b67d6cfd164/awx/main/models/__init__.py#L109) to that, I don't want to require such a thing. So this does the basic work involved to write a new alternative from scratch.

The SQL for this method:

```sql
SELECT "test_app_user"."id",
       "test_app_user"."password",
       "test_app_user"."last_login",
       "test_app_user"."is_superuser",
       "test_app_user"."username",
       "test_app_user"."first_name",
       "test_app_user"."last_name",
       "test_app_user"."email",
       "test_app_user"."is_staff",
       "test_app_user"."is_active",
       "test_app_user"."date_joined",
       "test_app_user"."modified",
       "test_app_user"."modified_by_id",
       "test_app_user"."created",
       "test_app_user"."created_by_id",
       (
        SELECT U0."ansible_id"
          FROM "dab_resource_registry_resource" U0
         WHERE (U0."content_type_id" = 17 AND U0."object_id" = ("test_app_user"."id")::varchar)
       ) AS "ansible_id_for_filter"
  FROM "test_app_user"
 WHERE (
        SELECT U0."ansible_id"
          FROM "dab_resource_registry_resource" U0
         WHERE (U0."content_type_id" = 17 AND U0."object_id" = ("test_app_user"."id")::varchar)
       ) = '0a4a242aa79b420c85847809eaa9cbcb'::uuid
 LIMIT 21
```

Some things to note:
- This is used in a `.get()`, so while it says LIMIT 21, anything other than 1 user returned will result in an error
- The code for obtaining the ansible_id is redundantly put in here twice, this is because we want the returned item to have `ansible_id_for_filter`, which can be referenced by the caller later. So this might not be the cleanest SQL, but there is a coherent reason for why it is the way it is.

### DIFF
--- a/ansible_base/lib/utils/auth.py
+++ b/ansible_base/lib/utils/auth.py
@@ -1,8 +1,13 @@
-from typing import Any, Type
+from typing import Any, Type, Union
+from uuid import UUID
 
 from django.apps import apps as django_apps
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.core.exceptions import ImproperlyConfigured
+from django.db.models import CharField, Model, OuterRef
+from django.db.models.functions import Cast
+from django.db.models.query import QuerySet
 
 from ansible_base.lib.abstract_models.organization import AbstractOrganization
 from ansible_base.lib.abstract_models.team import AbstractTeam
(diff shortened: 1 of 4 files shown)


ansible_base/lib/utils/auth.py
```diff
@@ -30,3 +35,18 @@ def get_team_model() -> Type[AbstractTeam]:
 
 def get_organization_model() -> Type[AbstractOrganization]:
     return get_model_from_settings('ANSIBLE_BASE_ORGANIZATION_MODEL')
+
+
+def get_object_by_ansible_id(qs: QuerySet, ansible_id: Union[str, UUID], annotate_as: str = 'ansible_id_for_filter') -> Model:
```
john-westcott-iv, line 40: We except this function to return 1 item. Because of that, do we need to feed this a `QuerySet` or could we just pass it in a `Model` object? If you named it cls you could get rid of line 43 and then on line 48 just do `return cls.objects.all().annotate...`

john-westcott-iv (approved):

ansible_base/lib/utils/auth.py
AlanCoding: I guess I was thinking that you would want to filter a queryset in related cases and get code reuse from that. But in that case you'd prefer a separate method for annotating, which would take the qs, and the get method could take the class. Although Django methods like get_object_or_404 tend to take a queryset as an argument.

PR #467, opened 2024-06-13
Add social auth exception handler middleware

bhavenst:
This middleware catches any exception from python social auth (failed authentication) and redirects the user to LOGIN_ERROR_URL.  The UI will receive something (query param, path var, slug, whatever) from this URL and display a login error of some sort.  So this PR should be updated when the UI knows how it will be handled.

### DIFF
--- a/ansible_base/authentication/middleware.py
+++ b/ansible_base/authentication/middleware.py
@@ -50,8 +50,5 @@ class SocialExceptionHandlerMiddleware(SocialAuthExceptionMiddleware):
         error_url = strategy.setting("LOGIN_ERROR_URL")
         backend = getattr(request, "backend", None)
         backend_name = getattr(backend, "name", "unknown-backend")
-        logger.error(f"Auth failure for backend {backend_name} - {repr(exception)}")
-        # UI expects auth_failed flag
-        url = urljoin(error_url, "/?auth_failed")
-        logger.info(f"Redirecting user to {url}")
-        return url
+        logger.error(f"Auth failure for backend {backend_name} - {repr(exception)}, redirecting to {error_url}")
+        return error_url
--- a/ansible_base/authentication/middleware.py
+++ b/ansible_base/authentication/middleware.py
@@ -1,5 +1,4 @@
 import logging
-from urllib.parse import urljoin
 
 from django.contrib.auth import BACKEND_SESSION_KEY
 from django.core.exceptions import ImproperlyConfigured
--- a/ansible_base/authentication/middleware.py
+++ b/ansible_base/authentication/middleware.py
@@ -1,8 +1,13 @@
+import logging
+
 from django.contrib.auth import BACKEND_SESSION_KEY
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.deprecation import MiddlewareMixin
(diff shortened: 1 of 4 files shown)


ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -139,6 +140,8 @@
     SOCIAL_AUTH_STRATEGY = "ansible_base.authentication.social_auth.AuthenticatorStrategy"
     SOCIAL_AUTH_LOGIN_REDIRECT_URL = "/"
 
+    # URL to send users when social auth login fails
+    LOGIN_ERROR_URL = "/"
```
bhavenst: Value for this TBD depending what UI does.  Also, maybe this should go in aap-gateway instead?

bhavenst: Agreed upon /?auth_failed

slemrmartin (commented):
What is difference between this middleware class and just setting `LOGIN_ERROR_URL="/?auth_failed"` ?

bhavenst:
There are two reasons I overrode the default middleware..  1) we can log the backend and the exception so it's visible in the log and 2) it allows for other flags to be added to the LOGIN_ERROR_URL for slightly different scenarios in the future if needed.

Here is the difference in logging:

aap_gw_1           | 2024-06-18 21:57:47,899 ERROR    ansible_base.authentication.middleware Auth failure for backend ansible_base-authentication-authenticator_plugins-oidc__external - AuthException()

aap_gw_1           | 2024-06-18 21:57:47,899 INFO     social 

They can have a message or not (AuthException does not), but at least it can give a bit more info.

But it is fine if we want to just use the provided middleware (social_django.middleware.SocialAuthExceptionMiddleware) without extending it if that is preferred..

ansible_base/authentication/middleware.py
```diff
@@ -36,3 +42,16 @@ def process_request(self, request):
...
+
+class SocialExceptionHandlerMiddleware(SocialAuthExceptionMiddleware):
+    def get_redirect_uri(self, request, exception):
+        strategy = getattr(request, "social_strategy", None)
+        error_url = strategy.setting("LOGIN_ERROR_URL")
+        backend = getattr(request, "backend", None)
+        backend_name = getattr(backend, "name", "unknown-backend")
+        logger.error(f"Auth failure for backend {backend_name} - {repr(exception)}")
+        # UI expects auth_failed flag
+        url = urljoin(error_url, "/?auth_failed")
```
slemrmartin: I would suggest to use urlparse and add the param. This prevents all apps to specify their URL with params. 
Or, the `/?auth_failed`  could be part of dynamic settings and is possible to redefine it (voting for this one).

slemrmartin (changes requested):

ansible_base/authentication/middleware.py
bhavenst: Oh yeah that makes sense.  No problem, I'll just add the whole thing in the settings.

slemrmartin (approved):
thanks for changes, it looks good now :+1:

PR #468, opened 2024-06-14
Allow null values for ansible_id fields

AlanCoding:
These were already non-required, allowing blank/omitting. But that might be difficult for some clients to adhere to. It also makes the DRF browseable API hard to use, giving:

```json
{
    "object_ansible_id": [
        "This field may not be null."
    ],
    "user_ansible_id": [
        "This field may not be null."
    ]
}
```

If you look at the serializer logic, it already assumes we might be getting blank values. This eliminates the error above, and just makes the assignment.


trucdg (approved):
LGTM!

PR #470, opened 2024-06-14
Revert superuser claims

slemrmartin:
Reverts this commit: https://github.com/ansible/django-ansible-base/pull/431/commits/903519bc54038ff6b18193c9f3e03d7214205920

Because `is_superuser` is handled by [update_user_claims()](https://github.com/ansible/django-ansible-base/blob/6a54df11c05ed58d20b3dd847fd2c0efbd6b04bf/ansible_base/authentication/utils/claims.py#L300)

### DIFF
--- a/ansible_base/authentication/utils/claims.py
+++ b/ansible_base/authentication/utils/claims.py
@@ -436,9 +436,6 @@ class ReconcileUser:
 
         self.apply_permissions()
 
-        # Superuser "role"
-        self.apply_superuser_permission()
-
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):

john-westcott-iv (commented):
I think there is already a bug with how update_user_claims is handeling is_superuser with the revoke flag and we likely need some logic like this down there specifically for that flag. But we can handle that in a seperate PR.

PR #472, opened 2024-06-14
Adding logging middleware

john-westcott-iv:
Adds a middleware that will extract the x-request-id headers and will automatically insert it into the log messages.

1. If we agree with this method I will need to change all of the getLogger calls in DAB to use this. 
2. I could easily be convinced to turn this into an application so that you can just add ansible_base.logging_middleware to your installed_apps and dynamic_config could add our middleware class.


ansible_base/lib/middleware/logging.py
```diff
@@ -0,0 +1,45 @@
...
+        self.http_id = None
+        super().__init__(logger, extra)
+
+    def process(self, msg, kwargs):
+        if self.http_id is None:
+            http_id = _ids.get(threading.current_thread().ident, kwargs.get('extra', {}).get('http_id', "None"))
+        else:
+            http_id = self.http_id
+
+        return f'[{http_id}] {msg}', kwargs
```
AlanCoding, line 36: Registering the sentiment here: I don't like to see this formatted here, as opposed to a python logging formatter.

AlanCoding:
Link https://github.com/ansible/django-ansible-base/pull/478, how is that one looking to you? I would prefer it.

PR #473, opened 2024-06-14
Add test model with no view permission

### LINKED ISSUE #457: Add test_app model that is visible to all users but has permissions for CUD (and test it)
Filing after reading comment:

https://github.com/ansible/galaxy_ng/pull/2162#discussion_r1635043038

This scenario seems perfectly coherent, and it should be perfectly compatible with DAB RBAC.
 - all records of a particular model are visible to all users in the system
 - only users who are specifically granted permission to modify that model can make changes - CUD - Create Update and Delete

I expect we have _a few_ cases where DAB RBAC assumed that a model has a "view" permission. We need docs to set expectations about configuration of this. For example, if the Meta doesn't list "view" permission is that _enough_ to assume that all users can view all entries? If so, that probably needs a special case in the logic going into `access_qs` and stuff. So there's at least 1 thing to do.

Also need to test that custom roles and role_metadata work as expected.

AlanCoding:
Fixes https://github.com/ansible/django-ansible-base/issues/457


ansible_base/rbac/validators.py
```diff
@@ -108,11 +108,12 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
 
     # Check that view permission is given for every model that has update/delete/special actions listed
     for cls, valid_model_permissions in permissions_by_model.items():
-        model_permissions = set(valid_model_permissions) & codename_list
-        non_add_model_permissions = {codename for codename in model_permissions if not is_add_perm(codename)}
-        if non_add_model_permissions and not any('view' in codename for codename in non_add_model_permissions):
-            display_perms = ', '.join(non_add_model_permissions)
-            raise ValidationError({'permissions': f'Permissions for model {role_model._meta.verbose_name} needs to include view, got: {display_perms}'})
+        if 'view' in cls._meta.default_permissions:
```
john-westcott-iv, line 111: This `if` initially threw me off but I think I might understand now. This is checking that `view` is an available permission is that correct?

If so, should we move that outside of the for loop (i.e. only run the for loop if its available). I tried to make a suggestion but its not possible because GH things there are deleted lines (but they are just indented). Can we also update the comment to somewhere say "if the vew permission is defined"

john-westcott-iv (approved):
Approving with one optional comment for clarification of intent.

ansible_base/rbac/validators.py
AlanCoding: I prefer your wording, "if `view` is an available permission" for the model.

But this checks that it is available for `cls`, which is the item of iteration in this loop. So it is being checked for every item in the loop

john-westcott-iv: Ah, yes, I see that now.

PR #474, opened 2024-06-14
Always clear ContentType cache to enforce determinism

AlanCoding:
I had a check failure on another PR, and I am pretty sure it is not from that patch, so I'm making this for first discover the test bugs we have, and then fix them on a case-by-case basis after identified.

### DIFF
--- a/test_app/tests/conftest.py
+++ b/test_app/tests/conftest.py
@@ -11,6 +11,7 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import rsa
 from django.apps import apps
+from django.contrib.contenttypes.models import ContentType
 from django.db.migrations.recorder import MigrationRecorder
 from django.db.models.signals import post_migrate
 from django.test.client import RequestFactory
@@ -63,6 +64,14 @@ def test_migrations_okay(*args, **kwargs):
 
 
 post_migrate.connect(test_migrations_okay)
+
+
+@pytest.fixture(autouse=True)
+def clear_content_type_cache():
+    """This enforces determinism between tests by deleting cached ContentType items from old tests"""
+    ContentType.objects.clear_cache()
+    yield
+    ContentType.objects.clear_cache()
 
 
 @pytest.fixture
--- a/test_app/tests/jwt_consumer/common/test_auth.py
+++ b/test_app/tests/jwt_consumer/common/test_auth.py
@@ -277,6 +277,7 @@ class TestJWTCommonAuth:
             request = mocked_http.mocked_parse_jwt_token_get_request('with_headers')
             common_auth.parse_jwt_token(request)
 
+    @pytest.mark.django_db
     def test_cert_fails_decryption_from_cached_key_but_key_is_invalid(self, mocked_http, expected_log, test_encryption_public_key):
         with override_settings(ANSIBLE_BASE_JWT_KEY=test_encryption_public_key):
             # Get the decryption key into the cache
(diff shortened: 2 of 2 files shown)


relrod (approved):

AlanCoding:
For records, this forces the following failure, and then fixes it:

```
def test_raise_an_exception_if_the_key_is_cached_but_the_new_key_is_the_same(self, random_public_key, mocked_http, create_mock_method):
        # We are going to:
        #     1. return a key which will not work with jwt_token provided by mocked_http
        #     2. Cache a key that is invalid
        #     3. Error when the new token is the same as the cached token
    
        mock_field_dicts = [
            {"key": random_public_key, "cached": True},
            {"key": random_public_key, "cached": False},
        ]
    
        # Pretend the key is coming from a URL
        url = 'https://example.com/'
        with override_settings(ANSIBLE_BASE_JWT_KEY=url):
            # 1. Make the get_decryption_key always return the random key (which is invalid) && 2. pretend the key is cached
            with mock.patch('ansible_base.jwt_consumer.common.auth.JWTCert.get_decryption_key', create_mock_method(mock_field_dicts)):
                # 3. Make the call.
                # This will attempt to use the cached key, recognize that its invalid, load the key again (which will be the same) and then error out
                request = mocked_http.mocked_parse_jwt_token_get_request('with_headers')
>               jwt_auth = JWTAuthentication()

test_app/tests/jwt_consumer/common/test_auth.py:457: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ansible_base/jwt_consumer/common/auth.py:283: in __init__
    self.common_auth = JWTCommonAuth(self.map_fields)
ansible_base/jwt_consumer/common/auth.py:40: in __init__
    self.team_content_type = ContentType.objects.get_for_model(apps.get_model(settings.ANSIBLE_BASE_TEAM_MODEL))
.tox/py39/lib/python3.9/site-packages/django/contrib/contenttypes/models.py:52: in get_for_model
    ct = self.get(app_label=opts.app_label, model=opts.model_name)
.tox/py39/lib/python3.9/site-packages/django/db/models/manager.py:87: in manager_method
.tox/py39/lib/python3.9/site-packages/django/db/models/query.py:633: in get
    num = len(clone)
.tox/py39/lib/python3.9/site-packages/django/db/models/query.py:380: in __len__
    self._fetch_all()
.tox/py39/lib/python3.9/site-packages/django/db/models/query.py:1881: in _fetch_all
    self._result_cache = list(self._iterable_class(self))
.tox/py39/lib/python3.9/site-packages/django/db/models/query.py:91: in __iter__
    results = compiler.execute_sql(
.tox/py39/lib/python3.9/site-packages/django/db/models/sql/compiler.py:1560: in execute_sql
    cursor = self.connection.cursor()
.tox/py39/lib/python3.9/site-packages/django/utils/asyncio.py:26: in inner
    return func(*args, **kwargs)
.tox/py39/lib/python3.9/site-packages/django/db/backends/base/base.py:330: in cursor
    return self._cursor()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <DatabaseWrapper vendor='postgresql' alias='default'>, name = None

    def _cursor(self, name=None):
        self.close_if_health_check_failed()
>       self.ensure_connection()
E       RuntimeError: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.
```

PR #475, opened 2024-06-15
Allow disabling object-level roles

### LINKED ISSUE #424: Allow disabling object-level permissions for a given model in the RBAC registry

AlanCoding:
Fixes https://github.com/ansible/django-ansible-base/issues/424

~Enabling work for AAP-25268~


ansible_base/rbac/api/views.py
```diff
@@ -60,7 +60,11 @@ def get(self, request, format=None):
             if cls is None:
                 cls_repr = 'system'
             else:
+                info = permission_registry.get_info(cls)
+                if not info.allow_object_roles:
+                    continue
```
john-westcott-iv, line 65: Should we warn or info if we hit this continue?

ansible_base/rbac/permission_registry.py
```diff
@@ -22,43 +22,59 @@
...
-                if model_name in self._name_to_model:
-                    raise RuntimeError(f'Two models registered with same name {model_name}')
-                self._name_to_model[model_name] = cls
-                if model_name != 'organization':
-                    self._parent_fields[model_name] = parent_field_name
+        for model in args:
+            if model._meta.model_name not in self._registry:
+                info = ModelPermissionInfo(model, **kwargs)
+                self._registry[info.model_name] = info
+            elif self._registry[model._meta.model_name] is model:
```
john-westcott-iv, line 55: This will account for the app_name because we are comparing the whole model right?

ansible_base/rbac/permission_registry.py
```diff
@@ -22,43 +22,59 @@
...
 
     def get_parent_model(self, model) -> Optional[type]:
-        model = self._name_to_model[model._meta.model_name]
-        parent_field_name = self.get_parent_fd_name(model)
-        if parent_field_name is None:
+        info = self._registry[model._meta.model_name]
+        if info.parent_field_name is None:
             return None
-        return model._meta.get_field(parent_field_name).related_model
+        return model._meta.get_field(info.parent_field_name).related_model
```
john-westcott-iv, line 70: Is it possible to register a model with a parent but not the parent?

ansible_base/rbac/permission_registry.py
```diff
@@ -22,43 +22,59 @@
...
 
     def get_parent_fd_name(self, model) -> Optional[str]:
```
john-westcott-iv, line 72: It would be really nice to rename this to `get_parent_field_name` for clarity.

john-westcott-iv: Not needed to merge this PR if its too much.

ansible_base/rbac/validators.py
```diff
@@ -100,6 +100,11 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
         role_model = content_type.model_class()
     permissions_by_model = permissions_allowed_for_role(role_model)
 
+    if not managed:
+        if not permission_registry.object_roles_enabled(role_model):
+            print_model = role_model._meta.verbose_name if role_model else 'global roles'
+            raise ValidationError({'content_type': f'Creating roles for the {print_model} model is disabled'})
```
john-westcott-iv, line 106: Is this the right termonology in the exception? `the global roles model is disabled`.
Also, we should likely translate these strings if they can get raised to the browser.

john-westcott-iv (approved):
I left a couple minor comments which likely don't need to be address if they are non-issues.

ansible_base/rbac/permission_registry.py
AlanCoding: Yes, very much so, `InstanceGroup` does not have a related organization. So you can have an object-level role, but no organization association. This call was made in fairly recent development, but before any work on the new RBAC system had started.

Even more interestingly, the `Instance` model is the one case where not being organization-scoped is truly defensible (an instance is a property of the cluster...), but object-level roles might also not make sense. It's more likely that you would _only_ want to delegate permissions with system-level roles. Because you're kind of either managing an AWX cluster or you're not... it's 1 thing, and sub-dividing it doesn't make a ton of sense.

ansible_base/rbac/permission_registry.py
AlanCoding: It will yell at you if you register 2 models with the same name, even if they're in different apps. Proxy models are fine, but don't register the original... and also register the proxy.

ansible_base/rbac/permission_registry.py
AlanCoding: I used that method name once, and only in eda-server tests, so it shouldn't be difficult to rename. But it is a cross-repo change.

ansible_base/rbac/validators.py
AlanCoding: Yeah, the flow of this isn't good. The method `object_roles_enabled` unconditionally returns `True` if passed `None`, and the `None` value signifies a global role. So, it would be more clear to have this block be `if (not managed) and role_model:`, so we're only considering this condition when it's an object-based (as opposed to system) role definition.

AlanCoding:
Retrospective:

The problem here is the creator roles. Permissions are an inter-connected system. If someone has "add" permission for a model, then after creating an object, they need the object-level admin role (or whatever role is configured for creators) after creation.

This means that we could only disable object-level roles created in the API, but they would still have to exist generally. This is an additional piece of information which would have to be communicated to clients, and increases confusion.

For overall system stability and sanity I am backing out of this approach. I may also close the issue as a won't-fix.

PR #478, opened 2024-06-17
Enhancements to request logging

relrod:
- Add middleware to inject the request into the log record
- Add a logging filter that can be used to pull out the request ID.

### DIFF
--- a/ansible_base/lib/logging/filters/request_id.py
+++ b/ansible_base/lib/logging/filters/request_id.py
@@ -1,0 +1,33 @@
+import logging
+import uuid
+
+from ansible_base.lib.logging import thread_local
+from ansible_base.lib.utils.collection import first_matching
+
+
+class RequestIdFilter(logging.Filter):
+    def filter(self, record):
+        # Always a default so we can use it in the logging formatter.
+        record.request_id = ""
+
+        request = getattr(thread_local, "request", None)
+
+        if request is None:
+            # request never got added to the thread local, so we can't add the request id to the log.
+            # But we still want to log the message.
+            return True
+
+        headers = request.META.keys()
+        request_id_header = first_matching(lambda x: x.lower().replace('-', '_') == "http_x_request_id", headers, default=None)
+        if request_id_header is None:
+            # We have a request, but no request id header. Still want to log the message.
+            return True
+        request_id = request.META.get(request_id_header)
+        try:
+            uuid.UUID(request_id)
+        except ValueError:
+            # Invalid request id, still want to log the message, but not with the invalid request id.
+            pass
+        else:
+            record.request_id = request_id
+        return True
--- a/ansible_base/lib/logging/filters/request_id.py
+++ b/ansible_base/lib/logging/filters/request_id.py
@@ -7,6 +7,12 @@ from ansible_base.lib.utils.collection import first_matching
 
 class RequestIdFilter(logging.Filter):
     def filter(self, record):
+        """
+        This "filter" is used to add the request id to the log record.
+        It will always return True, so that the message is always logged,
+        even if there is no request id.
+        """
+
         # Always a default so we can use it in the logging formatter.
         record.request_id = ""
 
--- a/ansible_base/lib/middleware/logging/log_request.py
+++ b/ansible_base/lib/middleware/logging/log_request.py
@@ -1,0 +1,23 @@
+import logging
+
+from ansible_base.lib.logging import thread_local
+
+logger = logging.getLogger(__name__)
+
+
+class LogRequestMiddleware:
+    """
+    Inject the request into the thread local so that it can be accessed by the logging filter.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        thread_local.request = request
+        return response
+
+    def process_request(self, request):
+        thread_local.request = request
+        return None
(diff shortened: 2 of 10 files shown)


ansible_base/lib/middleware/logging/log_request.py
```diff
@@ -0,0 +1,23 @@
...
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        thread_local.request = request
+        return response
+
+    def process_request(self, request):
+        thread_local.request = request
+        return None
```
AlanCoding, line 24: I think `process_request` is only called if subclassing `django.utils.deprecation.MiddlewareMixin`.

https://docs.djangoproject.com/en/5.0/topics/http/middleware/

If you're not, you could call `process_request` in the `__call__` method... or just write it like the modern examples.

ansible_base/lib/middleware/logging/log_request.py
```diff
@@ -0,0 +1,23 @@
...
+    Inject the request into the thread local so that it can be accessed by the logging filter.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        thread_local.request = request
+        return response
```
AlanCoding, line 20: I don't get it. Isn't this what you'd want?

```python
    def __call__(self, request):
        thread_local.request = request
        response = self.get_response(request)
        thread_local.request = None
        return response
```

I don't see how `.request` is set at the _beginning_ of the request-response cycle.

I also think setting `= None` at the _end_ of the cycle is good practice, so that garbage collection can do what it needs to do.

ansible_base/lib/logging/filters/request_id.py
```diff
@@ -0,0 +1,33 @@
...
+            return True
+        request_id = request.META.get(request_id_header)
+        try:
+            uuid.UUID(request_id)
+        except ValueError:
+            # Invalid request id, still want to log the message, but not with the invalid request id.
+            pass
+        else:
+            record.request_id = request_id
+        return True
```
AlanCoding, line 39: Does this ever return `False`? I would just like it to be a top-line doc string that this processes `record` to add information, and doesn't actually filter.

Abuse of the logging filter classes to add data [seems to be a common practice](https://stackoverflow.com/questions/43987624/how-to-add-custom-data-in-python-logs), so no blockers there.

john-westcott-iv (commented):
Since this has settings/uses settings, do we want to convert this into an app and have dynamic_settings set most required values?

ansible_base/lib/middleware/logging/log_request.py
relrod: I think you're right, I think what I was seeing in testing was the id from the previous request.

ansible_base/lib/middleware/logging/log_request.py
relrod: I only implemented this because we have code elsewhere that specifically calls `process_request()`, originally I didn't have this method. But yeah, could probably make `__call__` just use this method.

relrod:
> Since this has settings/uses settings, do we want to convert this into an app and have dynamic_settings set most required values?

I could go either way on this, but I have a slight inclination to _not_ make this an app.

I think either way before this merges, I need to add docs for how to set it up.

I'd be interested in @AlanCoding's thoughts. I'm not sure what kind of overhead comes with having a lot of tiny, composable apps, vs just providing a library. I'm sure there's some overhead as Django has to look for migrations, signal handlers, etc, in each app initialization. But maybe it's very minimal.

Without an app, setting this up looks something like:
- Add the middleware
- Add `RequestIdFilter` to `LOGGING['filters']`
- Add the filter to each `LOGGING['handlers']` that wants it
- Update the formatter(s) to have `%(request_id)s` somewhere

If we make it an app and do some magic mangling of `LOGGING` it becomes:
- Add the app
- Update the formatter(s) to have `%(request_id)s` somewhere

So the question is: Does any potential overhead that comes with an application outweigh the two extra steps that would otherwise be required?

AlanCoding:
> Since this has settings/uses settings, do we want to convert this into an app and have dynamic_settings set most required values?

Can I vote "no"?

I'm really not a fan of DAB making _deep_ changes to data structures in settings. The base-level expectation is that it adds top-level variables to the settings when you do the include. If it has to modify something, may it's better to document these as expectations for the app using DAB. Sure, we already broke that rule in a few cases, but `LOGGING` is a massively complicated data structure which is seriously next-level.

We often [modify](https://github.com/ansible/awx/blob/853af295d91cd0abde62b6c70787a036de150ed8/awx/settings/development.py#L32) `LOGGING` in multiple stages of settings. This can easily go wrong with unexpected `KeyError` type things, because of assumptions of what the partially-constructed data structure was at the start of your logic.

But what really kills :scream:  is this:

> Update the formatter(s) to have %(request_id)s somewhere

You're not just talking about modifying a dictionary 4-layers deep, but also string processing a format statement... and some formatters aren't even going to follow the rules you expect. What if you have a JSON formatter? That's going to need additional work on the app side, but if DAB is passing along an instruction to include `request_id`, then it has 100% finished its job, with no re-work on the DAB side needed. If you modify `LOGGING`, you will never stop adjusting.

relrod:
> but also string processing a format statement

Well no, I wouldn't do that from DAB, I'd leave it to the app to decide where in the format string to put the request id parameter. But yeah, I'm a bit squeamish about modifying `LOGGING` from DAB settings.

AlanCoding (approved):

mabashian (approved):

PR #480, opened 2024-06-18
Fix AuthMap groups and attributes rules

slemrmartin:
This PR fixes AuthMap permissions for `trigger_type='groups'` and `trigger_type='attributes'`

Before, it revoked permission to all users who don't met a trigger condition, regardless of the `revoke` attribute.
Now, it revokes permission only with `revoke: True`, otherwise it skips the auth map.


ansible_base/authentication/utils/claims.py
```diff
@@ -150,7 +150,7 @@ def _add_rbac_role_mapping(has_permission, role_mapping, role, organization=None
 
 def process_groups(trigger_condition: dict, groups: list, authenticator_id: int) -> Optional[bool]:
```
john-westcott-iv: I like where you are taking this to return None or True and then letting `revoke` convert None to False. 
```suggestion
def process_groups(trigger_condition: dict, groups: list, authenticator_id: int) -> Optional[Literal[True]]:
```

ansible_base/authentication/utils/claims.py
```diff
@@ -150,7 +150,7 @@ def _add_rbac_role_mapping(has_permission, role_mapping, role, organization=None
...
     """
-    Looks at a maps trigger for a group and users groups and determines if the trigger True or False
+    Looks at a maps trigger for a group and users groups and determines if the trigger True or None
```
john-westcott-iv:
```suggestion
    Looks at a maps trigger for a group and users groups and determines if the trigger True or None.
    Note, converting the returning None to False will happen later in the code.
```

ansible_base/authentication/utils/claims.py
```diff
@@ -178,7 +178,7 @@ def process_groups(trigger_condition: dict, groups: list, authenticator_id: int)
         else:
             has_access = True
 
-    return has_access
+    return True if has_access else None
```
john-westcott-iv: Since these are all elif/if should we just returning inside of these?

ansible_base/authentication/utils/claims.py
slemrmartin: It looks like this is not a valid type hint - there can't be value, only type, that's why I didn't update it

john-westcott-iv: Oh interesting, yea, because True is a value itself.... but its funny because its really not a bool. What about using Literal (updated suggestion).

slemrmartin: good one! yeah, that's what I missed

ansible_base/authentication/utils/claims.py
```diff
@@ -195,9 +188,9 @@ def has_access_with_join(current_access: Optional[bool], new_access: bool, condi
         return current_access and new_access
 
 
-def process_user_attributes(trigger_condition: dict, attributes: dict, authenticator_id: int) -> Optional[bool]:
+def process_user_attributes(trigger_condition: dict, attributes: dict, authenticator_id: int) -> Optional[Literal[True]]:
```
cutwater: The change to the function signature is confusing. While the former return type (i.e. `tribool` or `Optional[bool]`) seems logical, albeit  better to be avoided, the new signature `Optional[Literal[True]]` breaks general expectations. Is there a reason for it to return `True` or `None`, rather than `True` or `False` ?

slemrmartin: yes, that's a core of this change. If these methods return False, later processing can't distinguish if the `revoke` flag was used. 
By returning `None` with `revoke: False` means that the authentication map is ignored (and should be ignored)

cutwater: I understand the change, but I think that change in function signature is counterintuitive and obscure. In other circumstances I'd recommend avoid using `Optional[bool]` approach altogether, but here I think it'd be just better to leave type hint as `Optional[bool]` and not change it into `Optional[Literal[True]]` even considering that the function may never return `False` value.

slemrmartin: This is a requested change during review, so there are different needs between reviewers. 
So either there will be 3rd reviewer, or I vote for the current status, I find it useful - it's saying it *can't* return False

cutwater: First of all this usage of `Literal` type contradicts with it's intended usage. `Literal` type was added to describe function signatures that return different types depending on parameter value (e.g. `open(filename, mode)` ) and to describe vastly used approach in the existing code. See [https://peps.python.org/pep-0586/#motivation-and-rationale](url).

Second this signature on its own breaks the principle of least surprise. Separated from the context of usage the function that returns `None` or `True` doesn't make any sense. Any engineer that is not familiar with the code base will found this design at least illogical,  unless first presented with rationale behind this design and context of it's usage. While adding the comment would help to clarify intentions, the fact that interface must be additionally commented to clarify potential misunderstandings makes the interface questionable.

The change in logic here doesn't mean the signature has to be updated. If the requirements change again, the function may start return `False` again, and changing the signature once again to reflect this change implies that the contract of the function is volatile and cannot be relied on.

And the last usage of `Optional[bool]` in general is not a best practice, because it's error prone in Python, since False and None are both False-like, and the user of such interface must be extremely careful to explicitly expect either `None` or `False` values. Way better approach would be to declare an enum with 3 states:

```
class TriggerResult(Enum):
    ALLOW = auto()
    DENY = auto()
    SKIP = auto()
```

This is more expressive and less confusing than juggling with tri-state bool.

john-westcott-iv: The enum is an interesting option rather than True/False/None. I see where Literal could be being misused by my suggestion; but it does, to me, feel misleading to say `Optional[bool]`.  The function is currently intended to process the trigger itself which should be `True or None` the `False` comes into play later when looking at the results of the trigger and a property on the map. So I don't see this ever returning `False` unless we change the argument from the trigger to the map itself.

Given that I could go in one of two directions.
1. Leave the Optional[Literal[True]] - maybe suboptimal given the link above.
2. Defe some kind of enum or constant that would only contain ALLOW or SKIP (not DENY) and return one of those.

slemrmartin: > If the requirements change again, the function may start return False again

I don' t understand this argument - you can never define anything with this approach. Requirements of everything can change in the future, so with this logic you can't write any type checking (or even any method definition without `**kwargs` etc.)

slemrmartin: to the enum: as a developer reading the code without knowledge what is going on inside, I'll be totally lost with the Enum. It's conterproductive to document method, which can return True or None with the enum describing in can return `auto()` or `auto()`. I (and I'm convinced most people) won't understand it. 
So either let's go back to `Optional(bool)` or keep it as is.

cutwater: Using an enum doesn't return `auto()` or `auto()`, it returns either `TriggerResult.ALLOW` or `TriggerResult.SKIP`, which is self descriptive. The enum value is irrelevant and therefore the `auto()` is used, however you may define it as a string, doesn't really matter.

cutwater: > to the enum: as a developer reading the code without knowledge what is going on inside

do you truly believe that function returning `True` or `None`, but never `False` is more descriptive and intuitive than explicit `SKIP` or `ALLOW` values?

slemrmartin: clarified during meeting

cutwater (approved):
@slemrmartin Thank you for addressing my comments. LGTM!

ansible_base/authentication/utils/claims.py
```diff
@@ -60,21 +67,25 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
...
+        # Together with "revoke" flag => change permission to DENY
+        if auth_map.revoke and trigger_result == TriggerResult.SKIP:
+            trigger_result = TriggerResult.DENY
+
+        # If the trigger result is still SKIP, this auth map is not applicable to this user => no action needed
+        if trigger_result == TriggerResult.SKIP:
             rule_responses.append({auth_map.id: 'skipped'})
             continue
 
+        has_permission = trigger_result == TriggerResult.ALLOW
```
cutwater: Minor, FYI: When checking equality against enum instances you may use `is` operator, since enum instances are singletones:

```py
 has_permission = trigger_result is TriggerResult.ALLOW
```

slemrmartin: there can't be `is` or `==` because has_permission still needs to be 3-state. I'm working on the next commit

mabashian (approved):

PR #482, opened 2024-06-20
Allow role assignment queries to use .only

### LINKED ISSUE #449: `RecursionError` possible with `.only` on RBAC assignment models
### Steps to reproduce

Run bootstrap script.

Go into `python manage.py shell_plus`.

```python
RoleUserAssignment.objects.only('object_id')
```

### Expected behavior

Gives a list of object_ids, similar to the normal query

```python
RoleUserAssignment.objects.all()
```

### Actual Results

RecursionError: maximum recursion depth exceeded

Cycle is somehow related to:

```
File ~/repos/awx/testing/django-ansible-base/ansible_base/rbac/models.py:392, in AssignmentBase.__init__(self, *args, **kwargs)
    389 super().__init__(*args, **kwargs)
    391 # Cache fields from the associated object_role
--> 392 if self.object_role_id and not self.object_id:
    393     self.object_id = self.object_role.object_id
    394     self.content_type_id = self.object_role.content_type_id
```

AlanCoding:
Fixes https://github.com/ansible/django-ansible-base/issues/449


ansible_base/rbac/models.py
```diff
@@ -389,11 +389,15 @@ class Meta:
...
-        # Cache fields from the associated object_role
-        if self.object_role_id and not self.object_id:
-            self.object_id = self.object_role.object_id
-            self.content_type_id = self.object_role.content_type_id
-            self.role_definition_id = self.object_role.role_definition_id
+        # Fields from object_role are cached onto assignment objects, only when creating new assignments
+        # we must be very careful to avoid referencing deferred attributes to avoid RecursionError
+        def_fields = self.get_deferred_fields()
+        if not ({'id', 'object_id', 'object_role_id'} & def_fields):
+            if self.object_role_id and not self.object_id:
```
AlanCoding: This could still be wrong for system role assignments...

AlanCoding:
I was unhappy with any version that modifies `__init__` because for Django Models this is always brittle. Overriding the `save` method is a better practice, so I tried that and everything looks good to me. You should not have deferred fields when calling `save`... at least I don't think so...

Wait, that might be another test case.

AlanCoding:
No that concern is fine. Test case:

```python
@pytest.mark.django_db
def test_save_with_deferred_field(rando, inventory, inv_rd, global_inv_rd):
    assignment = inv_rd.give_permission(rando, inventory)
    reload_assignment = RoleUserAssignment.objects.filter(id=assignment.id).only('object_id').first()
    assert reload_assignment
    reload_assignment.save(update_fields=['role_definition_id'])
```

When this runs, the `id` field really is deferred. The important observation is that it's totally fine to load deferred fields when calling `save`, and also that deferred fields don't make any sense with _unsaved_ objects in the first place. So there's no loophole that can result in unintended errors. The above test case falls into the intended exception we intend to raise, so it's not an issue, and I'm not compelled to add this test case to the PR.

AlanCoding:
I know this was on hold as non-urgent, but I see very little risk in it, and we have been slowed down by this in other development.

relrod (approved):

PR #484, opened 2024-06-21
Split assignment setting from other setting

AlanCoding:
This will default to `True`, so the near-term plan to fix some problems we're having will be to let it assume the default value.

This will allow clients to assign permissions in the local (following) server so that it _can_ drift from the provider server's role assignments... for lack of a better solution.

AAP-25541

### DIFF
--- a/ansible_base/lib/dynamic_config/dynamic_settings.py
+++ b/ansible_base/lib/dynamic_config/dynamic_settings.py
@@ -199,7 +199,7 @@ if 'ansible_base.rbac' in INSTALLED_APPS:
     ANSIBLE_BASE_CACHE_PARENT_PERMISSIONS = False
 
     # API clients can assign users and teams roles for shared resources
-    ALLOW_LOCAL_RESOURCE_MANAGEMENT = True
+    ANSIBLE_BASE_ALLOW_LOCAL_RESOURCE_ASSIGNMENTS = True
 
     try:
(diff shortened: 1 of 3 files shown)


john-westcott-iv (approved):
Do we need follow up PRs for the services?
SOnar has coverage less than 80% but its such a minor change its likely just a partially covered conditional or something so I'm not worried about it.

bzwei:
Can this PR be merged? It blocks AAP-25541 which itself is a blocker.
@AlanCoding Does EDA needs to set ANSIBLE_BASE_ALLOW_LOCAL_RESOURCE_ASSIGNMENTS to True in its settings?

AlanCoding:
> Does EDA needs to set ANSIBLE_BASE_ALLOW_LOCAL_RESOURCE_ASSIGNMENTS to True in its settings?

By introducing a new setting here, eda-server will not have any override value, so it'll take the default of `True`, that's my intention.

AlanCoding:
Talking it over with @chrismeyersfsu and others, I agree I don't want to merge this. This will regress other things and pose much risk for the future.

AlanCoding:
Looking at John's comment

> Do we need follow up PRs for the services?

I think this led to a number of mis-understandings. We need to enable this setting, but we also need to disable it. There's a conflict, and this isn't fixing the core tension.

PR #486, opened 2024-06-21
jwt_consumer: remove unauthorized permissions

BrennanPaciorek:
Every time a user authenticates with JWT, all user role assignments
unlisted in the decrypted JWT are removed from the database.

AAP-25531

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -3,6 +3,7 @@ from typing import Optional, Tuple
 
 import jwt
 from django.apps import apps
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import Model
@@ -26,6 +27,18 @@ default_mapped_user_fields = [
     "email",
     "is_superuser",
 ]
+
+_permission_registry = None
+
+
+def permission_registry():
+    global _permission_registry
+
+    if not _permission_registry:
+        from ansible_base.rbac.permission_registry import permission_registry as permission_registry_singleton
+
+        _permission_registry = permission_registry_singleton
+    return _permission_registry
 
 
 class JWTCommonAuth:
@@ -184,9 +197,8 @@ class JWTCommonAuth:
         try:
             return RoleDefinition.objects.get(name=name)
         except RoleDefinition.DoesNotExist:
-            from ansible_base.rbac.permission_registry import permission_registry
 
-            constructor = permission_registry.get_managed_role_constructor_by_name(name)
+            constructor = permission_registry().get_managed_role_constructor_by_name(name)
             if constructor:
                 rd, _ = constructor.get_or_create(apps)
                 return rd
@@ -200,12 +212,20 @@ class JWTCommonAuth:
             logger.error("Unable to process rbac permissions because user or token is not defined, please call authenticate first")
             return
 
+        from ansible_base.rbac.models import RoleUserAssignment
+
+        role_diff = RoleUserAssignment.objects.filter(user=self.user, role_definition__name__in=settings.ANSIBLE_BASE_JWT_MANAGED_ROLES)
+
         for system_role_name in self.token.get("global_roles", []):
             logger.debug(f"Processing system role {system_role_name} for {self.user.username}")
             rd = self.get_role_definition(system_role_name)
             if rd:
-                rd.give_global_permission(self.user)
-                logger.info(f"Granted user {self.user.username} global role {system_role_name}")
+                if rd.name in settings.ANSIBLE_BASE_JWT_MANAGED_ROLES:
+                    assignment = rd.give_global_permission(self.user)
+                    role_diff = role_diff.exclude(pk=assignment.pk)
+                    logger.info(f"Granted user {self.user.username} global role {system_role_name}")
+                else:
+                    logger.error(f"Unable to grant {self.user.username} system level role {system_role_name} because it is not a JWT managed role")
             else:
                 logger.error(f"Unable to grant {self.user.username} system level role {system_role_name} because it does not exist")
                 continue
@@ -215,6 +235,9 @@ class JWTCommonAuth:
             if rd is None:
                 logger.error(f"Unable to grant {self.user.username} object role {object_role_name} because it does not exist")
                 continue
+            elif rd.name not in settings.ANSIBLE_BASE_JWT_MANAGED_ROLES:
+                logger.error(f"Unable to grant {self.user.username} object role {object_role_name} because it is not a JWT managed role")
+                continue
 
             object_type = self.token['object_roles'][object_role_name]['content_type']
             object_indexes = self.token['object_roles'][object_role_name]['objects']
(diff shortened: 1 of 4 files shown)


john-westcott-iv (approved):
@AlanCoding can you give this a second set of eyes for efficiency of code?

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -223,9 +227,13 @@ def process_rbac_permissions(self):
                 object_data = self.token['objects'][object_type][index]
                 resource, obj = self.get_or_create_resource(object_type, object_data)
                 if resource is not None:
-                    rd.give_permission(self.user, obj)
+                    assignment = rd.give_permission(self.user, obj)
+                    role_diff = role_diff.exclude(pk=assignment.pk)
                     logger.info(f"Granted user {self.user.username} role {object_role_name} to object {obj.name} with ansible_id {object_data['ansible_id']}")
 
+        # Remove all permissions not authorized by the JWT
+        role_diff.delete()
```
AlanCoding: Ways this won't work

1. The way DAB RBAC works is by filling in a "cache" table, and this logic only runs with `give_permission` and `remove_permission` calls, so this would not actually fully remove the permission. You'd need to call `remove_permission` for each case
2. For now, we support mixing and matching. Meaning, some role assignments will come from the resource provider service... but other permissions (which live in AWX/controller) are assigned directly in the AWX API. So `RoleUserAssignment.objects.filter(user=self.user)` is going to return a lot more than what you're interested to enforce... which makes me question, what _are_ we enforcing here? The JWT data only carries certain roles (5 by my last count). Since the JWT is the "authority" for those roles, you want to remove any assignments to those roles that are not in the JWT data. So this might look like `RoleUserAssignment.objects.filter(user=self.user, name__in=["Organization Member", ...])`. Exact implementation, I don't know, but that's looks like the general shape of the problem/solution.

BrennanPaciorek: Thank you for the constructive feedback. I believe the changes I have made address both of the reasons this PR previously did not work. I am not sure if it is as good as it can be in terms of efficiency though.

Can you take a look through the changes again for any oversights on my end and see if there are any apparent efficiency issues?

ansible_base/rbac/managed.py
```diff
@@ -151,6 +151,8 @@ class TeamMember(TeamMixin, ManagedActionBase):
     'action_base': ManagedActionBase,
 }
 
+MANAGED_ROLES = [role.name for role in managed_role_templates.values() if hasattr(role, "name")]
```
AlanCoding: This runs at import time, and there is a setting that allows changing the final value, which won't be ready until apps are ready. So you can't do it at import time. Here's the data structure I'd suggest from test_app:

```python
In [7]: [r.name for r in permission_registry._managed_roles.values()]
Out[7]: 
['Team Member',
 'Team Admin',
 'Organization Admin',
 'Organization Member',
 'Cow Admin',
 'Cow Mooer']
```

The Cow roles are intended.

BrennanPaciorek: Made some changes to use the permission registry instead. Do the new changes address the issue properly?

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -13,6 +13,7 @@
 from ansible_base.jwt_consumer.common.cert import JWTCert, JWTCertException
 from ansible_base.lib.utils.auth import get_user_by_ansible_id
 from ansible_base.lib.utils.translations import translatableConditionally as _
+from ansible_base.rbac.models import RoleUserAssignment
```
AlanCoding: I think this is unsafe to import at the top of the file. Elsewhere in this file `RoleDefinition` is imported in-line.

https://github.com/ansible/django-ansible-base/blob/2f8fcf18b95f975963fec893592ec33506421f78/ansible_base/jwt_consumer/common/auth.py#L195

The concern is that ansible/galaxy_ng is using the jwt_consumer app but is not using the RBAC app. Importing a model from an app not in INSTALLED_APP will error.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -200,11 +212,16 @@ def process_rbac_permissions(self):
             logger.error("Unable to process rbac permissions because user or token is not defined, please call authenticate first")
             return
 
+        role_diff = RoleUserAssignment.objects.filter(
+            user=self.user, role_definition__name__in=[role.name for role in permission_registry()._managed_roles.values()]
```
AlanCoding: I wish I had a better answer, but I think we're just going to have to hard-code these. There is a distinction here between "external" (from the JWT) and "managed". Even if everything made use of the "managed" role registry in the same way right now (which they don't), what we care about is "external" roles, where the JWT manages those roles. There's no standard registry of those at all.

It's probably best to put this into a setting, but that setting, I think is just going to have to list ["Organization Member", ...

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -223,9 +240,19 @@ def process_rbac_permissions(self):
...
+                    assignment = rd.give_permission(self.user, obj)
+                    role_diff = role_diff.exclude(pk=assignment.pk)
                     logger.info(f"Granted user {self.user.username} role {object_role_name} to object {obj.name} with ansible_id {object_data['ansible_id']}")
 
+        # Remove all permissions not authorized by the JWT
+        for role_assignment in role_diff:
+            rd = role_assignment.role_definition
+            object_role = role_assignment.object_role
+            if object_role:
+                rd.remove_permission(self.user, object_role.content_object)
```
AlanCoding: Performance-wise, you can also reference `role_assignment.content_object`. It's the same content_object. This would avoid an additional query.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -212,8 +211,10 @@ def process_rbac_permissions(self):
             logger.error("Unable to process rbac permissions because user or token is not defined, please call authenticate first")
             return
 
+        from ansible_base.rbac.models import RoleUserAssignment
+
         role_diff = RoleUserAssignment.objects.filter(
-            user=self.user, role_definition__name__in=[role.name for role in permission_registry()._managed_roles.values()]
+            user=self.user, role_definition__name__in=["System Auditor", "Organization Admin", "Organization Member", "Team Admin", "Team Member"]
```
AlanCoding:
```suggestion
            user=self.user, role_definition__name__in=["Platform Auditor", "Organization Admin", "Organization Member", "Team Admin", "Team Member"]
```

AlanCoding (changes requested):
Put in-line

AlanCoding (approved):

mabashian (approved):

BrennanPaciorek:
I have made more substantial changes since this was approved. @AlanCoding Can you review the changes in [this commit](https://github.com/ansible/django-ansible-base/pull/486/commits/2b1738731386670474c557a78917eadfef4f6c6a) in particular? 

That commit makes some changes to make sure that we don't add any roles via JWT auth which cannot be removed via JWT auth. It should be the last change to the application code I make before merging this.

test_app/settings.py
```diff
@@ -165,13 +165,15 @@
...
 ANSIBLE_BASE_MANAGED_ROLE_REGISTRY = {
+    'sys_auditor': {'name': "Platform Auditor"},
     'team_member': {},
     'team_admin': {},
     'org_admin': {},
     'org_member': {},
     'cow_admin': {'shortname': 'admin_base', 'model_name': 'test_app.cow', 'name': 'Cow Admin'},
     'cow_moo': {'shortname': 'action_base', 'model_name': 'test_app.cow', 'name': 'Cow Mooer', 'action': 'say_cow'},
 }
+ANSIBLE_BASE_JWT_MANAGED_ROLES.append("System Auditor")  # noqa: F821 this is set by dynamic settings for jwt_consumer
```
AlanCoding, line 176: I'm not worried about this line, but it does seem pointless. We don't create this role definition in test_app. Some _individual tests_ might create it. It doesn't seem to make sense in test_app settings, but it's just test_app, so non-blocking comment.

AlanCoding (approved):

test_app/settings.py
BrennanPaciorek: This is so other tests, which try to assign a role with the name `System Auditor` do not have their functionality interfered with.

I could just change everything to use `Platform Auditor` instead, but I think that should be addressed in a different PR.

Thank you for bearing with me for this PR, it must have been tedious with all the changes that needed to be made.

PR #488, opened 2024-06-24
Prevent N+1 queries problem when reconciling claims

cutwater:


slemrmartin (approved):
Nice catch! :+1:

ansible_base/authentication/utils/claims.py
```diff
@@ -424,7 +424,9 @@ def manage_permissions(self) -> None:
         """Processes the user claims (key `rbac_roles`)
         and adds/removes RBAC permissions (a.k.a. role_user_assignments)
         """
-        self.permissions_cache.cache_existing(self.user.role_assignments.all())
+        # NOTE(cutwater): Here `prefetch_related` is used to prevent N+1 problem when accessing `content_object`
```
slemrmartin, line 427: I think `prefetch_related` is always used to prevent N+1, thus the comment is not needed

cutwater: The comment here is actually not to explain what the `prefetch_related` does, but where the problem occurs, because the query and the code that relies on the query results are distanced far from each other. It may be really tricky to track where the attribute access happens, since it's in a separate class, several call stack nested levels away.

PR #489, opened 2024-06-24
Handle authenticator slug collision

bhavenst:
Adds random hex suffix to generated slug if it collides with an existing one (normally due to authenticator name change).

### DIFF
--- a/ansible_base/authentication/models/authenticator.py
+++ b/ansible_base/authentication/models/authenticator.py
@@ -1,3 +1,5 @@
+import secrets
+
 from django.db.models import SET_NULL, ForeignKey, JSONField, fields
 
 from ansible_base.authentication.authenticator_plugins.utils import generate_authenticator_slug, get_authenticator_plugin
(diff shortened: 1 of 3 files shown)


ansible_base/authentication/models/authenticator.py
```diff
@@ -39,8 +41,8 @@ def save(self, *args, **kwargs):
 
         if not self.slug:
             self.slug = generate_authenticator_slug(self.type, self.name)
-            # TODO: What happens if computed slug is not unique?
-            # You would have to create an adapter with a name, rename it and then create a new one with the same name
+            if Authenticator.objects.filter(slug=self.slug).count() > 0:
+                self.slug = generate_authenticator_slug(self.type, self.name, secrets.token_hex(2))
```
BrennanPaciorek: I think we should probably up the number of bytes on the slug suffix to at least 8 bytes if we do not want to loop and recheck for collisions. 1 in 65000 chance of collision is not bad, but I feel like we should probably reduce the collision odds by a little more.

BrennanPaciorek (approved):

ansible_base/authentication/models/authenticator.py
bhavenst: Hi Brennan, thanks for that.  Forgot to comment back. :)

PR #491, opened 2024-06-25
Parametrize visible users

slemrmartin:
RBAC method `visible_users` always adds superusers and request user. 

For related users views we need to show real members thus we need to be able to omit them

### DIFF
--- a/ansible_base/rbac/policies.py
+++ b/ansible_base/rbac/policies.py
@@ -11,28 +11,38 @@ from ansible_base.rbac.permission_registry import permission_registry
 from ansible_base.rbac.validators import permissions_allowed_for_role
 
 
-def visible_users(request_user, queryset=None) -> QuerySet:
-    "Gives a queryset of users that another user should be able to view"
+def visible_users(request_user, queryset=None, always_show_superusers=True, always_show_self=True) -> QuerySet:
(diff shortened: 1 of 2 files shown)


AlanCoding:
> For related users views we need to show real members thus we need to be able to omit them

I don't follow the problem statement here. If you are looking at a related view, the [originating queryset](https://github.com/ansible/django-ansible-base/blob/e2a75404d91a239762e046a7d8cdb5bc5c140e63/ansible_base/lib/routers/association_resource_router.py#L98) comes from a `getattr` to use the related manager. This is why `queryset` is already a parameter to `visible_users`. This method should take the originating queryset (which shouldn't include the superuser) and filters it further. I'd like to get more specifically what went wrong. Did the originating queryset have the superuser? Did `visible_users` somehow _add it back in_? This is the design, it should work, and I want to know specifically what didn't work before implementing a fix in the actual views.

AlanCoding (approved):
This is perfectly sensible enabling work. As a DAB change, total :+1:

Dostonbek1 (approved):

slemrmartin:
@AlanCoding the `visible_users` did always add superuser/request user by the `| .filter()` which is hard to undo later.
The affected use case is in the related PR

PR #492, opened 2024-06-25
Properly resolve the AuthenticatorUser for LDAP

john-westcott-iv:
When logging in through an LDAP that follows a different path than our development environment we were not properly resolving the gateway username. i.e. should it be the provided username or username-<hash>. This updated code will now properly call `determine_username_from_uid` before calling `get_or_create_authenticator_user`.

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/ldap.py
+++ b/ansible_base/authentication/authenticator_plugins/ldap.py
@@ -13,7 +13,7 @@ from django_auth_ldap.config import LDAPGroupType
 from rest_framework.serializers import ValidationError
 
 from ansible_base.authentication.authenticator_plugins.base import AbstractAuthenticatorPlugin, Authenticator, BaseAuthenticatorConfiguration
-from ansible_base.authentication.utils.authentication import get_or_create_authenticator_user
+from ansible_base.authentication.utils.authentication import determine_username_from_uid, get_or_create_authenticator_user
 from ansible_base.authentication.utils.claims import update_user_claims
 from ansible_base.lib.serializers.fields import BooleanField, CharField, ChoiceField, DictField, ListField, URLListField, UserAttrMap
 from ansible_base.lib.utils.validation import VALID_STRING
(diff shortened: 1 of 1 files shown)


trucdg (approved):
LGTM!

ansible_base/authentication/authenticator_plugins/ldap.py
```diff
@@ -410,7 +410,8 @@ def authenticate(self, request, username=None, password=None, **kwargs) -> (obje
             # In unit testing there were cases where the function we are in was being called before get_or_build_user.
             # Its unclear if that was just a byproduct of mocking or a real scenario.
             # Since this call is idempotent we are just going to call it again to ensure the AuthenticatorUser is created for update_user_claims
-            get_or_create_authenticator_user(username, self.database_instance, user_details={}, extra_data=user_from_ldap.ldap_user.attrs.data)
+            user_id = determine_username_from_uid(username, self.database_instance)
```
slemrmartin: variables are confusing 
- it takes uid and returns username
- there it looks like it takes username and returns int, which is not

ansible_base/authentication/authenticator_plugins/ldap.py
```diff
@@ -437,8 +438,9 @@ def get_or_build_user(self, username, ldap_user):
         """
         This gets called by _LDAPUser to create the user in the database.
         """
+        user_id = determine_username_from_uid(username, self.database_instance)
```
slemrmartin: the same, naming issue

slemrmartin (changes requested):

ansible_base/authentication/authenticator_plugins/ldap.py
john-westcott-iv: Yea, it’s kind of weird because it’s taking the uid from ldap and converting it into a username for the Django user table. But the authenticate function traditionally calls the parameter username. I’ll change this up a bit and see what I can come up with.

slemrmartin (approved):

PR #496, opened 2024-06-26
Silly attempt at running eda-server tests

relrod:
This adds EDA to the consumer test matrix.

### DIFF
--- a/.github/workflows/dab-consumers.yml
+++ b/.github/workflows/dab-consumers.yml
@@ -98,4 +98,5 @@ jobs:
           pipx install poetry
           poetry install -E all --only main,test
           poetry run pip install -e "$DAB_PATH"
+          poetry run pip show django-ansible-base
           poetry run pytest
--- a/.github/workflows/dab-consumers.yml
+++ b/.github/workflows/dab-consumers.yml
@@ -40,3 +40,62 @@ jobs:
         run: |
           cd awx
           DEVEL_IMAGE_NAME=awx-with-dab:latest AWX_DOCKER_CMD=/start_tests.sh make docker-runner
+
+  eda-server:
+    name: eda-server
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/dab-consumers.yml
```diff
@@ -40,3 +40,62 @@ jobs:
...
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Run eda-server tests
+        run: |
+          DAB_PATH="$PWD"
+          mv eda-server /tmp
+          cd /tmp/eda-server
```
AlanCoding, line 97: Why did you have to move it to `/tmp`?

.github/workflows/dab-consumers.yml
```diff
@@ -40,3 +40,62 @@ jobs:
...
+          pipx install poetry
+          poetry install -E all --only main,test
+          poetry run pip install -e "$DAB_PATH"
```
AlanCoding, line 100: Nice, so because we don't need to run these in a container, it's even easier than the monkey patching you did for AWX.

.github/workflows/dab-consumers.yml
relrod: It doesn't have to be `/tmp` but I don't want to "dirty" the clone of DAB in any way. So I just clone eda and then immediately move it to leave the DAB directory clean.

relrod:
@dhaustein @ttuffin

.github/workflows/dab-consumers.yml
AlanCoding: My only suggestion might be to add a line that did `pip show django-ansible-base` so we can have a correct sense that it is using the correct install for troubleshooting.

AlanCoding (approved):

Alex-Izquierdo (approved):
LGTM

PR #497, opened 2024-06-27
Require superuser permission to revoke global roles

AlanCoding:
Policy edit to be what AWX prior `is_system_auditor` policy was.

I gave more detail in a followup comment, but here are the key changes:
 - _Require_ superuser permission to revoke a system role assignment
 - _Allow_ removing an assignment if it applies to yourself
 - _Require_ that adding/revoking global role assignments only involve non-shared model permissions

AAP-26308

### DIFF
--- a/ansible_base/rbac/validators.py
+++ b/ansible_base/rbac/validators.py
@@ -192,7 +192,7 @@ def validate_assignment(rd, actor, obj) -> None:
         raise ValidationError(f'Role type {rd_model} does not match object {obj_ct.model}')
 
 
-def check_locally_managed(permissions_qs: Iterable[Model]) -> None:
+def check_locally_managed(permissions_qs: Iterable[Model], role_content_type: Optional[Model]) -> None:
     """Can the given role definition be managed here, or is it externally managed
 
     If the role definition manages permissions on any shared resources, then
@@ -203,9 +203,10 @@ def check_locally_managed(permissions_qs: Iterable[Model]) -> None:
     if settings.ALLOW_LOCAL_RESOURCE_MANAGEMENT is True:
         return
     for perm in permissions_qs:
-        # View permission for shared resources is interpreted as permission to view
+        # View permission for shared objects is interpreted as permission to view
         # the resource locally, which is needed to be able to view parent objects
-        if perm.codename.startswith('view'):
+        # For system roles, this exception is not allowed
+        if role_content_type and perm.codename.startswith('view'):
             continue
         model = perm.content_type.model_class()
         if permission_registry.get_resource_prefix(model) == 'shared':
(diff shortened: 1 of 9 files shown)


slemrmartin (approved):

AlanCoding:
This is flawed. Because it would prevent a system auditor from demoting themselves from being a system auditor. We did support that in the old AWX logic, at least for a user editing event.

AlanCoding:
Walking back through from the beginning:

Original story of concern:
 - A user who is a system auditor tries to revoke another user's system auditor role assignment
 - Previously this would have been allowed, but this patch changes it to a 403, which is consistent with prior AWX behavior

However, the obvious fix for this _would_ open up another user story (`test_remove_global_assignment_yourself`) that is likely to _regress_:
 - As a user who has the system auditor role, try to revoke your own assignment to that role
 - This should be allowed, which is consistent with prior AWX behavior.

The reason this needs to be stated is that our prior criteria ("you can remove a system role assignment if you have all its permissions") covered that case implicitly. So this needs to add an explicit criteria that you can remove assignments for a given user, regardless of the associated object, or regardless of it being a system role.

There is another object-role variant of this scenario (`test_remove_object_assignment_yourself`).
 - A user has "view" permission to an object
 - That user tries to revoke their role assignment (wants themselves to no longer be able to view the object)
 - Prior behavior looks like this would have given a 403, which is a bug. You should always be able to remove your own role assignments.

I saw another fairly nuanced problem with system roles (`test_auditor_for_external_models`). This is overlapping with recent merge https://github.com/ansible/django-ansible-base/pull/501, which added an exception for "view" permission to objects, but this adds an exception-to-the-exception for system roles.
 - A system-wide auditor role that comes from a permission-provider service. We say it is externally managed here if it includes any permissions for shared resources.
 - Someone tries to assign or revoke that role.
 - Prior behavior was that this would be allowed, but that is inconsistent with other rules for externally-managed roles. So this changes it to a 403.

slemrmartin (approved):
awesome work! 
What I'm not sure, how user can delete own global permission, but the test for it passes, so I'm good, approving

AlanCoding:
> What I'm not sure, how user can delete own global permission

Code-wise, on assignment DELETE requests `check_can_remove_assignment` is called for the permission check, and (given a user assignment) if they can make changes to the user, this unconditionally returns without error. You can make changes to a user if that user is yourself.

ansible_base/rbac/validators.py
```diff
@@ -192,7 +192,7 @@ def validate_assignment(rd, actor, obj) -> None:
         raise ValidationError(f'Role type {rd_model} does not match object {obj_ct.model}')
 
 
-def check_locally_managed(permissions_qs: Iterable[Model]) -> None:
+def check_locally_managed(permissions_qs: Iterable[Model], role_content_type: Optional[Model]) -> None:
```
AlanCoding, line 195: Now I'm getting concerned that this might give the wrong answer in the resource server (which manages the shared resources).

AlanCoding: No, it doesn't cause that issue. Why? Because `ALLOW_LOCAL_RESOURCE_MANAGEMENT` defaults to True.

mabashian (approved):

PR #500, opened 2024-06-27
Custom permission class for oauth token scope adherence

relrod:
Provide a DRF permission class for ensuring that token scopes are adhered to, based on the upstream DOT permission classes.

### DIFF
--- a/ansible_base/oauth2_provider/permissions.py
+++ b/ansible_base/oauth2_provider/permissions.py
@@ -1,0 +1,29 @@
+import oauth2_provider.models as oauth2_models
+from oauth2_provider.contrib.rest_framework import TokenHasReadWriteScope
+from rest_framework.permissions import BasePermission, IsAuthenticated
+
+# This is not in views/permissions.py because we import that in other oauth2_provider files
+# and when we try to specify this OAuth2ScopePermission in an app's settings.py, we get a
+# cyclic dependency.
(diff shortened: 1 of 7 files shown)


AlanCoding (changes requested):
Looking good, this should be mentioned in `docs/` somewhere, because otherwise it's only used in test_app right now.

AlanCoding:
I would model the docs after this:

https://github.com/ansible/eda-server/blob/main/src/aap_eda/settings/default.py#L325-L328

relrod:
@AlanCoding I reworked this a bit and made it require `IsAuthenticated` too. I am curious on your thoughts here. I guess the downside is that it places a restriction in that you can't just include this permission class everywhere -- but the views that don't require `IsAuthenticated` and should be fully public will need to set their permission_classes to avoid that anyway.

So I don't think there's anything lost by doing this, and I think there's an added security gain because I think requiring `IsAuthenticated` follows the principle of least surprise.

AlanCoding:
Sure, only concern is that, as in the eda-server case (link above), some apps may list both `IsAuthenticated` and other permission classes. So the DAB class here should make it clear that `IsAuthenticated` is redundant, and should not be listed.

AlanCoding (approved):

mabashian (approved):

PR #501, opened 2024-06-28
Allow managing org roles with view permission locally

AlanCoding:
Modifies work in https://github.com/ansible/django-ansible-base/pull/430

This replaces https://github.com/ansible/django-ansible-base/pull/484, which I put up because I did not have the full details of what was happening.

We are getting unintended errors because organization-level roles contain the "shared.view_organization" permission. So this change will wave those through.

This means that some access evaluations will give a different answer on different servers in some cases (can I view this organization?). If that is restricted to viewing permission of parent objects, that appears manageable.

AAP-25541


bzwei:
The logic makes sense. LGTM

rochacbruno (approved):

mabashian (approved):

PR #505, opened 2024-07-01
Adding get_relative_url and get_fully_qualified_url

john-westcott-iv:
AAP-26819

We somewhat arbitrarily used DRF reverse along with Django reverse. The difference is that Django reverse produces a relative URL and DRF reverse looks at the request and returns a fully qualified url.

This PR adds two functions: `get_fully_qualifed_url` and `get_relative_url` to `ansible_base.lib.utils.response`. This PR also converts all references to both `reverse` methods to our new method.

The `get_fully_qualified_url` will look for a setting called `FRONT_END_URL` and use that as the `protocol://server:port` portion of the URL if defined. If undefined it will allow DRF to attempt to do its normal reverse.

In most places in the API we should be using relative URLs. Hover, if we need a fully qualified URL for any reason we should attempt to use `get_fully_qualified_url`. These functions both take a view name and *args and **kwargs similar to the existing reverse methods.

In addition to being more explicit, this will help us to have a common method and seeing for constructing fully qualified URLs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,7 +130,7 @@ legacy_tox_ini = """
     [testenv:black]
     deps =
         black
-    commands = black {posargs:.}
+    commands = black --version && black {posargs:.}
 
     [testenv:isort]
     deps =
--- a/ansible_base/activitystream/models/entry.py
+++ b/ansible_base/activitystream/models/entry.py
@@ -3,11 +3,11 @@ import functools
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
 from django.db import models
-from django.urls import reverse
 from django.utils.http import urlencode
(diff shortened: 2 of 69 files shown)


BrennanPaciorek (approved):
lgtm.

Dostonbek1:
@john-westcott-iv not sure if I asked this before, but would this be a breaking change for services (e.g. EDA) using DAB?

mabashian (approved):

pyproject.toml
```diff
@@ -130,7 +130,7 @@ legacy_tox_ini = """
     [testenv:black]
     deps =
         black
-    commands = black {posargs:.}
+    commands = black --version && black {posargs:.}
```
relrod, line 133: @john-westcott-iv This disabled black. This is parsed as `black --version '&&' black .` which ends up the same as `black --version` and all other arguments get dropped.

PR #506, opened 2024-07-02
Optimize SQL queries when creating orgs and teams

cutwater:
The former SQL query was producing `WHERE` clause against python repr of tuples. Rewrting this exact query would likely require usage of raw PostgreSQL specific SQL.
Therefore instead of querying list teams by tuples of organization name and team name, we query all teams within set of organization ids. This queries slighly more date, but tallows us to avoid extra `JOIN` with the `Organization` table and complicated filtering.

Minor refactoring.

### DIFF
--- a/ansible_base/authentication/utils/claims.py
+++ b/ansible_base/authentication/utils/claims.py
@@ -1,3 +1,4 @@
+import contextlib
 import importlib
 import logging
 import re
@@ -9,7 +10,7 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AbstractUser
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ObjectDoesNotExist
-from django.db import models
+from django.db import IntegrityError, models
 from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
 from rest_framework.serializers import DateTimeField
@@ -69,11 +70,11 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
         for trigger_type, trigger in auth_map.triggers.items():
             if trigger_type == 'groups':
                 trigger_result = process_groups(trigger, groups, authenticator.pk)
-            if trigger_type == 'attributes':
+            elif trigger_type == 'attributes':
                 trigger_result = process_user_attributes(trigger, attrs, authenticator.pk)
-            if trigger_type == 'always':
+            elif trigger_type == 'always':
                 trigger_result = TriggerResult.ALLOW
-            if trigger_type == 'never':
+            elif trigger_type == 'never':
                 trigger_result = TriggerResult.DENY
 
         # If the trigger result is SKIP, auth map is not defined for this user.
@@ -108,7 +109,6 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
             _add_rbac_role_mapping(has_permission, rbac_role_mapping, auth_map.role, auth_map.organization)
         elif auth_map.map_type == 'role' and not is_empty(auth_map.role) and is_empty(auth_map.organization) and is_empty(auth_map.team):
             _add_rbac_role_mapping(has_permission, rbac_role_mapping, auth_map.role)
-
         else:
             logger.error(f"Map type {auth_map.map_type} of rule {auth_map.name} does not know how to be processed")
 
@@ -329,7 +329,6 @@ def update_user_claims(user: Optional[AbstractUser], database_authenticator: Aut
             reconcile_user_class.reconcile_user_claims(user, authenticator_user)
         except Exception as e:
             logger.exception("Failed to reconcile user claims: %s", e)
-
     return user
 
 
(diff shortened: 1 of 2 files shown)


ansible_base/authentication/utils/claims.py
```diff
@@ -346,52 +346,47 @@ def create_organizations_and_teams(results) -> None:
...
+            if org_name not in membership_map:
+                membership_map[org_name] = {'id': None, 'teams': []}
+            membership_map[org_name]['teams'].append(team_name)
+
+    # Create organizations
+    existing_orgs = dict(Organization.objects.filter(name__in=org_set | team_org_set).values_list("name", "id"))
+    for org_name in org_set:
+        org_id = existing_orgs.get(org_name)
+        if org_id is None:
+            org_id = Organization.objects.create(name=org_name).id
```
john-westcott-iv: If we drop to just create we should likely catch the object already exists incase two people are logging in at the same time and we have competing threads creating orgs.

ansible_base/authentication/utils/claims.py
```diff
@@ -346,52 +346,47 @@ def create_organizations_and_teams(results) -> None:
...
-    # make each team as necessary
+    team_org_ids = [membership_map[org_name]['id'] for org_name in team_org_set]
+    existing_teams = set(Team.objects.filter(organization__in=team_org_ids).order_by().values_list('organization', 'name'))
     for org_name, org_data in membership_map.items():
         org_id = org_data['id']
-        for team_name, is_member in org_data['teams'].items():
+        for team_name in org_data['teams']:
             if (org_id, team_name) not in existing_teams:
-                new_team, _ = Team.objects.get_or_create(name=team_name, organization_id=org_id)
+                Team.objects.create(name=team_name, organization_id=org_id)
```
john-westcott-iv: Same create comment here.

ansible_base/authentication/utils/claims.py
cutwater: The `get_or_create` method is not atomic and affected by the same race condition, which is usually ignored when writing code in Django. We just check organizations existence in batch above. There is no point of a redundant check just before creating a record.

Mitigating possible race condition might be necessary regardless of this change.

ansible_base/authentication/utils/claims.py
```diff
@@ -346,52 +346,47 @@ def create_organizations_and_teams(results) -> None:
...
+    team_org_set = set()
     # a structure for caching org+team,member info
     membership_map = {}
 
     # fill in the top level org membership data ...
     for org_name, is_member in results['claims']['organization_membership'].items():
-        if is_member:
-            org_list.add(org_name)
-            membership_map[org_name] = {'id': None, 'teams': {}}
+        if not is_member:
```
slemrmartin, line 365: I don't find this more readable, it's creating 4 lines from 3 lines, moreover introducing negative condition ( which is not recommended)
This would be useful if the following code will contain more lines, with more indentation

ansible_base/authentication/utils/claims.py
```diff
@@ -346,52 +346,47 @@ def create_organizations_and_teams(results) -> None:
...
-
-    # make each org as necessary or simply store the id
-    for org_name in org_list:
-        if org_name not in existing_orgs:
-            new_org, _ = Organization.objects.get_or_create(name=org_name)
-            membership_map[org_name]['id'] = new_org.id
-        else:
-            membership_map[org_name]['id'] = existing_orgs[org_name]
-
+            if not is_member:
```
slemrmartin, line 373: the same like above. Negative condition doesn't improve readability for few lines of code

ansible_base/authentication/utils/claims.py
```diff
@@ -346,52 +346,47 @@ def create_organizations_and_teams(results) -> None:
...
+                continue
+            team_org_set.add(org_name)
+            if org_name not in membership_map:
+                membership_map[org_name] = {'id': None, 'teams': []}
+            membership_map[org_name]['teams'].append(team_name)
+
+    # Create organizations
+    existing_orgs = dict(Organization.objects.filter(name__in=org_set | team_org_set).values_list("name", "id"))
+    for org_name in org_set:
```
slemrmartin: the meaning of variables above was changed, you splitted org_list to org_set and team_org_set. But this loop is still traversing only through org_set, thus all new team's organizations will be skipped

ansible_base/authentication/utils/claims.py
slemrmartin: so can you please handle this race condition?

ansible_base/authentication/utils/claims.py
cutwater: Unnecessary indentation is not recommended either.

Also would you give some reference, what makes you thinking introducing negative condition is not recommended? I think your understanding of the recommendation might be inaccurate.
What is not recommended is using double negation e.g.

```python
if not condition:
   # do something
else:
   # do something else
```

But here is a different case. It is an early `return` \ `continue`.

References:

1. https://dev.to/somedood/please-dont-write-confusing-conditionals-2n32
2. Code Complete, 2nd Edition by Steve McConnell
    Chapter 15. Using Conditionals

ansible_base/authentication/utils/claims.py
cutwater: ~~I disagree.~~

While above there are only two lines of code in the `if` block, this one is larger and has nested `if` clause. I'd prefer to keep the code less nested if possible. I can revert the code above, but in my opinion it would be better to keep it consistent.

ansible_base/authentication/utils/claims.py
cutwater: It depends on what exactly you mean by "fixing" it, and how exactly the code is expected to behave.
It can be mitigated by handling an `IntegrityError`, just like the `get_or_create` method does, but this only works as long as we execute every query in `autocommit` mode, which on its own negatively affects performance.

ansible_base/authentication/utils/claims.py
cutwater: This is a good catch, thank you.

ansible_base/authentication/utils/claims.py
john-westcott-iv: Yea, we talked about potentially looking at a bulk insert option to try and split the difference between transactions and performance. For this PR, catching `IntegrityError` is good enough for me. It should be a pretty low chance of hitting a collection with that in there.

john-westcott-iv (approved):
I'm good with this these changes but we should also get @slemrmartin approval as well.

ansible_base/authentication/utils/claims.py
slemrmartin: clarified during meeting, anyway this was more comment than request - resolving

slemrmartin (approved):
looks good now :+1:

mabashian (approved):

PR #509, opened 2024-07-02
Optimize SQL queries when assigning permissions on teams

cutwater:
Use similar approach to #506 to reduce number of queries that select teams by organization from N to 1 (where N is a number of organizations).


slemrmartin (approved):
nice work :+1:

mabashian (approved):

PR #510, opened 2024-07-02
Enforce CSRF mitigations for JWT authentication

BrennanPaciorek:
Impement enforce_csrf for JWT authenticatior, enabling django CsrfMiddleware.
This fixes disabled CSRF protections on sites that use jwt_consumer for authentication.

AAP-20597


BrennanPaciorek:
This PR will most likely be closed, since enforcing csrf makes little sense on JWTAuthentication, which does not use cookies on its own at any point during the authentication process.

PR #512, opened 2024-07-10
Make ansible_id match the JWT data if creating new user

AlanCoding:
Deep, deep, deep into a debugging hole, I thought I fixed the issue where the "admin" user had a resource conflict. But then it was still a conflict.

It was a conflict created by an entirely new mechanism. The JWT consumer was creating users from the JWT data that had a different ansible_id.

NOTE: We seem to have another problem where the `service_id` is not set for absolutely anything created in this file.

AAP-26659

### DIFF
--- a/test_app/tests/jwt_consumer/common/test_auth.py
+++ b/test_app/tests/jwt_consumer/common/test_auth.py
@@ -491,8 +491,9 @@ class TestJWTAuthentication:
                     assert 'check your key and generated token' in af
                     assert 'cached key was correct' in af
 
+    @pytest.mark.parametrize('new_user', [True, False])
     def test_correctly_authenticate_if_the_cached_key_is_invalid_but_the_new_key_is_correct(
-        self, random_public_key, mocked_http, test_encryption_public_key, django_user_model, jwt_token, create_mock_method
+        self, random_public_key, mocked_http, test_encryption_public_key, django_user_model, jwt_token, create_mock_method, new_user
     ):
         # Pretend the key is coming from a URL
         url = 'https://example.com'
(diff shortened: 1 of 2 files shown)


test_app/tests/jwt_consumer/common/test_auth.py
```diff
@@ -476,9 +477,14 @@ def test_correctly_authenticate_if_the_cached_key_is_invalid_but_the_new_key_is_
...
-                user = django_user_model.objects.create_user(username=jwt_token.unencrypted_token['user_data']['username'], password="password")
+                if not new_user:
+                    user = django_user_model.objects.create_user(username=jwt_token.unencrypted_token['user_data']['username'], password="password")
                 created_user, _ = jwt_auth.authenticate(request)
-                assert created_user == user
+                if not new_user:
+                    assert created_user == user
+                else:
+                    assert created_user.username == jwt_token.unencrypted_token['user_data']['username']
+                    assert str(created_user.resource.ansible_id) == jwt_token.unencrypted_token['sub']
```
AlanCoding, line 516: To clarify, this was failing before this patch, the ansible_id values were not the same. This is the source of a conflict in what I was trying to do.

AlanCoding:
Trying to give a skeptical self-review of this - the logic I'm building here starts to overlap significantly with `get_or_create_resource`, although that doesn't have the features I want to be thorough.

### Outline of the abstract case

We are given an `ansible_id` and `resource_data`. We should also have a `service_id`, but we don't, that that's another novel issue.

We prioritize looking up the object by `ansible_id`, and we may or may not find it. If we find it, we're great! Move on. If we don't find it, then we want to create it. Creating the object may end in conflict, due to a pre-existing resource conflict. A conflict doesn't necessarily imply this type of conflict, because it could be due to another unknown database constraint implemented by the application. In this last case we should prefer to let it error and just log what happens, maybe catching any DatabaseError generally.

### Problems with the existing method

The existing method would just error in the event of pre-existing resource conflicts. Should it?

There's no generalization of error handling in this respect, mainly because it doesn't do error handling. So we could tighten this code up in a couple of ways.

AlanCoding:
I tried to look into ways to make this better, and I just can't. There's too much model-specific behavior to do a light refactor which combines the get-by-ansible_id logic for user/teams/orgs. We could log the `created` attribute, but that's all we would be doing, or I could change the get_or_create to just a "create". I'm happy with this as it is now.

relrod (approved):

mabashian (approved):

jjwatt (commented):
LGTM

PR #513, opened 2024-07-11
Prevent disabling or deleting the last enabled authenticator

jctanner:
https://issues.redhat.com/browse/AAP-26867

### DIFF
--- a/ansible_base/authentication/serializers/authenticator.py
+++ b/ansible_base/authentication/serializers/authenticator.py
@@ -78,6 +78,14 @@ class AuthenticatorSerializer(NamedCommonModelSerializer):
         return parsed_data
 
     def validate(self, data) -> dict:
+        # prevent disabling the last enabled authenticator
+        if data.get('enabled') is False:
+            if self.instance and self.instance.enabled:
+                # count how many other authenticators are enabled ...
+                qs = Authenticator.objects.filter(enabled=True).exclude(id=self.instance.id)
+                if not qs.exists():
+                    raise ValidationError(_("At least one authenticator must be enabled"))
+
         validator_type = data.get('type', None)
         # if we didn't have a type, try to get the type of the existing object (if we have one)
         if not validator_type and self.instance:
--- a/ansible_base/authentication/serializers/authenticator.py
+++ b/ansible_base/authentication/serializers/authenticator.py
@@ -80,7 +80,7 @@ class AuthenticatorSerializer(NamedCommonModelSerializer):
     def validate(self, data) -> dict:
         # prevent disabling the last enabled authenticator
         if data.get('enabled') is False:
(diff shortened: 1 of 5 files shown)


mabashian:
I guess there's really two scenarios we need to account for here:

1. We should prevent a user from disabling the last enabled authenticator
2. We should prevent a user from deleting the last enabled authenticator

Maybe you had already thought through that @jctanner but I had only thought about scenario 2 initially

jctanner:
@mabashian we have a indirect gate in the view that throws an error if you try to delete an authenticator that is still attached to a user, which I guess would always get thrown due to the system user.

I can also extend this patch to directly prevent deletion of the last enabled authenticator.

ansible_base/authentication/views/authenticator.py
```diff
@@ -21,6 +21,12 @@ class AuthenticatorViewSet(AnsibleBaseDjangoAppApiView, ModelViewSet):
 
     def destroy(self, request, *args, **kwargs):
         instance = self.get_object()
+
+        # check if there is at least one other enabled authenticator...
+        if not Authenticator.objects.filter(enabled=True).exclude(id=instance.id).exists():
```
AlanCoding: I would prepend to this `if` condition, `instance.enabled and`. If somehow all enabled authenticators already got deleted, I don't want to block users doing cleanup of data which doesn't affect that situation (meaning, disabled authenticators).

ansible_base/authentication/serializers/authenticator.py
```diff
@@ -78,6 +78,14 @@ def to_internal_value(self, data):
         return parsed_data
 
     def validate(self, data) -> dict:
+        # prevent disabling the last enabled authenticator
+        if data.get('enabled') is False:
+            if self.instance and self.instance.enabled:
+                # count how many other authenticators are enabled ...
+                qs = Authenticator.objects.filter(enabled=True).exclude(id=self.instance.id)
```
AlanCoding: It might be good to combine the two querysets into an instance method `Authenticator.is_the_last_airbender`

ansible_base/authentication/serializers/authenticator.py
```diff
@@ -78,6 +78,11 @@ def to_internal_value(self, data):
         return parsed_data
 
     def validate(self, data) -> dict:
+        # prevent disabling the last enabled authenticator
+        if data.get('enabled') is False:
+            if self.instance and self.instance.enabled and self.instance.is_last_enabled:
```
AlanCoding: You should no longer need `and self.instance.enabled`

AlanCoding (approved):
Left one final cleanup suggestion. Overall the structure looks good here.

mabashian (approved):

PR #514, opened 2024-07-12
[oauth2_provider] Disable PKCE by default

relrod:
This is enabled by default in new DOT and breaks
compatibility with awx.


hasys (approved):
Looks good to me, thank you @relrod

mabashian (approved):

BrennanPaciorek (approved):
LGTM, though it is unfortunate that we cannot use PKCE

PR #517, opened 2024-07-12
[Low Prio] Add a sqlite3 check

AlanCoding:
I got 3 failures running this locally, and ability to address those failures is the key thing I'd like to get out of this. I would like to position us to be multi-vendor regarding the database as much as we can reasonably do with minimal effort, and this is a nice start.

### DIFF
--- a/test_app/tests/authentication/views/test_authenticator_map.py
+++ b/test_app/tests/authentication/views/test_authenticator_map.py
@@ -1,6 +1,8 @@
 import pytest
 from django.urls import reverse
 
+from django.db import connection
+
 
 def test_authenticator_map_list_empty_by_default(admin_api_client):
     """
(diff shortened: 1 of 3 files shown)


AlanCoding:
Locally I'm easily able to replicate:

```
DJANGO_SETTINGS_MODULE=test_app.sqlite3settings py.test test_app/tests/ -k test_authenticator_map_validate
```

AlanCoding:
Adding a link to the reason for the failure we're skipping https://github.com/django/django/pull/16601

test_app/tests/authentication/views/test_authenticator_map.py
```diff
@@ -84,6 +85,9 @@ def test_authenticator_map_invalid_map_type(admin_api_client, local_authenticato
     assert '"invalid" is not a valid choice.' in response.data['map_type'][0]
 
 
+@pytest.mark.skipif(
+    connection.vendor == 'sqlite', reason='PositiveIntegerField min_value was not discovered but fixed in Django 5 https://github.com/django/django/pull/16601'
```
relrod: Can we also check `django.VERSION[0] < 5`, so this test works whenever we upgrade?

relrod (approved):

test_app/tests/authentication/views/test_authenticator_map.py
AlanCoding: I was just being lazy. I'll push a commit.

PR #518, opened 2024-07-12
Fix incorrect mock object calls

AlanCoding:
The problem with mocked objects is that any properties will return a "fake" method. So `mockeddecode.called_with` just gives you a made-up method that does nothing. If you `assert mocked_obj.called_with('foo')`, then the fake method says "ok", returns some other made-up thing, and test passes.

I don't know what the consequence of these failures are, but I want to identify them and hopefully assign someone to either fix them or do something else.

### DIFF
--- a/test_app/tests/authentication/authenticator_plugins/test_oidc.py
+++ b/test_app/tests/authentication/authenticator_plugins/test_oidc.py
@@ -146,7 +146,7 @@ def test_user_data(mockedrequest, mockeddecode, mocksetting):
     mockeddecode.return_value = mr.json()
     data = ap.user_data("token")
 
-    assert mockeddecode.called_with("token", "VALUE", "VALUE", "VALUE")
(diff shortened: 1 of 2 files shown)


AlanCoding:
logging failures

FAILED test_app/tests/authentication/authenticator_plugins/test_radius.py::test_authenticator_plugin

```
E           AssertionError: expected call not found.
E           Expected: authenticate(username='user-73aa1c', password='radius_password-4aa69d')
E           Actual: authenticate(None, 'user-73aa1c', 'radius_password-4aa69d')
```


FAILED test_app/tests/authentication/authenticator_plugins/test_oidc.py::test_user_data

```
E           AssertionError: expected call not found.
E           Expected: decode('token', 'VALUE', 'VALUE', 'VALUE')
E           Actual: decode('token', key='-----BEGIN PUBLIC KEY-----\nVALUE\n-----END PUBLIC KEY-----', algorithms='VALUE', audience='VALUE')
```

relrod (approved):

PR #520, opened 2024-07-15
Error if Delete permission is issued without Change permission

thedoubl3j:
Currently, we have no logic to check if a combination of role permissions that includes delete is allowed which causes some issues when appropriately reflecting things in the UI and user_capabilities field. While at first we thought that this was caused because the permissions were not boiling up correctly, the system was actually working as intended but we were not warning the user of the bad combination of permissions. 

This PR addresses that by warning the user of the bad combination and alerting them that they need to include the `change` permission in order to appropriately assign the `delete` permission.

### DIFF
--- a/ansible_base/rbac/validators.py
+++ b/ansible_base/rbac/validators.py
@@ -116,19 +116,22 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
                 display_perms = ', '.join(non_add_model_permissions)
                 raise ValidationError({'permissions': f'Permissions for model {role_model._meta.verbose_name} needs to include view, got: {display_perms}'})
 
-    # check change edit permissions is given for every model that has update/delete/special actions listed
+    # Check change edit permissions is given for every model that has update/delete/special actions listed
     if settings.ANSIBLE_BASE_DELETE_REQUIRE_CHANGE:
-        for cls, valid_model_permissions in permissions_by_model.items():
-            if 'delete' in cls._meta.default_permissions and 'change' in cls._meta.default_permissions:
-                model_permissions = set(valid_model_permissions) & codename_list
-                non_add_model_permissions = {codename for codename in model_permissions if not is_add_perm(codename)}
-                if any('delete' in codename for codename in non_add_model_permissions) and not any(
-                    'change' in codename for codename in non_add_model_permissions
-                ):
-                    display_perms = ', '.join(non_add_model_permissions)
-                    raise ValidationError(
-                        {'permissions': f'Permissions for model {role_model._meta.verbose_name} needs to include change, got: {display_perms}'}
-                    )
+        # If role_model is a system-wide role (None means it is), skip this validation.
+        if role_model != None:
+            for cls, valid_model_permissions in permissions_by_model.items():
+                # import pdb; pdb.set_trace()
+                if 'delete' in cls._meta.default_permissions and 'change' in cls._meta.default_permissions:
+                    model_permissions = set(valid_model_permissions) & codename_list
+                    non_add_model_permissions = {codename for codename in model_permissions if not is_add_perm(codename)}
+                    if any('delete' in codename for codename in non_add_model_permissions) and not any(
+                        'change' in codename for codename in non_add_model_permissions
+                    ):
+                        display_perms = ', '.join(non_add_model_permissions)
+                        raise ValidationError(
+                            {'permissions': f'Permissions for model {role_model._meta.verbose_name} needs to include change, got: {display_perms}'}
+                        )
 
 
 def validate_codename_for_model(codename: str, model: Union[Model, Type[Model]]) -> str:
--- a/ansible_base/rbac/validators.py
+++ b/ansible_base/rbac/validators.py
@@ -117,14 +117,18 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
                 raise ValidationError({'permissions': f'Permissions for model {role_model._meta.verbose_name} needs to include view, got: {display_perms}'})
 
     # check change edit permissions is given for every model that has update/delete/special actions listed
-    for cls, valid_model_permissions in permissions_by_model.items():
-        if 'delete' in cls._meta.default_permissions:
-            # import pdb; pdb.set_trace()
-            model_permissions = set(valid_model_permissions) & codename_list
-            non_add_model_permissions = {codename for codename in model_permissions if not is_add_perm(codename)}
-            if 'delete' in non_add_model_permissions and not any('change' in codename for codename in non_add_model_permissions):
-                display_perms = ', '.join(non_add_model_permissions)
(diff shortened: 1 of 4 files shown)


ansible_base/rbac/validators.py
```diff
@@ -116,6 +116,16 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
                 display_perms = ', '.join(non_add_model_permissions)
                 raise ValidationError({'permissions': f'Permissions for model {role_model._meta.verbose_name} needs to include view, got: {display_perms}'})
 
+    # check change edit permissions is given for every model that has update/delete/special actions listed
+    for cls, valid_model_permissions in permissions_by_model.items():
+        if 'delete' in cls._meta.default_permissions:
+            # import pdb; pdb.set_trace()
```
thedoubl3j: remove this

ansible_base/rbac/validators.py
```diff
@@ -116,6 +116,16 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
                 display_perms = ', '.join(non_add_model_permissions)
                 raise ValidationError({'permissions': f'Permissions for model {role_model._meta.verbose_name} needs to include view, got: {display_perms}'})
 
+    # check change edit permissions is given for every model that has update/delete/special actions listed
+    for cls, valid_model_permissions in permissions_by_model.items():
+        if 'delete' in cls._meta.default_permissions:
```
AlanCoding: Nice thinking here. However, I'd like to request that we `and` this with `'change' in cls._meta.default_permissions`. The logic here only makes sense to me when both delete & change are valid permissions for the model type.

ansible_base/rbac/validators.py
```diff
@@ -116,6 +116,16 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
...
+            # import pdb; pdb.set_trace()
+            model_permissions = set(valid_model_permissions) & codename_list
+            non_add_model_permissions = {codename for codename in model_permissions if not is_add_perm(codename)}
+            if 'delete' in non_add_model_permissions and not any('change' in codename for codename in non_add_model_permissions):
```
AlanCoding: I don't think `'delete' in non_add_model_permissions` works, because in the case of credential, for example, the permission codename is "delete_credential".

And I'll volunteer that the current code is inconsistent and messy, in that `is_add_perm` is a method, but we should really have a similar method for any permission that is ever used in internal policy or validation logic. And it would probably be best to put that on the `DABPermission` model. I just haven't gotten around to doing that.

But in any case, I expect this condition to do a test "is this a delete codename" for any element in the list.

AlanCoding (changes requested):
My first comment is asking for additional gating of this validation. My second comment, I think, is likely the same issue causing your test to fail.

Third request - I would like this to be configurable via a setting. It can default to `True`, "require change permission to get delete permission", but I don't want to have to wait for a DAB change to turn this back off again. Really, we only need this for AWX. eda-server could turn this on or off right now, I don't have a preference. But soon-ish I'd like to turn it back off for both.

ansible_base/rbac/validators.py
```diff
@@ -116,6 +116,20 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
                 display_perms = ', '.join(non_add_model_permissions)
                 raise ValidationError({'permissions': f'Permissions for model {role_model._meta.verbose_name} needs to include view, got: {display_perms}'})
 
+    # check change edit permissions is given for every model that has update/delete/special actions listed
+    if settings.ANSIBLE_BASE_DELETE_REQUIRE_CHANGE:
+        for cls, valid_model_permissions in permissions_by_model.items():
+            if 'delete' and 'change' in cls._meta.default_permissions:
```
AlanCoding: The way this is written won't work, just by the way `and` and `in` combine

thedoubl3j:
@AlanCoding updated the logic to make check if the default perms has delete and change. this is causing 2 of the tests that you wrote for the view validator to fail since they have only `delete_publicdata` and or view. Do you want me to update those tests for now so that they don't fail?

thedoubl3j:
leaving this in the comments so that I can go fix awx test once this merges `FAILED awx/main/tests/functional/dab_rbac/test_dab_rbac_api.py::test_assign_custom_delete_role`

AlanCoding:
In the test output I see:

```
                        raise ValidationError(
>                           {'permissions': f'Permissions for model {role_model._meta.verbose_name} needs to include change, got: {display_perms}'}
                        )
E                       AttributeError: 'NoneType' object has no attribute '_meta'
```

This looks like a legitimate issue that should be fixed. We don't allow custom system-wide roles for anything that matters for now, so if `role_model` is None, which indicates a system-wide role, I think it's best to avoid this validation logic.

> @AlanCoding updated the logic to make check if the default perms has delete and change. this is causing 2 of the tests that you wrote for the view validator to fail since they have only delete_publicdata and or view. Do you want me to update those tests for now so that they don't fail?

I don't want to get into questioning whether the test would still be meaningful after making that change. For these failing tests, can you instead just wrap them in `@override_settings` to turn off the validation logic you added? That would be my preference.

ansible_base/rbac/validators.py
```diff
@@ -116,6 +116,23 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
                 display_perms = ', '.join(non_add_model_permissions)
                 raise ValidationError({'permissions': f'Permissions for model {role_model._meta.verbose_name} needs to include view, got: {display_perms}'})
 
+    # Check change edit permissions is given for every model that has update/delete/special actions listed
+    if settings.ANSIBLE_BASE_DELETE_REQUIRE_CHANGE:
+        # If role_model is a system-wide role (None means it is), skip this validation.
+        if role_model is not None:
+            for cls, valid_model_permissions in permissions_by_model.items():
+                # import pdb; pdb.set_trace()
```
AlanCoding: delete line

ansible_base/rbac/validators.py
```diff
@@ -116,6 +116,23 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
                 display_perms = ', '.join(non_add_model_permissions)
                 raise ValidationError({'permissions': f'Permissions for model {role_model._meta.verbose_name} needs to include view, got: {display_perms}'})
 
+    # Check change edit permissions is given for every model that has update/delete/special actions listed
+    if settings.ANSIBLE_BASE_DELETE_REQUIRE_CHANGE:
+        # If role_model is a system-wide role (None means it is), skip this validation.
+        if role_model is not None:
```
AlanCoding: If you look at the SonarCloud output, you'll probably see it recommend that 2 nested `if` statements should be combined, just as a code style thing

AlanCoding (approved):

thedoubl3j:
awx PR to fix failing test is up: https://github.com/ansible/awx/pull/15385

mabashian (approved):

PR #521, opened 2024-07-16
fix: Skip _system user when syncing resources

rochacbruno:
The endpoint `/service-index/resource-types/shared.user/manifest/` must not include the `_system` user so it is not include on the sync task.


ansible_base/resource_registry/views.py
```diff
@@ -141,6 +141,10 @@ def manifest(self, request, name, *args, **kwargs):
             return HttpResponseNotFound()
 
         resources = Resource.objects.filter(content_type__resource_type=resource_type).prefetch_related("content_object")
+
+        if name == "shared.user":  # Skip the _system user
+            resources = resources.exclude(name="_system")
```
relrod: Prefer `settings.SYSTEM_USERNAME` instead of hardcoding `_system`

rochacbruno: Interesting, galaxy_ng even when running with DAB profile does not have that `settings.SYSTEM_USERNAME` so I guess I must use `getattr(settings, "SYSTEM_USERNAME", "_system")`

AlanCoding: Makes sense since that is not in [dynamic settings](https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/lib/dynamic_config/dynamic_settings.py), and I don't think galaxy_ng _should_ define that setting, because like AWX, it is not using this construct of a system user.

Alternative to the `getattr` - it may be better to not exclude this user from the list if the setting is not set.

AlanCoding:
I'll paste this here, @TheRealHaoLiu and I discussed doing this on the resource consumer side (as opposed to the resource provider).

https://github.com/ansible/awx/compare/devel...AlanCoding:awx:no_thank_you?expand=1

Doing it in the resource provider would result in fewer patches, but I consider it less robust for the long-term. What if a resource consumer does use the system user construct? If it has the user, it should sync it from the resource provider, otherwise linkages will be broken.

rochacbruno:
@AlanCoding this

```python
class AWXSyncExecutor(SyncExecutor):
    def _process_manifest_item(self, manifest_item):
        if manifest_item.resource_data.get('username', None) == '_system':
            return SyncResult(SyncStatus.NOOP, manifest_item)
        return super()._process_manifest_item(manifest_item)
```

Is not going to work, at that point the manifest_item will be only `ansible_id, hash` the `resource_data` is fetched after the `return super()._process_manifest_item(manifest_item)` 

According to the conversations `settings.SYSTEM_USERNAME` should never sync, you are saying that there will be the case for syncing it?

ansible_base/resource_registry/views.py
```diff
@@ -141,6 +142,11 @@ def manifest(self, request, name, *args, **kwargs):
             return HttpResponseNotFound()
 
         resources = Resource.objects.filter(content_type__resource_type=resource_type).prefetch_related("content_object")
+
+        if name == "shared.user":
+            system_user = getattr(settings, "SYSTEM_USERNAME", "_system")
```
AlanCoding: If `not hasattr(settings, "SYSTEM_USERNAME")`, my preference is to not modify `resources`.

AlanCoding (approved):
Leaving one additional comment, not sure if it's actionable: This leaves the `ResourceViewSet` with the unfiltered queryset, meaning that streaming the csv manifest will give different content than a GET to the resource list. I understand that the csv data is used specifically by the resource sync, but the inconsistency would be interpreted as a bug if someone just saw it. So maybe a followup issue or documentation on that would be good.

mabashian (approved):

PR #522, opened 2024-07-16
Prepare 2 User managers to exclude _system user by default

slemrmartin:
This PR creates a `AbstractDABUser` who has 2 managers: 
- `User.objects....`
- `User.all_objects`

this adds an ability to exclude some objects by default by apps using DAB, for example to exclude `is_active=False`  or `managed=True`  in the default queryset - without the need to rewrite queries in dependent apps

AAP-27023


BrennanPaciorek (approved):
LGTM.

ansible_base/lib/abstract_models/user.py
```diff
@@ -0,0 +1,9 @@
+from django.contrib.auth.models import AbstractUser, UserManager
+
+
+class AbstractDABUser(AbstractUser):
```
AlanCoding, line 4: :+1: Let's plan to migrate more fields from downstream back up, and figure more potential points of integration. Shipping with an abstract org/team/user is going to present the most sane behavior for new apps integrating DAB.

AlanCoding (approved):

mabashian (approved):

PR #523, opened 2024-07-22
Fix new user getting created at every login

john-westcott-iv:
Fix `determine_username_from_uid()` being called too early resulting in us creating a new AuthenticatorUser at every login.

### DIFF
--- a/ansible_base/authentication/utils/authentication.py
+++ b/ansible_base/authentication/utils/authentication.py
@@ -4,6 +4,7 @@ from typing import Optional, Tuple
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AbstractUser
+from django.db.models import Q
 from django.utils.translation import gettext as _
 from social_core.exceptions import AuthException
 from social_core.pipeline.user import get_username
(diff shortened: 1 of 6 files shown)


mabashian (approved):

test_app/models.py
```diff
@@ -54,6 +54,12 @@ class Meta(AbstractDABUser.Meta):
     def summary_fields(self):
         return user_summary_fields(self)
 
+    def get_authenticator_ids(self) -> list[int]:
+        return list(self.authenticator_users.values_list('provider__id', flat=True))
+
+    def get_authenticator_uids(self) -> list[str]:
+        return list(self.authenticator_users.values_list('uid', flat=True).distinct())
+
```
john-westcott-iv, line 62: @relrod we should think if there is a way we can auto-add these methods to the default User object if dab_authentication is enabled.

ansible_base/authentication/utils/authentication.py
```diff
@@ -89,13 +96,11 @@ def determine_username_from_uid(uid: str = None, authenticator: Authenticator =
 
 
 def get_or_create_authenticator_user(
-    username: str, authenticator: Authenticator, user_details: dict = dict, extra_data: dict = dict
+    uid: str, authenticator: Authenticator, user_details: dict = dict, extra_data: dict = dict
```
AlanCoding, line 99: Can we add a doc blurb here about what `uid` is?

AlanCoding: https://github.com/python-social-auth/social-app-django/blob/af9ea55bbc057018056b1789c23c663e7c89d483/social_django/models.py#L45

relrod (approved):

PR #525, opened 2024-07-24
Fix SAML authenticator to set ACS url as CALLBACK_URL

john-westcott-iv:


mabashian (approved):

jctanner (approved):

rooftopcellist (approved):
We just tested this out on the hackathon openshift instance. It is working.

PR #526, opened 2024-07-26
Save simple default settings in a separate module

AlanCoding:
Working through https://github.com/ansible/galaxy_ng/pull/2202 with @jctanner I discovered it wasn't loading the settings. In short, I don't think we can rely on accurately having `INSTALLED_APPS` populated at the time we needed to load the defaults.

But we shouldn't need this. If an app wants to bypass our recommended system (which it is already doing), then it should have a constant data structure to import.

I wanted to do this with dictionaries too, but it seems we have special cases for each entry, which... I can't tough. So this is unfortunately incomplete.

### DIFF
--- a/ansible_base/lib/dynamic_config/dynamic_settings.py
+++ b/ansible_base/lib/dynamic_config/dynamic_settings.py
@@ -4,6 +4,8 @@
 #     Add a new requirements/requirements_<section>.in /even if its an empty file/
 #
 
+from ansible_base.lib.dynamic_config import default_settings
+
 # The org and team abstract models cause errors if not set, even if not used
 try:
     ANSIBLE_BASE_TEAM_MODEL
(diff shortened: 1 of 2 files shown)


rochacbruno (approved):
LGTM will make it easier to load the settings.

relrod (changes requested):
I'm not rejecting for any technical reason, but I just want to better understand the problem before this goes in.

i.e. What exactly is the problem with INSTALLED_APPS?

ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -26,26 +18,10 @@
...
-        'host_filter',
-        'count_disabled',
-        'no_truncate',
-        'limit',
-        'validate',
-    )
+for key, value in vars(_dab_default_settings.general).items():
+    if key in locals():
+        continue
+    locals()[key] = value
```
AlanCoding, line 24: FWIW, it is not suggested to do this

https://docs.python.org/3/library/functions.html#locals

> Note The contents of this dictionary should not be modified; changes may not affect the values of local and free variables used by the interpreter.

AlanCoding:
> i.e. What exactly is the problem with INSTALLED_APPS?

I'll link you the thing that ultimately needs to be fixed here

https://github.com/ansible/galaxy_ng/blob/b26816eaf9c1e34f4033835e09125013a5d182d3/galaxy_ng/app/dynaconf_hooks.py#L732

There's a method that needs to return a dictionary of the settings from DAB. It's probably already obvious to you that this poses an issue for any settings that have to get fine-grained merged. But I don't think galaxy_ng is using any part of DAB that requires that at this moment, and I'm not sure how it would work if we did have that situation.

But you mention `INSTALLED_APPS`. This would have to be included in local variables to pass the value, and that's just not the right way to do thing. I don't even really know if `split_settings` works within a function's variable scope, or if it might mess up something else.

AlanCoding:
For an immediate solution I prefer the other approach. We could do a version of this later (to define truly static settings), but it is not necessary.

PR #527, opened 2024-07-26
[WIP] Test for ansible_id in serialized assignments

AlanCoding:


john-westcott-iv:

AlanCoding:
I just rebased, if it passes the checks we should merge it to get that coverage.

AlanCoding:
Oh gosh, this does seem like a bug

```
    @pytest.mark.django_db
    def test_user_assignment_ansible_id(admin_api_client, inv_rd, rando, inventory):
        resource = Resource.objects.get(object_id=rando.pk, content_type=ContentType.objects.get_for_model(rando).pk)
        url = get_relative_url('roleuserassignment-list')
        data = dict(role_definition=inv_rd.id, content_type='aap.inventory', user_ansible_id=str(resource.ansible_id), object_id=inventory.id)
        response = admin_api_client.post(url, data=data, format="json")
        assert rando.has_obj_perm(inventory, 'change')
>       assert response.data['user_ansible_id'] == str(resource.ansible_id)
E       AssertionError: assert None == '60a212a1-6796-4e3b-a790-75a2d3e97bc8'
E        +  where '60a212a1-6796-4e3b-a790-75a2d3e97bc8' = str(UUID('60a212a1-6796-4e3b-a790-75a2d3e97bc8'))
E        +    where UUID('60a212a1-6796-4e3b-a790-75a2d3e97bc8') = <Resource: Resource object (350)>.ansible_id
```

If we don't include the user ansible_id, fine for now, I guess. But if we're returning None for its value, that would be a problem that should be fixed.

I should have better documented this before.

AlanCoding:
It looks like the list view also has null values

```
            "object_ansible_id": null,
            "role_definition": 2,
            "user": 5,
            "user_ansible_id": null
```

I thought this was only on creation. I guess this is consistent with what I wrote in the issue. That's a hot mess.

AlanCoding:
fixed elsewhere

PR #528, opened 2024-07-26
[get_fully_qualified_url] Fallback to crum request

relrod:
In the event where we a request isn't passed in to DRF `reverse()`, it will return a relative URL. This is unexpected, given the name of our wrapper, `get_fully_qualified_url()`.

In one case, in social auth completion, we weren't passing in a request and were getting a relative URL back. This is used as a default `CALLBACK_URL` in SAML. That field gets url-validated, and since we were ending up with a relative URL there, validation failed. This means that any attempt to create a SAML authenticator without a `CALLBACK_URL` explicitly defined, was failing validation and not saving.

With this patch, `get_fully_qualified_url()` defaults to django-crum's `get_current_request()`, if a request isn't passed in. It will still fall back to a relative URL if `get_current_request()` isn't able to identify a request (e.g. if `get_fully_qualified_url()` ends up being called somewhere in the lifespan of a CLI command where there's no request).

Test plan:
Added new test_app test to use the API to create a SAML authenticator with no `CALLBACK_URL` set. Before this patch, the test failed. After this patch, the test passes. Also added more test cases for `get_fully_qualified_url()`.

### DIFF
--- a/ansible_base/lib/utils/response.py
+++ b/ansible_base/lib/utils/response.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass
 from typing import Optional, Sequence
 from urllib.parse import urlparse
 
+from crum import get_current_request
 from django.http import StreamingHttpResponse
 from django.urls import reverse as django_reverse
 from rest_framework.reverse import reverse as drf_reverse
@@ -46,13 +47,16 @@ def get_fully_qualified_url(view_name: str, *args, **kwargs) -> str:
     Returns a fully qualified URL from the setting FRONT_END_URL or a DRF reverse if the setting is undefined
     NOTE: a DRF reverse could be relative or not depending on the request coming in
     '''
-    url = drf_reverse(view_name, *args, **kwargs)
     front_end_url = get_setting('FRONT_END_URL', None)
-    if not front_end_url:
-        return url
+    if front_end_url:
+        url = drf_reverse(view_name, *args, **kwargs)
+        url_pieces = urlparse(url)
+        return f"{front_end_url.rstrip('/')}{url_pieces.path}"
 
-    url_pieces = urlparse(url)
-    return f"{front_end_url.rstrip('/')}{url_pieces.path}"
+    if not kwargs.get('request'):
+        kwargs['request'] = get_current_request()
+
+    return drf_reverse(view_name, *args, **kwargs)
 
 
 def get_relative_url(view_name: str, *args, **kwargs) -> str:
--- a/test_app/tests/authentication/authenticator_plugins/test_saml.py
+++ b/test_app/tests/authentication/authenticator_plugins/test_saml.py
@@ -6,7 +6,7 @@ from django.conf import settings
 from ansible_base.authentication.authenticator_plugins.saml import AuthenticatorPlugin
 from ansible_base.authentication.session import SessionAuthentication
 from ansible_base.lib.utils.encryption import ENCRYPTED_STRING
-from ansible_base.lib.utils.response import get_relative_url
+from ansible_base.lib.utils.response import get_fully_qualified_url, get_relative_url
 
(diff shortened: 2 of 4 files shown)


ansible_base/lib/utils/response.py
```diff
@@ -46,13 +47,19 @@ def get_fully_qualified_url(view_name: str, *args, **kwargs) -> str:
...
+        url = drf_reverse(view_name, *args, **kwargs)
+        url_pieces = urlparse(url)
+        return f"{front_end_url.rstrip('/')}{url_pieces.path}"
+
+    if 'request' in kwargs and kwargs['request']:
+        request = kwargs.pop('request')
+    else:
+        request = get_current_request()
+
+    kwargs['request'] = request
```
john-westcott-iv: Non-blocking, could you just make this a `not` condition?
```suggestion
    if not kwargs.get('request', None):
        kwargs['request'] = get_current_request()
```

test_app/tests/authentication/authenticator_plugins/test_saml.py
```diff
@@ -212,8 +212,25 @@ def __init__(self):
...
+        "enabled": True,
+        "create_objects": True,
+        "remove_users": True,
+        "type": "ansible_base.authentication.authenticator_plugins.saml",
+        "configuration": saml_configuration,
+    }
+
+    url = get_relative_url("authenticator-list")
+    response = admin_api_client.post(url, data=authenticator_data, format="json", SERVER_NAME="dab.example.com")
+    assert response.status_code == 201, response.data
```
john-westcott-iv, line 229: Was this previously just a failure? Should we assert in here that the CALLBACK_URL is a fully qualified URL with a server we are expecting? No because we are testing it in test_get_fully_qualified_url is a fine answer for me.

john-westcott-iv (approved):

ansible_base/lib/utils/response.py
```diff
@@ -4,6 +4,7 @@
 from typing import Optional, Sequence
 from urllib.parse import urlparse
 
+from crum import get_current_request
```
AlanCoding, line 7: Talking it out loud - this import is okay because django-crum is in the DAB main requirements?

ansible_base/lib/utils/response.py
```diff
@@ -46,13 +47,19 @@ def get_fully_qualified_url(view_name: str, *args, **kwargs) -> str:
...
-    return f"{front_end_url.rstrip('/')}{url_pieces.path}"
+    if front_end_url:
+        url = drf_reverse(view_name, *args, **kwargs)
+        url_pieces = urlparse(url)
+        return f"{front_end_url.rstrip('/')}{url_pieces.path}"
+
+    if 'request' in kwargs and kwargs['request']:
+        request = kwargs.pop('request')
+    else:
+        request = get_current_request()
```
AlanCoding: This will still return None if `crum.CurrentRequestUserMiddleware` isn't in middleware right?

Perhaps a log statement to this effect would be nice to add.

AlanCoding (approved):
This is a crumy patch. But I guess that's okay.

mabashian (approved):

ansible_base/lib/utils/response.py
relrod: Yes, we list `django-crum` in requirements.in

test_app/tests/authentication/authenticator_plugins/test_saml.py
relrod: Done :+1: - and yes this failed previously :(

ansible_base/lib/utils/response.py
relrod: There are multiple ways this can return `None`. CLI/management commands is another. I didn't log here because I'm just assuming it can return `None` in some cases as normal behavior.

PR #529, opened 2024-07-26
Re-enable black and blacken files that got missed in later PRs

relrod:
It got disabled accidentally in #505


john-westcott-iv (approved):
Danke

PR #530, opened 2024-07-26
Enable/Disable authentication maps

dmzoneill:


ansible_base/authentication/utils/claims.py
```diff
@@ -62,9 +62,13 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
         trigger_result = TriggerResult.SKIP
         allowed_keys = TRIGGER_DEFINITION.keys()
         invalid_keys = set(auth_map.triggers.keys()) - set(allowed_keys)
+
+        if auth_map.enabled is False:
+            logger.warning(f"The AuthenticatorMap {auth_map.id} is disabled")
```
relrod: Maybe info instead? I feel like we're too quick to use warning sometimes - we should save that for things that aren't normal operating conditions. If we're supporting disabling authenticator maps, then that's a normal state for them to be in and warning about it seems excessive.

ansible_base/authentication/models/authenticator_map.py
```diff
@@ -88,3 +88,8 @@ class Meta:
             )
         ),
     )
+    enabled = models.BooleanField(
+        null=False,
+        default=False,
```
relrod: `False` here but `True` in the migration - is why CI is failing.

ansible_base/authentication/models/authenticator_map.py
```diff
@@ -88,3 +88,8 @@ class Meta:
             )
         ),
     )
+    enabled = models.BooleanField(
+        null=False,
+        default=True,
+        help_text=_("Enables or disables this authentication map"),
```
relrod:
```suggestion
        help_text=_("Enables or disables this authenticator map"),
```

test_app/tests/authentication/utils/test_claims.py
relrod, line 1: Can we have a test for what happens when it's disabled? Just something that checks that it doesn't get applied.

jay-steurer (approved):

john-westcott-iv (approved):

PR #531, opened 2024-07-29
Nest the dynamic settings inclusion into a method

AlanCoding:
Alternative to https://github.com/ansible/django-ansible-base/pull/526

Well, it's not really an alternative, we could do both of them (a module for static settings & a module for function-ified settings), but this would also address our immediate problem.

Overall idea:

This puts most of our existing logic around dynamic settings into a method `get_dab_settings`. This keeps the canonical method of including settings, this just adds another layer internally. Benefits:

 - maintains an _explicit_ interface, as opposed to an _implicit_ interface which is what split_settings `include` necessarily does
 - allows projects like galaxy_ng to import the method and just get a simple python dictionary as a result. In doing so, it can just invent whatever value it deems necessary for `INSTALLED_APPS` and it's up to it for any modified settings, which should be _more clear_ with how the `RuntimeError` is used here.
 - before, local variables were not really a thing, but now they are. For example, `drf_authentication_class` is just a variable that was never intended to be a setting. But the old dynamic include would have still put it in the settings. This makes things cleaner by maintaining inputs and outputs for the settings dynamic logic.


ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -25,229 +29,41 @@
...
+    installed_apps=INSTALLED_APPS,
+    rest_framework=REST_FRAMEWORK,
+    spectacular_settings=SPECTACULAR_SETTINGS,
+    authentication_backends=local_authentication_backends,
+    middleware=MIDDLEWARE,
+    oauth2_provider=OAUTH2_PROVIDER,
+).items():
+    if key in ANSIBLE_BASE_OVERRIDABLE_SETTINGS:
+        ANSIBLE_BASE_OVERRIDDEN_SETTINGS.append(key)
+        locals()[key] = value
```
rochacbruno, line 63: AFAIK, mutating `locals()` will not have side effects on local variables, is that the intention?

I have used `setattr(sys.modules[__name__], name, value)` to mutate local variables, but maybe you really want to avoid the global side effect it has.

ansible_base/lib/dynamic_config/settings_logic.py
```diff
@@ -0,0 +1,255 @@
...
+    spectacular_settings: Optional[dict] = None,
+    authentication_backends: Optional[list[str]] = None,
+    middleware: Optional[list[str]] = None,
+    oauth2_provider: Optional[dict] = None,
+) -> dict:
+    dab_data = {}
+
+    # The org and team abstract models cause errors if not set, even if not used
+    dab_data['ANSIBLE_BASE_TEAM_MODEL'] = 'auth.Group'
+    dab_data['ANSIBLE_BASE_ORGANIZATION_MODEL'] = 'auth.Group'
```
rochacbruno, line 23: This is already set on galaxy `settings` https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/settings.py#L390

So if we call `get_dab_settings` we need to ensure the existing variables are not overridden, and when the existing variable is a data structure like a `dict` or `list` we need to ensure it has the values merged.

On galaxy the merging is controlled by the merging markers.

What if

```python
def get_dab_settings(
    ...
    dab_data: Optional[dict] = None
): -> dict:
    if dab_data is None:
        dab_data = {}

    dab_data.setdefault("ANSIBLE_BASE_TEAM_MODEL", "auth.group")
    ...

    return dab_data
```

This does not solve the merging part, but at least allows us to pass a prepopulated `dab_data` that will be changed only if defaults doesn't exist yet.

That would also prevent override https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/settings.py#L226 and other variables like that one.

And for the merging parts we try to resolve on dynaconf side.

For authentication backends for example we cant simply override from DAB because we have very specific logic for that https://github.com/ansible/galaxy_ng/blob/b26816eaf9c1e34f4033835e09125013a5d182d3/galaxy_ng/app/settings.py#L333-L349

ansible_base/lib/dynamic_config/dynamic_settings.py
AlanCoding: Did you mean that it _can_ have side effects on local variables? I'm aware that those hazards exist, I'm just not sure if they matter in this context, where we won't be referencing them later. We only need to set variables for another module to include and pick them up.

ansible_base/lib/dynamic_config/settings_logic.py
AlanCoding: My intent is that `get_dab_settings` is a dictionary that _only represents_ default values to begin with. If you look at the dynamic_settings module, it skips any variables that are defined locally.

The current behavior is already fairly wrong/inconsistent on this point. For some variables, we reference it and only set it when getting a `NameError` exception, but for others we don't. So if you set a variable before doing the include, the dynamic_settings (currently, in devel) will overwrite it. This formalizes the practice so that it won't overwrite a user setting even if you set it before doing the include.

I would really prefer that `get_dab_settings` does not have knowledge of the entire settings. Because I want DAB developers to have to declare what variables from the user's settings it needs to see. I don't want DAB to be able to reference `dab_data["SOME_NEW_KEY"]` so that they get the value of SOME_NEW_KEY from the user settings. If that need does arise, that's fine, but it's a notable changelog-worthy kind of thing.

ansible_base/lib/dynamic_config/dynamic_settings.py
rochacbruno: nevermind, I just verified that `locals()[key] = value` has the same effect as `sys.modules[__name__].key = value`

ansible_base/lib/dynamic_config/settings_logic.py
rochacbruno: Ok, so after getting the `dab_data` we can get only the complementary keys to populate galaxy settings.

rochacbruno (approved):

rochacbruno:
Ci failing after rebasing.

The only diff was 
```python
dab_data['OAUTH2_BACKEND_CLASS'] = 'ansible_base.oauth2_provider.authentication.OAuthLibCore'
```

taken from #537

rochacbruno (approved):

relrod (approved):

mabashian (approved):

PR #535, opened 2024-07-30
Altering redis client - breaking change

john-westcott-iv:
Altering clustered and clustered_hosts parameters to mode and redis_hosts
Mode must be one of cluster or standalone or an ImproperlyConfigured exception is raised

### DIFF
--- a/ansible_base/lib/redis/client.py
+++ b/ansible_base/lib/redis/client.py
@@ -42,8 +42,8 @@ class RedisClient(DefaultClient):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         connection_kwargs = self._get_client_args()
-        self.clustered = connection_kwargs.get('clustered', False)
-        self.clustered_hosts = connection_kwargs.get('clustered_hosts', '')
+        self.mode = connection_kwargs.get('mode', 'standalone')
+        self.redis_hosts = connection_kwargs.get('redis_hosts', '')
 
     def connect(self, index: int = 0) -> Union[DABRedis, DABRedisCluster]:
         """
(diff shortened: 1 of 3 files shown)


docs/lib/redis.md
```diff
@@ -13,8 +13,8 @@ To leverage this client you can setup your Django cache like:
             "OPTIONS": {
                 "CLIENT_CLASS": "ansible_base.lib.redis.RedisClient",
                 "CLIENT_CLASS_KWARGS": {
-                    "clustered": <true or false>
-                    'clustered_hosts': "<host 1>:<port>,<host 2>:<port>,[...],<host N>:<port>",
+                    "mode": <cluser or standalone>
```
jay-steurer: cluster :police_car:

ansible_base/lib/redis/client.py
```diff
@@ -106,12 +106,12 @@ def _get_hosts(self) -> None:
 
         translated_generic_exception = ImproperlyConfigured(_('Unable to parse cluster_hosts, see logs for more details'))
```
bhavenst: cluster_hosts -> redis_hosts in exception text?

test_app/tests/lib/redis/test_client.py
```diff
@@ -67,8 +81,9 @@ def test_redis_client_confirm_connect_does_not_change_options():
     ],
 )
 def test_redis_client_location_parsing(url, host, port, username, password, db, raises, socket):
-    redis_cache = RedisCache(url, {})
-    client = RedisClient(url, {}, redis_cache)
+    args = {'OPTIONS': {'CLIENT_CLASS_KWARGS': {'mode': 'standalone'}}}
```
bhavenst, line 84: Looks like no mode specified should still result in standalone since it's defaulted that way earlier, so maybe test both here?

bhavenst (approved):
Just minor comments..

mabashian (approved):

rooftopcellist (approved):

ansible_base/lib/redis/client.py
john-westcott-iv: Great catch, fixed!

test_app/tests/lib/redis/test_client.py
john-westcott-iv: There are other tests up above which should cover this.

PR #541, opened 2024-07-31
Fix authenticator order

trucdg:
Fixes AAP-28034:
**Changes Made:**
- Added a `get_next_authenticator_order` function to find the next order value, which is equals to `max(order) + 1`
- Added unit test

**Testing Steps:**
1. Create a new authenticator, and confirm the followings:
- If given an order, the created authenticator's order = given order value
- If no order was given,  the created authenticator's order = max(order) + 1
2. Update/ Delete the order of the current authenticators
3. Back to step 1 and confirm the authenticator order is as expected


ansible_base/authentication/models/authenticator.py
```diff
@@ -5,6 +5,14 @@
 from ansible_base.lib.utils.models import prevent_search
 
 
+def get_next_authenticator_order():
+    """
+    Returns the next authenticator order, which is equals to max(order) + 1
+    """
+    largest_order_authenticator = Authenticator.objects.all().order_by('-order').first()
+    return largest_order_authenticator.order + 1 if largest_order_authenticator else 1
```
relrod:
```suggestion
    largest_order_authenticator = Authenticator.objects.values('order').order_by('-order').first()
    return largest_order_authenticator['order'] + 1 if largest_order_authenticator else 1
```

Small nitpick, but better to just select the field we need when we can.

relrod (approved):
One small optimization inline, otherwise lgtm!

mabashian (approved):

PR #544, opened 2024-08-01
Fix auto-creation of CALLBACK_URL for AzureAD

AparnaKarve:
Fix auto-creation of CALLBACK_URL for AzureAD authentication by subclassing `SocialAuthValidateCallbackMixin` in the AzureAD AuthenticatorPlugin class

Fixes https://issues.redhat.com/browse/AAP-27572

### DIFF
--- a/test_app/tests/authentication/authenticator_plugins/test_azuread.py
+++ b/test_app/tests/authentication/authenticator_plugins/test_azuread.py
@@ -1,5 +1,7 @@
 from unittest import mock
 
+import pytest
+
 from ansible_base.authentication.session import SessionAuthentication
 from ansible_base.lib.utils.response import get_relative_url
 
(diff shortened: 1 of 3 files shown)


test_app/tests/authentication/authenticator_plugins/test_azuread.py
```diff
@@ -23,6 +25,45 @@ def test_azuread_auth_successful(authenticate, unauthenticated_api_client, azure
...
+    ],
+)
+def test_azuread_endpoint_url_validation(
+    admin_api_client,
+    key,
+    secret,
+    expected_status_code,
+    expected_error,
+):
+    callback_url = "http://testserver/api/social/complete/ansible_base-authentication-authenticator_plugins-azuread__azuread-test/"  # noqa
```
john-westcott-iv: Instead of constructing this URL can we call `get_fully_qualifgied_url` like the validate method of SocialAuthValidateCallbackMixin would (or even just call validate itself)? That way if the function ever changes this test won't start failing. Because we are really testing that the URL got auto-generated and not what the content of the generated URL should be.

test_app/tests/conftest.py
```diff
@@ -80,7 +80,7 @@ def clear_content_type_cache():
 @pytest.fixture
 def azuread_configuration():
     return {
-        "CALLBACK_URL": "https://localhost/api/gateway/callback/github_test/",
+        "CALLBACK_URL": "https://localhost/api/gateway/callback/azuread_test/",
```
john-westcott-iv, line 83: Nice find!

john-westcott-iv (approved):
Approving with minor comment.

mabashian (approved):

AparnaKarve:
@john-westcott-iv Can you please review the last commit where I have addressed your feedback?
If that looks good, I will make a similar change to the other google oauth2 PR.

Thanks.

test_app/tests/authentication/authenticator_plugins/test_azuread.py
```diff
@@ -23,6 +25,46 @@ def test_azuread_auth_successful(authenticate, unauthenticated_api_client, azure
...
+
+    url = get_relative_url("authenticator-list")
+    response = admin_api_client.post(url, data=data, format="json")
+    assert response.status_code == expected_status_code
+    if expected_error:
+        assert response.json() == expected_error
+    else:
+        slug = response.data["slug"]
+        expected_path = get_fully_qualified_url('social:complete', kwargs={'backend': slug})
+        assert response.json()['configuration']['CALLBACK_URL'] == f"http://testserver{expected_path}"
```
john-westcott-iv: You can make get_fully_qualified_url auto-prepend a server by setting `FRONT_END_URL`. Something like this should work:

```suggestion
        slug = response.data["slug"]
        with override_settings(FRONT_END_URL='http://testserver/'):
            expected_path = get_fully_qualified_url('social:complete', kwargs={'backend': slug})
            assert response.json()['configuration']['CALLBACK_URL'] == expected_path
```

You would also need this import:
```
from django.test import override_settings
```

It would be bad to wrap your entire test in that as well because that should also effect the creation of the CALLBACK_URL so if the mock server changes from `http://testserver` to something else your function wouldn't care because of your override setting.

john-westcott-iv (approved):
Approving assuming tests work correctly.

PR #545, opened 2024-08-02
Add method to get status from redis

john-westcott-iv:
AAP-28156

TODO:
- [x] Add timeouts (when avaialable)
- [x] Unit tests
- [x] Test with different scenarios of Redis (nodes up/down/etc)

The easiest way to test the functionality is to go into shell_plus and execute the following statements:
```
from ansible_base.lib.redis.client import get_redis_status
url = settings.CACHES['default']['LOCATION']
kwargs = settings.CACHES['default']['OPTIONS']['CLIENT_CLASS_KWARGS']
get_redis_status(url, **kwargs)
```


jay-steurer (approved):

jay-steurer:
Fix redis cluster dev env on Linux
https://github.com/ansible/aap-gateway/pull/500

mabashian (approved):

jshimkus-rh (approved):

jay-steurer (approved):

ansible_base/lib/utils/timeout.py
```diff
@@ -0,0 +1,32 @@
+import signal
+
+
+class Timeout:
```
relrod: In general I think that's an okay way to time out - careful about threads though, I don't know how this works in the context of uwsgi worker threads...

> Python signal handlers are always executed in the main Python thread of the main interpreter, even if the signal was received in another thread.

- https://docs.python.org/3/library/signal.html#signals-and-threads

I'm okay with us trying this, just something to watch out for ^

relrod (approved):

test_app/tests/lib/utils/test_timeout.py
```diff
@@ -0,0 +1,16 @@
+from time import sleep
+
+import pytest
+
+from ansible_base.lib.utils.timeout import Timeout
+
+
+def test_timeout_success():
```
relrod: Maybe use `time.time()` to make sure this doesn't take 10 seconds when it shouldn't.

ansible_base/lib/utils/timeout.py
cutwater: What @relrod has said is correct. This will work only in the main thread of python interpreter. I would suggest rather determining and fixing the root cause instead of using an unstable workaround that may not work depending on how the application is executed.

john-westcott-iv: I was able to track down why it didn't work before. We had a retry mechanism baked in. Removing that has helped although there are still some edge cases where we don't get the 4 sec timeout behavior. I'm not in love with that but its safer overall.

test_app/tests/lib/utils/test_timeout.py
john-westcott-iv: This is going away.

PR #548, opened 2024-08-03
Reverse-syncing of resources to a resource server

relrod:
This implements reverse syncing from services to a resource server.

In the interest of trying to prevent as many failure scenarios as possible (though this is NOT failproof as @AlanCoding points out below in the comments), we wrap the REST call and the local commit in a transaction (unless a transaction is already present).

For modules which sync (users, orgs, teams), their `save()` method is monkeypatched to wrap the whole save in a transaction, if necessary.

A pre_delete and post_save signal is used for doing the actual sync.

In order to prevent sync loops, we sneak an internal attribute on instances that we create/fetch on requests from a resource server. In the syncing function, we look for this attribute and abort syncing if it's there and `True`.

Unit tests (somewhat gnarly ones) are provided. They attempt to ensure that we're calling the right REST client methods when we need to, bailing out early when we need to, rolling back when we need to, and committing when we need to. Coverage should be pretty darn close to 100%.

A new setting, `DISABLE_RESOURCE_SERVER_SYNC` works to disable the reverse syncing. We use it in `test_app` to avoid syncing on all but the specific reverse syncing tests (the ones written in this PR).


ansible_base/resource_registry/apps.py
```diff
@@ -103,6 +114,33 @@ def connect_resource_signals(sender, **kwargs):
...
+
+                # Avoid late binding issues
+                def save(self, *args, _original_save=cls._original_save, **kwargs):
+                    with transaction.atomic():
+                        # We need to know if this is a new object before we save it
+                        action = "create" if self._state.adding else "update"
+                        # Save so we get an ansible_id if it's a new object
+                        _original_save(self, *args, **kwargs)
+                        # Send the object to the resource server
+                        handlers.sync_to_resource_server(self, action)
```
AlanCoding: Why this, as opposed to just connecting your method to the `post_save` signal? Or even adding it on to the existing `handlers.update_resource` that already runs post-save?

You can pass a value from pre_save to post_save if you really need to... just stashing it on the object if you really need to.

That won't allow you to wrap it in a transaction, but I'd argue that's the responsibility of the app using DAB, and wrapping things in another transaction (really a save-point) is only going to sow more chaos.

AlanCoding:
Can we talk out the specific scope of this?

This is strictly for CRUD of a specific shared item, meaning POST/PATCH/DELETE of an org/team/user in a service.

When that CRUD happens, this is expected to fire off a request to POST/PATCH/DELETE to the corresponding endpoint under the `/service-index/resources/` endpoint in the resource provider? Good. It's really important to write that out and get everyone on board. One sentence, we should repeat it many times over.

To wrap up the scope, what I need to know is then, what is expected of the resource provider? Is it expected that the resource provider then _syncs that change to other services_? Either way, you have more explaining to do. This can't be left unsaid.

I'll continue with "no", the resource provider will not sync that change to the other services. This will leave the cluster out-of-sync until those other services run the 15-minute periodic resource sync, originally by @rochacbruno. I would agree that's the best answer, but it's not obvious, and to the best of my knowledge, this is the first description of such a plan.

ansible_base/resource_registry/apps.py
relrod: See this commit: https://github.com/ansible/django-ansible-base/pull/548/commits/e60196c850b02c03cf073ce8ca1c0962d9cdf0f8

Using a signal was the initial approach. As you mention, we can't wrap it in a transaction which is what we're really trying to do here. The original approach got close - it opened a transaction in pre_save, and closed it in post_save. But if post_save never gets called (if `save()` fails), then the transaction would never get closed.

So I moved to this approach of monkeypatching `save()` instead to wrap it in a transaction.

> but I'd argue that's the responsibility of the app using DAB

That puts a lot of onus on each app vs just doing it in one place and letting each app inherit the behavior.

> is only going to sow more chaos

Can you elaborate a bit here?

relrod:
@AlanCoding yes, this all needs to be documented, this is still a prototype and WIP, that's why it's still a draft :smile:

You're correct that, to the best of my knowledge, there's no plan to trigger an immediate sync from the resource server to the other services, so then yes there would be a period of time where the cluster will be out of sync until the services all check in for an update.

ansible_base/resource_registry/apps.py
AlanCoding: There's no benefit to nesting more than 1 transaction in a request, because the outcome can only be that it rolls back and returns an error code. I had to introduce a transaction for RBAC checking of related permissions for a few reasons:

https://github.com/ansible/django-ansible-base/blob/95f88964e2cea5af0712c91d9de758aa6f4651e4/ansible_base/rbac/api/related.py#L123-L125

I probably shouldn't have done it that way, but I knew where it would be used, eda-server, which was not enabling the ATOMIC_TRANSACTIONS setting, and AWX is. I believe it would be better that everywhere _DAB_ wants to introduce a transaction (including the code I wrote myself that I linked there), it first checks to see if it is already in a transaction, and if it is, doesn't.

But also, assume your transaction here is necessary, and think about what that implies. That means you POST to create an object and get a 400. Sure, that _could_ mean that we left a resource inconsistency. But there's a _much more basic_ bug here - that a request returning a 400 still created the object. There's really not much wiggle room there. That's a bug.

relrod: > it first checks to see if it is already in a transaction

Okay, we can check `connection.in_atomic_block`. It might not always be reliable (I think we can be in a transaction beyond being in an atomic block), but it might help in some or even most cases.

> that a request returning a 400 still created the object

I'm not sure I understand this, I think I need a concrete example. I'm assuming you mean if we get a 400 back from the resource server when we try to replicate the object to it. But that should always mean the request failed and the object didn't get saved on the resource server. Is there a case where that isn't true? I think I'm misunderstanding.

AlanCoding: > I'm not sure I understand this, I think I need a concrete example. I'm assuming you mean if we get a 400 back from the resource server when we try to replicate the object to it.

No, that's not what I'm talking about. Imagine there is no resource syncing at all.

Say you make a POST to create an organization in eda-server, for an example. Now imagine that you get an error code, that could be 400, 500, or something else. But not a 201. After getting that response, check the organization list. Was the organization created?

This question isn't irrelevant, because it relates strongly to the "why are you using a transaction?" question.

relrod: Well yes that would be a bug in the app. But I would argue that _if_ the app is broken and creates the org anyway, then it should also still do so in the resource server. The whole point here is to maintain _internal_ consistency across the cluster. Even if we do something stupid, let's make sure that stupid thing is synced.

relrod: I added a commit that looks at `connection.in_atomic_block` before deciding to wrap the block in a transaction. It's a util function that we can use elsewhere, as well.

AlanCoding: > But I would argue that if the app is broken and creates the org anyway, then it should also still do so in the resource server.

Then let's say it's a 500 error due to a `post_save` signal that errors. That, combined with requests being non-atomic, explain why the object gets created in the database when it shouldn't have.

But here's the thing - the object is incorrectly created, but it is also not synced to the resource server. The `post_save` signal fires as a part of `_original_save(self, *args, **kwargs)`, which happens before the syncing happens. Since it's an uncaught exception it doesn't ever bother syncing the object. It's created in the local server and not present in the resource server. That's _with_ the change here.

It's also possible that it had an error in the view or serializer finalization logic, in which case it would get synced to the resource server. There's no consistency in any case.

So I'm still lacking an original reason for the transaction here.

relrod: I'm really confused and missing something.

If post_save fails, the [original transaction from save()](https://github.com/django/django/blob/main/django/db/models/base.py#L989) will get rolled back, and the exception will get re-raised. At that point the outer transaction - the one I'm adding here, will see the exception that was re-raised and rollback and (itself) reraise, and that will happen before we ever call `sync_to_resource_server()`, right?

Here's my test:

1. Replace all of `sync_to_resource_server()` with this:
    ```python
    def sync_to_resource_server(instance, action):
        raise exception("sync_to_resource_server threw this")
    ```
2. In test_app:
```
In [1]: def bad_post_save(sender, **kwargs):
   ...:     raise Exception("oh no")
   ...: 

In [2]: from django.db.models.signals import post_save

In [3]: post_save.connect(bad_post_save, Organization)

In [4]: o = Organization.objects.first()

In [5]: o.save()
Exception                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 o.save()

File ~/dev/ansible/aap-gateway/django-ansible-base/ansible_base/resource_registry/apps.py:128, in connect_resource_signals.<locals>.save(self, _original_save, *args, **kwargs)
    126 action = "create" if self._state.adding else "update"
    127 # Save so we get an ansible_id if it's a new object
--> 128 _original_save(self, *args, **kwargs)
    129 # Send the object to the resource server
    130 sync_to_resource_server(self, action)

File ~/dev/ansible/aap-gateway/django-ansible-base/ansible_base/lib/abstract_models/common.py:155, in AbstractCommonModel.save(self, *args, **kwargs)
    152     field_value = getattr(self, field, None)
    153     setattr(self, field, ansible_encryption.encrypt_string(field_value))
--> 155 return super().save(*args, **kwargs)

File ~/dev/ansible/aap-gateway/django-ansible-base/ansible_base/lib/abstract_models/common.py:133, in CreatableModel.save(self, *args, **kwargs)
    130     if self.created_by is None:
    131         self.created_by = current_user_or_system_user()
--> 133 return super().save(*args, **kwargs)

File ~/dev/ansible/aap-gateway/django-ansible-base/ansible_base/lib/abstract_models/common.py:101, in ModifiableModel.save(self, *args, **kwargs)
     98     if has_update_fields:
     99         kwargs['update_fields'] = update_fields
--> 101 return super().save(*args, **kwargs)

File ~/dev/ansible/dab-venv/lib/python3.12/site-packages/django/db/models/base.py:814, in Model.save(self, force_insert, force_update, using, update_fields)
    811     if loaded_fields:
    812         update_fields = frozenset(loaded_fields)
--> 814 self.save_base(
    815     using=using,
    816     force_insert=force_insert,
    817     force_update=force_update,
    818     update_fields=update_fields,
    819 )

File ~/dev/ansible/dab-venv/lib/python3.12/site-packages/django/db/models/base.py:892, in Model.save_base(self, raw, force_insert, force_update, using, update_fields)
    890 # Signal that the save is complete
    891 if not meta.auto_created:
--> 892     post_save.send(
    893         sender=origin,
    894         instance=self,
    895         created=(not updated),
    896         update_fields=update_fields,
    897         raw=raw,
    898         using=using,
    899     )

File ~/dev/ansible/dab-venv/lib/python3.12/site-packages/django/dispatch/dispatcher.py:177, in Signal.send(self, sender, **named)
    170 if (
    171     not self.receivers
    172     or self.sender_receivers_cache.get(sender) is NO_RECEIVERS
    173 ):
    174     return []
    176 return [
--> 177     (receiver, receiver(signal=self, sender=sender, **named))
    178     for receiver in self._live_receivers(sender)
    179 ]

Cell In[1], line 2, in bad_post_save(sender, **kwargs)
      1 def bad_post_save(sender, **kwargs):
----> 2     raise Exception("oh no")

Exception: oh no

2024-08-06 19:22:15,205 DEBUG    []  asyncio Using selector: EpollSelector
In [6]: 
 
```

The exception in `sync_to_resource_server()` is never thrown because the function is never called.

So I think I still don't quite understand the issue. I'm trying to, I promise.

AlanCoding:
@newswangerd to document part of my hesitation here, AWX has a well-known category of bugs that only happen on commit. Example:

https://github.com/ansible/awx/issues/11165

```
tools_awx_1     | awx-uwsgi stdout |   File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/backends/base/base.py", line 262, in commit
tools_awx_1     | awx-uwsgi stdout |     self._commit()
tools_awx_1     | awx-uwsgi stdout |   File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/backends/base/base.py", line 240, in _commit
tools_awx_1     | awx-uwsgi stdout |     return self.connection.commit()
tools_awx_1     | awx-uwsgi stdout |   File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/utils.py", line 89, in __exit__
tools_awx_1     | awx-uwsgi stdout |     raise dj_exc_value.with_traceback(traceback) from exc_value
tools_awx_1     | awx-uwsgi stdout |   File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/backends/base/base.py", line 240, in _commit
tools_awx_1     | awx-uwsgi stdout |     return self.connection.commit()
tools_awx_1     | awx-uwsgi stdout | django.db.utils.IntegrityError: insert or update on table "main_host" violates foreign key constraint
```

This kind of bug might be racy, but it's absolutely not at all uncommon. I mention it because it's a time when errors happen _between_ the `sync_to_resource_server(self, action)` call and the closing of the transaction, and because it's intractably unavoidable. So it's reasonable to imagine that the sync is successful, and then the transaction still gets rolled back, leading to an inconsistency.

I just don't want promises that this is going to be fully transactionally robust, because it simply can never be.

newswangerd:
@AlanCoding This is a pretty big bummer. I had naively assumed that the transaction would check for foreign key constraints before commit.

Over all, I am more concerned about bugs that could cause the resource to get created in the services and NOT on the resource server. Anything that is created on the resource server will ultimately get pulled down to each service, so it will be eventually consistent. The big problem with this is that if you try to create the resource in the service, and it already was created in the resource server, the uniqueness constraint will fail on the resource server and prevent you from creating it locally.

I was thinking we could resolve this problem by switching the `sync_to_resource_server` to be a `get_or_create` operation. That way if the resource already exists on the resource server, the function will resolve in a sync instead of a create. There may be problems here that we have not considered, but I'd like the two of you (@AlanCoding and @relrod) to spend some time considering downsides for this approach.

One less common, but possibly nastier class of problems that could result from this is failures during updates. This will only matter in cases where names (or usernames) are updated, which is not a common thing to do, but the result would be that the resource is renamed in Gateway, but not on the service. I think this will get resolved during the periodic sync, but there will be a maximum of 15 minutes where the names will be out of sync on the two systems and potentially causing confusion.

ansible_base/resource_registry/utils/sync_to_resource_server.py
```diff
@@ -0,0 +1,64 @@
...
+        ansible_id=ansible_id,
+        service_id=instance.resource.service_id,
+        resource_data=data,
+    )
+
+    try:
+        if action == "create":
+            client.create_resource(body)
+        elif action == "update":
+            client.update_resource(ansible_id, body)
```
newswangerd, line 95: No need to do this now, but there are some optimizations we can make around updates. We only need to update the resource server if one of the shared attributes changes.

AlanCoding:
> Over all, I am more concerned about bugs that could cause the resource to get created in the services and NOT on the resource server. Anything that is created on the resource server will ultimately get pulled down to each service, so it will be eventually consistent.

Yes, I agree. The scenario I mentioned is actually relatively okay. That scenario is:

 - Object is created in transaction
 - Object is created in resource provider
 - transaction is rolled back so object no longer exists locally

In that case we would just wait for the periodic_resource_sync. And in the case of AWX, we could just schedule the sync task async in the request failure handler (hopefully based on some flag indicating that the Gateway resource was created). So this is actually all-okay.

> I was thinking we could resolve this problem by switching the `sync_to_resource_server` to be a `get_or_create` operation.

It doesn't do this now? Are you saying it doesn't create resources? Agree, I'd prefer it to create items from the resource provider if not locally present.

> possibly nastier class of problems that could result from this is failures during updates. This will only matter in cases where names (or usernames) are updated

I'd again point to the option I mentioned here, scheduling `periodic_resource_sync` in [finalize_response](https://github.com/ansible/awx/blob/37b7a693032edc346b20afbb32447dacd00476d6/awx/api/generics.py#L192). We probably can't do this in DAB shared base views or anything like that, because the means of scheduling an async task is completely service-specific.

Or maybe, instead of `finalize_response`, hook into

https://docs.djangoproject.com/en/5.0/ref/signals/#got-request-exception

Using signals might give us a better shot at sharing data to know whether or not the request has made a change in Gateway.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -120,10 +121,15 @@ def parse_jwt_token(self, request):
                 logger.warn(f"New user {self.user.username} created from JWT auth")
             except IntegrityError as exc:
                 logger.warning(f'Existing user {self.token["user_data"]} is a conflict with local user, error: {exc}')
-                self.user, created = get_user_model().objects.update_or_create(
-                    username=self.token["user_data"]['username'],
-                    defaults=user_defaults,
-                )
```
relrod, line 126: This is one place where we didn't use `Resource.*_resource()` which created problems since we need a way to inject an internal attribute so we don't try to sync back to the resource server in this case. The below is my attempt to replicate this logic and also be able to inject the attribute before we `.save()`.

ansible_base/resource_registry/utils/sync_to_resource_server.py
relrod: @newswangerd I'm actually going to say this is a blocker and we should absolutely do this. As it stands, if the sync fails for any reason, a user can't log in to AWX (because `last_login` gets updated, the sync fails, the request bombs out).

I'll get this fixed.

relrod: Done, I think.

ansible_base/resource_registry/utils/sync_to_resource_server.py
```diff
@@ -0,0 +1,92 @@
...
+
+    if action != "delete":
+        ansible_id = instance.resource.ansible_id
+
+    resource_type = instance.resource.content_type.resource_type
+    data = resource_type.serializer_class(instance).data
+    body = ResourceRequestBody(
+        resource_type=resource_type.name,
+        ansible_id=ansible_id,
+        service_id=instance.resource.service_id,
```
newswangerd: Don't set the service_id in the API call. Doing so will tell the resource server that it does not manage the resource anymore, which will prevent it from getting synced to other servers.

For `create`, we also need to read the `service_id` from the server's response and set it on the local copy of the `Resource` object. By default `service_id` is set to the service creates the resource, so any resources created on the service will have it set to their ID. Since resources created here are synced to the resource server, we need to indicate that it is managed by the resource server.

newswangerd (changes requested):

AlanCoding:
I would like to merge https://github.com/relrod/django-ansible-base/pull/1 in with the rest of this. Because that will avoid nested transaction in practice for eda-server requests specifically, which I'm pretty specific about not wanting.

newswangerd (approved):

mabashian (approved):

ansible_base/jwt_consumer/common/auth.py
AlanCoding: This is an error case of dubious functionality to begin with. It might be better to just fail the authentication at this point. And you really changed the intent here, going from `update_or_create` to just `.get`. This code is reached when the "admin" user already exists, but with a different ansible_id. There are no good options in that case, but what this ends up doing looks worse than doing nothing.

AlanCoding:
Whenever you get into full-cycle integration testing, what I'm predicting is:

Any logic connected to `post_migrate` signals that modify users/teams/orgs will try to sync any of the changes that it makes to the resource provider. Same for any logic ran manually after migration. This is a problem for installation ordering. Components must finish their migrations and then the resource provider must merge their data (`migrate_service_data`). What is expected run, order-wise, before the migration in the install story? This, for one:

https://github.com/ansible/awx/blob/devel/awx/main/management/commands/create_preload_data.py

Should we turn off the syncing while that is running?

ansible_base/resource_registry/rest_client.py
```diff
@@ -104,6 +104,8 @@ def _make_request(
             kwargs["stream"] = stream
 
         resp = requests.request(**kwargs)
+        logger.info(f"Response status code from {url}: {resp.status_code}")
```
AlanCoding: did you intend to leave this?

AlanCoding (approved):
I'm sure this is still going to break everything, but @relrod did a good job writing it defensively. I hate to see another introduction of something like `existing_instance = sender.objects.get(pk=instance.pk)` and I don't feel like we know yet how that will chain with existing model logic. But this is doing the minimal possible, because noisy signals is _less_ of a liability than making requests to an external system in the model layer.

At minimum I think we should card some work about the JWT auth, where I don't think it's doing what it should be doing for getting the user.

On the create_preload_data, I also anticipate we will very soon need more options for turning this off temporarily for a block of code beyond just setting it on the objects. Objects load other objects, load other objects, etc...

ansible_base/resource_registry/rest_client.py
relrod: I did, although I could be convinced to make it debug instead of info. I thought it could be useful for debugging later on.

AlanCoding (approved):

PR #558, opened 2024-08-07
Skip further validation on team parent model if it has no parent model.

jctanner:
https://issues.redhat.com/browse/AAH-3355

### DIFF
--- a/ansible_base/rbac/validators.py
+++ b/ansible_base/rbac/validators.py
@@ -183,15 +183,19 @@ def validate_team_assignment_enabled(content_type: Optional[Model], has_team_per
     if not team_team_allowed and content_type.model == team_model_name:
         raise ValidationError('Assigning team permissions to other teams is not allowed')
 
-    team_parent_model_name = permission_registry.get_parent_model(permission_registry.team_model)._meta.model_name
-    if not team_org_allowed and content_type.model == team_parent_model_name:
-        raise ValidationError(f'Assigning {team_parent_model_name} permissions to teams is not allowed')
+    # Not sure what the "parent" for a Team model would be, but it's unlikely to match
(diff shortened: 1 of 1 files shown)


ansible_base/rbac/validators.py
```diff
@@ -183,15 +183,19 @@ def validate_team_assignment_enabled(content_type: Optional[Model], has_team_per
...
-    team_parent_model_name = permission_registry.get_parent_model(permission_registry.team_model)._meta.model_name
-    if not team_org_allowed and content_type.model == team_parent_model_name:
-        raise ValidationError(f'Assigning {team_parent_model_name} permissions to teams is not allowed')
-
-    if (not team_org_member_allowed) and has_org_member:
-        raise ValidationError(f'Assigning {team_parent_model_name} member permission to teams is not allowed')
-
-    if not team_org_team_allowed and content_type.model == team_parent_model_name and has_team_perm:
-        raise ValidationError(f'Assigning {team_parent_model_name} permissions that manage other teams is not allowed')
+    # Not sure what the "parent" for a Team model would be, but it's unlikely to match
```
AlanCoding: Organization. The parent model for a team is organization in all familiar contexts. This is written weirdly generally so that it in theory, allows another type of parent, or no parent (your case).

ansible_base/rbac/validators.py
```diff
@@ -183,15 +183,19 @@ def validate_team_assignment_enabled(content_type: Optional[Model], has_team_per
...
+    # the content type model being sent for a role assignment ...
```
AlanCoding: It would very commonly match, as both the assignment model and the team parent model would be organization

AlanCoding (approved):
Good except for what the comment says.

AlanCoding (approved):

PR #559, opened 2024-08-08
Allow removing permissions even when they are invalid

AlanCoding:
Discovered in testing in https://github.com/ansible/galaxy_ng/pull/2202

```
pulp_1          |   File "/src/galaxy_ng/galaxy_ng/app/signals/handlers.py", line 313, in delete_pulp_user_role
pulp_1          |     rd.remove_global_permission(instance.user)
pulp_1          |   File "/usr/local/lib/python3.11/site-packages/ansible_base/rbac/models.py", line 184, in remove_global_permission
pulp_1          |     return self.give_or_remove_global_permission(actor, giving=False)
pulp_1          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pulp_1          |   File "/usr/local/lib/python3.11/site-packages/ansible_base/rbac/models.py", line 188, in give_or_remove_global_permission
pulp_1          |     raise RuntimeError('Role definition content type must be null to assign globally')
pulp_1          | RuntimeError: Role definition content type must be null to assign globally
```

The reason for opening this is that we can't honestly fix this issue in the signals there. If we caught the exception, then the permission wouldn't be removed. And that's wrong, as  fail-deadly kind of situation.

Removing permissions should never give an error in a situation where we are _able_ to remove the permission. Because this would _prevent_ the app from fixing validation violations, which is the opposite of what we want.

### DIFF
--- a/ansible_base/rbac/models.py
+++ b/ansible_base/rbac/models.py
@@ -198,7 +198,7 @@ class RoleDefinition(CommonModel):
             kwargs = dict(object_role=None, team=actor, role_definition=self)
             cls = RoleTeamAssignment
         else:
-            raise RuntimeError(f'Cannot give or remove permission to {actor}, must be a user or team')
+            raise RuntimeError(f'Cannot {giving and "give" or "remove"} permission for {actor}, must be a user or team')
 
         if giving:
             assignment, _ = cls.objects.get_or_create(**kwargs)
--- a/ansible_base/rbac/models.py
+++ b/ansible_base/rbac/models.py
@@ -184,11 +184,11 @@ class RoleDefinition(CommonModel):
         return self.give_or_remove_global_permission(actor, giving=False)
 
     def give_or_remove_global_permission(self, actor, giving=True):
-        if self.content_type is not None:
-            raise RuntimeError('Role definition content type must be null to assign globally')
+        if giving and (self.content_type is not None):
+            raise ValidationError('Role definition content type must be null to assign globally')
 
         if actor._meta.model_name == 'user':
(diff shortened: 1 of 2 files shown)


ansible_base/rbac/models.py
```diff
@@ -198,7 +198,7 @@ def give_or_remove_global_permission(self, actor, giving=True):
             kwargs = dict(object_role=None, team=actor, role_definition=self)
             cls = RoleTeamAssignment
         else:
-            raise RuntimeError(f'Cannot give permission to {actor}, must be a user or team')
+            raise RuntimeError(f'Cannot give or remove permission to {actor}, must be a user or team')
```
rochacbruno:
```suggestion
            action = "give" if giving else "remove"
            raise RuntimeError(f'Cannot {action} permission to {actor}, must be a user or team')
```

rochacbruno: I would do 
```py
f'Cannot {giving and "give" or "remove"} permission to {actor}, must be a user or team'
```

But others may think it is cryptic

rochacbruno (approved):

ansible_base/rbac/models.py
AlanCoding: fine by me

```
>>> giving = False
>>> actor = 'bob'
>>> f'Cannot {giving and "give" or "remove"} permission to {actor}, must be a user or team'
'Cannot remove permission to bob, must be a user or team'
>>> 
>>> giving = True
>>> f'Cannot {giving and "give" or "remove"} permission to {actor}, must be a user or team'
'Cannot give permission to bob, must be a user or team'
```

mabashian (approved):

PR #562, opened 2024-08-09
Use updated heuristic for identifying locally managed roles

AlanCoding:
See the prior criteria `if permission_registry.get_resource_prefix(model) == 'shared':`, I always knew this wasn't perfect and would, at some point, kind of not work.

This is hard to talk out - whether or not you can do permission management is going to be kind of divorced from whether the permissions that role gives are for shared resources.

I still don't think you should be able to create custom roles that give
 - team membership or change permission --> covered by `ANSIBLE_BASE_ALLOW_CUSTOM_TEAM_ROLES`
 - organization "change" permission --> kind of tough, added criteria back here to cover this

The PR https://github.com/ansible/django-ansible-base/pull/486 essentially introduced a new distinction of RoleDefinitions, where some are external, and some are local, determined by the setting. This distinction is completely separate from whether shared resources are managed by the roles.

### DIFF
--- a/ansible_base/rbac/validators.py
+++ b/ansible_base/rbac/validators.py
@@ -234,22 +234,13 @@ def validate_assignment(rd, actor, obj) -> None:
         raise ValidationError(f'Role type {rd_model} does not match object {obj_ct.model}')
 
 
-def check_locally_managed(permissions_qs: Iterable[Model], role_content_type: Optional[Model]) -> None:
+def check_locally_managed(rd: Model) -> None:
     """Can the given role definition be managed here, or is it externally managed
 
-    If the role definition manages permissions on any shared resources, then
-    those should be locked down locally.
     This rule is a bridge solution until the RoleDefinition model declares
     explicitly whether it is managed locally or by a remote system.
     """
     if settings.ALLOW_LOCAL_RESOURCE_MANAGEMENT is True:
         return
-    for perm in permissions_qs:
-        # View permission for shared objects is interpreted as permission to view
-        # the resource locally, which is needed to be able to view parent objects
-        # For system roles, this exception is not allowed
-        if role_content_type and perm.codename.startswith('view'):
-            continue
-        model = perm.content_type.model_class()
-        if permission_registry.get_resource_prefix(model) == 'shared':
-            raise ValidationError('Not managed locally, use the resource server instead')
+    if rd.name in settings.ANSIBLE_BASE_JWT_MANAGED_ROLES:
+        raise ValidationError('Not managed locally, use the resource server instead')
--- a/ansible_base/rbac/validators.py
+++ b/ansible_base/rbac/validators.py
@@ -156,6 +156,17 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
     if settings.ANSIBLE_BASE_DELETE_REQUIRE_CHANGE and role_model is not None:
         check_has_change_with_delete(codename_set, permissions_by_model)
 
+    if not managed:
+        for perm in permissions:
+            # View permission for shared objects is interpreted as permission to view
+            # the resource locally, which is needed to be able to view parent objects
+            # For system roles, this exception is not allowed
+            if content_type and perm.codename.startswith('view'):
+                continue
+            model = perm.content_type.model_class()
+            if permission_registry.get_resource_prefix(model) == 'shared':
+                raise ValidationError({'permissions', 'Not managed locally, local custom roles can only include view'})
+
 
 def validate_codename_for_model(codename: str, model: Union[Model, Type[Model]]) -> str:
     """Shortcut method and validation to allow action name, codename, or app_name.codename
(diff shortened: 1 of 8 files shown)


AlanCoding:
I added @relrod as a reviewer, but don't merge until @fosterseth has processed this all and is fully on the same page. The rest of it will be his work.

AlanCoding:
Feel free to merge this while I'm away next week.

thedoubl3j (approved):
understandable and looks good, would still gate on @fosterseth per Alan's comments.

fosterseth (approved):
this is clear to me!

ansible_base/rbac/validators.py
```diff
@@ -133,6 +132,17 @@ def validate_permissions_for_model(permissions, content_type: Optional[Model], m
...
+    if (not managed) and (not settings.ALLOW_SHARED_RESOURCE_CUSTOM_ROLES):
+        for perm in permissions:
+            # View permission for shared objects is interpreted as permission to view
+            # the resource locally, which is needed to be able to view parent objects
+            # For system roles, this exception is not allowed
+            if content_type and perm.codename.startswith('view'):
+                continue
+            model = perm.content_type.model_class()
+            if permission_registry.get_resource_prefix(model) == 'shared':
+                raise ValidationError({'permissions', 'Local custom roles can only include view permission for shared models'})
```
AlanCoding, line 167: We shouldn't raise this error if the local server is the provider for that shared resource.

AlanCoding: On second thought, no. That's what `ALLOW_SHARED_RESOURCE_CUSTOM_ROLES` is for, even if it's redundant with provider status.

AlanCoding:
Documenting, if I flip the default value of the setting (current allows, switch to not-allow), that results in the "Consumers" checks failing

### eda-server

```
ERROR tests/integration/api/test_user.py::test_create_superuser - rest_framework.exceptions.ValidationError: [ErrorDetail(string="{'permissions', 'Local custom roles can only include view permission for shared models'}", code='invalid')]
ERROR tests/integration/api/test_user.py::test_modify_superuser_as_org_admin - rest_framework.exceptions.ValidationError: [ErrorDetail(string="{'permissions', 'Local custom roles can only include view permission for shared models'}", code='invalid')]
ERROR tests/integration/api/test_user.py::test_organization_admin_can_create_user - rest_framework.exceptions.ValidationError: [ErrorDetail(string="{'permissions', 'Local custom roles can only include view permission for shared models'}", code='invalid')]
```

### AWX

```
FAILED awx/main/tests/functional/test_rbac_execution_environment.py::test_give_object_permission_to_ee
FAILED awx/main/tests/functional/test_rbac_execution_environment.py::test_give_org_permission_to_ee[new]
FAILED awx/main/tests/functional/test_rbac_execution_environment.py::test_give_org_permission_to_ee[old]
```

```
  File "/awx_devel/awx/main/tests/functional/test_rbac_execution_environment.py", line 136, in test_give_org_permission_to_ee
    check_user_capabilities(org_member, org_ee, {'edit': False, 'delete': False, 'copy': False})
  File "/awx_devel/awx/main/tests/functional/test_rbac_execution_environment.py", line 58, in _rf
    raise RuntimeError(f'Could not find expected object ({obj}) in EE list result: {r.data}')
RuntimeError: Could not find expected object (some user ee-1) in EE list result: {'count': 0, 'next': None, 'previous': None, 'results': []}
```

```
  File "/awx_devel/awx/main/tests/functional/test_rbac_execution_environment.py", line 103, in test_give_object_permission_to_ee
    assert access.can_read(org_ee)  # by virtue of being an org member
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: assert False
 +  where False = can_read(<ExecutionEnvironment: some user ee-1>)
 +    where can_read = <awx.main.access.ExecutionEnvironmentAccess object at 0x7f0530f6e010>.can_read
```

So that looks like this breaks some rule related to visibility of EEs.

AlanCoding:
The eda-server issue seems to be 100% from

```
    @pytest.fixture
    def org_admin_rd():
>       return RoleDefinition.objects.create_from_permissions(
            name="organization-admin",
            permissions=[
                "change_organization",
                "member_organization",
                "view_organization",
                "delete_organization",
            ],
            content_type=ContentType.objects.get_for_model(models.Organization),
        )
```

So we add `managed=True` and that should take care of this entirely.

AlanCoding:
Locally, confirmed these tests are passing with https://github.com/ansible/awx/pull/15453

AlanCoding:
This will be ready to merge when the dependent PRs are merged, which now includes https://github.com/ansible/eda-server/pull/1018

AlanCoding:
Checks all passing except flake8 :tada: 

```
./test_app/tests/rbac/api/test_rbac_validation.py:10:1: F811 redefinition of unused 'permission_registry' from line 8
```

HOWEVER, I may push another commit that introduces a new setting so we move away from the local resource management setting.

fosterseth (approved):

PR #564, opened 2024-08-09
Add SSO and partially migrated flag to Resource API

newswangerd:
This PR does two things:
- Adds a new `additional_data` field on users that returns the set of social_auth backends that a user has configured on their account along with their uid
- Adds a `is_partially_migrated` flag that allows the resource server to idempotently migrate data without updating the `service_id`


newswangerd:
I don't think I can increase the coverage any more than this. The missing code paths are all `try` and `if` blocks that handle failure cases for services that either use vanilla `social_auth` (instead of the authenticator framework) or don't have `social_auth` installed at all.

relrod:
We should be able to mock in that case. e.g. here, we can mock `load_strategy` to be `None` before we call the function: https://sonarcloud.io/code?id=ansible_django-ansible-base&pullRequest=564&selected=ansible_django-ansible-base%3Aansible_base%2Fresource_registry%2Futils%2Fsso_provider.py&line=26

Similar for the except on 45, we can mock `load_strategy` with something that raises an exception and ensure we log the message and return the right thing.

ansible_base/resource_registry/views.py
```diff
@@ -95,15 +94,15 @@ def get_serializer_class(self):
...
-            data = serializer.get_additional_data(obj.content_object)
-            if data is not None:
-                return Response(data.data)
+    # @action(detail=True, methods=['get'])
+    # def additional_data(self, *args, **kwargs):
+    #     obj = self.get_object()
+    #     if serializer := resource_type_cache(obj.content_type.pk).serializer_class:
+    #         data = serializer.get_additional_data(obj.content_object)
+    #         if data is not None:
+    #             return Response(data.data)
```
relrod: nix

relrod (approved):

PR #570, opened 2024-08-13
fix: coerce reading saml request as via https if REDIRECT_IS_HTTPS

BrennanPaciorek:
AAP-29226

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/saml.py
+++ b/ansible_base/authentication/authenticator_plugins/saml.py
@@ -3,6 +3,7 @@ import logging
 from django.http import HttpResponse, HttpResponseNotFound
 from django.urls import re_path
 from django.utils.translation import gettext_lazy as _
+from onelogin.saml2.auth import OneLogin_Saml2_Auth
 from onelogin.saml2.errors import OneLogin_Saml2_Error
 from onelogin.saml2.settings import OneLogin_Saml2_Settings
 from rest_framework.serializers import ValidationError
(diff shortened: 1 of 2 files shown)


BrennanPaciorek:
With local testing this seems to fix the problem, though I need to comment out [this line](https://github.com/ansible/django-ansible-base/blob/6769287d0960ad919b0ccb4c37344aeb9c32e7ed/ansible_base/authentication/authenticator_plugins/saml.py#L259C10-L259C76) to make it work with my local proxy config (because that line sets the port number to the actual listening port for my application)

Another fix should be attaching `X-Forwarded-Proto: https` to requests from the proxy to the server hosting this, and setting `X-Forwarded_Proto: https` to a SECURE_PROXY_SSL_HEADER in django.

https://docs.djangoproject.com/en/5.0/ref/settings/#secure-proxy-ssl-header

bhavenst (commented):
Does the addition of the header fix it without modifying the protocol validation?  That would be interesting to know.  But the change seems reasonable if there isn't a config only fix for it..

BrennanPaciorek:
> Does the addition of the header fix it without modifying the protocol validation? That would be interesting to know. But the change seems reasonable if there isn't a config only fix for it..

For the (untested) config only fix for this, we would need to make sure the proxy which TLS is offloaded to is capable of being configured to attach this (nonstandard) header that will not be dropped, which is probably less trivial than this fix is from my standpoint.

bhavenst (approved):
Fair enough, seems fine to me..

bhavenst:
I decided to test this, so I configured nginx to listen on 8080 without SSL and set up keycloak and the authenticator to use http on 8080.  With SOCIAL_AUTH_REDIRECT_IS_HTTPS = True, auth fails with an AuthFailed exception and with False it works fine, so I would say this is working properly.  This is actually a reverse test from what QE has, but it shows the flag it working as expected.

trucdg (approved):
I think it looks good :+1:

PR #572, opened 2024-08-14
Add fallback cache

bhavenst:
This is a fallback cache that kicks in when redis (assuming redis is used) is not available.  It clears the primary and switches back when it's available again.


ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -49,6 +57,11 @@
 except NameError:
     OAUTH2_PROVIDER = {}
 
+try:
+    CACHES  # noqa: F821
+except NameError:
+    CACHES = {}
```
bhavenst: Need to change this, I guess if CACHES is not defined at all, django is happy.  If it's defined but empty (or not containing a 'default' entry) then it isn't.  So if not defined, don't mess with it.

jay-steurer (approved):

ansible_base/lib/cache/fallback_cache.py
```diff
@@ -0,0 +1,109 @@
...
+                    _temp_file.touch()
+                response = getattr(self._fallback_cache, operation)(*args, **kwargs)
+
+        return response
+
+    def start_recovery(self):
+        with multiprocessing.Lock():
+            # Set single process/thread to do the recovery, but time out in case it dies
+            recoverer = self._fallback_cache.get_or_set('RECOVERY_THREAD_ID', id(self), timeout=60)
+            if recoverer == id(self):
```
john-westcott-iv, line 89: On thought here. id() is taking self which is this object. What happens if django deletes this object for some reason. We would never be able to check that redis is up again right? Should we clear this out in the same `finally` down below to let some other thread be able to check incase self ever "goes away"?

john-westcott-iv (approved):

ansible_base/lib/cache/fallback_cache.py
bhavenst: timeout=60 limits the lifetime to 60 sec (at least that is the idea :) )

john-westcott-iv: @bhavenst pointed out that the timeout=60 so worst case there would just be a 60 second delay.

PR #573, opened 2024-08-19
RBAC content filter should consider the app_label.

jctanner:
https://issues.redhat.com/browse/AAP-29363

Sometimes there are multiple models with the same name and different app labels. Example: container.containernamespace and galaxy.containernamespace

If there is only one matching model name, return that, otherwise filter on the app label as well so that only one object is returned.

### DIFF
--- a/ansible_base/rbac/api/serializers.py
+++ b/ansible_base/rbac/api/serializers.py
@@ -80,8 +80,11 @@ class ContentTypeField(ChoiceLikeMixin):
 
     def get_dynamic_object(self, data):
         app_label, model = data.rsplit('.')
-        if permission_registry.content_type_model.objects.get(model=model, app_label=app_label).exists():
-            return permission_registry.content_type_model.objects.get(model=model, app_label=app_label)
+        try:
+            if permission_registry.content_type_model.objects.get(model=model, app_label=app_label).exists():
+                return permission_registry.content_type_model.objects.get(model=model, app_label=app_label)
+        except Exception:
+            pass
         return permission_registry.content_type_model.objects.get(model=model)
 
     def to_representation(self, value):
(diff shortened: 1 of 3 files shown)


ansible_base/rbac/api/serializers.py
```diff
@@ -79,7 +79,9 @@ def get_dynamic_choices(self):
         return list(sorted((self.get_resource_type_name(cls), cls._meta.verbose_name.title()) for cls in permission_registry.all_registered_models))
 
     def get_dynamic_object(self, data):
-        model = data.rsplit('.')[-1]
+        app_label, model = data.rsplit('.')
```
AlanCoding: I don't think this will work, because it's not the `app_label` here, it's the resource_registry service label, like "awx" or "shared".

There are a couple of ways you could convert the server label to the app_label. You could get the model from either the permission_registry or the resource registry and then use the app_label from the model, then use `get_for_model` from contenttypes

jctanner: If all I have from the data is 'foo.bar' and 'bar' represents 2 models, and 'foo' isn't the app_label ... how do I go about determining which bar model the caller meant to use? I can get all of the models, but I still need something to determine which one is "right". For galaxy, if the payload sends `content_type:container.containernamespace` the "container" prefix that works for `permission_registry.content_type_model.objects.get(model=model, app_label=app_label)`

AlanCoding: you want the "bar" model that's in the `permission_registry._registry` or `permission_registry.all_registered_models`. This is just _slightly_ awkward because it's a set, but you get it. Then you could get the app label because you have the model class itself at that point. This will feel painful, but it will be robust.

AlanCoding: https://github.com/jctanner/django-ansible-base/pull/2

AlanCoding (approved):

PR #574, opened 2024-08-19
Only allow x-trusted-proxy header to be valid for so long

john-westcott-iv:


ansible_base/jwt_consumer/common/util.py
```diff
@@ -42,6 +44,15 @@ def validate_x_trusted_proxy_header(header_value: str, ignore_cache=False) -> bo
         logger.warning("Failed to validate x-trusted-proxy-header, malformed, expected value to contain a -")
         return False
 
+    # Validate that the header has been cut within the last 300ms (by default)
+    try:
+        if time.time_ns() - int(timestamp) > get_setting('trusted_header_timeout_in_ns', 300000000):
```
AlanCoding, line 49: This requires that the clocks from the proxy server and the local server (using this auth class) are synced. I just want to have that in writing here.

AlanCoding (approved):

ansible_base/jwt_consumer/common/util.py
john-westcott-iv: Docs issue opened.

PR #575, opened 2024-08-19
Only trust last entry in x-forwarded-for by default

john-westcott-iv:


ansible_base/lib/utils/requests.py
```diff
@@ -26,22 +35,30 @@ def get_remote_hosts(request: HttpRequest, get_first_only: bool = False) -> list
...
 
-    for header in headers:
-        for value in request.META.get(header, '').split(','):
-            value = value.strip()
-            if value:
-                if get_first_only:
-                    return [value]
-                remote_hosts.append(value)
+    if get_first_only:
+        return [remote_hosts[0]]
```
relrod, line 63: Are we guaranteed that there's at least one element in `remote_hosts` here?

john-westcott-iv: Good call out, we should be because REMOTE_ADDR should always be set. But I'll add a check just incase.

BrennanPaciorek (approved):

PR #577, opened 2024-08-20
Add tests for more than 1 team and org member roles

AlanCoding:
I was more worried about the revoking flow, than the granting flow.

But it turned out it worked just fine. So I'll go ahead and add these, in support of https://github.com/ansible/awx/pull/15445

AAP-24052


fosterseth (approved):
elegant and straightforward tests!

PR #565, opened 2024-08-09
Add service backed SSO

newswangerd:
Add the following:
- `redirect_to_resource_server` social auth pipeline stage that generates an auth code and redirects authenticated users to the resource server
- `get_user_auth_code` to generate auth codes

Usage: 
On a system with the `RESOURCE_SERVER` configured correctly set the following in `settings.py`

```python
SOCIAL_AUTH_PIPELINE = (
    ...,

    # Add to end of pipeline
    "ansible_base.resource_registry.utils.service_backed_sso_pipeline.redirect_to_resource_server"
)
```

When you successfully authenticate with a configure social auth backend you should be redirected to `<RESOURCE_SERVER>/api/gateway/v1/validate_auth_code?auth_code=<AUTH_CODE>`

PR #580, opened 2024-08-21
[resource registry] Allow for obtaining an auth code given username/password payload

relrod:
Built on #565.

### DIFF
--- a/ansible_base/lib/testing/fixtures.py
+++ b/ansible_base/lib/testing/fixtures.py
@@ -1,3 +1,4 @@
+import copy
 import os
 import uuid
 from collections import namedtuple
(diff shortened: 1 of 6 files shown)


ansible_base/lib/testing/fixtures.py
```diff
@@ -42,6 +43,33 @@ def _env(key, value):
...
+            # or even
+            delattr(settings, 'SOME_SETTING_VAR')  # for some reason normal "del" doesn't work here
+        # But out of the context manager the whole var is restored
+        assert settings.SOME_SETTING_VAR['and']['nested']['a']['bit'] == "old_value"
+    """
+
+    @contextmanager
+    def f(setting_key):
+        original = copy.deepcopy(getattr(settings, setting_key))
+        yield
```
AlanCoding: the docs say you should have a try-except

https://docs.python.org/3/library/contextlib.html

relrod: good catch, I fixed a few other fixtures too.

newswangerd (approved):

PR #582, opened 2024-08-21
[WIP] Don't require service path, and flip the bit for reverse sync by default

relrod:


AlanCoding (approved):

ansible_base/lib/dynamic_config/settings_logic.py
```diff
@@ -216,8 +216,7 @@ def get_dab_settings(
         dab_data['ORG_ADMINS_CAN_SEE_ALL_USERS'] = True
 
     if 'ansible_base.resource_registry' in installed_apps:
-        # Disable reverse syncing by default
-        dab_data['DISABLE_RESOURCE_SERVER_SYNC'] = True
+        dab_data['DISABLE_RESOURCE_SERVER_SYNC'] = False
```
AlanCoding, line 219: Did you want to rename this? In RBAC I have a lot of settings that start with `ALLOW_`, which would be same in function as settings that end in `_ENABLED`. I agree this it would be better to rename it, and now, while the deadline isn't near would be the _better_ time to make that kind of naming change.

Also, if you're basically supporting a default for `RESOURCE_SERVICE_PATH`, then this setting would be the only thing standing in the way of reverse-sync being enabled.

My intuition is that reverse-sync should not be enabled by default. You could have each app flip it in default settings (as opposed to installers). That seems reasonable to me.

Important NOTE:

https://github.com/ansible/eda-server/blob/main/tools/docker/docker-compose-dev.yaml#L38-L40

eda-server sets

```yaml
  EDA_RESOURCE_SERVER__URL: ${EDA_RESOURCE_SERVER__URL:-'@none None'}
```

which has the effect of making `settings.RESOURCE_SERVER` to `{"URL": None}`. The test you have of `getattr(settings, 'RESOURCE_SERVER', False)` fails in its intention for this case. For that reason, I don't consider a RESOURCE_SERVER test to be reliable (I'd rather not try to "fix" this), and this is the main reason I argue for the default to be to _not_ reverse-sync.

AlanCoding:
@fosterseth merged replacement

PR #583, opened 2024-08-22
[WIP] Use the same auth for galaxy as everything else

AlanCoding:
I tried to salvage the logic in the existing class... but I just can't do it.

Let's just apply the roles in the same way as all the other stuff. If there's a problem with that, we'll fix it in the galaxy_ng signals.

### DIFF
--- a/ansible_base/jwt_consumer/hub/auth.py
+++ b/ansible_base/jwt_consumer/hub/auth.py
@@ -1,6 +1,4 @@
 # Python
-import logging
-
 from ansible_base.jwt_consumer.common.auth import JWTAuthentication
 
 
--- a/ansible_base/jwt_consumer/hub/auth.py
+++ b/ansible_base/jwt_consumer/hub/auth.py
@@ -1,78 +1,8 @@
+# Python
 import logging
 
-from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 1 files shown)


john-westcott-iv:

AlanCoding:
Of course testing this will be challenging, but yes I think it makes sense to do this as cleanup.

john-westcott-iv:
@AlanCoding does this require changes on the Hub side as well?

AlanCoding:
Now leaning in the direction of abandoning all RBAC data in jwt_consumer in the near-term. This was never easy to maintain.

AlanCoding:
^ went back around and reversed that decision. However, some related work is going on in https://github.com/ansible/django-ansible-base/pull/787

AlanCoding:
> @AlanCoding does this require changes on the Hub side as well?

No. Even in the deleted code, it used the DAB RBAC models. It makes absolutely no sense as it is now.

I believe the only potential concern is that the "Organization Admin" role was never created in galaxy_ng. And if that's the case, the best thing to do is to ignore the case in the main jwt_consumer logic to process permissions.

AlanCoding:
Accomplished elsewhere.

PR #584, opened 2024-08-22
Ignore permissions given twice in migrations

AlanCoding:
I ran into this trying to test migrations in galaxy_ng. I need to test by running with current apps, but `IntegrityError` is obviously going to make that error, because all the data is filled in.

### DIFF
--- a/ansible_base/rbac/migrations/_utils.py
+++ b/ansible_base/rbac/migrations/_utils.py
@@ -30,7 +30,7 @@ def give_permissions(apps, rd, users=(), teams=(), object_id=None, content_type_
             RoleUserAssignment(object_role=object_role, user=user, **object_role_fields)
             for user in users
         ]
-        RoleUserAssignment.objects.bulk_create(user_assignments)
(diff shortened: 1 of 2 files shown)


john-westcott-iv (approved):

AlanCoding:
@Dostonbek1 I saw this in the eda-server downstream testing

```
FAILED tests/integration/dab_rbac/test_managed_roles.py::test_org_role_team_assignments - AssertionError: [ErrorDetail(string='Not managed locally, use the resource server instead', code='invalid')]
assert 400 == 201
 +  where 400 = <Response status_code=400, "application/json">.status_code
 +  and   201 = status.HTTP_201_CREATED
FAILED tests/integration/dab_rbac/test_managed_roles.py::test_org_role_user_assignments - AssertionError: [ErrorDetail(string='Not managed locally, use the resource server instead', code='invalid')]
assert 400 == 201
 +  where 400 = <Response status_code=400, "application/json">.status_code
 +  and   201 = status.HTTP_201_CREATED
```

I think this is addressed by PRs you have in progress. In any case, unrelated to this. I'll keep an eye on it.

PR #589, opened 2024-08-26
Several bug fixes for reverse sync

relrod:
- Don't instance.resource
- Don't sync an update of something owned by the local service
- Load ansible_id from gateway on create


ansible_base/resource_registry/utils/sync_to_resource_server.py
```diff
@@ -39,29 +39,32 @@ def sync_to_resource_server(instance, action, ansible_id=None):
...
-            logger.error(f"Resource {instance} does not have a resource or ansible_id")
-            return
+        resource = Resource.get_resource_for_object(instance)
     except Resource.DoesNotExist:
-        # The getattr() will raise a Resource.DoesNotExist if the resource doesn't exist.
         logger.error(f"Resource {instance} does not have a resource")
         return
 
+    if str(resource.service_id) == service_id() and action == "update":
+        # Don't sync if we're updating a resource that isn't owned by the resource server yet.
```
AlanCoding, line 54: > resource that isn't owned by the resource server

this seems to be opposite of

> `str(resource.service_id) == service_id()`

If we're here, doesn't that mean that it _is_ local?

ansible_base/resource_registry/utils/sync_to_resource_server.py
```diff
@@ -89,8 +92,9 @@ def sync_to_resource_server(instance, action, ansible_id=None):
             response = client.create_resource(body)
             json = response.json()
             if isinstance(json, dict):  # Mainly for tests... to avoid getting here with mock
-                instance.resource.service_id = json['service_id']
-                instance.resource.save()
+                resource.service_id = json['service_id']
+                resource.ansible_id = json['ansible_id']
+                resource.save()
```
AlanCoding, line 106: I must admit to not understanding what's going on here.

Yeah, it makes sense that `service_id` should be set to the resource provider. But wouldn't we know what that is _before_ creating the local `Resource` object, since we know where it's going? And if the request gives an `ansible_id` that's different on creation... that's a bug.

ansible_base/resource_registry/utils/sync_to_resource_server.py
relrod: Yes, and we're bailing out (returning) in that case.

If it is local, and it's an update, we don't sync. We wait until it is owned by the resource server before we allow syncing updates. At least that is my understanding of the requirement. @newswangerd

ansible_base/resource_registry/utils/sync_to_resource_server.py
relrod: Do we know the service_id beforehand? We know the URL of the resource server and how to talk to it, but I don't know that we know its service_id offhand?

From @newswangerd:
> on create we need to load the ansible_id from the gateway like we do with service_id

I don't know the scenario where the ansible_id can change, but this was the ask here.

AlanCoding: The `service_id` point just seems inefficient, but its function makes sense. I get why it happens, because while the resource provider is defined in the settings, the `service_id` of the server is not saved locally.

But the `ansible_id` from the resource server makes no sense to me. Here is a concrete reference:

https://github.com/ansible/django-ansible-base/blob/21c652631641e3adf5b9cdbb43bb79cbb04fb1ac/test_app/tests/resource_registry/test_resources_api.py#L207-L208

This test POSTs a body that includes the `ansible_id` to the resource list. It inspects the response data (`response.data` local variable), and assures that the returned `ansible_id` is the same as what was included in the POST data (`resource` local variable).

ansible_base/resource_registry/utils/sync_to_resource_server.py
AlanCoding: so... it's only local in the case that the creation signal went wrong, or the server is migrated and the data hasn't yet been migrated?

newswangerd: @AlanCoding that's correct.

ansible_base/resource_registry/utils/sync_to_resource_server.py
newswangerd: The scenario where I ran into this is the following:

1. Service A and Service B both try to create resource C at the same time.
2. Service A's request arrives first and receives the lock on the DB to create the resource and creates resource C with ID=1
3. Once the lock on the DB is released, Service B's request comes through. Because this endpoint does a create or update operation, Service B modifies resource C, and sets its ID=2.
4. Now resource C is out of sync. On the resource server and service B, the ID=2, but on service A the ID is now 1.

Fixing this problem is fairly easy. We just let the resource server set the ansible ID of the resource, rather than let each service pick their own random UUID.

newswangerd (approved):

ansible_base/resource_registry/utils/sync_to_resource_server.py
AlanCoding: Wouldn't that scenario be resolved by the periodic resource sync? And isn't fixing this kind of thing the job of the periodic resource sync?

AlanCoding (approved):
I would like more comments about the cases added here, because they are mostly corner cases. Without knowledge of those corner cases, they appear to be doing the wrong thing. Functionally I follow well enough.

AlanCoding (approved):

PR #590, opened 2024-08-26
Adding additional unit tests

john-westcott-iv:


ansible_base/authentication/utils/claims.py
```diff
@@ -543,8 +543,11 @@ def _give_permission(self, role_definition: CommonModel, obj: Union[AbstractOrga
     def _remove_permission(self, role_definition: CommonModel, obj: Union[AbstractOrganization, AbstractTeam, None] = None) -> None:
         if obj:
             logger.info(
-                _("Removing role '{rd}' from user '{username}' in '{object}").format(
-                    rd=role_definition.name, username=self.user.username, object=obj.__class__.__name__
+                _("Removing role '{rd}' from user '{username}' on '{object}' id {id}").format(
```
john-westcott-iv, line 546: Note, I found a missing `'` in the log and also thought I'd add the id (if its available) while I was in here changing things.

test_app/tests/authentication/utils/test_claims_reconcile_user.py
```diff
@@ -258,5 +259,10 @@ def test_remove_users_setting_set(org_member_rd, org_admin_rd, member_rd, admin_
...
     # Nothing was specified in orgs[2], teams[2]
-    assert not user.role_assignments.filter(object_id=str(orgs[2].id)).exists()
-    assert not user.role_assignments.filter(object_id=str(teams[2].id)).exists()
+    org_content_type = ContentType.objects.get_for_model(Organization)
+    user_role_assignments_for_org_2 = user.role_assignments.filter(content_type=org_content_type, object_id=str(orgs[2].id))
+    assert (
+        not user_role_assignments_for_org_2.exists()
+    ), f"Expected no assignments for org2 but got {', '.join([r.role_definition.name for r in user_role_assignments_for_org_2])}"
+    team_content_type = ContentType.objects.get_for_model(Team)
+    assert not user.role_assignments.filter(content_type=team_content_type, object_id=str(teams[2].id)).exists()
```
john-westcott-iv, line 268: There was something funky going on with this test after our changes. We were unable to replicate. What we found in our debugging (partially from the change above) was that another Team that the user had permissions to had the same object_id as `orgs[2]` so the assert was constantly failing. In the end, to get the test back to green we added a `content_type` to the filter so that we were only looking for the ID and the specific ContentType along with it.

test_app/tests/lib/cache/test_fallback_cache.py
```diff
@@ -0,0 +1,233 @@
...
+        self.__initialized = True
+
+    def add(self, key, value, timeout=300, version=None):
+        self.cache[key] = value
+
+    def get(self, key, default=None, version=None):
+        if self.working:
+            return self.cache.get(key, default)
+        else:
+            raise RuntimeError(f"Sorry, cache no worky {self}")
```
bhavenst, line 39: The {self} was for some debugging, but guess it doesn't really matter.

test_app/tests/lib/cache/test_fallback_cache.py
```diff
@@ -0,0 +1,233 @@
...
+        time.sleep(1)
+
+    # Ensure caches were cleared
+    assert cache.get('key') is None
+    assert fallback.get('key') is None
+    assert cache.get('tobecleared') is None
+
+    cache.set('key2', 'val3')
+
+    assert cache.get('key2') == 'val3'
```
bhavenst: This got changed several times and not sure it really makes sense to set/check like this here.  Maybe make line 113 do a get from primary instead?

bhavenst (approved):

test_app/tests/lib/cache/test_fallback_cache.py
john-westcott-iv: Changed!

PR #596, opened 2024-08-28
Remove is_superuser from resource API

fosterseth:
Do not sync is_superuser field on Users.


AlanCoding (approved):
so it is written, so it shall be done

AlanCoding:
Documenting for hazard assessment:

https://github.com/ansible/awx/blob/444af2b500517e0cc09f791b7775d4b5b5ed6cf4/awx/api/views/__init__.py#L753-L769

AWX code enforces `ALLOW_LOCAL_RESOURCE_MANAGEMENT`, and it will allow updates of non-shared fields even if "local management" is _not_ allowed. With this, `is_superuser` becomes a non-shared field, so it should work as intended.

AlanCoding:
But... I'm worried about eda-server and galaxy_ng

https://github.com/ansible/eda-server/blob/48b44cf06d4d0498d490810a043b4b3cf44578ac/src/aap_eda/api/serializers/mixins.py#L22-L27

and galaxy_ng

https://github.com/ansible/galaxy_ng/blob/b1c841132c4d3d27c3d0ae5ce5b744a656ef4bab/galaxy_ng/app/access_control/statements/standalone.py#L164-L167

shanemcd (dismissed):
Talking with Alan now, please hold up.

AlanCoding (changes requested):
yep, hold for investigation of last comments

AlanCoding (approved):
We have supporting patches to enable this. After some discussion, I'm not particular about merge order. Ping @elyezer

AlanCoding:
Dependent PRs have merged, https://github.com/ansible/galaxy_ng/pull/2255 and https://github.com/ansible/eda-server/pull/1043

PR #597, opened 2024-08-29
Dont require service path

fosterseth:
replaces https://github.com/ansible/django-ansible-base/pull/582

Rename `DISABLE_RESOURCE_SERVER_SYNC` to `RESOURCE_SERVER_SYNC_ENABLED`
set a default value for service path


ansible_base/resource_registry/apps.py
```diff
@@ -96,11 +96,9 @@ def proxies_of_model(cls):
...
 
 def _should_reverse_sync():
-    enabled = not getattr(settings, 'DISABLE_RESOURCE_SERVER_SYNC', False)
-    for setting in ('RESOURCE_SERVER', 'RESOURCE_SERVICE_PATH'):
-        if not getattr(settings, setting, False):
-            enabled = False
-            break
+    enabled = getattr(settings, 'RESOURCE_SERVER_SYNC_ENABLED', False)
+    if not getattr(settings, 'RESOURCE_SERVER', False):
+        enabled = False
```
AlanCoding, line 106: I'd argue that if `RESOURCE_SERVER_SYNC_ENABLED` is True and `RESOURCE_SERVER` is not set, this is invalid.

Near-term, the easiest way to properly treat this as invalid would be an error-level log. Likewise for not having the SECRET_KEY defined. I would just add a `logger.error` statement to both of these branches. This could be eliminated if we added a Django check, but that would be over-engineered for the moment.

ansible_base/resource_registry/utils/sync_to_resource_server.py
```diff
@@ -67,8 +67,9 @@ def sync_to_resource_server(instance, action, ansible_id=None):
     else:
         logger.error("No user found, syncing to resource server with jwt_user_id=None")
 
+    service_path = getattr(settings, "RESOURCE_SERVICE_PATH", "/api/gateway/v1/service-index/")
```
AlanCoding: I'd argue that this should be set as the default value in `ansible_base/lib/dynamic_config/settings_logic.py`, as opposed to here.

This is because the setting should be reused in this line, instead of having it hard-coded.

https://github.com/ansible/django-ansible-base/blob/fa97197950ee806659d61e7591183fa391468d00/ansible_base/resource_registry/tasks/sync.py#L68

Right?

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -65,7 +65,7 @@ def __iter__(self):
 
 def create_api_client() -> ResourceAPIClient:
     """Factory for pre-configured ResourceAPIClient."""
-    params = {"raise_if_bad_request": False, "service_path": "/api/gateway/v1/service-index/"}
+    params = {"raise_if_bad_request": False}
```
fosterseth, line 68: we are setting params RESOURCE_SERVICE_PATH below, so no need to set it here.

ansible_base/resource_registry/utils/sync_to_resource_server.py
fosterseth: I just removed that item from params on this line, as we set it below (as long as service_path is not None or empty)

hopefully that is the correct logic

ansible_base/resource_registry/utils/sync_to_resource_server.py
```diff
@@ -67,8 +67,9 @@ def sync_to_resource_server(instance, action, ansible_id=None):
     else:
         logger.error("No user found, syncing to resource server with jwt_user_id=None")
 
+    service_path = getattr(settings, "RESOURCE_SERVICE_PATH")
     client = get_resource_server_client(
-        settings.RESOURCE_SERVICE_PATH,
+        service_path,
```
AlanCoding: Code style: At this point just typing `settings.RESOURCE_SERVICE_PATH` should be the same effect

AlanCoding (approved):
Looks great, thanks. This should help to finalize the shape of reverse-sync adoption.

relrod (approved):

PR #598, opened 2024-08-30
📦 Implement tokenless publishing to PyPI

webknjaz:


AlanCoding (approved):

webknjaz:
FTR, there was one concern that Rich raised when we talked on the day I submitted the PR, which is that the current CalVer implementation only allows one release per day. I chose to ignore that in this PR, with the understanding that it can be improved by adding a new `version` field to the GH UI (via a `workflow_dispatch` input), in a follow-up.

If you agree that this is not an immediate concern in the scope of the PR, let's merge it.

john-westcott-iv:
We are going to merge this and then add follow up work to alter how DAB is using CalVer to allow for multiple releases in one day.

PR #601, opened 2024-09-02
Add option to automatically move users to a new authenticator

newswangerd:
This adds a new field on the authenticator model called `auto_migrate_users_to` which accepts a foreign key to another authenticator.  When this field is configured we will do the following:

- On login, collect any authenticator users from authenticators configured to migrate to the currently in use authenticator
- For each of the discovered authenticators, call the `move_authenticator_user_to` method on the authenticator plugin. This lets the authenticator plugin know that the user is being moved and gives it a chance to perform any cleanup before that happens.


ansible_base/authentication/models/authenticator.py
```diff
@@ -33,6 +33,14 @@ class Authenticator(UniqueNamedCommonModel):
     slug = fields.SlugField(max_length=1024, default=None, editable=False, unique=True, help_text="An immutable identifier for the authenticator")
     category = fields.CharField(max_length=30, default=None, help_text="The base type of this authenticator")
 
+    auto_migrate_users_to = ForeignKey(
+        "Authenticator",
+        help_text=("Automatically move users from this authenticator to the target authenticator when a matching user logs in via the targe authenticator"),
```
relrod:
```suggestion
        help_text=("Automatically move users from this authenticator to the target authenticator when a matching user logs in via the target authenticator"),
```

newswangerd:
Right now you can auto migrate users to any authenticator. I'm considering adding a field on the authenticator plugin where plugin authors can declare which authenticators are supported for auto migration. For example, we can declare that the generic `oidc` authenticator can be set to auto migrate users to `keycloak`, `github`, `google` etc (because these are all odic based authenticators), but can't be set to auto migrate users to `saml`.

Ultimately all of the authenticators that can be configured to use active directory as a backend (LDAP, OIDC, SAML, Keycloak) are potentially compatible with one another as long as they have the same user database as a backend.

ansible_base/authentication/authenticator_plugins/base.py
```diff
@@ -134,3 +135,45 @@ def add_related_fields(self, request, authenticator):
...
+                for x in getattr(old_user, name).all():
+                    # The only case where this might fail is if the relationship has a uniqueness
+                    # contraint on the user. In this case, all we can do is skip.
+                    try:
+                        # This atomic block is here to prevent a failure that is best described by
+                        # this stack overflow: https://stackoverflow.com/questions/21458387
+                        with transaction.atomic():
+                            getattr(new_user, name).add(x)
+                    except IntegrityError:
+                        continue
```
relrod, line 183: Should we log or do something else here? It seems concerning, like we can get into an inconsistent state, if only some of the relationships get copied but not all of them?

ansible_base/authentication/authenticator_plugins/keycloak.py
```diff
@@ -58,3 +58,6 @@ def extra_data(self, user, backend, response, *args, **kwargs):
 
     def get_user_groups(self, extra_groups=[]):
         return extra_groups
+
+    def get_alternative_uid(self, **kwargs):
```
relrod, line 62: Can we pull a default definition for this method out into some (new) base class somewhere and make {keycloak,oidc,saml} inherit from it?

Even if the default is just `raise Exception("This needs to be overridden before it's called")`. It feels weird, like there's no real API contract here, to have some plugins "randomly" have a public method defined on them.

And then below instead of `hasattr`, we can check for `issubclass` of that base class. That feels cleaner to me.

ansible_base/authentication/models/authenticator.py
```diff
@@ -33,6 +33,18 @@ class Authenticator(UniqueNamedCommonModel):
     slug = fields.SlugField(max_length=1024, default=None, editable=False, unique=True, help_text="An immutable identifier for the authenticator")
     category = fields.CharField(max_length=30, default=None, help_text="The base type of this authenticator")
 
+    auto_migrate_users_to = ForeignKey(
+        "Authenticator",
+        help_text=(
+            "Automatically move users from this authenticator to the target authenticator when a matching user logs in via the target authenticator. "
+            "For this to work, the field used for the user ID on both authenticators needs match up. This should only be used when migrating users "
```
relrod:
```suggestion
            "For this to work, the field used for the user ID on both authenticators needs to match up. This should only be used when migrating users "
```

But I'd maybe consider "needs to be the same" or "needs to have the same value" or something a bit more formal than "match up"

ansible_base/authentication/utils/authentication.py
```diff
@@ -40,19 +86,32 @@ def check_system_username(uid: str) -> None:
...
         raise AuthException(_('Unable to get associated username from: %(details)s') % {'details': kwargs.get("details", None)})
 
     authenticator = kwargs.get('backend')
     if not authenticator:
         raise AuthException(_('Unable to get backend from kwargs'))
 
-    return {"username": determine_username_from_uid(uid, authenticator)}
+    alt_uid = None
 
+    # TODO: add this to keycloak, oidc and SAML
```
relrod: this TODO can go away I think?

newswangerd:
@elyezer I can't add you as a reviewer, but I'd like it if you could take a look at this

ansible_base/authentication/migrations/0014_authenticator_auto_migrate_users_to.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('dab_authentication', '0013_alter_authenticator_order'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='authenticator',
+            name='auto_migrate_users_to',
+            field=models.ForeignKey(help_text='Automatically move users from this authenticator to the target authenticator when a matching user logs in via the targe authenticator', null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='auto_migrate_users_from', to='dab_authentication.authenticator'),
```
elyezer: s/targe/target

ansible_base/authentication/serializers/authenticator.py
```diff
@@ -94,6 +94,10 @@ def validate(self, data) -> dict:
         if not request or (request.method != 'PATCH' and configuration is None):
             raise ValidationError(_("You must specify configuration for the authenticator"))
 
+        if auto_migrate := data.get('auto_migrate_users_to'):
+            if auto_migrate.auto_migrate_users_to is not None:
+                raise ValidationError({"auto_migrate_users_to": _("The authenticator you have picked is already configured to auto migrate users.")})
```
john-westcott-iv:
```suggestion
                raise ValidationError({"auto_migrate_users_to": _("The authenticator you have picked is already configured to auto migrate users from a different authenticator.")})
```

ansible_base/authentication/utils/authentication.py
```diff
@@ -20,6 +21,51 @@ def setting(self, *args, **kwargs):
         return ["username", "email"]
 
 
+def migrate_from_existing_authenticator(uid, alt_uid, authenticator: Authenticator, preferred_username=None):
```
john-westcott-iv: Can we add more types to the parameters and a return type?

ansible_base/authentication/utils/authentication.py
```diff
@@ -20,6 +21,51 @@ def setting(self, *args, **kwargs):
...
+    uid_filter = [
+        uid,
+    ]
+    if alt_uid:
+        uid_filter = [uid, alt_uid]
+
+    if authenticator.auto_migrate_users_from.filter(type__endswith="saml").exists():
+        # SAML puts prepends all the UIDs with IdP. Adding this to the search criteria will
+        # allow us to find SAML accounts that match the UID.
+        uid_filter.append("IdP:" + uid)
```
john-westcott-iv: Would it hurt to just add this with out the exists filter to save a query?

ansible_base/authentication/utils/authentication.py
```diff
@@ -20,6 +21,51 @@ def setting(self, *args, **kwargs):
...
+        main_user = migrate_users[0].user
+
+    for migrate_user in migrate_users:
+        provider = migrate_user.provider
+        from_authenticator = get_authenticator_class(provider.type)(database_instance=provider)
+        old_user = from_authenticator.move_authenticator_user_to(main_user, migrate_user)
+        if old_user and not old_user.authenticator_users.exists():
+            old_user.delete()
+
+    # Now that we've potentially cleaned up any old user accounts, lets see if we have we can
```
john-westcott-iv:
```suggestion
    # Now that we've potentially cleaned up any old user accounts, lets see if we can
```

ansible_base/authentication/utils/authentication.py
```diff
@@ -116,7 +175,11 @@ def get_or_create_authenticator_user(
         logger.warning(f"AuthException: {e}")
         raise
 
-    username = determine_username_from_uid(uid, authenticator)
+    if migrated_username := migrate_from_existing_authenticator(uid=uid, alt_uid=None, authenticator=authenticator, preferred_username=uid):
+        username = migrated_username
+    else:
+        username = determine_username_from_uid(uid, authenticator)
```
john-westcott-iv, line 185: Should this be `determine_username_from_uid_social` or are you intending to do a recursive call here? I don't see authenticator or uid changing though.

john-westcott-iv (commented):
Can we squash the 2 migration files in the authentication app?

ansible_base/authentication/utils/authentication.py
newswangerd: No problem. That's how I had it originally.

ansible_base/authentication/utils/authentication.py
newswangerd: I think you might have this backwards? This is making the same call to `determine_username_from_uid` as it did before this patch. Changing it to `determine_username_from_uid_social` would be recursive.

ansible_base/authentication/authenticator_plugins/base.py
```diff
@@ -134,3 +135,56 @@ def add_related_fields(self, request, authenticator):
...
+            "role_assignments",
+        )
+
+        old_user = old_authenticator_user.user
+
+        # Delete the old authenticator user
+        old_authenticator_user.delete()
+
+        if new_user.pk == old_user.pk:
+            return
```
AlanCoding, line 163: It doesn't have to create the new `AuthenticatorUser`? That's happen through normal auth logic, and by deleting the `old_authenticator_user`, that is unblocked?

EDIT: in the case that old user is new user, we are not de-duplicating users.

ansible_base/authentication/authenticator_plugins/base.py
```diff
@@ -134,3 +135,56 @@ def add_related_fields(self, request, authenticator):
...
+
+        # Copy all of the relationships from the old user to the new one
+        for field in new_user._meta.get_fields():
+            if field.many_to_many is True or field.one_to_many is True:
```
AlanCoding, line 167: To write out what I expect to see here:

```python
In [4]: [field.name for field in User._meta.get_fields() if field.many_to_many is True or field.one_to_many is True]
Out[4]: 
['logentry',
 'social_auth',
 'authenticator_users',
 'role_assignments',
 'has_roles',
 'access_tokens',
 'applications',
 'dab_oauth2_provider_oauth2idtoken',
 'dab_oauth2_provider_oauth2refreshtoken',
 'member_of_organizations',
 'admin_of_organizations',
 'teams',
 'teams_administered',
 'related_fields_test_model_users',
 'animal',
 'animal_friends',
 'oauth2_provider_grant',
 'groups',
 'user_permissions']
```

ansible_base/authentication/authenticator_plugins/base.py
```diff
@@ -134,3 +135,56 @@ def add_related_fields(self, request, authenticator):
...
+        """
+        new_user: django User instance. User that we're moving this account to.
+        old_authenticator_user: AuthenticatorUser instance from this authenticator that is being removed.
+        """
+        exclude_fields = (
+            "social_auth",
+            "authenticator_users",
+            "groups",
+            "has_roles",
+            "role_assignments",
```
AlanCoding, line 153: @relrod I would consider maybe adding `logentry` to this list. Because re-associating the entire activity stream history of a user could time out the request.

I would also consider putting this into a setting, because relationships (similar to logentry, which are toxic in some way or another) will be added downstream.

ansible_base/authentication/utils/authentication.py
```diff
@@ -20,6 +21,57 @@ def setting(self, *args, **kwargs):
...
+    try:
+        main_user = AuthenticatorUser.objects.get(uid=uid, provider=authenticator).user
+    except AuthenticatorUser.DoesNotExist:
+        main_user = migrate_users[0].user
+
+    for migrate_user in migrate_users:
+        provider = migrate_user.provider
+        from_authenticator = get_authenticator_class(provider.type)(database_instance=provider)
+        old_user = from_authenticator.move_authenticator_user_to(main_user, migrate_user)
+        if old_user and not old_user.authenticator_users.exists():
```
AlanCoding, line 61: If the `old_user` is still associated with other, unrelated, authenticators, it will have already had its associations applied to the new user. But reading more closely, it seems that you (reasonably) did not remove those associations. That is probably the most correct behavior. But I think you should issue a warning log in an `else:` here because that's a fairly degenerate state.

I also see that you're not copying `role_assignments` from the prior file... so how are those ever going to get migrated? Because this logic seems to drop them on the floor.

test_app/tests/authentication/test_migrate_user_to_authenticator.py
```diff
@@ -0,0 +1,370 @@
...
+        things=[
+            "thing2",
+            "conflict_thing",
+        ],
+    ).user
+
+    kwargs = get_social_pipeline_kwargs(
+        uid="kc_user", username="kc_user", authenticator=keycloak_authenticator, response={"sub": "a9a3b88a-7c1a-4fd8-b735-f20c338debcc"}
+    )
+    username = determine_username_from_uid_social(**kwargs)
```
AlanCoding, line 84: This seems to call `migrate_from_existing_authenticator`, but it's really non-obvious from _this_ name that this is where the new login happens.

ansible_base/authentication/utils/authentication.py
```diff
@@ -40,19 +92,28 @@ def check_system_username(uid: str) -> None:
         raise AuthException(_('System user is not allowed to log in from external authentication sources.'))
 
 
-def determine_username_from_uid_social(**kwargs: dict) -> dict:
-    uid = kwargs.get('details', {}).get('username', None)
-    if not uid:
+def determine_username_from_uid_social(**kwargs) -> dict:
```
AlanCoding, line 97: Continuing from another comments, I would push for a different name for this method, because it does a whole lot more than what it looks like it does.

AlanCoding (approved):
I'm understanding high-level what this is doing and tests look fairly thorough.

ansible_base/authentication/authenticator_plugins/base.py
relrod: I think `Entry` (not logentry) is fine to not be included. Things that inherit `AuditableModel` get an `activity_stream_entries` property, but it's just that - a property, not a real relation. Since activity stream uses GFKs, there's no reverse relation, so I think we're okay?

ansible_base/authentication/authenticator_plugins/base.py
newswangerd: The pipeline here his a little wonky. If the user is the same, we just delete the authenticator user and then return the username. The username feeds into the rest of the auth pipeline and is used to create the new authenticator user. There might be a smarter way to do this, but the `determine_username_from_uid` is the only place that I can see which hooks into both the password and SSO auth pipelines.

ansible_base/authentication/authenticator_plugins/base.py
newswangerd: I ran into problems with log entry before. That's actually where this check comes into play:

```python
if not hasattr(old_user, name) or not hasattr(new_user, name):
    continue
```

For some reason, even though `logentry` shows up in fields, it's not actually an attribute on the model.

Either way, I'll add it to the exclude list just in case.

ansible_base/authentication/utils/authentication.py
newswangerd: I don't really want to mess with the name of this function, because it gets included in the SOCIAL_AUTH_PIPELINE setting. I know we configure this in the dynamic settings, but it would break any app that isn't using dynamic settings.

I could probably split this into a separate pipeline stage, but I know that this works right now from manual testing. We can't test the full pipeline in CI right now, so I'm not confident enough that I can make this change without breaking something else.

test_app/tests/authentication/test_migrate_user_to_authenticator.py
newswangerd: see my other note about renaming this function.

PR #604, opened 2024-09-03
Update social_auth.py to move log to debug

kdelee:
This was logging 100 times per request and flooding logs


BrennanPaciorek (approved):

BrennanPaciorek:
After fixing the mock on the failing test to mock `debug` instead of `info` this should be good.

john-westcott-iv (approved):

PR #609, opened 2024-09-08
[RR] Send auth backend in auth_code payload

relrod:


newswangerd (approved):

AlanCoding:
This gets serialized from the `/validate-local-account/` endpoint, right? Are we sure this gives a native JSON type?

newswangerd:
> This gets serialized from the `/validate-local-account/` endpoint, right? Are we sure this gives a native JSON type?

@AlanCoding I believe it does. Django stores this on the session when you log in.

Also, I just tested it.

AlanCoding (approved):
Ok, I think I remember it being a string like "keycloak". And it seems to come from

https://docs.djangoproject.com/en/5.1/topics/auth/default/#selecting-the-authentication-backend

> In cases 1 and 2, the value of the backend argument or the user.backend attribute should be a dotted import path string (like that found in [AUTHENTICATION_BACKENDS](https://docs.djangoproject.com/en/5.1/ref/settings/#std-setting-AUTHENTICATION_BACKENDS)), not the actual backend class.

PR #611, opened 2024-09-11
Add authentication backend for renamed user accounts

newswangerd:
This adds two things. The first is a modified version of the django ModelBackend authentication backend that prepends the prefix the user's username and checks it ModelBackend: `ansible_base.lib.backends.prefixed_user_auth.PrefixedUserAuthBackend`

The second is a mixin that allows this functionality to be added to other password based authentication backends, such as LDAP:

```python
from ansible_base.lib.middleware.prefixed_user_auth import PrefixedUserAuthenticationMixin
from django_auth_ldap.backend import LDAPBackend

class PrefixedLDAPBackend(PrefixedUserAuthenticationMixin, LDAPBackend)
    pass
```

Both the custom classes and the one provided by the library can be added to `AUTHENTICATION_BACKENDS`. This will ensure that a user can login with either their prefixed account username, or their original username. 

```python
AUTHENTICATION_BACKENDS = [
    'ansible_base.lib.backends.prefixed_user_auth.PrefixedUserAuthBackend',
    '...',
]

```


AlanCoding:
Should `validate-local-account` endpoint take into account the potential prefix?

https://github.com/ansible/django-ansible-base/blob/9365ce437182b940afa1831f3369215e0c16cf67/ansible_base/resource_registry/views.py#L187

See, "username" used without prefix there.

ansible_base/lib/middleware/prefixed_user_auth.py
relrod, line 1: Can we move this to `ansible_base.lib.backends.prefixed_user_auth` instead of putting it in middleware? middleware is more meant for actual django middleware.

test_app/tests/lib/test_prefixed_authentication_backend.py
```diff
@@ -0,0 +1,35 @@
...
+    # Check that the user can login with their unprefixed username
+    client = APIClient()
+    data = {"username": "foo", "password": "pass"}
+    resp = client.post(url, data=data, follow=True)
+    resp = client.get(me_url)
+
+    assert resp.status_code == 200
+    assert resp.data["username"] == "dab:foo"
+
+    # Check that the user can login with their prefixed username
```
AlanCoding: It might be good to break this up into 2 tests. Then rename `user` and make it a fixture in this same file.

AlanCoding (changes requested):
Agree with rick - move file. Asked to refactor tests slightly. Then ready to merge.

newswangerd:
> Should `validate-local-account` endpoint take into account the potential prefix?
> 
> https://github.com/ansible/django-ansible-base/blob/9365ce437182b940afa1831f3369215e0c16cf67/ansible_base/resource_registry/views.py#L187
> 
> See, "username" used without prefix there.

This just calls the AUTHENTICATION_BACKENDS under the hood, so it should be covered.

test_app/tests/lib/test_prefixed_authentication_backend.py
```diff
@@ -0,0 +1,43 @@
...
+@pytest.fixture
+@pytest.mark.django_db(transaction=True)
+def prefixed_user(local_authenticator):
+    user = User.objects.create(username='dab:foo')
+    user.set_password("pass")
+    user.save()
+
+    yield user
+
+    user.delete()
```
AlanCoding: You don't need this with transactional tests, and it causes side-effect errors when things go wrong.

AlanCoding (approved):

PR #613, opened 2024-09-13
Hide certain plugins from /authenticator_plugins/

relrod:
Be able to have some authenticator plugins be marked as "internal" and thus not shown in the UI as an option when configuring an authenticator.


bhavenst (approved):

relrod:
This is going to be handled by the UI instead.

PR #614, opened 2024-09-13
Pass requesting user into validate role assignment

fosterseth:
For credential and execution env, the callback will attempt to make extra requests to gateway, so it
needs the requesting user.


AlanCoding (approved):
Pasted my confirmation in AWX patch

PR #615, opened 2024-09-13
Force social-auth-app-django to 5.4.1

bhavenst:
* Fixes bug with Github SSO login
  * Compatible with social-auth-core 5.4.4 (latest) which is left unpinned
 
The combination previously in dev worked fine, but the combo in production did not (github social login broken).

It turns out that there was a defect fix partly in social auth core and partly in social-auth-app-django and updating both to latest again works fine.

PDE is making the same updates in production build environment.

### DIFF
--- a/requirements/requirements_all.txt
+++ b/requirements/requirements_all.txt
@@ -137,7 +137,7 @@ six==1.16.0
     #   isodate
     #   pyrad
     #   tacacs-plus
-social-auth-app-django==5.4.0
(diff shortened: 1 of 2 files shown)


relrod (approved):

simaishi:
requirements_all.txt should be updated to match?

bhavenst:
Yeah, good point.  It does get regenerated during build, but since we have a checked in copy, it should match.  Updated.  Thanks!

PR #616, opened 2024-09-16
Only return entries for local service_id

AlanCoding:
AAP-27844

This will avoid pulling entries for periodic resource sync if they are not marked as belonging to the resource server. I'm still unsure if this behavior is desirable.

### DIFF
--- a/ansible_base/resource_registry/views.py
+++ b/ansible_base/resource_registry/views.py
@@ -131,7 +131,15 @@ class ResourceTypeViewSet(
         if not resource_type.serializer_class:  # pragma: no cover
             return HttpResponseNotFound()
 
-        resources = Resource.objects.filter(content_type__resource_type=resource_type).prefetch_related("content_object")
+        if 'service_id' in request.query_params:
+            list_service_id = request.query_params['service_id']
+        else:
+            list_service_id = service_id()
+
+        resources = Resource.objects.filter(
+            content_type__resource_type=resource_type,
+            service_id=list_service_id
+        ).prefetch_related("content_object")
 
         if name == "shared.user" and (system_user := getattr(settings, "SYSTEM_USERNAME", None)):
             resources = resources.exclude(name=system_user)
--- a/ansible_base/resource_registry/views.py
+++ b/ansible_base/resource_registry/views.py
@@ -139,9 +139,7 @@ class ResourceTypeViewSet(
         else:
             service_filter = {'service_id': service_id()}
 
-        resources = Resource.objects.filter(
-            content_type__resource_type=resource_type, **service_filter
-        ).prefetch_related("content_object")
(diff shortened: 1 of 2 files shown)


rochacbruno (approved):

AlanCoding:
I pushed another change, because I'm concerned about a scenario where we want all records. It is very unlikely that it would be practical to get the `service_id` of something aside from the resource server or the local server. So I want to allow `?service_id=` in order to not do any service-specific filter. Whether we use this or not I'm cool, I just want to have this option in case things go badly we have this option to fix it.

test_app/tests/resource_registry/test_views.py
```diff
@@ -20,3 +24,47 @@ def test_validate_local_user(unauthenticated_api_client, admin_user, local_authe
...
+    url = get_relative_url('resourcetype-manifest', kwargs={'name': 'shared.user'})
+    r = client.get(url, data=data)
+    assert r.status_code == expect, f'request data: {data}, url: {url}, response: {r}'
+    return '\n'.join([str(line) for line in r])
+
+
+def test_resource_type_manifest_default(admin_api_client, user):
+    r_text = get_users_manifest(admin_api_client)
+    assert str(user.resource.ansible_id) in r_text
+    assert str(user.resource.ansible_id) in r_text
```
fosterseth: duplicate line

fosterseth (approved):

ansible_base/resource_registry/views.py
```diff
@@ -131,7 +131,17 @@ def manifest(self, request, name, *args, **kwargs):
         if not resource_type.serializer_class:  # pragma: no cover
             return HttpResponseNotFound()
 
-        resources = Resource.objects.filter(content_type__resource_type=resource_type).prefetch_related("content_object")
+        if 'service_id' in request.query_params:
+            if request.query_params['service_id']:
+                service_filter = {'service_id': request.query_params['service_id']}
+            else:
+                service_filter = {}
```
rochacbruno: IMO `?service_id=` is a bit strange, looks like a request error.

I would prefer explicitly `?service_id=all`

AlanCoding: ok, added that. Could use docs, but this should only matter internally for now.

elyezer (commented):
I talked to @AlanCoding and he explained to me the implications of these changes (documented [here](https://issues.redhat.com/browse/AAP-27844?focusedId=25596230&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25596230)). Based on the conversation these changes does not seem to pose any risk of breaking the build and will contribute to upcoming changes for event 2 where because of security concerns we can't upgrade the controller superuser to a Gateway superuser

PR #617, opened 2024-09-24
Redis flushdb

john-westcott-iv:
2 commits in here:
1.  Simplifying how exceptions are handled with a raised translated message. This allows us to catch other exceptions we weren't expecting and exiting with clean errors.
2. Overriding the flushdb method. Currently, in our version of Redis, the flushdb command does not respect ACLs. So when gateway switches from the fallback cache to the primary cache it calls `cache.clear()` this eventually does a `flushdb` command. The flushdb command will not check the connected users to see if they have access to  keys, it  just obliterates everything in the Redis DB. This causes issues in a shared Redis installation. This new method is less efficient but will load all of the keys and then attempt to delete them all and just ignores a `NoPermissionError` if it tried to delete something it didn't have access to.

### DIFF
--- a/ansible_base/lib/redis/client.py
+++ b/ansible_base/lib/redis/client.py
@@ -113,29 +113,30 @@ class RedisClientGetter:
         self.connection_settings.pop('port', None)
         self.connection_settings['startup_nodes'] = []
 
-        translated_generic_exception = ImproperlyConfigured(_('Unable to parse redis_hosts, see logs for more details'))
-
         # Make sure we have a string for redis_hosts
-        if not isinstance(self.redis_hosts, str):
-            logger.error(f"Specified redis_hosts is not a string, got: {self.redis_hosts}")
-            raise translated_generic_exception
+        try:
+            if not isinstance(self.redis_hosts, str):
+                logger.error(f"Specified redis_hosts is not a string, got: {self.redis_hosts}")
+                raise ValueError()
 
-        host_ports = self.redis_hosts.split(',')
-        for host_port in host_ports:
-            try:
-                node, port_string = host_port.split(':')
-            except ValueError:
-                logger.error(f"Specified cluster_host {host_port} is not valid; it needs to be in the format <host>:<port>")
-                raise translated_generic_exception
+            host_ports = self.redis_hosts.split(',')
+            for host_port in host_ports:
+                try:
+                    node, port_string = host_port.split(':')
+                except ValueError:
+                    logger.error(f"Specified cluster_host {host_port} is not valid; it needs to be in the format <host>:<port>")
+                    raise
 
-            # Make sure we have an int for the port
-            try:
-                port = int(port_string)
-            except ValueError:
-                logger.error(f'Specified port on {host_port} is not an int')
-                raise translated_generic_exception
+                # Make sure we have an int for the port
+                try:
+                    port = int(port_string)
+                except ValueError:
+                    logger.error(f'Specified port on {host_port} is not an int')
+                    raise
 
-            self.connection_settings['startup_nodes'].append(ClusterNode(node, port))
+                self.connection_settings['startup_nodes'].append(ClusterNode(node, port))
+        except ValueError:
+            raise ImproperlyConfigured(_('Unable to parse redis_hosts, see logs for more details'))
 
     def __init__(self, *args, **kwargs):
         self.url = ''
--- a/ansible_base/lib/redis/client.py
+++ b/ansible_base/lib/redis/client.py
@@ -1,7 +1,7 @@
 import copy
 import logging
 import os
-from typing import Dict, Union
+from typing import Any, Awaitable, Dict, Optional, Union
 from urllib.parse import parse_qs, urlparse
 
 from django.core.exceptions import ImproperlyConfigured
@@ -9,7 +9,7 @@ from django.utils.translation import gettext as _
 from django_redis.client import DefaultClient
 from redis import Redis
 from redis.cluster import ClusterNode, RedisCluster
-from redis.exceptions import RedisClusterException
+from redis.exceptions import NoPermissionError, RedisClusterException
 
 from ansible_base.lib.constants import STATUS_DEGRADED, STATUS_FAILED, STATUS_GOOD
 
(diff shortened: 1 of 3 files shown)


ansible_base/lib/redis/client.py
```diff
@@ -113,29 +134,30 @@ def _get_hosts(self) -> None:
...
-                port = int(port_string)
-            except ValueError:
-                logger.error(f'Specified port on {host_port} is not an int')
-                raise translated_generic_exception
-
-            self.connection_settings['startup_nodes'].append(ClusterNode(node, port))
+        try:
+            if not isinstance(self.redis_hosts, str):
+                logger.error(f"Specified redis_hosts is not a string, got: {self.redis_hosts}")
+                raise ValueError()
```
chrismeyersfsu, line 158: Took me a little bit to find the value error that the new except is trying to catch.

Consider using try/except/else and move the except that was added below to up here.

https://www.geeksforgeeks.org/try-except-else-and-finally-in-python/

ansible_base/lib/redis/client.py
```diff
@@ -20,11 +20,32 @@
 
 _INVALID_STANDALONE_OPTIONS = ['cluster_error_retry_attempts']
 
+# This will eventually be in the Redis package under redis.typing later
+ResponseT = Union[Awaitable, Any]
+
+
+class DABRedisRespectACLFlushMixin:
```
chrismeyersfsu, line 27: I love an explicit class name :+1: 

Maybe this class could include a comment that captures the digital detect work that went into finding this. Something like ...


FLUSHDB does NOT respect Redis ACL on keys. It flushes _all_ keys in a database. You think your keys are protected because the user you are running FLUSHDB as has write-only access to keys with the pattern `gateway*`, or hell, even read-only access  ...  WRONG, you aren't safe!

And if you are thinking of using Redis' command re-write feature to sweep this problem under the rug then god help the developer or support person that comes after you.

ansible_base/lib/redis/client.py
chrismeyersfsu: Talked with @john-westcott-iv offline. I was wrong with this suggestion.

ansible_base/lib/redis/client.py
john-westcott-iv: Incorporated

ansible_base/lib/redis/client.py
```diff
@@ -20,11 +20,46 @@
...
+
+    And if you are thinking of using Redis' command re-write feature to sweep this
+    problem under the rug then god help the developer or support person that comes after you.
+
+    At some point in the future we may also want to override flushall as well
+    """
+
+    def flushdb(self, asynchronous: Optional[bool] = None, **kwargs) -> ResponseT:
+        if asynchronous is not None:
+            logger.warning("DABRedis clients implement an ACL friendly FLUSHDB which can not be async")
```
chrismeyersfsu, line 44: Just reading over the tests and the question popped into my head. Would it be better to raise an exception here, `ValueError`, to notify the developer that `asynchronous` isn't supported?

test_app/tests/lib/redis/test_client.py
```diff
@@ -396,3 +398,89 @@ def test_redis_standalone_removes_cluster_settings():
...
+def test_ensure_flushdb_does_not_die_on_no_permission_exception():
+    args = {'mode': 'standalone'}
+    from redis.exceptions import NoPermissionError
+
+    from ansible_base.lib.redis.client import RedisClientGetter
+
+    with mock.patch('redis.Redis.__init__', return_value=None):
+        client_getter = RedisClientGetter()
+        client = client_getter.get_client('rediss://localhost', **args)
+        client.keys = mock.MagicMock(return_value=[])
```
chrismeyersfsu: Keys needs to return something so that `delete` gets called.

test_app/tests/lib/redis/test_client.py
```diff
@@ -396,3 +398,89 @@ def test_redis_standalone_removes_cluster_settings():
...
+        client.delete = mock.MagicMock(side_effect=NoPermissionError())
```
chrismeyersfsu, line 450: Consider adding an assertion on `client.delete()` getting called.

ansible_base/lib/redis/client.py
john-westcott-iv: I was concerned that if there was another library that was passing us the async it might be out of the devs control. i.e. if django Redis cache changed to pass async=True there is nothing a dev could really do about it.

test_app/tests/lib/redis/test_client.py
john-westcott-iv: For this particular test I was more concerned with how the code handles the NoPermisionError but that being said the failing test indicates that there are no keys in Redis to delete so I think I need to wrap the delete in an `if` condition to ensure there is 1 or more keys to be deleted so this test will need to change.

test_app/tests/lib/redis/test_client.py
```diff
@@ -396,3 +398,89 @@ def test_redis_standalone_removes_cluster_settings():
...
+    from ansible_base.lib.redis.client import RedisClientGetter
+
+    with mock.patch('redis.RedisCluster.__init__', return_value=None):
+        client_getter = RedisClientGetter()
+        client = client_getter.get_client('rediss://localhost', **args)
+        client.keys = mock.MagicMock(return_value=[])
+        client.delete = mock.MagicMock(return_value=None)
+        # This should just not raise if NoPermissionError is raised
+        client.flushdb()
+        client.keys.assert_called_with("*", target_nodes=client.ALL_NODES)
```
chrismeyersfsu, line 468: :+1:

test_app/tests/lib/redis/test_client.py
```diff
@@ -396,3 +398,89 @@ def test_redis_standalone_removes_cluster_settings():
...
+
+
+def test_flushdb_some_keys_delete_some_do_not(monkeypatch):
```
chrismeyersfsu: Seems like this test covers `test_ensure_flushdb_does_not_die_on_no_permission_exception` . Maybe delete the other test.

test_app/tests/lib/redis/test_client.py
john-westcott-iv: Done!

test_app/tests/lib/redis/test_client.py
john-westcott-iv: Yea, this one is harder because we don't have an actual Redis running with ACLs in the unit tests. The goal was to mock the delete  function so that one key would be deleted and one would not. But in the end, it's just raising the same exception. So I'll just delete this test. It's technically not even really working  because  key_name would be an array of `['key1', 'key2']`. The more I think about it, I don't think this test provides any value to us.

bhavenst (approved):
Looks good!  I don't see anything to comment on.  Ran all the unit tests as well with no issues locally.

PR #618, opened 2024-09-24
[Alt 1] Flip default of reverse sync to True

AlanCoding:
While revisiting https://github.com/ansible/eda-server/pull/1061, I realized that `RESOURCE_SERVER_SYNC_ENABLED` is not set to True.

This will not work to do what we want.

As of https://github.com/ansible/django-ansible-base/pull/603, for any practical system (excluding test_app specifically here), we will be safe to have a True value.

For some reason, we had planned to make this change in apps like AWX / eda-server, but this doesn't make sense to me. Making this change seems like the most reasonable.

### DIFF
--- a/ansible_base/lib/dynamic_config/settings_logic.py
+++ b/ansible_base/lib/dynamic_config/settings_logic.py
@@ -226,7 +226,10 @@ def get_dab_settings(
         dab_data['ORG_ADMINS_CAN_SEE_ALL_USERS'] = True
 
     if 'ansible_base.resource_registry' in installed_apps:
-        dab_data['RESOURCE_SERVER_SYNC_ENABLED'] = False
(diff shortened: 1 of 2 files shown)


AlanCoding:
First thing I wanted to establish is that
 - Testing DAB Consumers / AWX (pull_request) 
 - Testing DAB Consumers / eda-server (pull_request)

Are both passing.

We will still need https://github.com/ansible/eda-server/pull/1061 to actually use the feature, but the checks themselves don't have any coverage for it, so that's why they can pass now.

fosterseth (approved):

AlanCoding:
Votes were in favor of the other alternative

AlanCoding:
Another update - the other alternative is blocked because it requires more installer changes. So this is our only sensible option for now.

PR #619, opened 2024-09-25
Throw error if reverse-sync needed, but not allowed

AlanCoding:
This is a re-do of https://github.com/ansible/django-ansible-base/pull/618 that attempts to answer the question "what if someone runs `awx-manage createsuperuser` manually?"

If that command is ran with explicit disabling of reverse-sync, then it will be allowed. It _will be_ confusing that doing this is still permitted even through local_resource_management is off.

But what's the _most_ wrong is to have
 - ALLOW_LOCAL_RESOURCE_MANAGEMENT False
 - reverse sync enabled (multiple mechanisms involved)

The argument is that this configuration is so dangerously contradictory that we should throw an error so that we can front-load debugging of the situation.

### DIFF
--- a/ansible_base/resource_registry/utils/sync_to_resource_server.py
+++ b/ansible_base/resource_registry/utils/sync_to_resource_server.py
@@ -55,6 +55,16 @@ def sync_to_resource_server(instance, action, ansible_id=None):
         logger.info(f"Skipping sync of resource {instance} because its service_id is local")
         return
 
+    # By this point we have passed the exit conditions
+    # If an edit requires a resource sync, but we intended to not allow local edits
+    # that has led to a contradiction, normally this is enforced by the service API
+    # but for direct ORM actions we throw an error as a last-resort safety measure
+    if not getattr(settings, "ALLOW_LOCAL_RESOURCE_MANAGEMENT"):
+        raise ValidationError(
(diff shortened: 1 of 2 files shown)


ansible_base/resource_registry/utils/sync_to_resource_server.py
```diff
@@ -55,6 +55,13 @@ def sync_to_resource_server(instance, action, ansible_id=None):
         logger.info(f"Skipping sync of resource {instance} because its service_id is local")
         return
 
+    # By this point we have passed the exit conditions
+    # If an edit requires a resource sync, but we intended to not allow local edits
+    # that has led to a contradiction, normally this is enforced by he service API
```
relrod:
```suggestion
    # that has led to a contradiction, normally this is enforced by the service API
```

relrod:
Comments:

- Sonar can be ignored here, we use `Exception` above this anyway.
- I like this approach better than #618, this has my vote.
- Can we add a quick unit test for this in test_app/tests/resource_registry/test_utils.py

Beyond that, this LGTM.

fosterseth (dismissed):

AlanCoding:
I identified a conflict with awx-operator, and probably more. This would probably break things, and our viable path forward may be to go back to Alt 1 out of necessity.

AlanCoding:
Internal reference: AAP-33341

We should not solve the problem this way. We should solve this by principal of least-privilege, limiting scope of the token

PR #620, opened 2024-09-26
Update LDAP user search test

bhavenst:
* Blank UI field for user search results in the value []
 * Current test to ignore is "if field is None:", which fails since [] is not None
 * With "if field:", [] is False and properly ignored

LDAP auth should require one of User Search or User DN Template, but not both.  Currently LDAP auth is blocked if User Search is empty ([]), this change fixes this issue and allows auth when only User DN Template is specified.

This is for: https://issues.redhat.com/browse/AAP-28020

That Jira is now closed, but this is good to fix anyway.

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/ldap.py
+++ b/ansible_base/authentication/authenticator_plugins/ldap.py
@@ -374,7 +374,7 @@ class AuthenticatorPlugin(LDAPBackend, AbstractAuthenticatorPlugin):
         # Ensure USER_SEARCH and GROUP_SEARCH are converted into a search object
         for field, search_must_have_user in [('GROUP_SEARCH', False), ('USER_SEARCH', True)]:
             data = getattr(self.settings, field, None)
-            if data is None:
+            if not data:
                 setattr(self.settings, field, None)
             elif not isinstance(data, config.LDAPSearch):
                 try:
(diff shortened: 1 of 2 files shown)


tznamena:
Shouldn't we add a test for this as well?

bhavenst:
Yeah, was in progress..  Added a case for it now.

bhavenst:
Oops, test didn't hit new code..  Standby..

bhavenst:
OK, unit test added to cover this strange scenario.

ansible_base/authentication/authenticator_plugins/ldap.py
```diff
@@ -374,7 +374,7 @@ def authenticate(self, request, username=None, password=None, **kwargs) -> (obje
         # Ensure USER_SEARCH and GROUP_SEARCH are converted into a search object
         for field, search_must_have_user in [('GROUP_SEARCH', False), ('USER_SEARCH', True)]:
             data = getattr(self.settings, field, None)
-            if data is None:
+            if not data:
                 setattr(self.settings, field, None)
             elif not isinstance(data, config.LDAPSearch):
                 try:
```
art-tapin: Hey Bryan, I added this comment to make the condition clearer. Handling [] or None is a common pitfall, and this should help avoid confusion. Just a small change to make the intent more obvious! 😊 Other than that, LGTM!

```python
# Ignore None or empty (e.g., [])
if not data:
    setattr(self.settings, field, None)

art-tapin (commented):
:1st_place_medal:

ansible_base/authentication/authenticator_plugins/ldap.py
bhavenst: Very nice, thanks!

art-tapin (approved):

PR #626, opened 2024-10-07
[Looking For Feedback] Add feature app

john-westcott-iv:
Adds a DAB application whose goal is to give the platform a standard way to have "feature flags".
This would enable us to develop features which could be enabled/disabled by an admin user. 
There are likely many aspects of development which would need to be thought about for feature development.


chrismeyersfsu:
I like that this is a separate app. I could see this being its own microservice .. but not now .. for god sakes not now.

:+1: :heavy_plus_sign: 

I'm going to go into bigger picture thoughts. Features flagging is, basically, settings. The landscape of considerations that come with settings. Being able to change the value without restart the system and have them take effect, this is a cache problem. Particularly, invalidation when a value changes. Controller has a hacky feature for reloading rsyslog when external logging config changes. We run a service that listen to pgnotify. A signal is sent when the external logging settings are sent, the service re-writes the rsyslog config file and restart the service.

@TheRealHaoLiu has given thought to this topic.

docs/apps/features.md
```diff
@@ -0,0 +1,79 @@
...
+        'status': 'a',
+        'requires_restart': True,
+    }
+]
+```
+
+`short_name` str, an easy to reference name that will be used as an identifier inside the code
+`name` str, the display name for the feature (end user facing)
+`description` str, a brief description of the feature
+`status` one of ['a', 'b'], the status of the feature (Alpha or Beta, see Feature Status)
```
Ladas, line 43: so we should also have teach preview and GA? Because there will be normal supported features here?

matburt:
I kinda disagree with what this is doing... Feature Flags isn't a core capability of AAP, we'd have to maintain this. This is something we should pull off the shelf and my handbook FF proposal does just that.

PR #628, opened 2024-10-10
Groundwork for better DRF pages for services

relrod:
- Introduce an `is_proxied_request` for services to use
- Introduce a Django template tag that returns the result of the above
- Introduce a Django template tag for inlining files into templates


relrod:
This probably needs a few tests for the template tags, sonar will whine.

AlanCoding (approved):
This seems really clean, and all the utils here look very proper in the domain where they exist. I don't want AWX code telling whether it was a proxied request. And the custom template tag is :fire:

PR #629, opened 2024-10-11
They say sometimes I don't give an f. Fix that.

relrod:
Add missing f to format string.


art-tapin (approved):
F-antastic. 😂 Glad to see we’re finally giving ‘f’s where they’re due. The PR message deserves a place in the commit hall of fame. F-lawless! 🎖️

john-westcott-iv (approved):

PR #630, opened 2024-10-17
Improve start/stop of postgres container for UT

bhavenst:
Normal cases (for tox -e pyxyz):

dab_postgres container running:
  - just run tests

dab_postgres doesn't exist:
 - create/start container
 - run tests
 - destroy container

Odd case:
dab_postgres exists but not running:
 - start container
 - run tests
 - destroy container

This last one doesn't leave the system as found exactly.  Is that OK or should that situation also be handled?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ htmlcov
 *.tox
 venv/
 .venv/
+.pg_started
 
 # Mac OS X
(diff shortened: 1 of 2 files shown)


bhavenst:
Decided to take a better approach @ this.

PR #632, opened 2024-10-23
Auto start/stop postgres w/ tox

bhavenst:
Better implementation of postgres container handling for dab unit tests/tox.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,7 +109,8 @@ legacy_tox_ini = """
         -r{toxinidir}/requirements/requirements_all.txt
         -r{toxinidir}/requirements/requirements_dev.txt
     allowlist_externals = sh
-    commands = sh -c 'make postgres && pytest -n auto --cov=. --cov-report=xml:coverage.xml --cov-report=html --cov-report=json --cov-branch --junit-xml=django-ansible-base-test-results.xml {env:ANSIBLE_BASE_PYTEST_ARGS} {env:ANSIBLE_BASE_TEST_DIRS:test_app/tests} {posargs}'
+    commands = pytest -n auto --cov=. --cov-report=xml:coverage.xml --cov-report=html --cov-report=json --cov-branch --junit-xml=django-ansible-base-test-results.xml {env:ANSIBLE_BASE_PYTEST_ARGS} test_app/tests/test_demo_data.py {posargs}
+    docker = db
 
     [testenv:check]
     deps =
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -121,6 +122,19 @@ legacy_tox_ini = """
...
+    environment =
+        POSTGRES_PASSWORD=dabing
+        POSTGRES_USER=dab
+        POSTGRES_DB=dab_db
+    healthcheck_cmd = pg_isready -U dab -d dab_db
+    healthcheck_interval = 10
+    healthcheck_timeout = 5
+    healthcheck_retries = 5
+    expose =
+        DB_PORT=5432/tcp
```
AlanCoding, line 130: I would really rather not _duplicate_ the parameters. Even the port doesn't seem to be completely the same as

https://github.com/ansible/django-ansible-base/blob/devel/docker-compose.yml

But looking into tox-docker library, unfortunately, it looks like this may not be possible.

Re-phrasing, I would rather this _not_ list things like `image` and `environment`, but instead, reference the name of the service and the docker-compose.yml file.

https://github.com/tox-dev/tox-docker

Similar to:

```ini
[docker:app]
dockerfile = {toxinidir}/Dockerfile
```

but instead of Dockerfile, I'd like it to reference the docker-compose file, so we are only maintaining 1 thing.

bhavenst: Yeah I had the same thought.  This is really an entirely different DB though, that only exists during the tox run (unless the flag is added to hold it for debugging) and for different purposes so I wasn't sure it is a big deal.  I could make the values totally different so it doesn't seem like duplication. :)

The port difference is because tox-docker handles it automatically, allowing parallel tox execution by using a different port for each instance.  It sets DB_PORT for the test session to whatever it chose.  But not sure if this is valuable for us (doubt it).

I also looked into the tool docs as well and one method I found that could consolidate them would be to use an environment file called .devdata.env which is loaded by tox-envfile plugin and can be also loaded by docker compose.  As you found, tox-docker does not support docker compose specs.

I suppose another method could be to extract a dockerfile and use the dockerfile directive in docker compose for the db container (and also use it in tox).  There we would ~~lose the auto-port function~~, but I'm not sure if we would ever run parallel tox sessions on the same host anyway.

AlanCoding: > I also looked into the tool docs as well and one method I found that could consolidate them would be to use an environment file called .devdata.env which is loaded by tox-envfile plugin and can be also loaded by docker compose.

Re-stating for posterity, `tox-envfile` would be adding an entirely new plugin. There is a certain hackiness that appears to me with this approach too.

Anyway, it's not that I want to block this, I just want to ask the question and perform due diligence. If I'm really the first one to ask the question, then I want to file it in the appropriate place. Then we may accept the duplication as an upstream issue.

https://github.com/tox-dev/tox-docker/issues/190

bhavenst: Hi @AlanCoding Please take a look now.  I used the common Dockerfile approach and that seems to work well for both tools.  If tox-docker ever decides to support docker-compose then it can be moved back into the docker-compose.yml.

AlanCoding (approved):
Checked out the branch, and local dev processes still seem to work.

PR #633, opened 2024-10-24
Adding requirements file for resource_registry

john-westcott-iv:
In creating a new Django application I found that the resource registry was missing a requirements file. I went through all of the imports and created one specifically for the registry.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ optional-dependencies.all = { file = [
     "requirements/requirements_testing.in",
     "requirements/requirements_redis_client.in",
     "requirements/requirements_oauth2_provider.in",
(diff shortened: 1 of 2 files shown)


ansible_base/resource_registry/imports.txt
```diff
@@ -0,0 +1,6 @@
+from social_django.utils import load_strategy
+from asgiref.sync import sync_to_async
+from crum import get_current_user
+import jwt
+import requests
+import urllib3
```
AlanCoding: text file? what's going on here?

john-westcott-iv: Thanks for catching this, fixed!

newswangerd (approved):

PR #636, opened 2024-10-30
feat(oauth2_provider): check for OAuth2ScopePermission on all APIViews

BrennanPaciorek:
Add a check for OAuth2ScopePermission on all active views in an application.
- ANSIBLE_BASE_OAUTH2_PROVIDER_PERMISSIONS_CHECK_IGNORED_VIEWS django setting
  for setting ignores by class path.
- Uses django's check framework, should be compatible with the related
  functionality.
- Set as a deployment check, so it will not block app startup.

AAP-26507

### DIFF
--- a/ansible_base/lib/dynamic_config/dynamic_urls.py
+++ b/ansible_base/lib/dynamic_config/dynamic_urls.py
@@ -10,7 +10,8 @@ url_types = ['api_version_urls', 'root_urls', 'api_urls']
 for url_type in url_types:
     globals()[url_type] = []
 
-for app in getattr(settings, 'INSTALLED_APPS', []):
+installed_apps = getattr(settings, 'INSTALLED_APPS', [])
+for app in installed_apps:
     if app.startswith('ansible_base.'):
         if not importlib.util.find_spec(f'{app}.urls'):
             logger.debug(f'Module {app} does not specify urls.py')
@@ -18,4 +19,5 @@ for app in getattr(settings, 'INSTALLED_APPS', []):
         url_module = __import__(f'{app}.urls', fromlist=url_types)
         logger.debug(f'Including URLS from {app}.urls')
         for url_type in ['api_version_urls', 'root_urls', 'api_urls']:
-            globals()[url_type].extend(getattr(url_module, url_type, []))
+            urls = getattr(url_module, url_type, [])
+            globals()[url_type].extend(urls)
--- a/ansible_base/lib/dynamic_config/settings_logic.py
+++ b/ansible_base/lib/dynamic_config/settings_logic.py
@@ -283,6 +283,8 @@ def get_dab_settings(
 
         dab_data['ALLOW_OAUTH2_FOR_EXTERNAL_USERS'] = False
 
+        dab_data['ANSIBLE_BASE_OAUTH2_PROVIDER_PERMISSIONS_CHECK_DEFAULT_IGNORED_VIEWS'] = []
+
     if caches is not None:
         dab_data['CACHES'] = copy(caches)
         # Ensure proper configuration for fallback cache
--- a/ansible_base/lib/utils/views/permissions.py
+++ b/ansible_base/lib/utils/views/permissions.py
@@ -1,5 +1,25 @@
 from rest_framework.permissions import SAFE_METHODS, BasePermission
 
+from ansible_base.lib.utils.settings import get_setting
+
+oauth2_provider_installed = "ansible_base.oauth2_provider" in get_setting("INSTALLED_APPS", [])
+
+
+def try_add_oauth2_scope_permission(permission_classes: list):
+    """
+    Attach OAuth2ScopePermission to the provided permission_classes list
+
+    :param permission_classes: list of rest_framework permissions
+    :return: A list of permission_classes, including OAuth2ScopePermission
+        if ansible_base.oauth2_provider is installed; otherwise the same
+        permission_classes list supplied to the function
+    """
+    if oauth2_provider_installed:
+        from ansible_base.oauth2_provider.permissions import OAuth2ScopePermission
+
+        return [OAuth2ScopePermission] + permission_classes
+    return permission_classes
+
 
 class IsSuperuser(BasePermission):
     """
--- a/ansible_base/lib/utils/views/urls.py
+++ b/ansible_base/lib/utils/views/urls.py
@@ -1,0 +1,24 @@
+from typing import Type
+
+from rest_framework.schemas.generators import EndpointEnumerator
+from rest_framework.views import APIView
+
+
+def get_api_view_functions(urlpatterns=None) -> set[Type[APIView]]:
+    """
+    Extract view classes from a urlpatterns list using the show_urls helper functions
+
+    :param urlpatterns: django urlpatterns list
+    :return: set of all view classes used by the urlpatterns list
+    """
+    views = set()
+
+    enumerator = EndpointEnumerator()
+    # Get all active APIViews from urlconf
+    endpoints = enumerator.get_api_endpoints(patterns=urlpatterns)
+    for _, _, func in endpoints:
+        # ApiView.as_view() breadcrumb
+        if hasattr(func, 'cls'):
+            views.add(func.cls)
+
+    return views
--- a/ansible_base/oauth2_provider/checks/permisssions_check.py
+++ b/ansible_base/oauth2_provider/checks/permisssions_check.py
@@ -1,0 +1,153 @@
+from typing import Iterable, Optional, Type, Union
+
+from django.apps import AppConfig
+from django.conf import settings
+from django.core.checks import CheckMessage, Debug, Error, Warning
+from rest_framework.permissions import AllowAny, OperandHolder, OperationHolderMixin, SingleOperandHolder
+from rest_framework.views import APIView
+
+from ansible_base.lib.utils.views.urls import get_api_view_functions
+from ansible_base.oauth2_provider.permissions import OAuth2ScopePermission
+
+
+class OAuth2ScopePermissionCheck:
+    """
+    Class containing logic for checking view classes for the
+    OAuth2ScopePermission permission_class, and aggregating
+    CheckMessage's for django system checks.
+
+    :param ignore_list: List of python import path strings for view classes exempt from the check logic.
+    :type ignore_list: list
+    """
+
+    def __init__(self, ignore_list: Iterable[str], generate_check_messages=True):
+        self.messages: list[CheckMessage] = []
+        self.current_view: Optional[Type[APIView]] = None
+        self.ignore_list = ignore_list
+        self.generate_check_messages = generate_check_messages
+
+    def check_message(self, message: CheckMessage):
+        if self.generate_check_messages:
+            self.messages.append(message)
+
+    # These are all warning or error conditions, this function is mostly saying to not invert OAuth2ScopePermissions
+    # Returns False.
+    def process_single_operand_holder(self, operand_holder: SingleOperandHolder) -> bool:
+        # The only unary operand for permission classes provided by rest_framework is ~ (not)
+        if self.parse_permission_class(operand_holder.op1_class):
+            self.check_message(
+                Warning(
+                    "~ (not) operand used on OAuth2ScopePermission, probably a bad idea.",
+                    id="ansible_base.oauth2_provider.W001",
+                    obj=self.current_view,
+                )
+            )
+
+        return False
+
+    def process_operand_holder(self, operand_holder: OperandHolder) -> bool:
+        return self.parse_permission_class(operand_holder.op1_class) or self.parse_permission_class(operand_holder.op2_class)
+
+    # Check if permission class is present in nested operands
+    # Sort of recursive? Reasonably this should not be an issue, so long as we don't recurse on an unknown OperationHolderMixin type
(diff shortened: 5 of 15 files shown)


ansible_base/oauth2_provider/checks/permisssions_check.py
```diff
@@ -0,0 +1,154 @@
...
+from ansible_base.oauth2_provider.permissions import OAuth2ScopePermission
+
+# Fallback value in the event that ANSIBLE_BASE_OAUTH2_PROVIDER_PERMISSIONS_CHECK_IGNORED_VIEWS setting is undefined
+FALLBACK_OAUTH2_SCOPE_PERMISSION_CHECK_IGNORE_LIST = set(
+    [
+        "drf_spectacular.views.SpectacularAPIView",
+        "drf_spectacular.views.SpectacularRedocView",
+        "drf_spectacular.views.SpectacularSwaggerView",
+    ]
+)
```
BrennanPaciorek: Can probably be done away with in favor of a dynamic setting for oauth2_provider.

BrennanPaciorek:
Manual testing info:

Start the test_app, run `django-admin check --deploy --tag oauth2_permissions`. Test this for any other client services you'd like. You may need to set ignores for non-DAB APIView's by appending to the ANSIBLE_BASE_OAUTH2_PROVIDER_PERMISSIONS_CHECK_IGNORED_VIEWS setting.

ansible_base/lib/dynamic_config/dynamic_urls.py
```diff
@@ -10,12 +10,19 @@
...
         url_module = __import__(f'{app}.urls', fromlist=url_types)
         logger.debug(f'Including URLS from {app}.urls')
         for url_type in ['api_version_urls', 'root_urls', 'api_urls']:
-            globals()[url_type].extend(getattr(url_module, url_type, []))
+            urls = getattr(url_module, url_type, [])
+            if oauth2_provider_installed:
+                from ansible_base.oauth2_provider.utils import attach_oauth2_permission_class_to_urlpatterns
+
+                attach_oauth2_permission_class_to_urlpatterns(urls)
+            globals()[url_type].extend(urls)
```
john-westcott-iv: I think I would remove this from here. If people are using DAB apps we expect them to be using dynamic_urls but they technically don't have to. If we are "fixing" views at runtime in this and someone is not using dynamic_urls they will not get this benefit. Instead we should explicitly add the permission to the DAB views and raise CI errors if a new endpoint is missing the permission class.

ansible_base/lib/dynamic_config/settings_logic.py
```diff
@@ -283,6 +283,17 @@ def get_dab_settings(
 
         dab_data['ALLOW_OAUTH2_FOR_EXTERNAL_USERS'] = False
 
+        dab_data['ANSIBLE_BASE_OAUTH2_PROVIDER_PERMISSIONS_CHECK_IGNORED_VIEWS'] = [
+            'drf_spectacular.views.SpectacularAPIView',
+            'drf_spectacular.views.SpectacularRedocView',
+            'drf_spectacular.views.SpectacularSwaggerView',
+            'ansible_base.oauth2_provider.views.authorization_root.ApiOAuthAuthorizationRootView',
```
john-westcott-iv: I'd like to try an minimize the need to put views in this setting. If possible it would be nice for the check to just ignore AllowAny views rather than having to add them to this list.

ansible_base/lib/dynamic_config/settings_logic.py
```diff
@@ -283,6 +283,17 @@ def get_dab_settings(
 
         dab_data['ALLOW_OAUTH2_FOR_EXTERNAL_USERS'] = False
 
+        dab_data['ANSIBLE_BASE_OAUTH2_PROVIDER_PERMISSIONS_CHECK_IGNORED_VIEWS'] = [
```
john-westcott-iv: Add `if ansible_base.api_docs in installed_apps:`. Also. if drf_spectacular has its own settings for permissions classes we could add a second check to make sure those aren't set and then add these entries.

ansible_base/lib/dynamic_config/settings_logic.py
john-westcott-iv: Also, I think we should change this so that there are the default DAB views that will be ignored and then a service can add additional views it wants to ignore but it wouldn't be able to delete the items from DAB itself.

ansible_base/lib/utils/views/urls.py
```diff
@@ -0,0 +1,29 @@
+from typing import Type, Union
+
+from django_extensions.management.commands.show_urls import Command as ShowUrlsCommand
+from rest_framework.views import APIView, View
+
+
+def get_view_functions(urlpatterns) -> set[Union[Type[APIView], Type[View]]]:
```
john-westcott-iv: In other services we use Rest Service UrlEnumerator  ti get the active URLs for the service differently from the `ShowUrlsCommand`, this might enable you to not have to pass in urlpatters and only need to run this once for the entire application.

ansible_base/lib/utils/views/urls.py
```diff
@@ -0,0 +1,29 @@
...
+
+    # Get all active views
+    view_functions = ShowUrlsCommand().extract_views_from_urlpatterns(urlpatterns)
+    for func, _, _ in view_functions:
+        view = None
+        # ApiView.as_view() breadcrumb
+        if hasattr(func, 'cls'):
+            view = func.cls
+        if hasattr(func, "view_class"):
+            view = func.view_class
```
john-westcott-iv: Lets double check that you can't change the permissions classes for a specific view method. i.e. have [AllowAny] just on an options view.

ansible_base/lib/dynamic_config/settings_logic.py
BrennanPaciorek: Leaving the addition of the ignore up to the check users, since by default api_docs uses AllowAny as far as I'm aware.

ansible_base/lib/dynamic_config/settings_logic.py
BrennanPaciorek: I've got it down to 0 for DAB views, split the setttings into `ANSIBLE_BASE_OAUTH2_PROVIDER_DEFAULT_IGNORED_VIEWS` (for use in dab) and `ANSIBLE_BASE_OAUTH2_PROVIDER_IGNORED_VIEWS` (for use in applications consuming dab).

ansible_base/lib/utils/views/urls.py
BrennanPaciorek: Testing it out, does not seem like the permission class decorator influences permissions for individual functions on `APIView`s

BrennanPaciorek: My test:

```python
from rest_framework.decorators import permission_classes as mutate_permission_classes
# ...
class ApiRootView(AnsibleBaseView):
    permission_classes = (AllowAny,)

    @mutate_permission_classes([IsSuperuserOrAuditor])
    def get(self, request, format=None):
        # ...
        return Response(data)
```

ansible_base/lib/utils/views/urls.py
BrennanPaciorek: The `EndpointEnumerator` appears to work, but it filters out non-APIViews. The result is cleaner though.

BrennanPaciorek: I would say that this tradeoff is worth it, so I'm applying the changes. Feel free to un-resolve this if you disagree.

ansible_base/lib/utils/views/decorators.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    Prepend OAuth2ScopePermission to all APIVews that do not have
+    blank permission classes or AllowAny in its permission classes.
+
+    Intended to be used as a decorator for APIView classes.
+
+    :param view_class: rest_framework APIView instance
+    """
+    if oauth2_provider_installed:
+        from ansible_base.oauth2_provider.checks.permisssions_check import OAuth2ScopePermissionCheck
```
relrod: So, I'm not sure about this for a few reasons. Generally, using `inspect` like the check does in production code is something we want to avoid. But also I feel like having a decorator here makes it less clear what is going on.

I think it is better for views and viewsets to explicitly add `OAuth2ScopePermission` to their `permission_classes`, then it's clear at a glance what is happening. Particularly around security/permissions code, I think being explicit and clear is important.

I also feel weird about using a Django "check" in production code in general. I think it's better to just catch this stuff at CI time.

Thoughts on that?

BrennanPaciorek: `inspect` has been removed from the check in general now.

I've also changed the mechanism for adding `OAuth2ScopePermission` to other django apps that do not require `oauth2_provider` to be installed.

One thing that I want to avoid is writing this in every time we reassign permission classes:
```python
if "ansible_base.oauth2_provider" in settings.INSTALLED_APPS:
    # import OAuth2ScopePermission
    permission_classes += OAuth2ScopePermission
```

ansible_base/lib/utils/views/permissions.py
```diff
@@ -1,5 +1,24 @@
+from django.conf import settings
 from rest_framework.permissions import SAFE_METHODS, BasePermission
 
+oauth2_provider_installed = "ansible_base.oauth2_provider" in getattr(settings, "INSTALLED_APPS", [])
```
john-westcott-iv: There is be an `ansible_base.lib.utils.settings get_setting` function which will search in both the services settings as well as the actual settings. It might be "safer" to use that here (for now).

ansible_base/oauth2_provider/views/application.py
```diff
@@ -3,10 +3,11 @@
...
 from ansible_base.oauth2_provider.models import OAuth2Application
+from ansible_base.oauth2_provider.permissions import OAuth2ScopePermission
 from ansible_base.oauth2_provider.serializers import OAuth2ApplicationSerializer
 
 
 class OAuth2ApplicationViewSet(AnsibleBaseDjangoAppApiView, ModelViewSet):
     queryset = OAuth2Application.objects.all()
     serializer_class = OAuth2ApplicationSerializer
-    permission_classes = [IsSuperuserOrAuditor]
+    permission_classes = [OAuth2ScopePermission, IsSuperuserOrAuditor]
```
john-westcott-iv, line 13: Should this be wrapped in`try_add_oauth2_scope_permission`?

ansible_base/oauth2_provider/views/token.py
```diff
@@ -54,4 +55,4 @@ def create_token_response(self, request):
 class OAuth2TokenViewSet(ModelViewSet, AnsibleBaseDjangoAppApiView):
     queryset = OAuth2AccessToken.objects.all()
     serializer_class = OAuth2TokenSerializer
-    permission_classes = [OAuth2TokenPermission]
+    permission_classes = [OAuth2ScopePermission, OAuth2TokenPermission]
```
john-westcott-iv, line 77: Should this also be wrapped in `try_add_oauth2_scope_permission`?

test_app/settings.py
```diff
@@ -185,6 +185,7 @@
 ANSIBLE_BASE_JWT_MANAGED_ROLES.append("System Auditor")  # noqa: F821 this is set by dynamic settings for jwt_consumer
 ANSIBLE_BASE_ALLOW_SINGLETON_USER_ROLES = True
 ANSIBLE_BASE_ALLOW_SINGLETON_TEAM_ROLES = True
+ANSIBLE_BASE_OAUTH2_PROVIDER_PERMISSIONS_CHECK_IGNORED_VIEWS = ["drf_spectacular.views.SpectacularSwaggerView"]
```
john-westcott-iv, line 188: Should this be added by default in the dynamic settings? That way if anything implements dab_api_documentation they "get it for free"?

test_app/views.py
```diff
@@ -57,7 +58,7 @@ class TeamViewSet(TestAppViewSet):
 
 class UserViewSet(DABOAuth2UserViewsetMixin, TestAppViewSet):
     queryset = models.User.objects.all()
-    permission_classes = [AnsibleBaseUserPermissions]
+    permission_classes = [OAuth2ScopePermission, AnsibleBaseUserPermissions]
```
john-westcott-iv, line 61: Should this be wrapped in try_add_oauth2_scope_permission or do we expect services to directly include this permission?

ansible_base/oauth2_provider/views/application.py
BrennanPaciorek: I don't think so, this ViewSet is in `ansible_base.oauth2_provider` so there is not enough risk of this being used without the needed dependencies to use the less definitive `try_add_oauth2_scope_permission`

ansible_base/oauth2_provider/views/token.py
BrennanPaciorek: I think no for the same reason as the other one in `ansible_base.oauth2_provider`

test_app/settings.py
BrennanPaciorek: I left this ignore in for testing purposes (to reach the ignore code path), but the defaults should not need it for the view to pass.

test_app/views.py
BrennanPaciorek: I think we can expect services to directly include the permission, since they have the app installed and hopefully read the documentation to know that the permission class is necessary, and know about the check to find views where the permission class is not present.

john-westcott-iv (approved):

PR #637, opened 2024-11-04
Fix issue where RADIUS auth logged to root logger

AlanCoding:
I was looking for this, because I made the same kind of mistake in another library.

TBH, I think SonarCloud should create a rule to find these cases. `logging.info` goes to the _root_ logger. That is almost never what you want. Any configuration of the `ansible_base` logger will be ignored. I don't think we even customize the root logger in most cases, so the messages would likely be lost.

The python standard library is deceptively permissive of this. Even though these cases are almost certainly a typo/goof, python accepts it as the programmer's intention.


john-westcott-iv (commented):
Very nice catch!
We could likely do our own CI check as well. Something like `grep -r ' logging\.[^(getLogger)]' ansible_base` might get us 90% of the way there.

john-westcott-iv (approved):

PR #646, opened 2024-11-12
Model import fix

Kuppjaerk:
This fixes issue [443](https://github.com/ansible/django-ansible-base/issues/443). 
Imports of AbstractUser are now inline in the functions that use it, also added new import "annotations" in order to handle the functions which had AbstractUser in type hints.

Please let me know if I need to add some additional information or if something needs fixing, this is my first PR here!

### DIFF
--- a/ansible_base/lib/utils/models.py
+++ b/ansible_base/lib/utils/models.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
+
 import logging
 from dataclasses import asdict, dataclass
 from itertools import chain
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 from crum import get_current_user
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


john-westcott-iv:
@Kuppjaerk sorry for the delay in getting back to this. I've rebase the PR onto the latest devel but it looks like there are some linting issues. Would you be able to fix those? If not I can help get this PR over the line.

Kuppjaerk:
@john-westcott-iv Hi John! No worries, thanks for the help! I believe I've fixed the isort and flake8 issues, these were all of the issues right? I tried to see what caused the other checks to fail (django-ansible-base - py39 for instance) but did not quite understand this - were these also due to the aforementioned linting issues?

john-westcott-iv (approved):

john-westcott-iv:
Thanks for this contribution!

PR #647, opened 2024-11-13
Adding missing help_text and check

john-westcott-iv:
This adds a help_text check, validating that every model has all of its help_text defined.

It also addresses missing help_text from existing models.


ansible_base/rbac/models.py
```diff
@@ -30,9 +30,13 @@
 class DABPermission(models.Model):
     "This is a minimal copy of auth.Permission for internal use"
 
-    name = models.CharField("name", max_length=255)
-    content_type = models.ForeignKey(ContentType, models.CASCADE, verbose_name="content type")
-    codename = models.CharField("codename", max_length=100)
+    name = models.CharField("name", max_length=255, help_text=_("The name of this permission."))
```
AlanCoding, line 33: You'll find that in a lot of these, they are not exposed in the API.

This model is also a mirror of `auth.Permission` so why doesn't it have help_text? Because the model it mirrors doesn't have help text.

The permission name is more of a python / admin pages display oriented thing. Hopefully we don't ever use it. If you wanted to get more specific about that, you could. Again, it shouldn't be outward API-facing.

ansible_base/rbac/models.py
```diff
@@ -30,9 +30,13 @@
...
+    content_type = models.ForeignKey(ContentType, models.CASCADE, verbose_name="content type", help_text=_("The content type this permission will apply to."))
```
AlanCoding, line 34: Let's say we have a permission statement "John can pet the dog named Fido". That would be an instantiation of `Permission` (note that a permission entry it not an instantiation), the content type of the permission being dog.

ansible_base/rbac/models.py
john-westcott-iv: Should I take the `help_text` off these and add them to the `.help_text.ignore` file or just leave it on for now?

ansible_base/rbac/models.py
```diff
@@ -30,9 +30,13 @@
...
+    codename = models.CharField(
+        "codename",
+        max_length=100,
+        help_text=_("The name of the permission, giving the action and the model, from the Django Permission model."),
```
AlanCoding: This is the only one that matters, and unfortunately, you can't say the format rule _always_ applies, but this does conform to the existing Django auth app pattern.

So the format of the permission is usually in the form of `{action}_{model_name}`, but can be customized for a specific entry in a model's Meta class. The model name is well-defined formally, and the action is usually / probably the DRF viewset action. This is actually under-specified and in some internal Django uses the actually unique combination of `(app_label, model_name, action)` is used. But in DAB RBAC, it is expected to be unique without app_label, due to additional enforcement.

ansible_base/rbac/models.py
```diff
@@ -330,8 +336,8 @@ class Meta:
         abstract = True
 
     # role_definition set on child models to set appropriate help_text and related_name
-    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
-    object_id = models.TextField(null=False)
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE, help_text=_("The content type this object role applies to."))
+    object_id = models.TextField(null=False, help_text=_("The object id this role applies to."))
```
AlanCoding: You're actually going to be wrong, unexpectedly, here.

This is an abstract model used for several down the chain, notably assignments. This is written for it being on a Role-like model. But that's actually... not, for what matters. The effect we care most about would apply this to _assignment_ models.

These _will_ be user-facing.

Unfortunately, I think that means you need to be intentionally vague.

Say something like

> The content type of the subject of permission assignments. Duplicated from related RoleDefinition.

> The database primary key of the subject of permission assignments.

And we could go on like a lawyer about the definition of "subject". If "John has permission to pet dog Fido", then that looks like "_Actor_ has permission to _action_ (_content-type_ _identifier_) --> (_subject_)". This is effectively a permission "evaluation", but really, everything else in the system is just mechanisms for bulk assigning, which introduce intermediaries by necessity but come back to these primitives.

ansible_base/rbac/models.py
```diff
@@ -375,11 +381,11 @@ class AssignmentBase(ImmutableCommonModel, ObjectRoleFields):
     both models are immutable, making caching easy.
     """
 
-    object_role = models.ForeignKey('dab_rbac.ObjectRole', on_delete=models.CASCADE, editable=False, null=True)
+    object_role = models.ForeignKey('dab_rbac.ObjectRole', on_delete=models.CASCADE, editable=False, null=True, help_text=_("The object role this applies to."))
```
AlanCoding: Not API-facing. I would say

> A roll-up of the fields (role_definition, content_type)

ansible_base/rbac/models.py
AlanCoding: Writing more comments further down, I realize "subject" is a hazardous word. "object" might be better. English SVO grammar structure would imply "subject" is what is referred to as "actor" elsewhere.

ansible_base/rbac/models.py
```diff
@@ -430,7 +438,9 @@ class RoleTeamAssignment(AssignmentBase):
         help_text=_("The role definition which defines permissions conveyed by this assignment"),
         related_name='team_assignments',
     )
-    team = models.ForeignKey(settings.ANSIBLE_BASE_TEAM_MODEL, on_delete=models.CASCADE, related_name='role_assignments')
+    team = models.ForeignKey(
+        settings.ANSIBLE_BASE_TEAM_MODEL, on_delete=models.CASCADE, related_name='role_assignments', help_text=_("The team this role will be applied to.")
```
AlanCoding: Careful here, because a team can be both an actor and an object (can get permission, or have permission to it given to another actor).

This field is a team-as-actor field.

So this is the team that receives the permissions of the related role definition, to the related object. You can cut that down to

> The team that receives permissions

ansible_base/rbac/models.py
```diff
@@ -705,7 +715,7 @@ class Meta(RoleEvaluationMeta):
     role = models.ForeignKey(
         ObjectRole, null=False, on_delete=models.CASCADE, related_name='permission_partials', help_text=_("The object role that grants this form of permission")
     )
-    object_id = models.PositiveIntegerField(null=False)
+    object_id = models.PositiveIntegerField(null=False, help_text=_("The object ID this role evaluation will be applied to."))
```
AlanCoding: Grammar seems a little backwards. This is a statement of role-object-permission.

> The id of the object that the related role gives the related permission to

ansible_base/resource_registry/models/resource.py
```diff
@@ -40,28 +42,28 @@ def get_resource_config(self):
...
         null=False,
         default=service_id,
-        help_text="ID of the service responsible for managing this resource.",
+        help_text=_("ID of the service responsible for managing this resource."),
     )
 
     # we're not using this as the primary key because the ansible_id can change if the object is
     # externally managed.
-    ansible_id = models.UUIDField(default=uuid.uuid4, db_index=True, unique=True)
+    ansible_id = models.UUIDField(default=uuid.uuid4, db_index=True, unique=True, help_text=_("A unique ID identifying this resource in AAP."))
```
AlanCoding: in AAP --> by the resource server

ansible_base/resource_registry/models/resource.py
```diff
@@ -40,28 +42,28 @@ def get_resource_config(self):
...
 
     # human readable name for the resource
-    name = models.CharField(max_length=512, null=True)
+    name = models.CharField(max_length=512, null=True, help_text=_("The name of this resource."))
```
AlanCoding, line 62: It might not be the name...

Remind me again why we have this? Oof, ask @newswangerd

ansible_base/rbac/models.py
AlanCoding: The permission model, itself, maybe one day (soon-ish) be surfaced in the API in a read-only way. So it might be good to have these, although I realize I probably just have to make a PR.

ansible_base/activitystream/models/entry.py
```diff
@@ -32,17 +32,26 @@ class Meta:
...
-    related_object_id = models.TextField(null=True, blank=True)
+    related_content_type = models.ForeignKey(
+        ContentType,
+        on_delete=models.DO_NOTHING,
+        null=True,
+        blank=True,
+        related_name='related_content_type',
+        help_text=_("The content type if this entry is recording an m2m change."),
+    )
+    related_object_id = models.TextField(null=True, blank=True, help_text=_("The object id of the related model if this entry is for an m2m change."))
```
BrennanPaciorek: I think the phrasing "m2m change", abbreviating Many-to-Many, is only apparent when viewing the help text in code.

ansible_base/activitystream/models/entry.py
```diff
@@ -32,17 +32,26 @@ class Meta:
...
     related_content_object = GenericForeignKey('related_content_type', 'related_object_id')
-    related_field_name = models.CharField(max_length=64, null=True, blank=True)
+    related_field_name = models.CharField(max_length=64, null=True, blank=True, help_text=_("The related field name if this entry is for an m2m relationship."))
```
BrennanPaciorek: The phrasing "m2m relationship" probably leaves more breadcrumbs for users (as in a search engine will probably help them out a little more), so I'm alright with this one.

ansible_base/rbac/models.py
BrennanPaciorek: I don't see much purpose in dropping the help_text and making an ignore here, so I think we should keep this one, save us an extra migration if or when this gets exposed to the API.

BrennanPaciorek (changes requested):
More migrations appear to be necessary.
```
check: exit 1 (1.02 seconds) /home/runner/work/django-ansible-base/django-ansible-base> python3 manage.py makemigrations --check
```

The other comments are just recommendations and lightly-held opinions.

ansible_base/activitystream/models/entry.py
john-westcott-iv: I think we are supposed to be spelling out acronyms and then adding it in `()`. So I'll change this one too.

ansible_base/rbac/models.py
john-westcott-iv: How does this sound?
```
The name of the permission. This is in the format {action}_{model_name}. Where action is typically the view set action (view/list/etc) from Django rest framework. 
```

BrennanPaciorek (approved):

ansible_base/rbac/models.py
tznamena: I don't think we would want to call this `The name of the permission....` because that is the text used for `name` on line 33.
As to the comment about the format, we should only include the format if the model class itself sets format.

ansible_base/resource_registry/models/resource.py
```diff
@@ -40,28 +42,28 @@ def get_resource_config(self):
...
+    ansible_id = models.UUIDField(default=uuid.uuid4, db_index=True, unique=True, help_text=_("A unique ID identifying this resource by the resource server."))
 
     # human readable name for the resource
-    name = models.CharField(max_length=512, null=True)
+    name = models.CharField(max_length=512, null=True, help_text=_("The name of this resource."))
 
     is_partially_migrated = models.BooleanField(
         default=False,
-        help_text="This gets set to True when a resource has been copied into the resource server, but the service_id hasn't been updated yet.",
+        help_text=_("This gets set to True when a resource has been copied into the resource server, but the service_id hasn't been updated yet."),
```
tznamena: I would suggest to avoid starting help text with `This ...`. 

Instead, let's just call it what it is:
`Flag indicating that  the resource has been copied into resource server` 
or 
`Indicator of ...`

ansible_base/rbac/models.py
john-westcott-iv: @tznamena Good callout

ansible_base/resource_registry/models/resource.py
newswangerd: @AlanCoding to make it human readable/easier to search. You're right, it can be anything identifiable. It's the username for users.

PR #648, opened 2024-11-13
Fix: Implement LDAP specific authenticator_map group check

BrennanPaciorek:
This alleiviates case-sensitivity problems when processing groups

Other change:
re-factor: move logic for evaluating group claims into the authenticator plugin

Current logic moved to the BaseAuthenticatorPlugin, but this allows other
plugins to override these check with their own logic for evaluating these
strings.

Preliminary change for introducing a comparison class to the
LDAPAuthenticatorPlugin.

AAP-33304


BrennanPaciorek:
Closing in favor of a documentation change #649

PR #650, opened 2024-11-14
[WIP] Add feature_flags app

john-westcott-iv:


john-westcott-iv:
This was just an experiment, closing.

AlanCoding:
This seems sketchy since it appears to be duplicated effort with https://github.com/ansible/awx/pull/15660, which is using an external library. I'm not sure if that was a good idea, but by the time I was involved it seems like everything had already been decided.

PR #651, opened 2024-11-15
Add example svc to service_type

TheRealHaoLiu:
https://github.com/ansible/platform-service-template/ establish a minimal example svc that can work with the platform and we want to add the example as a service_type


AlanCoding (approved):

TheRealHaoLiu:
@john-westcott-iv another place that will need to loosen if we want service to be configurable at runtime

PR #652, opened 2024-11-20
fix(lib.validation): properly validate ca-signed keypairs

BrennanPaciorek:
Change the mechanism we use to validate keypairs to not assume that a 
certificate is self-signed. This causes some valid keypairs to fail.

Testing steps:
- tox -e py311 -- -n0 -k test_validate_cert_with_signed_certificate
- verify passing
- drop commit "fix(lib.validation): Do not treat ca-signed keypairs as invalid"
  - Save the commit digest from most recent commit before doing this
  - Or branch before doing this, so you can just switch back to the actual PR branch
- tox -e py311 -- -n0 -k test_validate_cert_with_signed_certificate
- verify failing test
- Cleanup: git reset --hard ${remote-branch or commit you reset from)

AAP-33294

### DIFF
--- a/ansible_base/lib/testing/fixtures.py
+++ b/ansible_base/lib/testing/fixtures.py
@@ -235,6 +235,72 @@ def rsa_keypair(rsa_keypair_factory):
 
 @copy_fixture(copies=3)
 @pytest.fixture
+def rsa_keypair_with_signed_cert(rsa_keypair_factory):
+    root_keypair = rsa_keypair_factory()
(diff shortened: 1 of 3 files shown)


BrennanPaciorek:
Sonar cloud output seems bugged, maybe it expects test to be committed after? If it fails again, I'll reorder the commits.

john-westcott-iv:
We might want to exclude lib/testsing/fixtures.py from the sonar analysis. This can be done in= `sonar-project.properties`. @relrod, any thoughts on this?

john-westcott-iv (approved):

PR #657, opened 2024-11-26
Revert "📦 Implement tokenless publishing to PyPI"

AlanCoding:
Reverts ansible/django-ansible-base#598

The release job has not completed since we merged this:

https://github.com/ansible/django-ansible-base/actions/workflows/release.yml

no releases have been published

https://pypi.org/project/django-ansible-base/

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,9 +9,12 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
+  stage:
     runs-on: ubuntu-latest
-    timeout-minutes: 2
+    timeout-minutes: 90
+    permissions:
+      packages: write
+      contents: write
     steps:
       - name: Checkout dab
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


john-westcott-iv:
@webknjaz It looks like PyPi does not like the version we are pushing up:

> ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         The use of local versions in <Version('24.11.22.1.dev1+g7c71771')> is  
         not allowed. See                                                       
         https://packaging.python.org/specifications/core-metadata for more     
         information.       

Is this something you might want to try and look at before we revert?

webknjaz:
@john-westcott-iv yes, PEP 440 explicitly instructs the indexes to reject local versions. I however don't remember changing the build process. So I wouldn't expect that reverting the publishing change would help you. I may take a look, though.

webknjaz:
#663 should do it. Perhaps, ping me when I'm back, next year.

PR #660, opened 2024-12-02
Migrate AWX utility for slow method logging to DAB

AlanCoding:
Followup from https://github.com/ansible/django-ansible-base/pull/356

The intent is that AWX will delete this method and import from DAB.


chrismeyersfsu:
Is this the equivalent in awx  https://github.com/ansible/awx/blob/adc2162bac1c9832e8df39df8dc0a4c04126cc78/awx/main/utils/common.py#L1208 ?

AlanCoding:
Yes, https://github.com/ansible/awx/pull/15675 would remove it from there and import from DAB

john-westcott-iv (approved):

PR #661, opened 2024-12-02
Add advisory_lock utility from AWX

AlanCoding:
This migrates `advisory_lock` from AWX to DAB. This is similar in spirit to https://github.com/ansible/django-ansible-base/pull/660

Importantly, I went the route of _not_ adding a new dependency, because there was only a single method from the library we were using, and I'm pretty confident that the MIT license lets you do whatever you want. In doing this, I deleted the use of `six` for python2 compatibility which further reduced the complexity of the code I ported.

Additionally requested reviewers @Alex-Izquierdo @TheRealHaoLiu

### DIFF
--- a/ansible_base/lib/utils/db.py
+++ b/ansible_base/lib/utils/db.py
@@ -1,6 +1,7 @@
 from contextlib import contextmanager
+from zlib import crc32
 
-from django.db import connection, transaction
+from django.db import DEFAULT_DB_ALIAS, connection, connections, transaction
 from django.db.migrations.executor import MigrationExecutor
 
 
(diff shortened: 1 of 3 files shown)


test_app/tests/lib/utils/test_db.py
```diff
@@ -1,9 +1,65 @@
...
+            with advisory_lock(['test_invalid_tuple_name', 'foo']):
+                pass
+
+    @pytest.mark.django_db
+    def test_lock_session_timeout_milliseconds(self):
+        with pytest.raises(OperationalError) as exc:
+            # uses miliseconds units
+            with advisory_lock('test_lock_session_timeout_milliseconds', lock_session_timeout_milliseconds=2):
+                time.sleep(3)
+        assert 'the connection is lost' in str(exc)
```
AlanCoding, line 65: @TheRealHaoLiu I was really uncomfortable with this `lock_session_timeout_milliseconds` when it went into AWX. Historically, we have accumulated untestable code, and then gotten issues reported due to some esoteric circumstance, and gotten into a pickle about how or if we will address it.

Here, this is offering true functional testing. It sets the timeout for the current session and actually watches it time out.

I would like to convert AWX to run checks with postgres, but now is not the time for that. For this overused / abused utility in AWX, holding it to the higher code quality standard in DAB is a good idea, particularly since I'm basically saying we should reuse it in eda-server.

ansible_base/lib/utils/db.py
```diff
@@ -25,3 +26,114 @@ def migrations_are_complete() -> bool:
...
+                cur.execute("SET idle_in_transaction_session_timeout = %s", (lock_session_timeout_milliseconds,))
+                cur.execute("SET idle_session_timeout = %s", (lock_session_timeout_milliseconds,))
+        with django_pglocks_advisory_lock(*args, **kwargs) as internal_lock:
+            yield internal_lock
+            if lock_session_timeout_milliseconds > 0:
+                with connection.cursor() as cur:
+                    cur.execute("SET idle_in_transaction_session_timeout = %s", (idle_in_transaction_session_timeout,))
+                    cur.execute("SET idle_session_timeout = %s", (idle_session_timeout,))
+    else:
+        yield True
```
john-westcott-iv: Can we make this an:
```
elif connection.vendor == 'sqllite':
    yield True
else:
   rise Exception("Advisory locks are not implemented for your DB vendor")
```

john-westcott-iv (commented):
It might be nice to have vendor specific files in ansible_base/lib/utils/db i.e. something for postgres/mssql/sqlite and then the `db.py` would have "Generic methods" which would check the vendor and then call items from the DB specific files.

ansible_base/lib/utils/db.py
```diff
@@ -25,3 +26,114 @@ def migrations_are_complete() -> bool:
...
+            release_params = (release_function_name,) + params
+
+            command = base % release_params
+            cursor.execute(command)
+
+        cursor.close()
+
+
+@contextmanager
+def advisory_lock(*args, lock_session_timeout_milliseconds=0, **kwargs):
```
tznamena, line 121: Can we add some comments here to explain the logic?

AlanCoding:
> It might be nice to have vendor specific files in ansible_base/lib/utils/db i.e. something for postgres/mssql/sqlite

Maybe sometime, I just don't feel like we're really there yet. AWX has probably the most postgres-specific code, and it's just lying around scatter-shot.

It might be nice to attempt sqlite solution, probably like this:

https://stackoverflow.com/questions/6931342/system-wide-mutex-in-python-on-linux

```python
class Locker:
    def __enter__ (self):
        self.fp = open("./lockfile.lck")
        fcntl.flock(self.fp.fileno(), fcntl.LOCK_EX)

    def __exit__ (self, _type, value, tb):
        fcntl.flock(self.fp.fileno(), fcntl.LOCK_UN)
        self.fp.close()
```

Then that would share a lot of feature and not share a lot of features. So then you have a lot of param handling logic to error if it's 1 vendor and not another. Definitely going in the opposite direction of django-pglocks where this comes from. And I do think that would be good for when we introduce a sqlite version, but that just seems too arbitrary and premature until we're at that point.

ansible_base/lib/utils/db.py
```diff
@@ -25,3 +26,126 @@ def migrations_are_complete() -> bool:
...
+
+# NOTE: the django_pglocks_advisory_lock context manager was forked from the django-pglocks v1.0.4
+# that was licensed under the MIT license
+
+
+@contextmanager
+def django_pglocks_advisory_lock(lock_id, shared=False, wait=True, using=None):
+
+    if using is None:
+        using = DEFAULT_DB_ALIAS
```
PabloHiro, line 39: Why the if block here instead of making `DEFAULT_DB_ALIAS` the default in the function?

ansible_base/lib/utils/db.py
```diff
@@ -25,3 +26,126 @@ def migrations_are_complete() -> bool:
...
+
+        tuple_format = True
+    elif isinstance(lock_id, str):
+        # Generates an id within postgres integer range (-2^31 to 2^31 - 1).
+        # crc32 generates an unsigned integer in Py3, we convert it into
+        # a signed integer using 2's complement (this is a noop in Py2)
+        pos = crc32(lock_id.encode("utf-8"))
+        lock_id = (2**31 - 1) & pos
+        if pos & 2**31:
+            lock_id -= 2**31
```
PabloHiro, line 82: I would have put the content of this if block into a separate function, so it can be tested separately and maybe be reused?

ansible_base/lib/utils/db.py
AlanCoding: This comes from the library:

https://github.com/Xof/django-pglocks/blob/master/django_pglocks/__init__.py#L10

And the reason is clearly because of a hesitancy to import from `django.db` at the top level of the file. For many imports this is a valid concern, due to the "apps are not loaded yet" error. But DAB imports from paths in Django in many places in a way that relies on those specific paths not requiring app setup in advance. So I agree, we could move the import to the top of the file and make that the default.

ansible_base/lib/utils/db.py
AlanCoding: The library seems to have preferred full functional tests...

https://github.com/Xof/django-pglocks/blob/master/django_pglocks/tests.py#L41-L46

ansible_base/lib/utils/db.py
chrismeyersfsu: If you are feeling frisky you can use `get_connection()` and dodge reaching for a possibly not stable constant. But that doesn't solve the underlying import hairyness, it also deviates from the copy paste.
 
https://github.com/django/django/blob/3ee4c6a27ad520d4ecc3cded260d47cbccafb144/django/db/transaction.py#L18

ansible_base/lib/utils/db.py
```diff
@@ -25,3 +26,126 @@ def migrations_are_complete() -> bool:
...
+            lock_id[1],
+        )
+    else:
+        base = "SELECT %s(%d)"
+        params = (lock_id,)
+
+    acquire_params = (function_name,) + params
+
+    command = base % acquire_params
+    cursor = connections[using].cursor()
```
chrismeyersfsu, line 99: could replace with `get_connection(using=using)`. I don't think it's worth it.

chrismeyersfsu (approved):

chrismeyersfsu:
I'll mention a real-world issue I've had.

We had a lock that was held and not let go. Which lock? That was hard to find out. It was easy enough to surface the set of held locks from Postgres. The hard part was taking the integer (`lock_id` in this code I think) --> lock name as a string.

ansible_base/lib/utils/db.py
AlanCoding: It's not quite the same, as `get_connection` returns the wrapper for the connection, but `connection` returns a proxy for the wrapper.

```
In [6]: connection
Out[6]: <django.utils.connection.ConnectionProxy at 0x7fd28ddc4530>

In [7]: get_connection()
Out[7]: <DatabaseWrapper vendor='sqlite' alias='default'>
```

But if you look at the code for `ConnectionProxy`, its life goal is to pretend to be the thing it wraps (the `DatabaseWrapper`), while still allowing configuration changes to change out the specific `DatabaseWrapper` to change out from under it.

So end then end, they are different, but as long as you are only referencing its properties (like `connection.cursor`), it shouldn't matter. It is designed to not matter.

AlanCoding: Filed

https://github.com/ansible/django-ansible-base/issues/679

I do generally agree, but we'd need to approach this incrementally.

PR #664, opened 2024-12-04
AAP-37017 Fix how we build AuthExceptions

john-westcott-iv:
The `AuthException` class from `social_core.exceptions` takes a first parameter of a backend and a second of a message. Since we only had one parameter when we constructed our AuthException, printing it may be black:

```
In [6]: from social_core.exceptions import AuthException
In [7]: try:
   ...:     raise AuthException(None, 'Testing')
   ...: except AuthException as e:
   ...:     print(f'AuthException: {e}')
   ...: 
AuthException: Testing

In [8]: try:
   ...:     raise AuthException('Testing')
   ...: except AuthException as e:
   ...:     print(f'AuthException: {e}')
   ...: 
AuthException: 
```

This PR fixes this by passing in `None` as the first backend parameter to the `AuthException`.

### DIFF
--- a/ansible_base/authentication/utils/authentication.py
+++ b/ansible_base/authentication/utils/authentication.py
@@ -21,6 +21,10 @@ class FakeBackend:
         return ["username", "email"]
 
 
+def raise_auth_exception(message: str, backend: Optional[Authenticator] = None):
+    raise AuthException(backend, message)
+
+
 def migrate_from_existing_authenticator(
     uid: str, alt_uid: Optional[str], authenticator: Authenticator, preferred_username: Optional[str] = None
 ) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):

john-westcott-iv:
This PR needs a rebase and some tests added around this.

ansible_base/authentication/utils/authentication.py
```diff
@@ -21,6 +21,10 @@ def setting(self, *args, **kwargs):
         return ["username", "email"]
 
 
+def raise_auth_exception(message):
+    raise AuthException(None, message)
```
BrennanPaciorek: Any reason for why we aren't making any best-effort attempts at fetching the authentication backend?

Iirc we can get the backend from the pipeline details dictionary.

BrennanPaciorek (commented):
Maybe we should make a best effort attempt at getting the authenticator for the AuthException

ansible_base/authentication/utils/authentication.py
john-westcott-iv: I'm not sure having it makes a ton of difference but I pushed a change to try and get that.

BrennanPaciorek (approved):

Issue #211, opened 2024-03-12
Convert to CommandError


john-westcott-iv:
### Bug Summary

In ansible_base/authentication/management/commands/authenticators.py we call system.exit if there is no admin user defined. Instead we should raise a CommandError and let django deal with it accordingly. See comments in https://github.com/ansible/django-ansible-base/pull/209.

Kuppjaerk:
Hi, I just submitted a PR for this issue. Please let me know if anything else needs to be done apart from this simple fix!

PR #666, opened 2024-12-05
Convert to CommandError

Kuppjaerk:
This fixes issue [211](https://github.com/ansible/django-ansible-base/issues/211).

Replaced sys.exit() call with a CommandError raise, and removed sys import since unused.


AlanCoding:
Link another issue this addresses https://github.com/ansible/django-ansible-base/issues/443

AlanCoding (changes requested):
We looked through this in PR review, and had consensus that the patches should be split up for different issues.

Also, I would prefer that we type hint `Model` from `django.db` as opposed to `AbstractUser`, because that'll avoid the messiness here. But those comments are a good reason to have this split into 2 patches.

Kuppjaerk:
> We looked through this in PR review, and had consensus that the patches should be split up for different issues.
> 
> Also, I would prefer that we type hint `Model` from `django.db` as opposed to `AbstractUser`, because that'll avoid the messiness here. But those comments are a good reason to have this split into 2 patches.

It seems the changes I made in my previous PR which solved [443](https://github.com/ansible/django-ansible-base/issues/443) snuck in here as well, my bad! This PR now only contain the changes relevant to the issue at hand.

AlanCoding (approved):

PR #667, opened 2024-12-05
update dependencies

jctanner:
https://issues.redhat.com/browse/AAH-3262

### DIFF
--- a/requirements/requirements_all.txt
+++ b/requirements/requirements_all.txt
@@ -1,20 +1,21 @@
 asgiref==3.8.1
     # via
+    #   -r requirements/requirements_resource_registry.in
     #   channels
     #   django
-attrs==23.2.0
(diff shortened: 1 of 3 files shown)


AlanCoding (approved):

requirements/requirements_all.txt
```diff
@@ -1,20 +1,21 @@
 asgiref==3.8.1
     # via
+    #   -r requirements/requirements_resource_registry.in
```
AlanCoding, line 3: These added comments are correct.

tznamena:
This now breaks gateway, we should make the dependency upgrades in more organized fashion. In my previous attempts to upgrade dependencies I was told that we should start at tower repo and only then upgrade DAB and GW. Maybe this has changed since?

PR #672, opened 2024-12-09
fix(authenticator.utils): get ID_KEY from authenticator configuration, not defaults

BrennanPaciorek:
- **tests(authentication): fix UID_KEY tests for authentication.determine_username_from_uid_social, introduce failing test**
- **fix(authenticator.utils): get ID_KEY from authenticator configuration, not defaults**

AAP-37203

### DIFF
--- a/ansible_base/authentication/utils/authentication.py
+++ b/ansible_base/authentication/utils/authentication.py
@@ -101,7 +101,7 @@ def determine_username_from_uid_social(**kwargs) -> dict:
     selected_username = kwargs.get('details', {}).get(uid_field, None)
     if not selected_username:
         raise AuthException(
-            _('Unable to get associated username from attribute {uid_field}: %(details)s') % {'uid_field': uid_field, 'details': kwargs.get("details", None)}
+            _('Unable to get associated username from attribute %(uid_field)s: %(details)s') % {'uid_field': uid_field, 'details': kwargs.get("details", None)}
         )
 
     authenticator = kwargs.get('backend')
--- a/ansible_base/authentication/utils/authentication.py
+++ b/ansible_base/authentication/utils/authentication.py
@@ -98,8 +98,8 @@ def determine_username_from_uid_social(**kwargs) -> dict:
     backend = kwargs.get('backend', None)
     uid_field = 'username'
     if backend:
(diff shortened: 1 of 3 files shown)


ansible_base/authentication/utils/authentication.py
```diff
@@ -95,13 +95,14 @@ def check_system_username(uid: str) -> None:
 
 
 def determine_username_from_uid_social(**kwargs) -> dict:
-    uid_field = getattr(kwargs.get('backend', None), 'ID_KEY', 'username')
-    if uid_field is None:
-        uid_field = 'username'
+    backend = kwargs.get('backend', None)
+    uid_field = 'username'
+    if backend and hasattr(backend, "setting"):
+        uid_field = backend.setting('ID_KEY', default='username')
```
john-westcott-iv: Is there a case where `ID_KEY` might be directly in backend that we need to cover as well?

john-westcott-iv: For example: https://github.com/python-social-auth/social-core/blob/master/social_core/backends/battlenet.py

BrennanPaciorek: You are correct. Updating this now with more favorable logic. This one setting fetch seems to have a ton of cases to consider.

ansible_base/authentication/utils/authentication.py
```diff
@@ -95,13 +95,21 @@ def check_system_username(uid: str) -> None:
...
 
 def determine_username_from_uid_social(**kwargs) -> dict:
-    uid_field = getattr(kwargs.get('backend', None), 'ID_KEY', 'username')
-    if uid_field is None:
-        uid_field = 'username'
+    backend = kwargs.get('backend', None)
+    uid_field = 'username'
+    if backend:
+        # Update our fallback if the authenticator has an ID_KEY field
+        if hasattr(backend, "ID_KEY"):
```
BrennanPaciorek:
```suggestion
        # Update our fallback if the authenticator has an ID_KEY field, ignore if ID_KEY is None
        if getattr(backend, "ID_KEY", None) is not None:
```
This is probably something we want to do.

john-westcott-iv (approved):

PR #676, opened 2024-12-11
[POC] Manage static settings with dynaconf

rochacbruno:
P.O.C - Manage settings with Dynaconf


ansible_base/lib/dynamic_config/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+    caller_path = os.path.dirname(inspect.getfile(frame))
+
+    settings = Dynaconf(
+        env_switcher=f"{name}_MODE",
+        envvar=f"{name}_SETTINGS",
+        envvar_prefix=name,
+        root_path=caller_path,
+        **options
+    )
+    # settings.update(get_dab_settings(...))
```
AlanCoding: yes... did this not work? Otherwise I generally follow this.

What about precedence? You're loading the default dab settings after the other stuff, but I expect that env vars take higher precedence than these file defaults.

rochacbruno: this works, I commented out because I was debugging.

The envvars will have higher precedence because the order will be defined on each component.

PR #682, opened 2025-01-07
AAP-41636 Support extracting the groups claim from `id_token` when using social auth

### LINKED ISSUE #681: Unable to extract groups claim when using Social Auth
### Bug Summary

Currently when using social auth, we can only successfully extract the groups claim if it is included in the `UserInfo` endpoint. 

Some IDPs do not support this (EntraID) and only include the groups claim in the `id_token`.

This means that group mapping does not work properly in these cases.

markafarrell:
# Support extracting the groups claim from `id_token` when using social auth

## Description
Currently when using social auth we can only extract the groups claim if it it included in the UserInfo endpoint. 
Some IDPs do not support this (EntraID) and instead only include the groups claim in the `id_token`.

This add the capability to first look for the groups claim in the information returned by the UserInfo endpoint and if it is not found to then attempt to extract it from the `id_token`.

Fixes #681 

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. 
2. 
3. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/authentication/social_auth.py
+++ b/ansible_base/authentication/social_auth.py
@@ -195,6 +195,18 @@ def create_user_claims_pipeline(*args, backend, response, **kwargs):
     groups_claim = backend.groups_claim if backend.groups_claim is not None else "Group"
 
     extra_groups = response[groups_claim] if groups_claim in response else []
+    logger.debug(f'groups extracted from UserInfo. "{groups_claim}": {extra_groups}')
+
+    # attempt to get group claims from id token if current backend has it
+    if hasattr(backend, "id_token"):
+        extra_groups_from_id_token = backend.id_token.get(groups_claim, [])
+        logger.debug(f'groups extracted from id_token. "{groups_claim}": {extra_groups_from_id_token}')
+
+        # Merge groups from UserInfo and groups_claim in id_token and remove duplicates
+        extra_groups = list(set(extra_groups + extra_groups_from_id_token))
+
+    logger.debug(f'updating user claims with groups claim "{groups_claim}": {extra_groups}')
+
     user = update_user_claims(kwargs["user"], backend.database_instance, backend.get_user_groups(extra_groups))
     if user is None:
         return SOCIAL_AUTH_PIPELINE_FAILED_STATUS
--- a/test_app/tests/authentication/test_social_auth.py
+++ b/test_app/tests/authentication/test_social_auth.py
@@ -89,17 +89,49 @@ def test_social_auth_validate_callback_mixin(mocked_generate_slug, mocked_revers
         assert mocked_reverse.called
 
 
+id_token_no_groups = {
+    "ver": "2.0",
+    "iss": "https://login.microsoftonline.com/9122040d-6c67-4c5b-b112-36a304b66dad/v2.0",
+    "sub": "AAAAAAAAAAAAAAAAAAAAAIkzqFVrSaSaFHy782bbtaQ",
+    "aud": "6cb04018-a3f5-46a7-b995-940c78f5aef3",
+    "exp": 4073899721,
+    "iat": 1536274711,
+    "nbf": 1536274711,
+    "name": "Abe Lincoln",
+    "preferred_username": "AbeLi@microsoft.com",
+    "email": "AbeLi@microsoft.com",
+    "oid": "00000000-0000-0000-66f3-3332eca7ea81",
+    "tid": "9122040d-6c67-4c5b-b112-36a304b66dad",
+    "nonce": "123523",
+    "aio": "Df2UVXL1ix!lMCWMSOJBcFatzcGfvFGhjKv8q5g0x732dR5MB5BisvGQO7YWByjd8iQDLq!eGbIDakyp5mnOrcdqHeYSnltepQmRp6AIZ8jY",
+}
+
+id_token = {**id_token_no_groups, "groups": ["myidtokengroup"]}
+
+id_token_duplicate_group = {**id_token_no_groups, "groups": ["mygroup", "myidtokengroup"]}
+
+
 @pytest.mark.parametrize(
-    "groups_claim,returned_groups,expected_groups",
+    "groups_claim,user_info_groups,id_token,expected_groups",
     [
-        (None, ["mygroup"], ["mygroup"]),
-        ("groups", ["mygroup"], ["mygroup"]),
-        (None, None, []),
-        ("groups", None, []),
+        (None, ["mygroup"], {}, ["mygroup"]),
+        ("groups", ["mygroup"], {}, ["mygroup"]),
+        (None, None, {}, []),
+        ("groups", None, {}, []),
+        # Check extracting groups claim from id_token
+        ("groups", None, id_token, ["myidtokengroup"]),
+        # Test extracting groups claim from id_token when groups claim does not exist
+        (None, None, id_token, []),
+        # Test merging groups from UserInfo and id_token.
+        ("groups", ["mygroup"], id_token, ["myidtokengroup", "mygroup"]),
+        # Test merging groups from UserInfo and id_token where we have duplicate groups.
+        ("groups", ["mygroup"], id_token_duplicate_group, ["myidtokengroup", "mygroup"]),
+        # Test where id_token has no groups-claim.
+        ("groups", ["mygroup"], id_token_no_groups, ["mygroup"]),
     ],
 )
 @mock.patch("ansible_base.authentication.utils.claims.update_user_claims")
-def test_create_user_claims_pipeline(mock_update_user_claims, groups_claim, returned_groups, expected_groups):
+def test_create_user_claims_pipeline(mock_update_user_claims, groups_claim, user_info_groups, id_token, expected_groups):
     '''
     We are testing to see if extracting groups from a claim is working correctly
     '''
@@ -107,22 +139,24 @@ def test_create_user_claims_pipeline(mock_update_user_claims, groups_claim, retu
     class MockBackend(SocialAuthMixin):
         database_instance = None
 
-        def __init__(self, groups_claim=None):
+        def __init__(self, groups_claim=None, id_token=None):
             if groups_claim is not None:
                 self.groups_claim = groups_claim
+            if id_token is not None:
+                self.id_token = id_token
 
         def get_user_groups(self, extra_groups=[]):
             return extra_groups
 
-    backend = MockBackend(groups_claim=groups_claim)
+    backend = MockBackend(groups_claim=groups_claim, id_token=id_token)
 
     rData = {}
-    if returned_groups is not None:
-        rData[backend.groups_claim] = returned_groups
+    if user_info_groups is not None:
+        rData[backend.groups_claim] = user_info_groups
 
     user = {
         'auth_time': "2024-11-07T05:19:08.224936Z",
-        'id_token': "asdf",
+        'id_token': id_token,
         'refresh_token': None,
         'id': "ccd2cf13-d927-41ad-cd8c-adb18b2e5f78",
         'access_token': "asdf",
@@ -134,4 +168,6 @@ def test_create_user_claims_pipeline(mock_update_user_claims, groups_claim, retu
     assert mock_update_user_claims.called
     call_args = mock_update_user_claims.call_args
 
-    assert call_args == ((user, None, expected_groups),)
+    assert call_args[0][0] == user
+    assert call_args[0][1] is None
+    assert call_args[0][2].sort() == expected_groups.sort()


ansible_base/authentication/social_auth.py
```diff
@@ -193,7 +194,21 @@ def create_user_claims_pipeline(*args, backend, response, **kwargs):
 
     groups_claim = backend.groups_claim if backend.groups_claim is not None else "Group"
 
-    extra_groups = response[groups_claim] if groups_claim in response else []
+    try:
+        id_token_claims = jwt.decode(
+            response['id_token'], options={"verify_signature": False})
```
john-westcott-iv: Should we  make the `verify_signature` a setting so that it could be True or False depending on someones needs?

ansible_base/authentication/social_auth.py
```diff
@@ -193,7 +194,21 @@ def create_user_claims_pipeline(*args, backend, response, **kwargs):
...
+    except jwt.exceptions.InvalidTokenError:
+        id_token_claims = {}
```
jshimkus-rh: Performing this decode only if UserInfo response does not include the claims is preferable.

Performing it before allows an unexpected exception to abort the entire process even in the case that UserInfo provides the claims.

ansible_base/authentication/social_auth.py
markafarrell: This should be done earlier in the authentication pipeline so its probably not strictly necessary.

However, its probably not a bad idea to verify the signature anyway.

ansible_base/authentication/social_auth.py
markafarrell: Agreed.

Its probably also catching a bare `Exception` as well so we don't abort the process for an unexpected exception

ansible_base/authentication/social_auth.py
markafarrell: I looked into this and its going to be a bit complicated to do the signature verification.

I'm also not sure its really necessary. 

The `id_token` is verified earlier in the pipeline.

[`BaseOAuth2.auth_complete`](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/oauth.py#L424) ->[`OpenIdConnectAuth.request_access_token`](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/open_id_connect.py#L253) ->[`OpenIdConnectAuth.validate_and_return_id_token`](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/open_id_connect.py#L218)

so if we get to this point in the pipeline we can assume that the signature is valid.

ansible_base/authentication/social_auth.py
markafarrell: I made changes to catch the base pyjwt exception `PyJWTError` and also move the decoding logic to only occur if we don't locate the claim in `UserInfo`

john-westcott-iv:
@markafarrell, thanks again for your work on this! We've reviewed your updates and run the unit tests but they are failing. Can you address the issues in there? Let us know if you need any assistance in doing this.

tznamena (changes requested):
This change also needs corresponding test. Such test can go into: `test_app/tests/authentication/test_social_auth.py` file.

ansible_base/authentication/social_auth.py
jshimkus-rh: Looks good.

markafarrell:
> This change also needs corresponding test. Such test can go into: `test_app/tests/authentication/test_social_auth.py` file.

Tests added

markafarrell:
> @markafarrell, thanks again for your work on this! We've reviewed your updates and run the unit tests but they are failing. Can you address the issues in there? Let us know if you need any assistance in doing this.

Fixed changes do handle when `id_token` is missing

markafarrell:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=ansible_django-ansible-base&pullRequest=682) **Quality Gate failed**
> 
> Failed conditions  [1 New issue](https://sonarcloud.io/project/issues?id=ansible_django-ansible-base&pullRequest=682&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [D Security Rating on New Code](https://sonarcloud.io/dashboard?id=ansible_django-ansible-base&pullRequest=682) (required ≥ A)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=ansible_django-ansible-base&pullRequest=682)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

@john-westcott-iv is this an issue? Or can we just ignore as we are doing signature verification earlier in the pipeline

markafarrell:
> > ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=ansible_django-ansible-base&pullRequest=682) **Quality Gate failed**
> > Failed conditions  [1 New issue](https://sonarcloud.io/project/issues?id=ansible_django-ansible-base&pullRequest=682&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [D Security Rating on New Code](https://sonarcloud.io/dashboard?id=ansible_django-ansible-base&pullRequest=682) (required ≥ A)
> > [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=ansible_django-ansible-base&pullRequest=682)
> >  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)
> 
> @john-westcott-iv is this an issue? Or can we just ignore as we are doing signature verification earlier in the pipeline

I added signature verification in ea4310b597bd6d1500528b303c3945b69af7cb06 we can revert this if we think its better to do without verification

john-westcott-iv:
@markafarrell Sorry for the late response. We recently made a change to the repo where we need to address issues identified by Sonar before we can merge. Can you take a look at the issues reported and see if you can address them? If you need assistance please let us know.

markafarrell:
> @markafarrell Sorry for the late response. We recently made a change to the repo where we need to address issues identified by Sonar before we can merge. Can you take a look at the issues reported and see if you can address them? If you need assistance please let us know.

@john-westcott-iv 

These should all be addressed now.

1. I moved claim extraction for the jwt a separate function
2. I added tests to validate behavior when jwt signature verification fails
3. remove unused variables

Hopefully SonarQube is happy with that

john-westcott-iv:
@markafarrell things are looking better except some linting issues.

markafarrell:
> @markafarrell things are looking better except some linting issues.

@john-westcott-iv Linting fixed and squashed into a single commit

tznamena (approved):

ansible_base/authentication/social_auth.py
```diff
@@ -189,12 +192,57 @@ def validate(self, serializer, data):
         return data
 
 
+def extract_id_token_claims(backend, response):
+    try:
```
john-westcott-iv: In your code below you wrap this function in this if. Lets make this method deal with that case itself incase we ever wanted to call it from somewhere else:

```suggestion
   if 'id_token' not in response:
       return {}
    try:
```

ansible_base/authentication/social_auth.py
```diff
@@ -189,12 +192,57 @@ def validate(self, serializer, data):
...
+    if groups_claim in response:
+        # Attempt to extact groups claim from data returned by UserInfo endpoint
+        extra_groups = response[groups_claim]
+        logger.debug(f"extra_groups extracted from UserInfo: {extra_groups}")
+    else:
+        # If groups_claim isn't in UserInfo attempt to extract it from id_token
+        if 'id_token' in response:
+            id_token_claims = extract_id_token_claims(backend, response)
+        else:
+            id_token_claims = {}
```
john-westcott-iv: Based on the change above this gets simplified:

```suggestion
        id_token_claims = extract_id_token_claims(backend, response)
```

ansible_base/authentication/social_auth.py
```diff
@@ -189,12 +192,57 @@ def validate(self, serializer, data):
...
 
     groups_claim = backend.groups_claim if backend.groups_claim is not None else "Group"
 
-    extra_groups = response[groups_claim] if groups_claim in response else []
+    if groups_claim in response:
+        # Attempt to extact groups claim from data returned by UserInfo endpoint
+        extra_groups = response[groups_claim]
+        logger.debug(f"extra_groups extracted from UserInfo: {extra_groups}")
+    else:
+        # If groups_claim isn't in UserInfo attempt to extract it from id_token
```
john-westcott-iv: Would there ever be a condition where we could get `groups_claims` and also have groups in `id_token` and want to get both?

ansible_base/authentication/social_auth.py
```diff
@@ -189,12 +192,57 @@ def validate(self, serializer, data):
...
+        if 'id_token' in response:
+            id_token_claims = extract_id_token_claims(backend, response)
+        else:
+            id_token_claims = {}
+
+        if groups_claim in id_token_claims:
+            extra_groups = id_token_claims[groups_claim]
+            logger.debug(f"extra_groups extracted from id_token: {extra_groups}")
+        else:
+            extra_groups = []
```
john-westcott-iv: What are your thoughts on simplifying this block a bit:
```suggestion
        extra_groups = id_token_claims.get(groups_claim, [])
        logger.debug(f"extra_groups extracted from id_token: {extra_groups}")
```

This would also give us the benefit of always logging from this function so we would know that we got into here even if we ended up with no groups.

ansible_base/authentication/social_auth.py
```diff
@@ -189,12 +192,57 @@ def validate(self, serializer, data):
...
 
+def extract_id_token_claims(backend, response):
+    try:
+        id_token = response['id_token']
+        client_id, _ = backend.get_key_and_secret()
+
+        key = backend.find_valid_key(id_token)
+
+        if not key:
+            raise AuthTokenError(backend, "Signature verification failed")
```
john-westcott-iv: Out of curiosity, why `raise` here instead of just logging and `return {}`? Is it possible that other calls in this try might also raise `AuthTokenError` so you want to catch that anyway?

ansible_base/authentication/social_auth.py
```diff
@@ -189,12 +192,57 @@ def validate(self, serializer, data):
...
+            algorithms=backend.setting("JWT_ALGORITHMS", backend.JWT_ALGORITHMS),
+            audience=client_id,
+            issuer=backend.id_token_issuer(),
+            options=backend.setting("JWT_DECODE_OPTIONS", backend.JWT_DECODE_OPTIONS),
+        )
+    except PyJWTError as e:
+        logger.error(f"Unable to decode id_token response JWT: {e}")
+        id_token_claims = {}
+    except AuthTokenError as e:
+        logger.error(f"Unable to decode id_token response JWT: {e}")
```
john-westcott-iv: Lets change these two messages slightly. If we were ever trying to debug why we got this message it will make our lives a bit simpler. Granted that having `e` in the response should give us an indication of where we hit it. But typically I would start by searching the code base for the fixed part of the error message and having them just slightly different would take me right there.

john-westcott-iv (commented):
This is looking great. I have just a couple minor code structure comments and then we should be good to merge this. Thanks again!

ansible_base/authentication/social_auth.py
markafarrell: I don't think this is really a valid use case. 

We could handle it though by doing something like:

```python
extra_groups = list(set(extra_groups_from_user_info + extra_groups_from_id_token)))
```

ansible_base/authentication/social_auth.py
markafarrell: seems reasonable

ansible_base/authentication/social_auth.py
markafarrell: I was essentially copying the implementation as closely as possible from https://github.com/python-social-auth/social-core/blob/master/social_core/backends/open_id_connect.py#L203

Wanted to change as little as possible so I didn't introduce any additional security vulnerabilities

ansible_base/authentication/social_auth.py
markafarrell: fixed

ansible_base/authentication/social_auth.py
markafarrell: fixed

ansible_base/authentication/social_auth.py
markafarrell: fixed

ansible_base/authentication/social_auth.py
markafarrell: fixed

markafarrell:
> This is looking great. I have just a couple minor code structure comments and then we should be good to merge this. Thanks again!

@john-westcott-iv I believe it should be all good now.

BrennanPaciorek:
I'll do a test run of this PR today.

test_app/tests/authentication/test_social_auth.py
```diff
@@ -89,37 +90,226 @@ def test_social_auth_validate_callback_mixin(mocked_generate_slug, mocked_revers
...
 
+        def id_token_issuer(self):
+            # must match iss claim in jwt
+            return 'https://login.microsoftonline.com/9122040d-6c67-4c5b-b112-36a304b66dad/v2.0'
+
+        def get_key_and_secret(self):
+            # client_id must match aud claim in jwt
+            return ('6cb04018-a3f5-46a7-b995-940c78f5aef3', '')
+
+        def find_valid_key(self, id_token):
```
BrennanPaciorek: What classes provide this method normally? Cannot find it in any superclasses of authentication plugins, though my search for it has not been exhaustive.

ansible_base/authentication/social_auth.py
```diff
@@ -189,12 +192,53 @@ def validate(self, serializer, data):
...
 
+def extract_id_token_claims(backend, response):
+    if 'id_token' not in response:
+        return {}
+
+    try:
+        id_token = response['id_token']
+        client_id, _ = backend.get_key_and_secret()
+
+        key = backend.find_valid_key(id_token)
```
BrennanPaciorek: Will this always be defined for all authenticator plugins calling this? No base class that I can locate in my dependencies seems to have `find_valid_key` defined.

test_app/tests/authentication/test_social_auth.py
markafarrell: Its implemented in `social_core.backends.open_id_connect::OpenIdConnectAuth` https://github.com/python-social-auth/social-core/blob/master/social_core/backends/open_id_connect.py#L175

ansible_base/authentication/social_auth.py
markafarrell: It's implemented in `social_core.backends.open_id_connect::OpenIdConnectAuth` https://github.com/python-social-auth/social-core/blob/master/social_core/backends/open_id_connect.py#L175.

This is essentially only the case for the `oidc` authenticator plugin.

I think we probably actually want to move this into the authenticator plugin instead of here.

Let me have a go at that

john-westcott-iv:
We got a report of this raising an exception from a user. We need to investigate this further before merging.

tznamena:
I'm looking at this again, it is related to a customer reported issue I'm investigating.

tznamena:
@markafarrell I have found that we can simplify changes in this PR because the authentication backend(s) will hold on to already decoded `id_token` and therefore we don't need to find the keys and decode the token again in the pipeline function.
Also, with the simplified code, even the tests are much simpler.
Please have a look at the proposed changes and respond here if you find any issues integrating this into your PR or have other questions.

Changes for `ansible_base/authentication/social_auth.py`:
- if the backend holds on to `id_token` then the groups claim is pulled from that, or `[]` otherwise
```diff
@@ -195,6 +195,13 @@ def create_user_claims_pipeline(*args, backend, response, **kwargs):
     groups_claim = backend.groups_claim if backend.groups_claim is not None else "Group"
 
     extra_groups = response[groups_claim] if groups_claim in response else []
+
+    # attempt to get group claims from id token if current backend has it
+    if hasattr(backend, "id_token"):
+        extra_groups_from_id_token = backend.id_token.get(groups_claim, [])
+        extra_groups = list(set(extra_groups + extra_groups_from_id_token))
+
+    logger.debug(f'updating user claims with groups claim "{groups_claim}": {extra_groups}')
     user = update_user_claims(kwargs["user"], backend.database_instance, backend.get_user_groups(extra_groups))
     if user is None:
         return SOCIAL_AUTH_PIPELINE_FAILED_STATUS
```

Changes for `test_app/tests/authentication/test_social_auth.py`:
- added more parameters and added `id_token` to the mocked backend
```diff
@@ -90,16 +90,18 @@ def test_social_auth_validate_callback_mixin(mocked_generate_slug, mocked_revers
 
 
 @pytest.mark.parametrize(
-    "groups_claim,returned_groups,expected_groups",
+    "groups_claim, returned_groups, id_token_groups, expected_groups",
     [
-        (None, ["mygroup"], ["mygroup"]),
-        ("groups", ["mygroup"], ["mygroup"]),
-        (None, None, []),
-        ("groups", None, []),
+        (None, ["mygroup"], None, ["mygroup"]),
+        ("groups", ["mygroup"], None, ["mygroup"]),
+        (None, None, None, []),
+        ("groups", None, None, [],),
+        ("Group", None, ["group-1"], ["group-1"]),
+        ("Group", ["group-1"], ["group-2"], ["group-1", "group-2"]),
     ],
 )
 @mock.patch("ansible_base.authentication.utils.claims.update_user_claims")
-def test_create_user_claims_pipeline(mock_update_user_claims, groups_claim, returned_groups, expected_groups):
+def test_create_user_claims_pipeline(mock_update_user_claims, groups_claim, returned_groups, id_token_groups, expected_groups):
     '''
     We are testing to see if extracting groups from a claim is working correctly
     '''
@@ -120,9 +122,14 @@ def test_create_user_claims_pipeline(mock_update_user_claims, groups_claim, retu
     if returned_groups is not None:
         rData[backend.groups_claim] = returned_groups
 
+    id_token = {'claim': "value"}
+    if id_token_groups is not None:
+        id_token[groups_claim] = id_token_groups
+        backend.id_token = id_token
+
     user = {
         'auth_time': "2024-11-07T05:19:08.224936Z",
-        'id_token': "asdf",
+        'id_token': id_token,
         'refresh_token': None,
         'id': "ccd2cf13-d927-41ad-cd8c-adb18b2e5f78",
         'access_token': "asdf",
```

tznamena (changes requested):
Changes proposed in a comment.

markafarrell:
@tznamena I think it actually makes more sense to have this logic added to `authentication/authenticator_plugins/oidc.py::AuthenticatorPlugin.user_data` 

I will do some commits for you to review

tznamena:
Putting the change in one authentication plugin solves the problem **only** for that one authentication. The original location in the pipeline function was better as it solves the problem for any plugin that uses `id_token`. Let's move it back to that function.

tznamena (changes requested):
Let's keep the changes in the `social_auth.create_user_claims_pipeline()` as that will be applicable to all authentication methods that include `id_token`.

test_app/tests/authentication/test_social_auth.py
```diff
@@ -89,36 +89,127 @@ def test_social_auth_validate_callback_mixin(mocked_generate_slug, mocked_revers
...
+
+        def __init__(self, groups_claim=None, id_token=None, user_info_groups=None):
+            self._groups_claim = groups_claim
+            self.id_token = id_token
+            self.user_info_groups = user_info_groups
 
-        def get_user_groups(self, extra_groups=[]):
-            return extra_groups
+        def setting(self, a, b=None):
+            settings = {"JWT_ALGORITHMS": "RS256", "JWT_DECODE_OPTIONS": None, "GROUPS_CLAIM": self._groups_claim, "USERINFO_URL": "https://asdf"}
```
bhavenst: Use constants defined above here?

test_app/tests/authentication/test_social_auth.py
```diff
@@ -89,36 +89,127 @@ def test_social_auth_validate_callback_mixin(mocked_generate_slug, mocked_revers
...
+    "preferred_username": "AbeLi@microsoft.com",
+    "email": "AbeLi@microsoft.com",
+    "groups": ["myidtokengroup"],
+    "oid": "00000000-0000-0000-66f3-3332eca7ea81",
+    "tid": "9122040d-6c67-4c5b-b112-36a304b66dad",
+    "nonce": "123523",
+    "aio": "Df2UVXL1ix!lMCWMSOJBcFatzcGfvFGhjKv8q5g0x732dR5MB5BisvGQO7YWByjd8iQDLq!eGbIDakyp5mnOrcdqHeYSnltepQmRp6AIZ8jY",
+}
+
+id_token_duplicate_group = {
```
bhavenst: id_token_no_groups = {...}
id_token = {**id_token_no_groups, "groups": ["myidtokengroup"]}
id_token_duplicate_group = {**id_token, "groups": ["mygroup", "myidtokengroup"]}

Can save some space. :)

bhavenst (commented):
Looks like the UT is failing due to mock returning string instead of json object..

markafarrell:
> Putting the change in one authentication plugin solves the problem **only** for that one authentication. The original location in the pipeline function was better as it solves the problem for any plugin that uses `id_token`. Let's move it back to that function.

Moved back to `create_user_claims_pipeline`

test_app/tests/authentication/test_social_auth.py
markafarrell: not relevant anymore as this has been removed

test_app/tests/authentication/test_social_auth.py
markafarrell: fixed.

tznamena (approved):
Pulled down, ran the unit tests and tested with OIDC and Azure AD(with changes from PR #708 ).

markafarrell:
@tznamena are you able to approve the CI workflow so it can run?

john-westcott-iv (approved):

bhavenst (approved):

test_app/tests/authentication/test_social_auth.py
bhavenst: Great, thanks

test_app/tests/authentication/test_social_auth.py
bhavenst: Cool

john-westcott-iv:
@markafarrell thank you so much for working with us on this. I know it was quite a long process!

PR #684, opened 2025-01-16
AAP-39232 Use ghcr mirror for image

AlanCoding:
Was done in AWX https://github.com/ansible/awx/pull/15743/files

### DIFF
--- a/tools/dev_postgres/Dockerfile
+++ b/tools/dev_postgres/Dockerfile
@@ -1,4 +1,4 @@
-FROM postgres:15
+FROM mirror.gcr.io/library/postgres:15
 
 ENV POSTGRES_DB=dab_db
 ENV POSTGRES_USER=dab
--- a/compose/ingress/Dockerfile
+++ b/compose/ingress/Dockerfile
@@ -1,4 +1,4 @@
-FROM nginx:latest
+FROM mirror.gcr.io/library/nginx:latest
 
 COPY entrypoint.sh /entrypoint.sh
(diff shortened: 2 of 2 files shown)


tools/dev_postgres/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM postgres:15
+FROM mirror.gcr.io/library/postgres:15
```
TheRealHaoLiu, line 1:
```suggestion
FROM quay.io/sclorg/postgresql-15-c9s
```

AlanCoding: The quay image uses different environment variables, and causes problems with parallel tests because the default user doesn't have permission to create new databases.

TheRealHaoLiu: ok got it... i suppose this is fine since this is not a productized component...
just to note that sclorg image is the most compatible with the postgres image we ship downstream

TheRealHaoLiu (approved):

tools/dev_postgres/Dockerfile
AlanCoding: yeah, and I would convert to it, but I need docs or something that says "this is how you create a database". From https://quay.io/repository/sclorg/postgresql-15-c9s?tab=info, I guess we could do everything as the admin account.

Or, better yet, we could create all users for all projects, and give all users permission to create all the things, throw all this in a Dockerfile and push the result to ghcr so we don't ever have to configure anything ever again.

chrismeyersfsu (approved):

PR #685, opened 2025-01-23
Add Feature Flag List state API

zkayyali812:
# [[AAP-39135](https://issues.redhat.com/browse/AAP-39135)] Add Feature Flag List API

## Description
- What is being changed?
Adds a get/list API for feature flags to the DAB for consumption by other components
- Why is this change needed?
To provide a common API for other components to get/modify the FLAGS variable
Adds a list/get feature flags API

## Type of Change

- [x] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

## Testing Instructions

Deploy local environment and visit - 
1. http://127.0.0.1:8000/api/v1/feature_flags_state/
2. Also run unit tests

### Steps to Test
1. 
5. 
6. 

### Expected Results
Feature flags get/list API should work. In test-app, we should see 3 dummy feature flags.

### Required Actions

- [ ] Blocked by PR/MR: #XXX


ansible_base/feature_flags/migrations/0001_initial.py
```diff
@@ -0,0 +1,20 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='FeatureFlag',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+            ],
+        ),
+    ]
```
rochacbruno: Can we skip adding the Model and migration for now?

it looks like for this phase the model is not needed

ansible_base/feature_flags/urls.py
```diff
@@ -0,0 +1,13 @@
+from django.urls import path, re_path
+
+from ansible_base.feature_flags import views
+from ansible_base.feature_flags.apps import FeatureFlagsConfig
+
+app_name = FeatureFlagsConfig.label
+
+api_version_urls = [
+    path('feature-flags/', views.FeatureFlagsListView.as_view(), name='featureflags-list'),
+    re_path(r'feature-flags/(?P<category_slug>[a-zA-Z0-9_]+)/$', views.FeatureFlagDetailView.as_view(), name='featureflags-detail'),
```
rochacbruno: Is this a detail for a specific single FLAG or a category of flags?

Is there a requirement to filter by category? if so, can we have another endpoint for getting a specific flag?

or if it is a single flag detail rename to `<flag_name>` (unless django-flags requires that naming, in that case we can just add a comment explaining)

ansible_base/feature_flags/utils.py
```diff
@@ -0,0 +1,7 @@
+from django.conf import settings
+
+
+def get_django_flags():
+    if hasattr(settings, 'FLAGS'):
+        return settings.FLAGS
+    return {}
```
rochacbruno: I suggest using `getattr` with a default here, historically there were [problems](https://hynek.me/articles/hasattr/) with hasattr, and getattr is more efficient anyway.

```suggestion
    return getattr(settings, 'FLAGS', {})
```

ansible_base/lib/dynamic_config/settings_logic.py
```diff
@@ -297,4 +299,28 @@ def get_dab_settings(
             if PRIMARY_CACHE not in caches or FALLBACK_CACHE not in caches:
                 raise RuntimeError(f'Cache definitions with the keys {PRIMARY_CACHE} and {FALLBACK_CACHE} must be defined when DABCacheWithFallback is used.')
 
+    if 'ansible_base.feature_flags' in installed_apps:
+        dab_data.setdefault('INSTALLED_APPS', copy(installed_apps))
+        dab_data['INSTALLED_APPS'].append('flags')
```
rochacbruno: Services already added `flags` so we need to avoid it being added twice
```suggestion
        dab_data.setdefault('INSTALLED_APPS', copy(installed_apps))
        if "flags" not in dab_data["INSTALLED_APPS"]:
            dab_data['INSTALLED_APPS'].append('flags')
```

ansible_base/feature_flags/urls.py
zkayyali812: Hey Bruno,
This is for a specific flag, not a category of flags. Will rename this

ansible_base/feature_flags/utils.py
zkayyali812: Great note! Will update

ansible_base/lib/dynamic_config/settings_logic.py
zkayyali812: Great note! Will update

ansible_base/feature_flags/views.py
```diff
@@ -0,0 +1,24 @@
...
+    serializer_class = FeatureFlagSerializer
+    filter_backends = []
+    name = _('Feature Flags')
+    http_method_names = ['get', 'head']
+
+    def get(self, request, format=None):
+        return Response(get_django_flags())
+
+    def get_queryset(self):
+        return get_django_flags()
```
AlanCoding, line 25: What permissions should be set on this view strikes me as a shared problem with `/api/v2/ping/` in AWX, as well as others. This is what https://github.com/ansible/awx/pull/15724 was seeking to address. Doing a pure-AWX solution is not the right answer here, but I don't have a good answer as to what the answer should be.

zkayyali812: Hey @AlanCoding ,
Is there anything that I can do to better represent this now? Sounds like we may want to open another issue to address this later?

AlanCoding: I'm assuming the default and expectation would be that all authenticated users can view, but I'd first ask the PM driving this. Otherwise, no, there's nothing better I could do.

AlanCoding (changes requested):
The tests seem to be limited to GET type actions. I'd like to see `override_settings` used to test switching the value (simulating an admin taking this action).

AlanCoding (approved):
I would still like to see a solution to the FLAG_SOURCES issue, but this doesn't try to use that at all. I'll ping you on another patch soon about that. This seems to do the intended with the settings test added.

ansible_base/feature_flags/views.py
zkayyali812: This API endpoint is only valid for authenticated users currently. Resolving this convo, but please feel free to reopen if you think this should be discussed more

zkayyali812:
Hey @AlanCoding ,
Currently I would recommend leaving FLAG_SOURCES defined by each component, if they would like. Thinking it would be good to keep the DAB implementation as minimalist as possible. I'd love to learn more about the affects you're seeing with this, but from my understanding currently I think we can reliably use the default.

ansible_base/lib/dynamic_config/dynamic_settings.py
```diff
@@ -57,6 +57,11 @@
 except NameError:
     OAUTH2_PROVIDER = {}
 
+try:
+    TEMPLATES  # noqa: F821
+except NameError:
+    TEMPLATES = []
```
john-westcott-iv, line 63: Should we do the same thing for `FLAGS` (i.e. make sure it at least an ampty dict) or will the flags library work fine with everything above if its not defined?

ansible_base/lib/dynamic_config/settings_logic.py
```diff
@@ -297,4 +298,29 @@ def get_dab_settings(
...
+                for context_processor in template['OPTIONS']['context_processors']:
+                    if context_processor == 'django.template.context_processors.request':
+                        found_context_processor = True
+                # If we didn't find the context processor we care about append it
+                if not found_context_processor:
+                    template['OPTIONS']['context_processors'].append('django.template.context_processors.request')
+
+        # If we never even found the backend, add one
+        if not found_template_backend:
+            dab_data['TEMPLATES'].append({'BACKEND': 'django.template.backends.django.DjangoTemplates', 'OPTIONS': {'context_processors': []}})
```
john-westcott-iv: Do we not need the `django.template.context_processors.request` in this default setting?

ansible_base/lib/dynamic_config/dynamic_settings.py
zkayyali812: It will work fine without it defined. It will use the default if so - `{}`

ansible_base/lib/dynamic_config/settings_logic.py
zkayyali812: Good catch, I can update to include this

zkayyali812: Updated

AlanCoding:
I agree that DAB should not modify FLAG_SOURCES. That would only make our problems worse.

It may be harmless to leave the default as it is. But I want to avoid accidentally committing us to maintaining something that was not expected. If flags are allowed to be set in the local (component) database, then we would need to migrate those values to the shared API later. This is extremely non-trivial and introduces potential conflicts.

So I just want a statement, yeah, people might be using django-flags in a way that allows modification of the values via the database. But if anyone actually sets the database value, that's not something we're going to respect in the future versions, and we _will_ break it with current plans. Our current plans are extremely clear to toggle enablement via settings are restart services. This will change in phase 2, involving the shared API.

zkayyali812:
Yes, that is correct. Database flag sources will not be supported in this first phase. Only settings sources.
Phase 2, if and when prioritized, will likely involve runtime toggling, via the shared API.

ansible_base/lib/dynamic_config/settings_logic.py
```diff
@@ -297,4 +298,35 @@ def get_dab_settings(
...
+        # Look through all of the tmplates
+        for template in dab_data['TEMPLATES']:
+            # If this template has the BACKEND we care about...
+            if template['BACKEND'] == 'django.template.backends.django.DjangoTemplates':
+                found_template_backend = True
+                # Look through all of its context processors
+                found_context_processor = False
+                for context_processor in template['OPTIONS']['context_processors']:
+                    if context_processor == template_context_processor:
+                        found_context_processor = True
```
jshimkus-rh, line 318:
```suggestion
                        found_context_processor = True
                        break
```

zkayyali812: Good suggestion, will capture this

zkayyali812:
Please feel free to merge this when ready @john-westcott-iv and team.
I do not have merge rights here.

BrennanPaciorek:
Sure, I'm alright with merging this.

PR #687, opened 2025-01-28
AAP-37820: Hide python paths of authenticators from ui_auth

BrennanPaciorek:
change social auth urls

change ui auth infomration shown on authenticators

Make new authenticator slugs choosable by the user, defaulting to a uuid otherwise.

Fixes AAP-37820

# [AAP-37820] Brief Title Describing the Change

## Description
- What is being changed?

Restriction of information present in ui_auth, more random authenticator slugs, to make the information removed not derivable from the urls

- Why is this change needed?

## Type of Change
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites
The backend for the created authenticator needs to be set up

## Testing Instructions

- create an authenticator
- enable it
- Try to login with that authenticator
- Go to ui_auth endpoint

### Expected Results

you should not be able to derive the import path of the authenticator plugin from the ui auth endpoint and the urls contained

## Additional Context

For keycloak SAML authenticator, you will need to update the keycloak realms allowed redirect-uris to match the new social auth url formats

### Required Actions
- [X] Requires documentation updates
  
- [X] Requires downstream repository changes
  
- [X] Requires infrastructure/deployment changes
  
- [X] Requires coordination with other teams
  
- [ ] Blocked by PR/MR: #XXX


bhavenst:
Wouldn't this also require update of OIDC and Keycloak(OIDC) servers for the same reason as SAML?

BrennanPaciorek:
> Wouldn't this also require update of OIDC and Keycloak(OIDC) servers for the same reason as SAML?

Yes, confirmed identical changes are needed to the realm redirect uri for keycloak OIDC and SAML.

bhavenst (commented):
This looks good.  The dev keycloak will need to be updated somehow, since it hard codes the redirect URLs based on the old behavior in the realm template..  It might take some restructuring though since you don't know the redirect URLs until the authenticators are created after this change.  That or anyone who uses it will have to manually update the redirect URLs after starting it but that is a pain.  I don't know if it's possible to update the URLs via the keycloak admin API after it's started or not..
Screenshot From 2025-01-28 17-21-54

ansible_base/authentication/views/ui_auth.py
```diff
@@ -40,7 +40,6 @@ def generate_ui_auth_data():
             response['passwords'].append(
                 {
                     'name': authenticator.name,
-                    'type': authenticator.type,
                 }
```
john-westcott-iv, line 42: Lets check with the UI team as to how they are using this. We likely don't even need to expose the name here as the logic should be something like `if len(reponse['passwords']: show the password on the login form`. So maybe we could just do something like `response['passwords'].append(True)`. Ideally, if possible, we could just convert this to a boolean but the UI would likely need to make a change for that unless they are already doing something `if response['passwords']`.

ansible_base/authentication/authenticator_plugins/utils.py
```diff
@@ -53,7 +53,6 @@ def get_authenticator_urls(authenticator_type: str) -> list:
     return []
 
 
-def generate_authenticator_slug(type: str, name: str, suffix: Optional[str] = None) -> str:
-    clean_type = f"{type.replace('.', ' ')}"
-    slug_template = f"{clean_type}__{name}" if not suffix else f"{clean_type}__{name}__{suffix}"
+def generate_authenticator_slug(name: str) -> str:
+    slug_template = f"{name}__{uuid.uuid4()}"
```
john-westcott-iv: We really just need a unique identifier so that we can figure out which authenticator should get the response from an SSO attempt. One of the problems we have is a bit of a chicken and egg where some providers want the information to make the client secret but this information can be pre-determined easily so you have to save the client (which needs the client secret). Instead of having a uuid in here, which is going to be a bit more cringy for pre-determining the endpoint names, what if we just made a unique string field with a default of a uuid? The user could decide on a random phrase like `ardvark-sandwich` so that they could pre-configure the SSO authenticator. It could be optional in the UI but if we get nothing we either makeup some words or insert a UUID in there. Thoughts?

bhavenst: I love aardvark sandwiches.

That seems reasonable though to allow the (extension portion?) to be specified optionally.  As long as it validates there aren't any collisions then it sounds good and should make the dev env update easier..

BrennanPaciorek: I'm okay with this. so it would be provide a slug or have one generated as a uuid?

As far as random strings go, uuid is hard to beat.

ansible_base/authentication/authenticator_plugins/utils.py
```diff
@@ -53,7 +53,5 @@ def get_authenticator_urls(authenticator_type: str) -> list:
     return []
 
 
-def generate_authenticator_slug(type: str, name: str, suffix: Optional[str] = None) -> str:
-    clean_type = f"{type.replace('.', ' ')}"
-    slug_template = f"{clean_type}__{name}" if not suffix else f"{clean_type}__{name}__{suffix}"
-    return slugify(slug_template)
+def generate_authenticator_slug() -> str:
+    return slugify(uuid.uuid4())
```
john-westcott-iv: Instead of using slugify in the serializer lets have this method take an optional string and return a slug from it. Obviously you will have to import Optional again.

```suggestion
def generate_authenticator_slug(value: Optional[str] = None) -> str:
    if value is None:
        value = uuid.uuid4()
    return slugify(value)
```

ansible_base/authentication/serializers/authenticator.py
```diff
@@ -1,27 +1,46 @@
...
+class AuthenticatorSerializer(NamedCommonModelSerializer, ImmutableFieldsMixin):
     type = ChoiceField(get_authenticator_plugins())
 
     def validate_type(self, value):
         if self.instance and self.instance.type != value:
             raise ValidationError(_("Cannot change authenticator type after it has been created."))
         return value
 
+    def validate_slug(self, value):
+        value = slugify(value)
```
john-westcott-iv: If we alter the get_authenticator_slug we can call it here instead. This would also get rid of the else statement.

ansible_base/lib/serializers/mixins.py
```diff
@@ -0,0 +1,24 @@
...
+class ImmutableFieldsMixin(serializers.ModelSerializer):
+    def get_extra_kwargs(self):
+        kwargs = super().get_extra_kwargs()
+        immutable_fields = getattr(self.Meta, "immutable_fields", None)
+
+        # Make field read_only if instance already exists
+        if self.instance and immutable_fields:
+            for field in immutable_fields:
+                kwargs.setdefault(field, {})
+                kwargs[field]["read_only"] = True
```
john-westcott-iv: What does this do, if anything, to the options page? Would it be read_only = False on a list page and read_only = True on view page? Does this mess with the OpenAPI schema at all?

john-westcott-iv (commented):
Can we alter the help_text on the Authenticator.slug field to include:
```
If unspecified, a UUID will be genrated and used for the slug.
```

ansible_base/lib/serializers/mixins.py
BrennanPaciorek: Good catch. It seems the options page does not take into account what endpoint the `metadata` action is being called on, so it treats it as read_only="false" on `/api/authenticators/{id}`. OpenAPI schema also treats `slug` as a valid option on `PUT` and `PATCH` actions.

Looking into ways to fix this. OpenAPI schema should be an easier fix, but for the options page, we may need to tweak `rest_framework.metadata.SimpleMetadata.get_serializer_info` in a subclass to make the `metadata` action feed us more dynamic metadata.

BrennanPaciorek: For the options page, the fix is simple but not super clean.

It seems to involve either doing as directed in my previous comment, which leads to duplicated code, or overriding `get_serializer` to return an instanced one if possible and if the action is `metadata`.

ansible_base/authentication/views/ui_auth.py
BrennanPaciorek: ui team does not seem to have any usage or intentions to use `response['passwords']`. It seems password will be true.

ansible_base/lib/serializers/mixins.py
BrennanPaciorek: I'm throwing in the towel on the OpenAPI spec, it should generate working API specs, just cannot seem to get the schema to change based on what context the serializer is generated. managed to return different values depending on the context, but the end result didn't consume the modified introspection.

ansible_base/authentication/views/ui_auth.py
BrennanPaciorek: Forgot that making passwords True would break typing of UIAuth model. Any recommendations there?

ansible_base/authentication/views/ui_auth.py
```diff
@@ -36,11 +35,9 @@ def generate_ui_auth_data():
 
     for authenticator in authenticators:
         if authenticator.category == 'password':
-            response['show_login_form'] = True
             response['passwords'].append(
                 {
                     'name': authenticator.name,
```
bhavenst, line 41: So this is sticking around after all?  thought it was just going to be passwords: true..

ansible_base/lib/serializers/mixins.py
```diff
@@ -0,0 +1,35 @@
...
+        # Make field writable if no instance yet exists
+        elif immutable_fields:
+            for field in immutable_fields:
+                kwargs.setdefault(field, {})
+                kwargs[field]["read_only"] = False
```
bhavenst: Dumb that python doesn't have a one line way to set dict[x][y] or this whole thing could be one line.

OK, found one. :) 

new_kwargs = dict.fromkeys(immutable_fields, {"read_only": True if self.instance else False})

Ha ha

test_app/tests/authentication/models/test_authenticator.py
```diff
@@ -37,16 +36,14 @@ def test_authenticator_order_on_create_update():
 
 @pytest.mark.django_db
 def test_dupe_slug(ldap_authenticator):
-    ldap_auth = Authenticator.objects.first()
```
bhavenst, line 40: Hmm, clearly I didn't know how those fixtures worked. :)

test_app/tests/authentication/serializers/test_authenticator.py
```diff
@@ -1,11 +1,38 @@
...
+        assert uuid.UUID(serializer.validate_slug(""))
+    except ValueError:
+        pytest.fail("uuid not generated if a blank slug is provided")
+
+
+def test_validate_new_authenticator_slug(shut_up_logging):
+    slug = slugify(uuid.uuid4())
+
+    serializer = AuthenticatorSerializer()
+    assert serializer.validate_slug(slug) == slugify(slug)
```
bhavenst, line 23: Double slugification? :)

ansible_base/authentication/views/ui_auth.py
bhavenst: If they don't use it at all maybe it can be removed entirely?

bhavenst (commented):
I will test w/ gateway too

test_app/tests/authentication/serializers/test_authenticator.py
BrennanPaciorek: Good point. I missed that.

BrennanPaciorek: Fixed, sort of. typeguard will whine if I don't wrap it in `str()` though.

ansible_base/lib/serializers/mixins.py
BrennanPaciorek: I've made a change to dedup the logic. still not a one-liner, but I think its more readable this way.

bhavenst:
I don't get the login buttons for keycloak oidc/saml with these changes, but not sure why.  Maybe the UI doesn't like the new format of ui_auth.  I'm terrible w/ javascript/patternfly but here I see it wanting name, login_url, type: https://gitlab.cee.redhat.com/ansible/aap-ui/-/blob/main/frontend/common/SocialAuthLogin.tsx?ref_type=heads

ansible_base/lib/serializers/mixins.py
BrennanPaciorek: Can still make more changes actually. Running tests then pushing them up here.

ansible_base/lib/serializers/mixins.py
```diff
@@ -0,0 +1,30 @@
...
+
+    #        See ansible_base.authentication.views.AuthenticatorViewSet for an example.
+    #    2. The generated OpenAPI spec will treat immutable fields as valid parameters on PUT and PATCH endpoints
+
+    def get_extra_kwargs(self):
+        kwargs = super().get_extra_kwargs()
+        immutable_fields = getattr(self.Meta, "immutable_fields", None)
+
+        # Make field read_only if instance already exists
+        if immutable_fields:
```
bhavenst: yeah, the if is extraneous here I think.

BrennanPaciorek:
It seems its relying on `type` to fetch icons. I'll obfuscate it for now regardless, since it does not seem to be the thing breaking ui

bhavenst:
> Can we alter the help_text on the Authenticator.slug field to include:
> 
> ```
> If unspecified, a UUID will be genrated and used for the slug.
> ```

@john-westcott-iv Do we have any aversion to migrations in general?  This causes one just to update help text..  Fine if it's fine, just not sure if migrations are painful to anyone.

ansible_base/lib/serializers/mixins.py
bhavenst: Ah yeah, it wasn't but now is, nice.

BrennanPaciorek:
> I don't get the login buttons for keycloak oidc/saml with these changes

Seems thats what `show_login_form` is used for. We were convinced it was useless, but that appears to be its use case. It shows the buttons when its on.

Not sure why its switched on when there is a password authenticator, and not when there's an sso authenticator present.

BrennanPaciorek:
It seems re-adding types as I did satisfies the necessary conditions for icons to work, so I'll do that.

It was never a requirement to remove them entirely anyways.

ansible_base/authentication/serializers/authenticator.py
```diff
@@ -3,25 +3,40 @@
...
 
     def validate_type(self, value):
         if self.instance and self.instance.type != value:
             raise ValidationError(_("Cannot change authenticator type after it has been created."))
         return value
 
+    def validate_slug(self, value):
+        value = generate_authenticator_slug(value)
+        if self.instance and self.instance.slug != value:
+            raise ValidationError(_("Cannot change slug after it has been created."))
```
bhavenst: Testing this from GW API I just see the slug not changed and the config returned as it was with a 200 and no entry in the log, so not sure if that is proper or if this validationerror should be visible.  Same for PUT and PATCH.

bhavenst (commented):
Just the one question..

ansible_base/authentication/serializers/authenticator.py
BrennanPaciorek: That makes sense, I'll toss that validation error, leaving a comment describing why. It seems `read_only` prevents the field from being included in PUT/PATCH inputs.

bhavenst (approved):

ansible_base/authentication/social_auth.py
```diff
@@ -179,7 +179,9 @@ def validate(self, serializer, data):
         configuration = data['configuration']
         if not configuration.get('CALLBACK_URL', None):
             if not serializer.instance:
-                slug = generate_authenticator_slug(data['type'], data['name'])
+                if not data.get("slug"):
+                    data["slug"] = generate_authenticator_slug()
```
john-westcott-iv: We could skip this if now. Just do `data['slug'] = generate_authenticator_slug(data['slug'])`

ansible_base/authentication/views/ui_auth.py
```diff
@@ -25,9 +25,9 @@ def get(self, request, format=None):
 def generate_ui_auth_data():
     authenticators = Authenticator.objects.filter(enabled=True)
     response = {
-        'show_login_form': False,
         'passwords': [],
         'ssos': [],
+        'show_login_form': False,
```
john-westcott-iv, line 30: We are not going to nix this one? If not, can we move it back up to prevent the "random" change.

ansible_base/authentication/views/ui_auth.py
```diff
@@ -36,22 +36,16 @@ def generate_ui_auth_data():
...
             try:
-                response['ssos'].append(
-                    {
-                        'name': authenticator.name,
-                        'login_url': authenticator.get_login_url(),
-                        'type': authenticator.type,
-                    }
-                )
+                response['ssos'].append({'name': authenticator.name, 'login_url': authenticator.get_login_url(), 'type': authenticator.type.split('.')[-1]})
+                response["show_login_form"] = True
```
john-westcott-iv, line 48: These seem like more erroneous changes.

ansible_base/authentication/social_auth.py
BrennanPaciorek: Will do this.

ansible_base/authentication/views/ui_auth.py
BrennanPaciorek: Yeah, apparently it controls if the login buttons appear on UI

ansible_base/authentication/views/ui_auth.py
BrennanPaciorek: show login form change is intentional, black flattened the dict.

BrennanPaciorek: last component of type is needed because the last component chooses the icon.

john-westcott-iv (approved):

PR #693, opened 2025-02-05
AAP-39124: Correct required param list in openapi spec

bhavenst:
## Description

This just adds a param that tells DRF spectacular not to mark read-only items as required.  This results in only those params that are truly required being marked that way, which will help ATF client auto-generation.

For User:

BEFORE:
      required:
      - created
      - created_by
      - id
      - is_platform_auditor
      - last_login
      - managed 
      - modified
      - modified_by
      - related
      - summary_fields
      - url
      - username

AFTER:
      required: 
      - username

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites
None.

### Steps to Test
1. From gateway open gateway/api/v1/docs/schema
2. See that only those params actually required by the models are listed as required in the schema
3. 

### Expected Results
schema is correct

### Required Actions
  
- [ ] Blocked by PR/MR: #XXX


tznamena (approved):

ansible_base/lib/dynamic_config/settings_logic.py
```diff
@@ -61,6 +61,7 @@ def get_dab_settings(
             'DESCRIPTION': 'Open API',
             'VERSION': 'v1',
             'SCHEMA_PATH_PREFIX': '/api/v1/',
+            'COMPONENT_NO_READ_ONLY_REQUIRED': True,
```
john-westcott-iv, line 64: I am writing this assuming the default is False.

I don't have a problem if we want to change this to True here but we should make sure no other services are using this app yet or we will end up changing their behavior. 
If no one is using it, we can approve and merge this.
If someone is using it I am fine with changing it if we have consensus with other teams using it.
If we don't have consensus, we should default this to what it would be (again presumably False) and then individual services could override it to True in their own settings.

bhavenst: Yeah, default is False.  I'll post in the DAB channel and give people a chance to complain..

AlanCoding: https://github.com/ansible/eda-server/blob/85cfadcf8846473e3a5a700c81108f7acd8af400/src/aap_eda/settings/default.py#L565-L576

Re-stating, what @john-westcott-iv is saying is that `SPECTACULAR_SETTINGS` in eda-server will be modified by this patch. What consequences that'll have, we don't really know.

Likewise for galaxy_ng

https://github.com/ansible/galaxy_ng/blob/b525b5f1d030b3d16c89212e92eb975958439685/galaxy_ng/app/settings.py#L229

AWX is using an older schema generator library, so not affected.

bhavenst: Yeah, that is true for sure.  I asked on the DAB channel for some feedback, but it's also fine to just make this change in gateway and let other DAB users deal with it as they desire or don't desire.

relrod: Yeah I'd like some confirmation from galaxy and eda before I +1 this. Chances are they are just facing the same bug we are solving and they probably want this fix too, but we need to check and not assume.

relrod: @rochacbruno, @Alex-Izquierdo, @jneedle for eyes.

Alex-Izquierdo: Hi @AlanCoding @john-westcott-iv @bhavenst 
EDA is not using `ansible_base.api_documentation` 
Unless I'm missing something, if we don't define the app you would not be overriding the spectacular settings. Am I wrong?

rochacbruno: +1 galaxy is not using api_documentation app, so not affected.

AlanCoding:
Just for due diligence, did you look at `COMPONENT_SPLIT_REQUEST` for this same problem?

https://drf-spectacular.readthedocs.io/en/latest/client_generation.html

The writing there suggests that the split-request solution would be more maintainable in the longer term. But depending on what you need this for, using read-only-not-required might be necessary.

ansible_base/lib/dynamic_config/settings_logic.py
bhavenst: Hi @Alex-Izquierdo I am trying to test w/ EDA, just having troubles getting it started.. :)

Can you clarify if you aren't using openapi spec gen at all vs you are using spectacular but without the (minimal) extra config that api_documentation brings?  But yes, you are correct that your SPECTACULAR_SETTINGS will not be modified by DAB if you aren't including the api_documentation app.

bhavenst:
Hi @AlanCoding yes I actually started with COMPONENT_SPLIT_REQUEST since I thought maybe it would split request/response and fix the required field issue for request only.  But that was not the case.  It does split them, but they come out the same.  So then I tried COMPONENT_NO_READ_ONLY_REQUIRED with and without COMPONENT_SPLIT_REQUEST and did not see any benefit to including COMPONENT_SPLIT_REQUEST since it just creates identical separate schema objects for request.  I'm fine including it though, since it doesn't seem to do any harm.

The purpose for this change was just to unblock ATF client generation, so perhaps I can work with Tomer who filed the bug to see if it works better with or without the split.

ansible_base/lib/dynamic_config/settings_logic.py
bhavenst: Got EDA up and running and answered my own question.  So yes you do use spectacular, cool and I see that your solution takes care of the issue this is trying to fix, so it doesn't benefit or hurt you at all..

john-westcott-iv (approved):

PR #694, opened 2025-02-05
Fix the docker-compose serving

AlanCoding:
This fixes a grab bag of problems for me, but still doesn't get it to usability. The static files makes this render.

Screenshot from 2025-02-05 15-04-56

And then the other changes address the errors in logs

```
dab_postgres  | 2025-02-05 18:39:53.062 UTC [58] FATAL:  role "$${POSTGRES_USER}" does not exist
```

and

```
test_app-1    | uwsgi_check_logrotate()/lseek(): Illegal seek [core/logging.c line 494]
```

This is addressed by the uwsgi config file change, and to the startup script.

I think this is some progress, but unfortunately, I'm still blocked on using the docker-compose environment as I get a CSRF error when trying to log in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ tools/generated/*
 
 # generated ssl certs
 compose/ingress/certs/*
+
+# static files
+test_app/static_collected/
--- a/test_app/scripts/container_startup_uwsgi.sh
+++ b/test_app/scripts/container_startup_uwsgi.sh
@@ -19,6 +19,7 @@ echo "Read the custom settings file data just as a test"
 $DYNACONF -i test_app.settings.DYNACONF inspect -k JUST_A_TEST
 
 $PYTHON manage.py migrate
+$PYTHON manage.py collectstatic --clear --noinput
 DJANGO_SUPERUSER_PASSWORD=password DJANGO_SUPERUSER_USERNAME=admin DJANGO_SUPERUSER_EMAIL=admin@stuff.invalid $PYTHON manage.py createsuperuser --noinput || true
 $PYTHON manage.py authenticators --initialize
 $PYTHON manage.py create_demo_data
--- a/test_app/uwsgi.ini
+++ b/test_app/uwsgi.ini
@@ -7,7 +7,7 @@ chmod-socket = 660
 vacuum = true
 
 # Log to stdout
-logto = /dev/stdout
+# logto = /dev/stdout
 log-master = true
 #disable-logging = true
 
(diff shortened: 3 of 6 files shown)


test_app/uwsgi.ini
```diff
@@ -7,7 +7,7 @@ chmod-socket = 660
 vacuum = true
 
 # Log to stdout
-logto = /dev/stdout
+# logto = /dev/stdout
```
AlanCoding, line 10: This seemed to cause the problem related to seeking, that it was logging to a file that didn't allow seeking (which was /dev/stdout). It still logged without this, and I think we weren't really getting uwsgi logs without it, maybe.

.gitignore
```diff
@@ -49,3 +49,6 @@ tools/generated/*
 
 # generated ssl certs
 compose/ingress/certs/*
+
+# static files
+test_app/static_collected/
```
AlanCoding, line 54: I put this in /static/ first, but someone had already put 1 file in that folder. Using a different dir requires more settings to manage that but is then non-conflicting with this existing file. I tested it and things are working.

test_app/scripts/container_startup_uwsgi.sh
```diff
@@ -12,6 +12,7 @@ echo "settings.DATABASE ..."
 $PYTHON manage.py shell -c 'from django.conf import settings; print(settings.DATABASES)'
 
 $PYTHON manage.py migrate
+$PYTHON manage.py collectstatic --clear --noinput
```
AlanCoding, line 22: The `--clear` option would blow away the existing file in the test_app/static/ dir when I had it there, but using the new folder resolves this, so that'll only contain collected static files.

tznamena:
Can we get a list of the problems that are supposed to be fixed by this PR?

AlanCoding:
@tznamena 
 - serving static files was not working, so templates were kind of functional but didn't have DRF styling
 - postgres health_check in docker-compose didn't work when running other services with it
 - The intent of the docker compose was to use from port 443 (IIRC), so `https://localhost`, but you couldn't log in from that URL, and this fixes that CRSF related error so you can log in

john-westcott-iv:
@AlanCoding since this is just dev environment we will let you all review and and merge it. If you need anything from platform services please let us know.

john-westcott-iv (approved):

tznamena:
@AlanCoding can you rebase this one, it is ready to be merge then.

john-westcott-iv:
@AlanCoding Are you still intending on merging this?

AlanCoding:
Reconfirmed this is still working after resolving conflicts (predictable due to dynaconf merge).

 - Run `docker compose build` then `docker compose up`
 - Log in from http://localhost/ as admin/password
 - Can poke around at like http://localhost/api/v1/activitystream/10/

relrod (approved):

PR #696, opened 2025-02-17
A whole bucket of fixes to resource sync

newswangerd:
# 40296 Fixes for the sync task

## Description

This PR includes fixes for the following:
- https://issues.redhat.com/browse/AAP-39958
- https://issues.redhat.com/browse/AAP-40123
- https://issues.redhat.com/browse/AAP-40446
- https://issues.redhat.com/browse/AAP-40298

To summarize those tickets these changes are:
- Removing the check on the user.created field because Hub doesn't have it (AAP_40123)
- Fixing the delete function to use `Resource.delete_resource()` so that we don't accidentally trigger reverse sync
- Updating service owned resources (ie resources where the service ID is set to one of hub, controller or eda) so that they are kept in sync with the Gateway (AAP-40298)
  - The majority of the fixes are related to this. One big one is that the `service_id` query param now returns the resources owned by the service represented by `service_id` and the resource server.
- Fixing the user not getting loaded from their ansible_id on initial login (AAP-40446)
- Adding better error handling so that if a username conflicts during login, we log the user in with their old username.
  - The user's username should be fixed next time sync runs.
- Adding better error handling so that the sync task can continue if individual resources fail

Should these all be separate PRs? Probably. In my defense, they are all small fixes, and I need them to all be in place in order to adequately test this.

## Type of Change

- [ ] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites

Use the new tests in ATF.

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -137,7 +137,7 @@ class JWTCommonAuth:
 
                     resource.ansible_id = self.token['sub']
                     resource.service_id = self.token['service_id']
-                    resource.save()
+                    resource.save(update_fields=['ansible_id', 'service_id'])
 
         setattr(self.user, "resource_api_actions", self.token.get("resource_api_actions", None))
 
@@ -171,7 +171,6 @@ class JWTCommonAuth:
         if user_needs_save:
             with no_reverse_sync():
                 logger.info(f"Saving user {self.user.username}")
-
                 self.user.save()
 
     def validate_token(self, unencrypted_token, decryption_key):
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -102,6 +102,7 @@ class JWTCommonAuth:
             self.token = self.validate_token(token_from_header, cert_object.key)
 
         # Let's see if we have the same user info in the cache already
+        # Note: we're not using the "_, user_defaults=" trick here because _ is used for our translation function.
         user_defaults = self.cache.check_user_in_cache(self.token)[1]
 
         self.user = None
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -102,7 +102,7 @@ class JWTCommonAuth:
             self.token = self.validate_token(token_from_header, cert_object.key)
 
         # Let's see if we have the same user info in the cache already
-        is_cached, user_defaults = self.cache.check_user_in_cache(self.token)
+        user_defaults = self.cache.check_user_in_cache(self.token)[1]
 
         self.user = None
         try:
@@ -156,6 +156,12 @@ class JWTCommonAuth:
             old_value = getattr(self.user, attribute, None)
             new_value = self.token.get('user_data', {}).get(attribute, None)
             if old_value != new_value:
+                if attribute == "username" and get_user_model().objects.filter(username=new_value).exists():
+                    logger.warning(
+                        f"Renaming user {old_value} to {new_value} would result in a duplicate key error. "
+                        "Please make sure the sync task is running to prevent this warning in the future."
+                    )
+                    continue
                 if attribute == 'is_superuser' and new_value is False:
                     continue
                 logger.debug(f"Changing {attribute} for {self.user.username} from {old_value} to {new_value}")
@@ -164,6 +170,7 @@ class JWTCommonAuth:
         if user_needs_save:
             with no_reverse_sync():
                 logger.info(f"Saving user {self.user.username}")
+
                 self.user.save()
 
     def validate_token(self, unencrypted_token, decryption_key):
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -105,13 +105,10 @@ class JWTCommonAuth:
         is_cached, user_defaults = self.cache.check_user_in_cache(self.token)
 
         self.user = None
-        if is_cached:
-            try:
-                self.user = get_user_by_ansible_id(self.token['sub'])
-            except ObjectDoesNotExist:
-                # ooofff... I'm sorry, you user was in the cache but deleted from the database?
-                # but now you have to pay the price to continue logging in
-                pass
+        try:
+            self.user = get_user_by_ansible_id(self.token['sub'])
+        except ObjectDoesNotExist:
+            pass
 
         if not self.user:
             # Either the user wasn't cached or the requested user was not in the DB so we need to make a new one
@@ -134,6 +131,12 @@ class JWTCommonAuth:
                         username=self.token["user_data"]['username'],
                         defaults=user_defaults,
                     )
+
+                    resource = Resource.get_resource_for_object(self.user)
+
+                    resource.ansible_id = self.token['sub']
+                    resource.service_id = self.token['service_id']
+                    resource.save()
 
         setattr(self.user, "resource_api_actions", self.token.get("resource_api_actions", None))
 
@@ -159,8 +162,9 @@ class JWTCommonAuth:
                 setattr(self.user, attribute, new_value)
                 user_needs_save = True
         if user_needs_save:
-            logger.info(f"Saving user {self.user.username}")
-            self.user.save()
+            with no_reverse_sync():
+                logger.info(f"Saving user {self.user.username}")
+                self.user.save()
 
     def validate_token(self, unencrypted_token, decryption_key):
         validated_body = None
--- a/ansible_base/resource_registry/management/commands/resource_sync.py
+++ b/ansible_base/resource_registry/management/commands/resource_sync.py
@@ -19,7 +19,6 @@ Optional parameters::
 
     `--retries number` to retry failed syncs
     `--retrysleep seconds` to set interval between retries
-    `--retain_seconds` to set how much seconds to retain deleted resources
     `--asyncio` Flag to enable asyncio executor
 """
 
@@ -60,13 +59,6 @@ class Command(BaseCommand):  # pragma: no cover
             type=int,
             default=30,
             help="Interval between retries",
-            required=False,
-        )
-        parser.add_argument(
-            "--retain_seconds",
-            type=int,
-            default=120,
-            help="Seconds to retain orphan resources from deletion.",
             required=False,
         )
         parser.add_argument("--asyncio", action="store_true", default=False, help="Enable asyncio executor")
--- a/ansible_base/resource_registry/management/commands/resource_sync.py
+++ b/ansible_base/resource_registry/management/commands/resource_sync.py
@@ -65,7 +65,7 @@ class Command(BaseCommand):  # pragma: no cover
 
     def handle(self, *args, **options):
         """Handle RESOURCE_PROVIDER sync"""
-        arguments = ["resource_type_names", "retries", "retrysleep", "retain_seconds", "asyncio"]
+        arguments = ["resource_type_names", "retries", "retrysleep", "asyncio"]
         options = {k: v for k, v in options.items() if k in arguments}
         try:
             executor = SyncExecutor(**options, stdout=self.stdout)
--- a/ansible_base/resource_registry/tasks/sync.py
+++ b/ansible_base/resource_registry/tasks/sync.py
@@ -2,6 +2,7 @@ from __future__ import annotations  # support python<3.10
 
 import asyncio
 import csv
+import logging
 import time
 from collections import defaultdict
 from dataclasses import dataclass, field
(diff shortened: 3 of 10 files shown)


ansible_base/jwt_consumer/common/auth.py
```diff
@@ -105,13 +105,10 @@ def parse_jwt_token(self, request):
         is_cached, user_defaults = self.cache.check_user_in_cache(self.token)
```
BrennanPaciorek:
```suggestion
        _, user_defaults = self.cache.check_user_in_cache(self.token)
```
If we're dropping the `is_cached` check

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -317,8 +316,9 @@ def _cleanup_orphans(self, resource_type, manifest_list):
         if self.deleted_count:
             self.write(f"Deleting {self.deleted_count} orphaned resources")
             for orphan in resources_to_cleanup:
-                # If it was created in the latest X seconds, ignore it.
-                if orphan.content_object.created >= timezone.now() - timedelta(seconds=self.retain_seconds):
+                # Skip items that owned by this service, but are partially migrated
+                if orphan.service_id == service_id() and orphan.is_partially_migrated == True:
+                    # TODO: how do I log skip?
```
BrennanPaciorek: If you want to log to the application log, I'd imagine you `import logging` and create a logger for this file.

If you mean write to stderr, you'd need to do something similar to how we track stdout, defining a `stderr` field for the `SyncExecutor` dataclass, and update the `write` method.

BrennanPaciorek (commented):
After taking the time to read through it, I cannot think of an error case, so doing some dev environment testing.

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -317,8 +316,9 @@ def _cleanup_orphans(self, resource_type, manifest_list):
         if self.deleted_count:
             self.write(f"Deleting {self.deleted_count} orphaned resources")
             for orphan in resources_to_cleanup:
-                # If it was created in the latest X seconds, ignore it.
-                if orphan.content_object.created >= timezone.now() - timedelta(seconds=self.retain_seconds):
+                # Skip items that owned by this service, but are partially migrated
+                if orphan.service_id == service_id() and orphan.is_partially_migrated == True:
```
trucdg: case: is_partially_migrated == False
- Can items owned by a service, but hasn't been migrated to Gateway, be deleted? 
- Do gateway items have is_partially_migrated field? If yes, I assume is_partially_migrated in gateway is always False. This means that we always delete Gateway items if they're not in the manifest?

The old code had a check to avoid deleting newly created items, that's why I asked.

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -317,8 +316,9 @@ def _cleanup_orphans(self, resource_type, manifest_list):
         if self.deleted_count:
             self.write(f"Deleting {self.deleted_count} orphaned resources")
             for orphan in resources_to_cleanup:
```
trucdg, line 318: If we skip some resources below, the number of deleted resources should = resources_to_cleanup.count() - number of skipped resources.

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -317,8 +316,9 @@ def _cleanup_orphans(self, resource_type, manifest_list):
         if self.deleted_count:
             self.write(f"Deleting {self.deleted_count} orphaned resources")
```
trucdg, line 336: If we skip some resources below, the number of deleted resources should = resources_to_cleanup.count() - number of skipped resources.

ansible_base/resource_registry/tasks/sync.py
newswangerd: I'm removing the time check because Hub doesn't have a created field for users :/

We do have `is_partially_migrated` in gateway, but it's not really used for anything.

ansible_base/resource_registry/tasks/sync.py
newswangerd: I ended up filtering out the skipped items in the queryset, so the counts should all line up nicely now!

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -153,14 +156,22 @@ def map_user_fields(self):
...
+                    continue
                 if attribute == 'is_superuser' and new_value is False:
                     continue
                 logger.debug(f"Changing {attribute} for {self.user.username} from {old_value} to {new_value}")
                 setattr(self.user, attribute, new_value)
                 user_needs_save = True
         if user_needs_save:
-            logger.info(f"Saving user {self.user.username}")
-            self.user.save()
+            with no_reverse_sync():
```
john-westcott-iv, line 172: Is the `no_reverse_sync` here just for login performance so we aren't pushing back to gateway with what just came down from the JWT?

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -102,16 +102,13 @@ def parse_jwt_token(self, request):
             self.token = self.validate_token(token_from_header, cert_object.key)
 
         # Let's see if we have the same user info in the cache already
-        is_cached, user_defaults = self.cache.check_user_in_cache(self.token)
+        user_defaults = self.cache.check_user_in_cache(self.token)[1]
```
tznamena, line 106: A comment might be helpful here to explain the hard-coded index of 1.

ansible_base/resource_registry/management/commands/resource_sync.py
```diff
@@ -62,13 +61,6 @@ def add_arguments(self, parser):
             help="Interval between retries",
             required=False,
         )
-        parser.add_argument(
```
john-westcott-iv, line 65: If we are going to get rid of this, can you also remove it from the arguments on line 76?

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -89,9 +91,8 @@ def fetch_manifest(
 
     resp_metadata = api_client.get_service_metadata()
     resp_metadata.raise_for_status()
-    service_id = resp_metadata.json()["service_id"]
 
-    manifest_stream = api_client.get_resource_type_manifest(resource_type_name)
+    manifest_stream = api_client.get_resource_type_manifest(resource_type_name, filters={"service_id": str(service_id())})
```
john-westcott-iv: Do we have to wrap service_id in str? It looks like the method does it for us already.

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -102,36 +103,41 @@ def fetch_manifest(
...
 def get_orphan_resources(
     resource_type_name: str,
     manifest_list: list[ManifestItem],
 ) -> QuerySet:
     """QuerySet with orphaned managed resources to be deleted."""
-    return Resource.objects.filter(
-        service_id=manifest_list[0].service_id,
-        content_type__resource_type__name=resource_type_name,
-    ).exclude(ansible_id__in=[item.ansible_id for item in manifest_list])
+    return (
```
john-westcott-iv, line 114: Do you need this extra paren here?

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -317,9 +335,9 @@ def _cleanup_orphans(self, resource_type, manifest_list):
         if self.deleted_count:
             self.write(f"Deleting {self.deleted_count} orphaned resources")
             for orphan in resources_to_cleanup:
-                # If it was created in the latest X seconds, ignore it.
-                if orphan.content_object.created >= timezone.now() - timedelta(seconds=self.retain_seconds):
-                    continue
+                # Skip items that owned by this service, but are partially migrated
+                # if orphan.service_id == service_id() and orphan.is_partially_migrated is True:
+                #     continue
```
john-westcott-iv: Can we delete these lines if we're not using them anymore?

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -135,6 +132,12 @@ def parse_jwt_token(self, request):
                         defaults=user_defaults,
                     )
 
+                    resource = Resource.get_resource_for_object(self.user)
```
tznamena, line 136: With this code, the method name no longer matches the implementation, it is no longer just `parse_jwt_token` it is more like `parse_jwt_token_and_create_users_if_needed`. Can this be split up somehow?

test_app/tests/resource_registry/test_resource_sync.py
```diff
@@ -114,8 +123,56 @@ def test_noop_existing_resource(admin_api_client, static_api_client, stdout):
...
+            "username": "theceo",
+            "email": "theceo@other-email.com",
+            "first_name": "A Different",
+            "last_name": "Other Name",
+        },
+    }
+    response = admin_api_client.post(url, resource, format="json")
+    assert response.status_code == 201
+
+    with mock.patch("ansible_base.resource_registry.models.resource.Resource.update_resource", side_effect=Error("Something went wrong")):
```
tznamena, line 143: Just wondering why we need to setup the resource above when all we testing below is that we see an error in std out and that error is injected into `update_resource`.

ansible_base/jwt_consumer/common/auth.py
newswangerd: Exactly

ansible_base/resource_registry/tasks/sync.py
newswangerd: Probably not, but I've been burned by this so many times, that I just do it reflexively now

ansible_base/resource_registry/tasks/sync.py
newswangerd: Yes. That's what allows you to break the qs into multiple lines in a readable manner.

ansible_base/jwt_consumer/common/auth.py
newswangerd: I think that's probably the right thing to do, but I'm not going to do it because this is a library that gets imported into lots of different places, and I don't want to accidentally break something else. We have most of the integrations in DAB, but there are other projects that are using this now, and I don't want to risk breaking changes.

test_app/tests/resource_registry/test_resource_sync.py
newswangerd: We can only trigger the update, create and delete methods in the sync task by actually making changes that will get cause one of those operations to fire. In this case I'm lazy so I just copied ones that Bruno had in the other tests, but I suppose a fixture would make more sense.

newswangerd: I've fix(tured) it :D

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -135,6 +133,12 @@ def parse_jwt_token(self, request):
...
+
+                    resource.ansible_id = self.token['sub']
+                    resource.service_id = self.token['service_id']
+                    resource.save()
```
john-westcott-iv: Do you want to specify `update_fields` on this save? Just thinking we know exactly what fields we need to change here so if there is any kind of race condition with the sync process this might help.
```suggestion
                    resource.save(update_fields=['ansible_id', 'service_id'])
```

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -153,14 +157,22 @@ def map_user_fields(self):
...
                 logger.debug(f"Changing {attribute} for {self.user.username} from {old_value} to {new_value}")
                 setattr(self.user, attribute, new_value)
                 user_needs_save = True
         if user_needs_save:
-            logger.info(f"Saving user {self.user.username}")
-            self.user.save()
+            with no_reverse_sync():
+                logger.info(f"Saving user {self.user.username}")
+
+                self.user.save()
```
john-westcott-iv, line 174: We could do the same update_fields here as well by making user_needs_saves an array instead of a bool.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -153,14 +157,22 @@ def map_user_fields(self):
             old_value = getattr(self.user, attribute, None)
             new_value = self.token.get('user_data', {}).get(attribute, None)
             if old_value != new_value:
+                if attribute == "username" and get_user_model().objects.filter(username=new_value).exists():
```
john-westcott-iv, line 160: Would it be faster to not check here and just try/catch the save below and drop this error on a duplicate key error? Although we would have to try the save a second time so maybe not? But it should be easy to recover if we used the update_fields and just droped the username on a duplicate key error.

john-westcott-iv (approved):

ansible_base/jwt_consumer/common/auth.py
newswangerd: I think this is fine. We only trigger the second conditional if the first is false, and I don't expect usernames to change very often.

ansible_base/jwt_consumer/common/auth.py
newswangerd: I'd rather not mess with that. This is how it was before and we know that worked.

zkayyali812:
Noting that this fix is also expected to address - https://issues.redhat.com/browse/AAP-40523

PR #700, opened 2025-02-20
AAP-39202 Fix update for SAML keys

bhavenst:
## Description
Fixes a bug that prevented the public cert from being updated for SAML authenticators.

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Deploy gateway with keycloak/saml enabled 
2. Generate a new certificate/keypair
3. Update the SAML authenticator with the cert and private key
4. See that the update is accepted (was always blocked with error before)
5. If desired, configure the new key in keycloak and log in with an SSO user

### Expected Results
SAML authenticators can have their key/cert updated successfully

### Required Actions
- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/.github/workflows/dab-consumers.yml
+++ b/.github/workflows/dab-consumers.yml
@@ -28,6 +28,7 @@ jobs:
           FROM ghcr.io/ansible/awx_devel:devel
           COPY . /opt/dab/
           RUN pip install -e /opt/dab/
+          RUN pip install jq
           EOF
           docker build -t awx-with-dab:latest -f awx-dockerfile/Dockerfile .
 
--- a/ansible_base/authentication/authenticator_plugins/saml.py
+++ b/ansible_base/authentication/authenticator_plugins/saml.py
@@ -162,10 +162,10 @@ class SAMLConfiguration(BaseAuthenticatorConfiguration):
     def validate(self, attrs):
         # attrs is only the data in the configuration field
         errors = {}
-        # pull the cert_info out of the existing object (if we have one)
+        # pull the cert_info out of provided attrs or the existing object (if we have one)
         cert_info = {
             "SP_PRIVATE_KEY": getattr(self.instance, 'configuration', {}).get('SP_PRIVATE_KEY', None),
-            "SP_PUBLIC_CERT": getattr(self.instance, 'configuration', {}).get('SP_PUBLIC_CERT', attrs.get('SP_PUBLIC_CERT', None)),
+            "SP_PUBLIC_CERT": attrs.get('SP_PUBLIC_CERT', getattr(self.instance, 'configuration', {}).get('SP_PUBLIC_CERT', None)),
         }
 
         # Now get the SP_PRIVATE_KEY out of the passed in attrs (if there is any)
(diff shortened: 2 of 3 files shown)


ansible_base/authentication/authenticator_plugins/saml.py
```diff
@@ -162,10 +162,10 @@ class SAMLConfiguration(BaseAuthenticatorConfiguration):
     def validate(self, attrs):
         # attrs is only the data in the configuration field
         errors = {}
-        # pull the cert_info out of the existing object (if we have one)
+        # pull the cert_info out of provided attrs or the existing object (if we have one)
```
tznamena: I wonder if it would be cleaner or more obvious if we were to split this up into two parts:
First, pull info from configuration:
```
# create a cert_info dict from existing configuration
cert_info = {
            "SP_PRIVATE_KEY": getattr(self.instance, 'configuration', {}).get('SP_PRIVATE_KEY', None),
            "SP_PUBLIC_CERT": getattr(self.instance, 'configuration', {}).get('SP_PUBLIC_CERT', None)
        }
```
Then, update with data from attributes:
```
public_cert = attrs.get('SP_PUBLIC_CERT', None)
if public_cert is not None:
        cert_info['SP_PUBLIC_CERT'] = public_cert
```

bhavenst: It could be done like that.  My understanding is that it was done like that for the private key to deal with the encrypted/unencrypted state, otherwise both would have been one-liners.  But I'm fine with changing it unless that will jeopardize getting this into the release.

bhavenst: OK, updated.

BrennanPaciorek:
I have confirmed the erroneous behavior and that said behavior has been fixed.

In unpatched code:  `[ErrorDetail(string='The certificate and private key do not match', code='invalid')]`
In patched code: Authenticator config successfully updates with the provided keypair.

BrennanPaciorek (approved):

.github/workflows/dab-consumers.yml
```diff
@@ -28,6 +28,7 @@ jobs:
           FROM ghcr.io/ansible/awx_devel:devel
           COPY . /opt/dab/
           RUN pip install -e /opt/dab/
+          RUN pip install jq
```
john-westcott-iv, line 31: Is this still necessary or was it just part of your testing?

bhavenst: Not related to my change, just trying to get the awx job to pass..

john-westcott-iv (approved):

.github/workflows/dab-consumers.yml
BrennanPaciorek, line 31: Bugs me that this was merged still despite it not doing what it intended to do.

Reason it did not work though is because it probably sources a venv to run the tests. Not sure when the venv is created and installs its dependencies though.

https://github.com/ansible/django-ansible-base/actions/runs/13504381487/job/37730418006#step:6:41

bhavenst: It did work as far as fixing the test run..  They all ran but one test failed.

BrennanPaciorek: Oh in that case i probably messed up in tab management. Sorry about this one.

PR #703, opened 2025-02-28
Resource sync fixes: org and team edition.

newswangerd:
## Description
Fixes for:
- AAP-38123: Update sync task to try and handle conflicts.
- AAP-32044: Gracefully handle duplicate name issues for orgs and teams during login

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Create an org with a random ansible ID in Gateway using the name "Conflict"
2. Create an org in any aservice with a randome ansible ID using the name "Conflict"
3. Add a user to the org in Gateway
4. Attempt to login to the service

### Expected Results

You should be able to login no problem, but you won't be marked as a member of the organization until the sync runs.

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -298,8 +298,11 @@ class JWTCommonAuth:
                 object_data = self.token['objects'][object_type][index]
                 try:
                     resource, obj = self.get_or_create_resource(object_type, object_data)
-                except IntegrityError:
-                    logger.warning(f"Got integrity error on {object_data}. Skipping {object_type} assignment.")
+                except IntegrityError as e:
+                    logger.warning(
+                        f"Got integrity error ({e}) on {object_data}. Skipping {object_type} assignment. "
+                        "Please make sure the sync task is running to prevent this warning in the future."
+                    )
                     continue
 
                 if resource is not None:
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -296,7 +296,12 @@ class JWTCommonAuth:
 
             for index in object_indexes:
                 object_data = self.token['objects'][object_type][index]
-                resource, obj = self.get_or_create_resource(object_type, object_data)
+                try:
+                    resource, obj = self.get_or_create_resource(object_type, object_data)
+                except IntegrityError:
+                    logger.warning(f"Got integrity error on {object_data}. Skipping {object_type} assignment.")
+                    continue
+
                 if resource is not None:
                     assignment = rd.give_permission(self.user, obj)
                     role_diff = role_diff.exclude(pk=assignment.pk)
--- a/ansible_base/jwt_consumer/hub/auth.py
+++ b/ansible_base/jwt_consumer/hub/auth.py
@@ -1,6 +1,7 @@
 import logging
 
 from django.contrib.contenttypes.models import ContentType
+from django.db.models.query import IntegrityError
 
 from ansible_base.jwt_consumer.common.auth import JWTAuthentication
 from ansible_base.jwt_consumer.common.exceptions import InvalidService
@@ -47,7 +48,11 @@ class HubJWTAuth(JWTAuthentication):
                     try:
                         team = Resource.objects.get(ansible_id=ansible_id).content_object
                     except Resource.DoesNotExist:
-                        team = self.common_auth.get_or_create_resource('team', team_data)[1]
+                        try:
+                            team = self.common_auth.get_or_create_resource('team', team_data)[1]
+                        except IntegrityError:
+                            logger.warning(f"Got integrity error on {team_data}. Skipping team assignment.")
+                            continue
 
                     if role_name == 'Team Admin':
                         admin_teams.append(team)
--- a/ansible_base/jwt_consumer/hub/auth.py
+++ b/ansible_base/jwt_consumer/hub/auth.py
@@ -50,8 +50,11 @@ class HubJWTAuth(JWTAuthentication):
                     except Resource.DoesNotExist:
                         try:
                             team = self.common_auth.get_or_create_resource('team', team_data)[1]
-                        except IntegrityError:
-                            logger.warning(f"Got integrity error on {team_data}. Skipping team assignment.")
+                        except IntegrityError as e:
+                            logger.warning(
+                                f"Got integrity error ({e}) on {team_data}. Skipping team assignment. "
+                                "Please make sure the sync task is running to prevent this warning in the future."
+                            )
                             continue
 
                     if role_name == 'Team Admin':
--- a/ansible_base/resource_registry/models/resource.py
+++ b/ansible_base/resource_registry/models/resource.py
@@ -50,7 +50,7 @@ class ResourceType(models.Model):
         qfilter = {}
 
         if not self.can_be_managed:
-            raise Exception(f"Resource type {self.name} does not have a managed serializer.")
+            raise UnamangedResourceException(f"Resource type {self.name} does not have a managed serializer.")
 
         serializer = self.serializer_class(data=resource_data)
         serializer.is_valid(raise_exception=True)
--- a/ansible_base/resource_registry/models/resource.py
+++ b/ansible_base/resource_registry/models/resource.py
@@ -4,6 +4,7 @@ from typing import Union
 
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ObjectDoesNotExist
 from django.db import models, transaction
 from django.utils.translation import gettext_lazy as _
 from rest_framework.serializers import ValidationError
@@ -40,6 +41,21 @@ class ResourceType(models.Model):
     def get_resource_config(self):
         return self.resource_registry.get_config_for_model(model=ContentType.objects.get_for_id(self.content_type_id).model_class())
 
+    def get_conflicting_resource(self, resource_data):
+        filter = {}
+
+        serialized_data = self.serializer_class(data=resource_data)
+        serialized_data.is_valid(raise_exception=True)
+        serialized_data = serialized_data.validated_data
+
+        for field in self.serializer_class.UNIQUE_FIELDS:
+            filter[field] = serialized_data[field]
+
+        try:
+            return Resource.get_resource_for_object(self.content_type.get_object_for_this_type(**filter))
+        except ObjectDoesNotExist:
+            return None
+
 
 class Resource(models.Model):
     content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE, related_name="resources", help_text=_("The content type for this resource."))
@@ -72,6 +88,10 @@ class Resource(models.Model):
     @property
     def resource_type(self):
         return resource_type_cache(self.content_type.pk).name
+
+    @property
+    def resource_type_obj(self):
+        return resource_type_cache(self.content_type.pk)
 
     class Meta:
         unique_together = ('content_type', 'object_id')
--- a/ansible_base/resource_registry/models/resource.py
+++ b/ansible_base/resource_registry/models/resource.py
@@ -17,6 +17,10 @@ def resource_type_cache(content_type_id):
     return ContentType.objects.get_for_id(content_type_id).resource_type
 
 
+class UnamangedResourceException(Exception):
+    pass
+
+
 class ResourceType(models.Model):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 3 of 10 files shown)


ansible_base/jwt_consumer/common/auth.py
```diff
@@ -296,7 +296,12 @@ def process_rbac_permissions(self):
 
             for index in object_indexes:
                 object_data = self.token['objects'][object_type][index]
-                resource, obj = self.get_or_create_resource(object_type, object_data)
+                try:
+                    resource, obj = self.get_or_create_resource(object_type, object_data)
+                except IntegrityError:
+                    logger.warning(f"Got integrity error on {object_data}. Skipping {object_type} assignment.")
```
john-westcott-iv: Should we also split out what the `IntegrityError` actually is or will it always be something like `duplicate key`?

john-westcott-iv: Also, if I saw this I'd be like why is this happening, how do I fix this? Should we add something like "This will be resolved in the next periodic sync" or something along those lines?

ansible_base/jwt_consumer/hub/auth.py
```diff
@@ -47,7 +48,11 @@ def process_permissions(self):
                     try:
                         team = Resource.objects.get(ansible_id=ansible_id).content_object
                     except Resource.DoesNotExist:
-                        team = self.common_auth.get_or_create_resource('team', team_data)[1]
+                        try:
+                            team = self.common_auth.get_or_create_resource('team', team_data)[1]
+                        except IntegrityError:
+                            logger.warning(f"Got integrity error on {team_data}. Skipping team assignment.")
```
john-westcott-iv: Same comment on this log about should we dump the error and/or give the user more details on how to resolve.

ansible_base/resource_registry/models/resource.py
```diff
@@ -40,6 +41,21 @@ def can_be_managed(self):
     def get_resource_config(self):
         return self.resource_registry.get_config_for_model(model=ContentType.objects.get_for_id(self.content_type_id).model_class())
 
+    def get_conflicting_resource(self, resource_data):
+        filter = {}
+
+        serialized_data = self.serializer_class(data=resource_data)
+        serialized_data.is_valid(raise_exception=True)
+        serialized_data = serialized_data.validated_data
```
john-westcott-iv: Do we ever have to consider a case in here where the data isn't valid like the ones with the user IDs and wrap this in a try/catch to prevent an exception?

ansible_base/resource_registry/models/resource.py
```diff
@@ -40,6 +41,21 @@ def can_be_managed(self):
     def get_resource_config(self):
         return self.resource_registry.get_config_for_model(model=ContentType.objects.get_for_id(self.content_type_id).model_class())
 
+    def get_conflicting_resource(self, resource_data):
+        filter = {}
+
+        serialized_data = self.serializer_class(data=resource_data)
+        serialized_data.is_valid(raise_exception=True)
```
john-westcott-iv: Should we rename these to `serializer` or something else instead of reassigning to it below for clarity?

ansible_base/resource_registry/models/resource.py
```diff
@@ -40,6 +41,21 @@ def can_be_managed(self):
     def get_resource_config(self):
         return self.resource_registry.get_config_for_model(model=ContentType.objects.get_for_id(self.content_type_id).model_class())
 
+    def get_conflicting_resource(self, resource_data):
+        filter = {}
+
+        serialized_data = self.serializer_class(data=resource_data)
```
john-westcott-iv: Do you need to check `can_be_manged` to confirm there is a serializer here? Or does something else do that before calling this method?

ansible_base/resource_registry/models/resource.py
```diff
@@ -40,6 +41,21 @@ def can_be_managed(self):
...
+        serialized_data.is_valid(raise_exception=True)
+        serialized_data = serialized_data.validated_data
+
+        for field in self.serializer_class.UNIQUE_FIELDS:
+            filter[field] = serialized_data[field]
+
+        try:
+            return Resource.get_resource_for_object(self.content_type.get_object_for_this_type(**filter))
```
john-westcott-iv: Any possibility that we could get a different type of exception here? Or if we do raising is ok?

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -148,22 +148,105 @@ def get_resource_type_names() -> list[str]:
...
+
+    # If the resource does exist, lets update it first. Hopefully this desn't also result
+    # in a duplicate key error. If it does, we're cooked.
+    elif resp.status_code == 200:
+        data = resp.json()
+        conflict_resource.update_resource(
+            resource_data=data["resource_data"], service_id=data["service_id"], is_partially_migrated=data["is_partially_migrated"]
+        )
+    else:
+        raise ResourceSyncHTTPError
```
john-westcott-iv: Should we give some details in this exception as I imagine it would end up getting logged somewhere?

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -148,22 +148,105 @@ def get_resource_type_names() -> list[str]:
...
+    # If the conflicting resource doesn't exist on the server, just go ahead and delete it.
+    # This will most likely happen with resources that are ignored by get_orphan_resources.
+    if resp.status_code == 404:
+        delete_resource(conflict_resource)
+
+    # If the resource does exist, lets update it first. Hopefully this desn't also result
+    # in a duplicate key error. If it does, we're cooked.
+    elif resp.status_code == 200:
+        data = resp.json()
+        conflict_resource.update_resource(
```
john-westcott-iv, line 179: Do we want to try/catch here or otherwise validate that we succeeded?

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -148,22 +148,105 @@ def get_resource_type_names() -> list[str]:
...
     resource_data: dict,
+    api_client: ResourceAPIClient,
     **kwargs,
 ) -> SyncResult:
     """Try to update existing resource."""
     try:
         resource.update_resource(resource_data, partial=True, **kwargs)
     except IntegrityError:  # pragma: no cover
-        return SyncResult(SyncStatus.CONFLICT, manifest_item)
+        # This typically means that there was a duplicate key error. To mitigate this
```
john-westcott-iv, line 197: Is there any chance it wouldn't be a duplicate key and we could up trying to _handle_conflict for something else?

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -148,22 +148,105 @@ def get_resource_type_names() -> list[str]:
...
+        # we will attempt to hanlde the conflicting resource and perform the operation
+        # again.
+        try:
+            _handle_conflict(resource_data, resource.resource_type_obj, api_client)
+            resource.update_resource(resource_data, partial=True, **kwargs)
+        except (ResourceDeletionError, IntegrityError) as e:
+            logger.error(f"Failed to gracefully handle conflict for {resource_data}. Got error {e}.")
+            return SyncResult(SyncStatus.CONFLICT, manifest_item)
```
john-westcott-iv, line 205: Would this def be a CONFLICT and not potentally an an actual ERROR?

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -148,22 +148,105 @@ def get_resource_type_names() -> list[str]:
...
     except Error as e:
+        # Something happened with the database. We don't know what it is. Instead of failing the whole
+        # sync, we'll raise an error and skip this for now.
         logger.error(f"Failed to update resource {resource.ansible_id}. Received error: {e}")
```
john-westcott-iv: Can we alter this message to say something like "Will try again on the nest sync"?

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -148,22 +148,105 @@ def get_resource_type_names() -> list[str]:
...
+    resource_service_id: str,
+    api_client: ResourceAPIClient,
+) -> SyncResult:
+    try:
+        Resource.create_resource(
+            resource_type=resource_type,
+            resource_data=resource_data,
+            ansible_id=manifest_item.ansible_id,
+            service_id=resource_service_id,
+        )
```
john-westcott-iv, line 228: The pattern of try and try again is duplicated. It would be really sweet if we could figure out a way of abstracting this so we could work on the "try again" logic in one place instead of two.

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -148,22 +148,105 @@ def get_resource_type_names() -> list[str]:
...
+            _handle_conflict(resource_data, resource_type, api_client)
+            Resource.create_resource(
+                resource_type=resource_type,
+                resource_data=resource_data,
+                ansible_id=manifest_item.ansible_id,
+                service_id=resource_service_id,
+            )
+        except (ResourceDeletionError, IntegrityError) as e:
+            logger.error(f"Failed to gracefully handle conflict for {resource_data}. Got error {e}.")
+            return SyncResult(SyncStatus.CONFLICT, manifest_item)
```
john-westcott-iv, line 243: Same comments about updating the exception with "Will try again" or whatever and also, is this def guaranteed a conflict here?

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -148,22 +148,105 @@ def get_resource_type_names() -> list[str]:
...
+    except Error as e:
+        # Something happened with the database. We don't know what it is. Instead of failing the whole
+        # sync, we'll raise an error and skip this for now.
+        logger.error(f"Failed to update create {manifest_item.ansible_id}. Received error: {e}")
```
john-westcott-iv: `update create` is that correct?

ansible_base/jwt_consumer/common/auth.py
newswangerd: I think it will always be duplicate key, but I added the error message to the log just in case.

ansible_base/resource_registry/models/resource.py
newswangerd: I think I prefer to let the caller catch this one. I'm not sure what we'd do if it was not valid. We'd just have to raise another exception.

ansible_base/resource_registry/models/resource.py
newswangerd: Potentially "too many objects returned", but if that's the case, then the model is misconfigured and I'd rather raise it.

ansible_base/resource_registry/tasks/sync.py
newswangerd: If this does raise an integrity error, or any other DB related error it will get caught in `_attempt_create_resource` or `_attempt_update_resource`.

ansible_base/resource_registry/tasks/sync.py
newswangerd: Maybe... It's an error, but it is caused by a conflict. I think I'd leave it at conflict.

ansible_base/resource_registry/tasks/sync.py
newswangerd: I thought about making this function recursive, but that got complicated very quickly. This repeats itself, but it's much more understandable.

ansible_base/resource_registry/tasks/sync.py
newswangerd: I'm not sure. I don't know how to tell.

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -148,22 +149,105 @@ def get_resource_type_names() -> list[str]:
...
+    #   incredibly difficult to pull off.
+
+    conflict_resource = resource_type.get_conflicting_resource(resource_data)
+
+    if conflict_resource is None:
+        return
+
+    resp = api_client.get_resource(conflict_resource.ansible_id)
+
+    # If the conflicting resource doesn't exist on the server, just go ahead and delete it.
```
tznamena, line 170: How are we deleting a resource that does not exist?

newswangerd: It doesn't exist on the resource server, but it does exist on the service where the task is running.

bhavenst: If you just delete it and sync over a new version, might there be loss of assigned roles?

bhavenst (approved):
I don't see any glaring issues and John already gave lots of feedback.

tznamena (approved):

PR #707, opened 2025-03-26
[AAP-41963] Fix SonarCloud security hotspots

trucdg:
# [AAP-41963] Fix SonarCloud security hotspots

## Description

- What is being changed?
Dockerfile to build image for dab postgres service and ingress service
- Why is this change needed?
Currently, the `postgres` and `ingress` containers are run as root, about which SonarCloud is complaining.
1. in `tools/dev_postgres/Dockerfile`: add `USER` directive to run the container as `postgres` non-root user; fix `$${} ` error for environment variables expansion
2. in `compose/ingress/Dockerfile`: modify the `nginx.conf` file to add appropriate configurations for non-root nginx user (reference: https://hub.docker.com/_/nginx), and add `USER` directive
## Type of Change

- [ ] Documentation update
- [ ] Test update
- [x] Development environment change
- [x] Configuration change

## Self-Review Checklist

### Steps to Test
1. remove existing `django-ansible-base-postgres`, `nginx` and `django-ansible-base-ingress` container images in your local
2. in `django-ansible-base` dir, run command `make docker compose build`, then run `make docker compose up`
3. check the logs of all containers and confirm that the services are running fine.

### Expected Results

- All services in dab should run fine.
- `django-ansible-base-ingress-1` container is run as `nginx` user.
- `django-ansible-base-nginx-1` container is run as `nginx` user.
- `django-ansible-base-postgres` container is run as `postgres` user.

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/compose/ingress/Dockerfile
+++ b/compose/ingress/Dockerfile
@@ -2,3 +2,17 @@ FROM mirror.gcr.io/library/nginx:latest
 
 COPY entrypoint.sh /entrypoint.sh
 COPY nginx.conf /etc/nginx/conf.d/default.conf
+
+# Modify the default Nginx configuration to allow it to run as a non-root user
+# This follows Nginx Docker Hub instructions: https://hub.docker.com/_/nginx
+# By default, Nginx writes its PID file to /var/run/nginx.pid, which is a restricted location.
+# We change it to /tmp/nginx.pid so that it can be accessed by a non-root user.
+RUN sed -i 's|pid        /var/run/nginx.pid;|pid        /tmp/nginx.pid;|' /etc/nginx/nginx.conf && \
+    sed -i '/http {/a \
+    client_body_temp_path /tmp/client_temp;\n\
+    proxy_temp_path       /tmp/proxy_temp;\n\
+    fastcgi_temp_path     /tmp/fastcgi_temp;\n\
+    uwsgi_temp_path       /tmp/uwsgi_temp;\n\
+    scgi_temp_path        /tmp/scgi_temp;' /etc/nginx/nginx.conf
+
+USER nginx
--- a/compose/ingress/Dockerfile
+++ b/compose/ingress/Dockerfile
@@ -10,8 +10,8 @@ COPY nginx.conf /etc/nginx/conf.d/default.conf
 # This follows Nginx Docker Hub instructions: https://hub.docker.com/_/nginx
 # By default, Nginx writes its PID file to /var/run/nginx.pid, which is a restricted location.
 # We change it to /tmp/nginx.pid so that it can be accessed by a non-root user.
-RUN sed -i '/^user  nginx;/d' /etc/nginx/nginx.conf && \
-    sed -i 's|pid        /var/run/nginx.pid;|pid        /tmp/nginx.pid;|' /etc/nginx/nginx.conf && \
+RUN sed -i -E '/^user\s+nginx;/d' /etc/nginx/nginx.conf && \
+    sed -i -E 's|pid\s+/var/run/nginx.pid;|pid        /tmp/nginx.pid;|' /etc/nginx/nginx.conf && \
     sed -i '/http {/a \
     client_body_temp_path /tmp/client_temp;\n\
     proxy_temp_path       /tmp/proxy_temp;\n\
--- a/compose/ingress/Dockerfile
+++ b/compose/ingress/Dockerfile
@@ -1,6 +1,6 @@
 # Dockerfile to build the image for dab_ingress service
 
-FROM mirror.gcr.io/library/nginx:latest
+FROM mirror.gcr.io/library/nginx:1.27
 
 RUN mkdir -p /etc/nginx/ssl && chown -R nginx:nginx /etc/nginx/ssl
 COPY entrypoint.sh /entrypoint.sh
--- a/compose/nginx/Dockerfile
+++ b/compose/nginx/Dockerfile
@@ -1,9 +1,7 @@
-# Dockerfile to build the image for dab_ingress service
+# Dockerfile to build the image for dab_nginx service
 
 FROM mirror.gcr.io/library/nginx:latest
 
-RUN mkdir -p /etc/nginx/ssl && chown -R nginx:nginx /etc/nginx/ssl
-COPY entrypoint.sh /entrypoint.sh
 COPY nginx.conf /etc/nginx/conf.d/default.conf
 
 # Modify the default Nginx configuration to allow it to run as a non-root user
--- a/compose/nginx/Dockerfile
+++ b/compose/nginx/Dockerfile
@@ -1,6 +1,6 @@
 # Dockerfile to build the image for dab_nginx service
 
-FROM mirror.gcr.io/library/nginx:latest
+FROM mirror.gcr.io/library/nginx:1.27
 
 COPY nginx.conf /etc/nginx/conf.d/default.conf
 
--- a/compose/nginx/Dockerfile
+++ b/compose/nginx/Dockerfile
@@ -8,8 +8,8 @@ COPY nginx.conf /etc/nginx/conf.d/default.conf
 # This follows Nginx Docker Hub instructions: https://hub.docker.com/_/nginx
 # By default, Nginx writes its PID file to /var/run/nginx.pid, which is a restricted location.
 # We change it to /tmp/nginx.pid so that it can be accessed by a non-root user.
-RUN sed -i '/^user  nginx;/d' /etc/nginx/nginx.conf && \
-    sed -i 's|pid        /var/run/nginx.pid;|pid        /tmp/nginx.pid;|' /etc/nginx/nginx.conf && \
+RUN sed -i -E '/^user\s+nginx;/d' /etc/nginx/nginx.conf && \
+    sed -i -E 's|pid\s+/var/run/nginx.pid;|pid        /tmp/nginx.pid;|' /etc/nginx/nginx.conf && \
     sed -i '/http {/a \
     client_body_temp_path /tmp/client_temp;\n\
     proxy_temp_path       /tmp/proxy_temp;\n\
--- a/tools/dev_postgres/Dockerfile
+++ b/tools/dev_postgres/Dockerfile
@@ -4,6 +4,9 @@ ENV POSTGRES_DB=dab_db
 ENV POSTGRES_USER=dab
 ENV POSTGRES_PASSWORD=dabing
 
+USER postgres
+
 EXPOSE 5432
 
-HEALTHCHECK --interval=10s --timeout=5s --retries=5 CMD ["pg_isready", "-U", "$${POSTGRES_USER}", "-d", "$${POSTGRES_DB}"]
+HEALTHCHECK --interval=10s --timeout=5s --retries=5 CMD sh -c 'pg_isready -U "$POSTGRES_USER" -d "$POSTGRES_DB"'
+
--- a/tools/dev_postgres/Dockerfile
+++ b/tools/dev_postgres/Dockerfile
@@ -9,5 +9,3 @@ USER postgres
 EXPOSE 5432
 
 HEALTHCHECK --interval=10s --timeout=5s --retries=5 CMD ["sh", "-c", "pg_isready -U \"$POSTGRES_USER\" -d \"$POSTGRES_DB\""]
-
-
--- a/tools/dev_postgres/Dockerfile
+++ b/tools/dev_postgres/Dockerfile
@@ -8,5 +8,6 @@ USER postgres
 
 EXPOSE 5432
 
-HEALTHCHECK --interval=10s --timeout=5s --retries=5 CMD sh -c 'pg_isready -U "$POSTGRES_USER" -d "$POSTGRES_DB"'
+HEALTHCHECK --interval=10s --timeout=5s --retries=5 CMD ["sh", "-c", "pg_isready -U \"$POSTGRES_USER\" -d \"$POSTGRES_DB\""]
+
 
--- a/compose/README.md
+++ b/compose/README.md
@@ -1,0 +1,4 @@
+## `compose` Directory
+This directory contains the Dockerfile and configuration files for Nginx and Ingress services used in the `test_app`.
+At the moment, this code is only for testing and development purposes, and is not used in production.
+
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,7 @@ services:
   # This is the intermediate application reverse proxy without ssl
   nginx:
     build: ./compose/nginx
-    image: "dab_nginx:latest"
+    image: "dab_nginx:1.27"
     ports:
       - '80:80'
     depends_on:
@@ -46,7 +46,7 @@ services:
   # This is the ssl terminated "ingress" reverse proxy
   ingress:
     build: ./compose/ingress
-    image: "dab_ingress:latest"
+    image: "dab_ingress:1.27"
     command: './entrypoint.sh'
     ports:
       - "443:443"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,9 +36,8 @@ services:
 
   # This is the intermediate application reverse proxy without ssl
   nginx:
-    image: "nginx:latest"
-    volumes:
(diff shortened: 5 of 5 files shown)


bhavenst:
Hmm..  I could run this fine, but I do not see the nginx configuration applied in the container.  It's still in /var/run and root owned/running and I don't see the content in nginx.conf..

tznamena:
There are several issues here, but before we get into those, a bit of context (for other reviewers as well). 
The issue reported in SonarCloud is that it figured out that the container image used (in `FROM` instruction) does not change a user to a non-root user. Therefore, this PR is attempting to make the image use `nginx` user that the official image does include. However, there are further complications if we really want to do that.

The change `compose/ingress/Dockerfile` is fine for running nginx by itself, however, that's not enough for this setup: 
1. the `docker-compose.yml` file that orchestrates all the containers is telling the image to run a script `./entrypoint.sh` at the start of the container and that script attempts to generate the SSL certificate (with its private key).
2. the directory where the certificate and key are to be stored is a local directory that is mounted into the image. This is where the real problems start. (this can be different on different machines, like Linux or Macs) The main issue is the ownership of that directory once it is mounted into the container. It will be owned by `root` user **HOWEVER** our image is now running as `nginx` user, including the startup script so we can't generate the the ssl certificates there.

Another issue is the one pointed out by Bryan above. There are two instances of nginx running, one for ingress (doing SSL termination and proxying to service named "nginx") and another doing simple proxying to test app. Why we have the two is beyond the scope of this PR but it is strange that we need two. Bryan is correct that we still have the "master" nginx process running in the second container running as root. And the reason why SonarCloud did not flag that one is probably that it does not have means to detect it, i.e., it only looks at "Dockerfile" files.

How do we fix this all?
I think there are two options:
1. We do NOT fix it. The nginx image is implemented in such way that only its "master" process runs as root and the workers run as `nginx` user. We can use this as justification for marking the SonarCloud findings as safe.
2. If we really want to fix it, and fix it for both containers, we can do following steps:
    1. Do not mount the `certs` folder into the nginx container
    2. Add a following instruction into `compose/ingress/Dockerfile` file: 
        - `RUN mkdir -p /etc/nginx/ssl && chown -R nginx:nginx /etc/nginx/ssl` 
        -  the net effect of this is that the certificate is generated fresh every time the container starts, hence the start is a tiny bit slower. 
    3. Add a `Dockerfile` for the service named "nginx" instead of just using latest nginx image and mounting the config file `/compose/nginx/nginx.conf` into the container. It would be similar to the other docker file(without the instruction for SSL directory)

BrennanPaciorek:
```
bpaciore@bpaciore-thinkpadp16vgen1:~/src/aap-gateway/django-ansible-base$ docker compose up -d
[+] Running 5/5
 ✔ Network django-ansible-base_default       Created                                                                                                                                                                                     0.2s 
 ✔ Container dab_postgres                    Healthy                                                                                                                                                                                    11.1s 
 ✔ Container django-ansible-base-test_app-1  Started                                                                                                                                                                                    12.0s 
 ✔ Container django-ansible-base-nginx-1     Started                                                                                                                                                                                    12.3s 
 ✔ Container django-ansible-base-ingress-1   Started                                                                                                                                                                                    12.5s 
bpaciore@bpaciore-thinkpadp16vgen1:~/src/aap-gateway/django-ansible-base$ docker ps -a | grep nginx
63ee1697dfed   nginx:latest                                  "/docker-entrypoint.…"   17 seconds ago   Up 4 seconds               0.0.0.0:80->80/tcp, :::80->80/tcp               django-ansible-base-nginx-1
bpaciore@bpaciore-thinkpadp16vgen1:~/src/aap-gateway/django-ansible-base$ docker ps -a | grep nginx
63ee1697dfed   nginx:latest                                  "/docker-entrypoint.…"   20 seconds ago   Up 7 seconds               0.0.0.0:80->80/tcp, :::80->80/tcp               django-ansible-base-nginx-1
bpaciore@bpaciore-thinkpadp16vgen1:~/src/aap-gateway/django-ansible-base$ docker ps -a | grep nginx
63ee1697dfed   nginx:latest                                  "/docker-entrypoint.…"   20 seconds ago   Up 7 seconds               0.0.0.0:80->80/tcp, :::80->80/tcp               django-ansible-base-nginx-1
bpaciore@bpaciore-thinkpadp16vgen1:~/src/aap-gateway/django-ansible-base$ docker ps -a | grep nginx
63ee1697dfed   nginx:latest                                  "/docker-entrypoint.…"   21 seconds ago   Up 8 seconds               0.0.0.0:80->80/tcp, :::80->80/tcp               django-ansible-base-nginx-1
bpaciore@bpaciore-thinkpadp16vgen1:~/src/aap-gateway/django-ansible-base$ docker ps -a | grep nginx
63ee1697dfed   nginx:latest                                  "/docker-entrypoint.…"   22 seconds ago   Up 9 seconds               0.0.0.0:80->80/tcp, :::80->80/tcp               django-ansible-base-nginx-1
bpaciore@bpaciore-thinkpadp16vgen1:~/src/aap-gateway/django-ansible-base$ docker exec -it django-ansible-base-nginx-1 /bin/bash
nginx@63ee1697dfed:/$ 
```

Seems to work on Fedora without problems.

tools/dev_postgres/Dockerfile
```diff
@@ -4,6 +4,10 @@ ENV POSTGRES_DB=dab_db
...
 ENV POSTGRES_PASSWORD=dabing
 
+USER postgres
+
 EXPOSE 5432
 
-HEALTHCHECK --interval=10s --timeout=5s --retries=5 CMD ["pg_isready", "-U", "$${POSTGRES_USER}", "-d", "$${POSTGRES_DB}"]
+HEALTHCHECK --interval=10s --timeout=5s --retries=5 CMD ["sh", "-c", "pg_isready -U \"$POSTGRES_USER\" -d \"$POSTGRES_DB\""]
+
+
```
BrennanPaciorek: We probably want to drop these trailing lines.

BrennanPaciorek (approved):
Seems to address the sonarcloud warning without any adverse effects.

BrennanPaciorek:
@tznamena ~We appear to only be exposing HTTP traffic ports anyways. Not sure where SSL termination would occur in our current docker-compose config.~ My bad, scratch that entire train of thought.

BrennanPaciorek:
Oh I notice the problem now. No perms to revoke my approval though.
```
bpaciore@bpaciore-thinkpadp16vgen1:~/src/aap-gateway/django-ansible-base$ docker logs django-ansible-base-ingress-1 
+ [[ ! -f /etc/nginx/ssl/selfsigned.key ]]
+ nginx -g 'daemon off;'
2025/03/28 16:21:37 [warn] 7#7: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
2025/03/28 16:21:37 [emerg] 7#7: cannot load certificate key "/etc/nginx/ssl/selfsigned.key": BIO_new_file() failed (SSL: error:8000000D:system library::Permission denied:calling fopen(/etc/nginx/ssl/selfsigned.key, r) error:10080002:BIO routines::system lib)
nginx: [emerg] cannot load certificate key "/etc/nginx/ssl/selfsigned.key": BIO_new_file() failed (SSL: error:8000000D:system library::Permission denied:calling fopen(/etc/nginx/ssl/selfsigned.key, r) error:10080002:BIO routines::system lib)
```

BrennanPaciorek:
As Tomas stated, we probably do not need to change the container user, the nginx process spawns its workers under the `nginx` user. To illustrate this, I modified the container so I could use the `ps` utility:
```
bpaciore@bpaciore-thinkpadp16vgen1:~/src/aap-gateway/django-ansible-base$ docker exec -it django-ansible-base-ingress-1 /bin/bash
root@ccda9d3b31a1:/# ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.2  0.0   3928  2944 ?        Ss   16:54   0:00 /bin/bash -x ./entrypoint.sh
root           7  0.0  0.0  11588  8192 ?        S    16:54   0:00 nginx: master process nginx -g daemon off;
nginx          8  0.0  0.0  12016  3272 ?        S    16:54   0:00 nginx: worker process
nginx          9  0.0  0.0  12016  3272 ?        S    16:54   0:00 nginx: worker process
nginx         10  0.0  0.0  12016  3272 ?        S    16:54   0:00 nginx: worker process
nginx         11  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         12  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         13  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         14  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         15  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         16  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         17  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         18  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         19  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         20  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         21  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         22  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
nginx         23  0.0  0.0  12016  3276 ?        S    16:54   0:00 nginx: worker process
root          24  1.2  0.0   4192  3328 pts/0    Ss   16:54   0:00 /bin/bash
root          30  0.0  0.0   8104  4352 pts/0    R+   16:54   0:00 ps aux
root@ccda9d3b31a1:/# 
```

We can probably safely resolve this without any changes to our container definition since the nginx container has its own way of working around this specific problem.

compose/ingress/Dockerfile
```diff
@@ -2,3 +2,17 @@ FROM mirror.gcr.io/library/nginx:latest
...
+# We change it to /tmp/nginx.pid so that it can be accessed by a non-root user.
+RUN sed -i 's|pid        /var/run/nginx.pid;|pid        /tmp/nginx.pid;|' /etc/nginx/nginx.conf && \
+    sed -i '/http {/a \
+    client_body_temp_path /tmp/client_temp;\n\
+    proxy_temp_path       /tmp/proxy_temp;\n\
+    fastcgi_temp_path     /tmp/fastcgi_temp;\n\
+    uwsgi_temp_path       /tmp/uwsgi_temp;\n\
+    scgi_temp_path        /tmp/scgi_temp;' /etc/nginx/nginx.conf
+
+USER nginx
```
BrennanPaciorek:
```suggestion
```

BrennanPaciorek (changes requested):
See the above comment.

compose/ingress/Dockerfile
trucdg: @BrennanPaciorek ~~Is it me or I'm not sure where your suggested change is different from the above code. 😮~~

- I see that you suggested deleting the above block. I've discussed the option of marking the above issue as safe on SonarCloud and apply no fix with John. However, I think it is preferred that we apply a fix and add a `USER nginx` directive in the Dockerfile. That way it's apparently clear that the container is run as a non-root user and SonarCloud doesn't complain about it as a security hotspots.

BrennanPaciorek: I think the best fix for this is no fix, and a comment in the dockerfile. We could do some runtime generation of the certificates by user `nginx`, but I think this is not a great idea if we want to run containers that may require a copy the cert being used by the nginx container.

If we want to lower the container's privileges further, running the containers using podman, or restricting the users capabilities via the compose file may be the best option. 

https://www.redhat.com/en/blog/secure-your-containers-one-weird-trick

BrennanPaciorek: Even considering the above, I would still approve if you decide to go the route of changing the container image's use. If you strictly angling for the best container security, that is probably the best solution.
 (EDIT just noticed the new commits and approved)

(Other edit: You'll need to resolve this one Truc. I do not have the permissions to.)

BrennanPaciorek (approved):

jay-steurer (approved):

compose/ingress/Dockerfile
```diff
@@ -1,4 +1,22 @@
...
+# By default, Nginx writes its PID file to /var/run/nginx.pid, which is a restricted location.
+# We change it to /tmp/nginx.pid so that it can be accessed by a non-root user.
+RUN sed -i '/^user  nginx;/d' /etc/nginx/nginx.conf && \
+    sed -i 's|pid        /var/run/nginx.pid;|pid        /tmp/nginx.pid;|' /etc/nginx/nginx.conf && \
+    sed -i '/http {/a \
+    client_body_temp_path /tmp/client_temp;\n\
+    proxy_temp_path       /tmp/proxy_temp;\n\
+    fastcgi_temp_path     /tmp/fastcgi_temp;\n\
+    uwsgi_temp_path       /tmp/uwsgi_temp;\n\
+    scgi_temp_path        /tmp/scgi_temp;' /etc/nginx/nginx.conf
```
jshimkus-rh: Suggest using `sed -i -E` which enables extended regex.  This allows the specification of each search regex to be written as, for example, `'/^user\s+nginx;/d'`.  This makes it applicable to any non-empty amount of whitespace as opposed to a specific amount.

compose/nginx/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# By default, Nginx writes its PID file to /var/run/nginx.pid, which is a restricted location.
+# We change it to /tmp/nginx.pid so that it can be accessed by a non-root user.
+RUN sed -i '/^user  nginx;/d' /etc/nginx/nginx.conf && \
+    sed -i 's|pid        /var/run/nginx.pid;|pid        /tmp/nginx.pid;|' /etc/nginx/nginx.conf && \
+    sed -i '/http {/a \
+    client_body_temp_path /tmp/client_temp;\n\
+    proxy_temp_path       /tmp/proxy_temp;\n\
+    fastcgi_temp_path     /tmp/fastcgi_temp;\n\
+    uwsgi_temp_path       /tmp/uwsgi_temp;\n\
+    scgi_temp_path        /tmp/scgi_temp;' /etc/nginx/nginx.conf
```
jshimkus-rh: Suggest using `sed -i -E` which enables extended regex.  This allows the specification of each search regex to be written as, for example, `'/^user\s+nginx;/d'`.  This makes it applicable to any non-empty amount of whitespace as opposed to a specific amount.

bhavenst (approved):

relrod:
Awesome work here -- but having thought more about this, I'm skeptical that this is worth doing. Or at least, I would like a more convincing reason than "SonarCloud no likey root".

What specific attack vector does this actually mitigate in our setup?

First off - if we do this, I would suggest a simpler approach: Instead of `sed` at build time which can be hard to follow at a glance, we could copy `/etc/nginx/nginx.conf` from the base image, modify it as needed, and `COPY` it into our new image. Perhaps we could rename our existing `nginx.conf` to `default.conf` and name the new one `nginx.conf` - as it will be named in the image.

That said, this adds ongoing maintenance overhead. Now we're maintaining an entire nginx stack (image/dockerfile/config set) as part of our supported dev environment for DAB. In practice, nginx changes fairly slowly and this probably isn't a huge deal, but it could set a less than ideal precedent.

The thing - and I assume the context that SonarCloud is lacking - is that we aren't running these as privileged containers and we aren't giving the container any additional [capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html). So yes, the nginx master process itself runs as root and the workers drop to `nginx`.

But what is the actual risk here? Are we assuming an upstream supply chain attack on the official nginx image? Even if that happened, how would this container (without extra privileges or capabilities) break out and affect the host?

I get and admire the general instinct toward least privilege, but in this case I don't immediately see the benefit. I'm open to this, but I'd at least like to understand what it is actually solving.

john-westcott-iv (approved):
Approving pending addition of comments being implemented from PR review.

PR #708, opened 2025-03-26
AAP-41636 Add configurable group claim and reading groups from backend

tznamena:
# [AAP-41636] Allow configurable name of "groups" claim for Azure AD and add support for reading groups from this claim

**NOTE:** This PR depends on PR #682 .

## Description

- What is being changed?
  - The AzureAD plugin gets a new configuration option and the inherited function for reading groups is overriden so it does return the groups populated from the claim
- Why is this change needed?
  - Without this change the authentication with Azure AD always defaults the name of groups claim to "Group" however that is not what Azure AD sends the groups in. In our tests we noticed it always been "groups". To avoid hard-coding and giving the users option to change the claim name, it needs to be configurable.
  - New configuration option shown in UI and user can use correct claim name

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites

To test fully (with authentication) you will need to configure an application in Azure and you must use Microsoft Id for that (hotmail id works too). 
To test only the configuration, you don't need Azure at all.

### Steps to Test
1. Pull down the PR
2. Start AAP
3. Navigate to configuration of authentication methods and create one for Azure AD type.
4. (optional, when PR #682 is included as well) Perform login with Azure AD

### Expected Results

Configuration of Azure AD shows new field - default is empty string which effectively means "Group".

Screenshot_20250326_164609

### Required Actions
- [x] Blocked by PR/MR: #682

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/azuread.py
+++ b/ansible_base/authentication/authenticator_plugins/azuread.py
@@ -46,7 +46,6 @@ class AzureADConfiguration(BaseAuthenticatorConfiguration):
     )
 
 
-
 class AuthenticatorPlugin(SocialAuthMixin, SocialAuthValidateCallbackMixin, AzureADOAuth2, AbstractAuthenticatorPlugin):
     configuration_class = AzureADConfiguration
     type = "azuread"
@@ -57,6 +56,6 @@ class AuthenticatorPlugin(SocialAuthMixin, SocialAuthValidateCallbackMixin, Azur
     @property
     def groups_claim(self):
         return self.setting('GROUPS_CLAIM')
-    
+
     def get_user_groups(self, extra_groups=[]):
         return extra_groups
--- a/ansible_base/authentication/authenticator_plugins/azuread.py
+++ b/ansible_base/authentication/authenticator_plugins/azuread.py
@@ -40,7 +40,7 @@ class AzureADConfiguration(BaseAuthenticatorConfiguration):
     GROUPS_CLAIM = CharField(
         help_text=_("The JSON key used to extract the user's groups from the ID token or userinfo endpoint."),
         required=False,
-        allow_null=True,
+        allow_null=False,
         default="Group",
         ui_field_label=_("Groups Claim"),
     )
--- a/ansible_base/authentication/authenticator_plugins/azuread.py
+++ b/ansible_base/authentication/authenticator_plugins/azuread.py
@@ -37,6 +37,15 @@ class AzureADConfiguration(BaseAuthenticatorConfiguration):
         ui_field_label=_('OIDC Secret'),
     )
 
+    GROUPS_CLAIM = CharField(
+        help_text=_("The JSON key used to extract the user's groups from the ID token or userinfo endpoint."),
+        required=False,
+        allow_null=True,
+        default="Group",
+        ui_field_label=_("Groups Claim"),
+    )
+
+
 
 class AuthenticatorPlugin(SocialAuthMixin, SocialAuthValidateCallbackMixin, AzureADOAuth2, AbstractAuthenticatorPlugin):
     configuration_class = AzureADConfiguration
@@ -44,3 +53,10 @@ class AuthenticatorPlugin(SocialAuthMixin, SocialAuthValidateCallbackMixin, Azur
     logger = logger
     category = "sso"
     configuration_encrypted_fields = ['SECRET']
+
+    @property
+    def groups_claim(self):
+        return self.setting('GROUPS_CLAIM')
+    
+    def get_user_groups(self, extra_groups=[]):
+        return extra_groups
--- a/test_app/tests/authentication/authenticator_plugins/test_azuread.py
+++ b/test_app/tests/authentication/authenticator_plugins/test_azuread.py
@@ -101,11 +101,11 @@ def test_groups_setting_and_user_groups(keycloak_authenticator):
 
     ad = get_authenticator_plugin("ansible_base.authentication.authenticator_plugins.azuread")
     ad.database_instance = MockedDb(custom_groups_claim)
-    
+
     # assert that groups claim setting is there and AD has the expected groups claim
     assert ad.strategy.get_setting('GROUPS_CLAIM', backend) == custom_groups_claim
     assert ad.groups_claim == custom_groups_claim
-    
+
     # assert that AD returns expected user groups
     assert ad.get_user_groups() == []
-    assert ad.get_user_groups(["a","b"]) == ["a","b"]
+    assert ad.get_user_groups(["a", "b"]) == ["a", "b"]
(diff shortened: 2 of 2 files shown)


bhavenst (commented):
Add unit test?  I did this a while back: https://github.com/ansible/django-ansible-base/pull/683/files

tznamena:
Good point, I will add similar unit test.

ansible_base/authentication/authenticator_plugins/azuread.py
```diff
@@ -37,10 +37,25 @@ class AzureADConfiguration(BaseAuthenticatorConfiguration):
         ui_field_label=_('OIDC Secret'),
     )
 
+    GROUPS_CLAIM = CharField(
+        help_text=_("The JSON key used to extract the user's groups from the ID token or userinfo endpoint."),
+        required=False,
+        allow_null=True,
```
john-westcott-iv: Do we want to actually allow null? Does the code handle it properly if this is unset? If so can we add a unit test for it? If not can we set this to False?

tznamena: That's a good point, I just took it from OIDC one. I will check how the OIDC handles it.

tznamena: Even though this is the same as in other authenticators, I will change the flag to disallow `null`. 
Allowing `null`  leads to another issue if API call is used to create authenticator with the claim field set explicitly to null and then using UI to edit the authenticator.

test_app/tests/authentication/authenticator_plugins/test_azuread.py
```diff
@@ -79,3 +81,31 @@ def test_azuread_auth_failed(authenticate, unauthenticated_api_client, azuread_a
...
+    ad = get_authenticator_plugin("ansible_base.authentication.authenticator_plugins.azuread")
+    ad.database_instance = MockedDb(custom_groups_claim)
+
+    # assert that groups claim setting is there and AD has the expected groups claim
+    assert ad.strategy.get_setting('GROUPS_CLAIM', backend) == custom_groups_claim
+    assert ad.groups_claim == custom_groups_claim
+
+    # assert that AD returns expected user groups
+    assert ad.get_user_groups() == []
+    assert ad.get_user_groups(["a", "b"]) == ["a", "b"]
```
john-westcott-iv, line 111: This is a positive test, is it worth having a negative test (i.e. if group_claims is set to something that shouldn't be in the token? Or do we already have that somewhere?

ansible_base/authentication/authenticator_plugins/azuread.py
```diff
@@ -37,10 +37,25 @@ class AzureADConfiguration(BaseAuthenticatorConfiguration):
...
 class AuthenticatorPlugin(SocialAuthMixin, SocialAuthValidateCallbackMixin, AzureADOAuth2, AbstractAuthenticatorPlugin):
     configuration_class = AzureADConfiguration
     type = "azuread"
     logger = logger
     category = "sso"
     configuration_encrypted_fields = ['SECRET']
+
+    @property
+    def groups_claim(self):
+        return self.setting('GROUPS_CLAIM')
```
john-westcott-iv, line 58:
```suggestion
        return getattr(self.setting, 'GROUPS_CLAIM', 'Groups')
```

Trying to think if there is a posibility of "I have an AD Authenticator already" so its never been saved with the new configuration schema so it might not have GROUP_CLAIM? Is this something you can test to see if we need to maybe change this to return a default if not found?

test_app/tests/authentication/authenticator_plugins/test_azuread.py
tznamena: The method under test is overriding the method from super class where the method always returns empy list. With the implementation here it is testing both cases: 
- no parameter, return empty list 
- a list passed in a parameter, return the list as is

As to testing of something that should not be in the token, I would say that sort of test should be in the library that deals with decoding and validating the tokens.

ansible_base/authentication/authenticator_plugins/azuread.py
tznamena: The `GROUPS_CLAIM` field does not allow blanks so existing users can not have the authenticator configured without it. They will either have a "Group" value or `null` in there (but they have to explicitly do this via API to get `null` in there as there is no way to set it to `null` via UI). 
The `null` will become `None` in the python code and we already check for that in the pipeline function:
```
groups_claim = backend.groups_claim if backend.groups_claim is not None else "Group"
```

john-westcott-iv: I was thinking more of an exiting authenticator. 
i.e. Before your patch I have AzureAD configured and the "configuration" for that will be stored in the database (with no knowledge of GROUPS_CLAIM).
Then I apply your patch and someone tries to login with the configurator. IDK how `self.setting('GROUPS_CLAIM')` will behave in this scenario.

test_app/tests/authentication/authenticator_plugins/test_azuread.py
john-westcott-iv: Ok, I am ok with this response if this test gives us fill coverage.

ansible_base/authentication/authenticator_plugins/azuread.py
tznamena: I have tested that scenario even though it is an edge case because users can only do it via API.

First, tried creating the authenticator without claim specified:
Since the field "GROUPS_CLAIM does not allow blanks, the default value is used and the field is set to `Group`.

Then I relaxed the field constraints to allow `null` because that's what existing deployments would have and added a logging call to see what's returned from that getter. I've changed the `GROUPS_CLAIM` to `null` in the configuration of the authenticator.
Logging in with that authenticator worked fine, the getter returned None and our code in `social_auth.py` checks for that and uses `Group` in that case.

bhavenst (approved):
I did not test this w/ Azure but it's basically the same as the other similar backends so good w/ it.

ansible_base/authentication/authenticator_plugins/azuread.py
tznamena: I realized my test above was **NOT** testing the same behavior because I was always creating the authenticator configuration with updated code.

Today I have retested this as follows:
1. Created the Azure AD authentication without my changes and ended up with configuration that does not include the "GROUPS_CLAIM" field
2. Added back my changes
3. Logged in with Azure AD and the code in `social_auth.py` checked for None in group claims and used the default of "Groups" since it was `None`. I confirmed this by checking the logs
4. Updated the the authentication to set the "Groups Claim" to "groups" and added mapping that will make the test user an admin/superuser when an expected group are present.
5. Logged in again and the test user became admin/superuser

newswangerd (approved):

PR #713, opened 2025-04-15
AAP-43763 Fix incomplete error handling in advisory_lock wrapper

AlanCoding:
## Description
In AWX we relied on this behavior from `__clean_on_fork__`

https://github.com/ansible/awx/blob/49240ca8e835561f9bc514e37dd70b616d3c922e/awx/conf/settings.py#L267-L281

And we ran this before starting every task. This actually masked a lot of error handling bugs internal to the `advisory_lock` we were using. I recently moved that method from AWX to DAB, and this puts it in a situation not really safe for others to use.

This adds tests which are honest to the bug - they fail without the corresponding fix, and fixes those bugs.

The condition we have breaks _all_ queries after the error. It is very toxic.

Ping @bzwei for review

and @TheRealHaoLiu 

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

## Testing Instructions
The tests are in the tests.

### Steps to Test
What the test does

1. obtain advisory lock
2. time out
3. make a query after that

### Expected Results
subsequent queries work

## Additional Context
in Jira

https://github.com/ansible/eda-server/pull/1262

### Required Actions
- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/lib/utils/db.py
+++ b/ansible_base/lib/utils/db.py
@@ -1,8 +1,12 @@
 from contextlib import contextmanager
 from zlib import crc32
+import logging
 
-from django.db import DEFAULT_DB_ALIAS, connection, connections, transaction
+from django.db import DEFAULT_DB_ALIAS, connection, connections, transaction, OperationalError
 from django.db.migrations.executor import MigrationExecutor
+
+
+logger = logging.getLogger(__name__)
 
 
 @contextmanager
@@ -129,6 +133,7 @@ def advisory_lock(*args, lock_session_timeout_milliseconds=0, **kwargs):
     @param: lock_session_timeout_milliseconds Postgres-level timeout
     @param: using django database identifier
     """
+    internal_error = False
     if connection.vendor == "postgresql":
         cur = None
         idle_in_transaction_session_timeout = None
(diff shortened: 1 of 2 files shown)


bzwei (approved):

chrismeyersfsu:
I can't find any software project that calls or uses `__clean_on_fork__`. I thought maybe it was a uwsgi thing.

ansible_base/lib/utils/db.py
```diff
@@ -139,12 +143,23 @@ def advisory_lock(*args, lock_session_timeout_milliseconds=0, **kwargs):
...
+        except OperationalError:
+            # Suspected case is that timeout happened due to the given timeout
+            # this is _expected_ to leave the connection in an unusable state, so dropping it is better
+            logger.info('Dropping connection due to suspected timeout inside advisory_lock')
+            connection.close_if_unusable_or_obsolete()
+            internal_error = True
+            raise
+        finally:
+            if (not internal_error) and lock_session_timeout_milliseconds > 0:
                 with connection.cursor() as cur:
```
chrismeyersfsu, line 159: even these operations _could_ fail :/

chrismeyersfsu:
Usage of `lock_session_timeout_milliseconds` https://github.com/ansible/awx/blob/db6e8b9bad660bdc028f48e8461e6794b6c3cef1/awx/main/scheduler/task_manager.py#L148 Note `wait=False` and the timeout is about 6 minutes. Practically speaking the timeout would only be hit when things are sideway. What action can be possibly take to recover in this scenario? Retry and establish a new database connection :+1: 

For the above case, this feature makes sense. I would be confusing if I used `advisory_lock` with `wait=True`. With a well behaved health system and connection, I wouldn't expect a database error to be thrown when my timeout is hit.

Now I'm going to read the test cases harder to see if what I described is the behavior.

edit: Ya, it's the behavior.

I'm not going to block this PR on this fact. Someone could still use this function to differentiate a "clean" timeout vs. a failure timeout since you reraise the underlying error. Ideally, we would differentiate those two cases for the consumer in `advisory_lock`

chrismeyersfsu (approved):

test_app/tests/lib/utils/test_db.py
```diff
@@ -14,12 +14,16 @@ def test_migrations_are_complete():
     assert migrations_are_complete()
 
 
-class TestAdvisoryLock:
+class SkipIfSqlite:
```
john-westcott-iv, line 17: Out of curiosity did you have to make a parent class here? Could you have just made a fixture at the "root" of the python file that both classes would have picked up? Also, could this be useful somewhere else (i.e. should we make this available from a normal fixture)?

john-westcott-iv (approved):

test_app/tests/lib/utils/test_db.py
AlanCoding: > Could you have just made a fixture at the "root" of the python file that both classes would have picked up?

If making it module scope, but I'm not a big fan of module scope. I've used it sometimes, but it's better to be explicit IMO, even if more verbose. Not everything in this module should skip sqlite, at least philosophically.

PR #702, opened 2025-02-25
[POC] AAP-40527 DAB tasking app

AlanCoding:
# [AAP-39817] DAB app for background tasks via pg_notify

## Description
This gives a higher-level interface to the [dispatcher library](https://github.com/ansible/dispatcher), formerly the AWX dispatcher. As DAB exists to provide a framework for Django things, this splits out current practices for this in AWX to provide a framework for a background task service.

Apps, at their choosing, may use either dispatcher (lower-level) tasks, or durable tasks tracked by this app. By "lower-level", we mean, specifically, the `psycopg` client library for postgres. Since the dispatcher library only speaks `psycopg`, and not Django, this, predictably, contains a number of utils for translating between Django and `psycopg` language.

Also covers AAP-39484

~Also covers AAP-40527~ Now this is the primary issue

Investigation spike was AAP-39817

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [x] Test update
- [ ] Refactoring (no functional changes)
- [x] Development environment change
- [ ] Configuration change

## Self-Review Checklist
- [ ] I have performed a self-review of my code
- [ ] I have added relevant comments to complex code sections
- [ ] I have updated documentation where needed
- [x] I have considered the security impact of these changes
- [x] I have considered performance implications
- [x] I have thought about error handling and edge cases
- [x] I have tested the changes in my local environment

## Testing Instructions

From my manual demo instructions, in a tab, run `honcho start`.

In another tab, you can run these snippets

```
from test_app.tasks import sleep
from ansible_base.task.tasks import manage_lost_tasks
from uuid import uuid4
from django.utils.timezone import now
import time

sleep.delay(seconds=60)

Task.objects.create(name='footask', state='running', wrapper_uuid=uuid4(), started_at=now())

time.sleep(1)

manage_lost_tasks(grace_period=0)
```

```
from test_app.tasks import *

hello_world.delay()
hello_world_other_queue.delay()
```

### Steps to Test
1. you can run the test_app content by `honcho start`
2. `/api/v1/trigger_task/` can trigger by the API

PR #686, opened 2025-01-23
[WIP] Move psycopg param generator to DAB utility

AlanCoding:
An attempt to take the core utilities from https://github.com/ansible/awx/pull/15736 and polish them up

PR #715, opened 2025-04-18
AAP-43949 Add psycopg db utils for dispatcherd use

AlanCoding:
## Description
We have pushed out our objective merge date for DAB task in the general, which is at https://github.com/ansible/django-ansible-base/pull/702, and doing this creates a need to be able to feature-flag a release using https://github.com/ansible/dispatcherd

Using dispatcherd by itself doesn't need most of the content from DAB task, but it needs _some utility_ type methods. As we flip-flopped on the exact order of operations for this, I had previously put up https://github.com/ansible/django-ansible-base/pull/686, and closed it _in favor of_ DAB task. Some content was added while I worked on DAB task, so I created a new branch (this one) and put the updated content there.

Why?

Merging this will allow the EDA-server and AWX work to de-duplicate code that was copy+pasted there.

https://github.com/ansible/awx/compare/devel...feature_dispatcher#diff-60fa214144754e9e3cf12cc5c7054d8b0ce04f43da33181c0fe9f0d927627038

https://github.com/ansible/eda-server/compare/main...dispatcher-poc-v2#diff-ff14e1e048588b54cbac15edc61d4c830abc8a87cdae26b7fb603d52f3fe1960

Those were always intended to go live in DAB.

Next question - why can't we just put them in the dispatcherd repo? Because _it does not have Django_. This is very important. I do intend to write some docs in dispatcherd that will reference this content.

Ping @bzwei for review

## Type of Change
- [ ] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

## Testing Instructions
This probably needs to be tested with the respective EDA or AWX branches. They have been working, this just consolidates these methods.

### Steps to Test
1. 
2. 
3. 

### Required Actions
- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/lib/utils/db.py
+++ b/ansible_base/lib/utils/db.py
@@ -1,8 +1,13 @@
 import logging
 from contextlib import contextmanager
+from copy import deepcopy
+from typing import Union
 from zlib import crc32
 
+import psycopg
+from django.conf import settings
 from django.db import DEFAULT_DB_ALIAS, OperationalError, connection, connections, transaction
+from django.db.backends.postgresql.base import DatabaseWrapper as PsycopgDatabaseWrapper
 from django.db.migrations.executor import MigrationExecutor
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


ansible_base/lib/utils/db.py
```diff
@@ -149,3 +154,78 @@ def advisory_lock(*args, lock_session_timeout_milliseconds=0, **kwargs):
...
+def get_pg_notify_params(alias: str = DEFAULT_DB_ALIAS, **extra_options) -> dict:
+    """Returns a dictionary that can be used as kwargs to create a psycopg.Connection
+
+    This should use the same connection parameters as Django does.
+    However, this also allows overrides specified by the setting
+    PG_NOTIFY_DATABASES and then the LISTENER_DATABASES, deprecated AWX setting.
+    """
+    pg_notify_overrides = {}
+    if hasattr(settings, 'PG_NOTIFY_DATABASES'):
+        pg_notify_overrides = settings.PG_NOTIFY_DATABASES.get(alias, {})
```
AlanCoding, line 241: I'm lacking a citation for this at the moment, but I believe this setting was a result of prior back-and-fourth with @Alex-Izquierdo, where he suggested this `PG_NOTIFY_DATABASES` as the preferred setting, and I was totally cool with that. We still might have people out there using `LISTENER_DATABASES`. That's the only reason I have to include it.

ansible_base/lib/utils/db.py
```diff
@@ -149,3 +154,78 @@ def advisory_lock(*args, lock_session_timeout_milliseconds=0, **kwargs):
...
+
+
+def psycopg_connection_from_django(**kwargs) -> psycopg.Connection:
+    """Compatibility with dispatcherd connection factory, just returns the Django connection
+
+    dispatcherd passes config info as kwargs, but in this case we just want to ignore then.
+    Because the point of this it to not reconnect, but rely on existing Django connection management.
+    """
+    if connection.connection is None:
+        connection.ensure_connection()
```
AlanCoding, line 185: I had to add this condition, as it presumably fixes test failure in https://github.com/ansible/awx/pull/15949

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/var/lib/awx/venv/awx/lib64/python3.11/site-packages/dispatcherd/registry.py:60: in delay
    return self.apply_async(args, kwargs)
/var/lib/awx/venv/awx/lib64/python3.11/site-packages/dispatcherd/registry.py:101: in apply_async
    broker.publish_message(channel=queue, message=json.dumps(obj))
/var/lib/awx/venv/awx/lib64/python3.11/site-packages/dispatcherd/brokers/pg_notify.py:308: in publish_message
    connection = self.get_connection()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <dispatcherd.brokers.pg_notify.Broker object at 0x7f3834d929d0>

    def get_connection(self) -> psycopg.Connection:
        # Check if the cached connection is either None or closed.
        if not self._sync_connection or getattr(self._sync_connection, "closed", 0) != 0:
            if self._sync_connection_factory:
                factory = resolve_callable(self._sync_connection_factory)
                if not factory:
                    raise RuntimeError(f'Could not import connection factory {self._sync_connection_factory}')
                connection = factory(**self._config)
            elif self._config:
                self.owns_sync_connection = True
                connection = create_connection(**self._config)
            else:
                raise RuntimeError('Could not construct connection for lack of config or factory')
            self._sync_connection = connection
>       assert self._sync_connection is not None
E       AssertionError
```

It is expected that this resolves this. Just documenting.

chrismeyersfsu (approved):

BrennanPaciorek (approved):

ansible_base/lib/utils/db.py
```diff
@@ -149,3 +154,78 @@ def advisory_lock(*args, lock_session_timeout_milliseconds=0, **kwargs):
...
+    Example return value: "dbname=postgres user=postgres"
+    """
+    conn_params = psycopg_kwargs_from_settings_dict(settings_dict)
+    return psycopg.conninfo.make_conninfo(**conn_params)
+
+
+def combine_settings_dict(settings_dict1: dj_db_dict, settings_dict2: dj_db_dict, **extra_options) -> dj_db_dict:
+    """Given two Django database settings dictionaries, combine them and return a new settings_dict"""
+    settings_dict = deepcopy(settings_dict1)
+    settings_dict['OPTIONS'] = deepcopy(settings_dict.get('OPTIONS', {}))
```
john-westcott-iv: Is this actually needed? I thought `deepcopy` would already ensure that nested items were copied?

ansible_base/lib/utils/db.py
```diff
@@ -149,3 +154,78 @@ def advisory_lock(*args, lock_session_timeout_milliseconds=0, **kwargs):
...
+    # These extra options are used by AWX to set application_name
+    settings_dict['OPTIONS'].update(extra_options)
+
+    # Apply overrides specifically for the listener connection
+    for k, v in settings_dict2.items():
+        if k != 'OPTIONS':
+            settings_dict[k] = v
+
+    for k, v in settings_dict2.get('OPTIONS', {}).items():
+        settings_dict['OPTIONS'][k] = v
```
john-westcott-iv: This was probably copied in from AWX code but I'm wondering if we want to make this a more generic recursive function that could be reused somewhere?

```
def merge_dicts(dict1, dict2):
    ### Performs a deep merge of dict1 and dict2 with dict2 taking precedence over dict1

    # Make a deep copy of dict1, this should also deep copy any nested dicts or arrays
    results = deep_copy(dict1)
        
    for k, v in settings_dict2.items():
        if type(v) == dict:
            results[k] = merge_dicts(results[k], dict2[k])
        else:
            results[k] = v

    return results
```

ansible_base/lib/utils/db.py
```diff
@@ -149,3 +154,78 @@ def advisory_lock(*args, lock_session_timeout_milliseconds=0, **kwargs):
...
+
+    return settings_dict
+
+
+def get_pg_notify_params(alias: str = DEFAULT_DB_ALIAS, **extra_options) -> dict:
+    """Returns a dictionary that can be used as kwargs to create a psycopg.Connection
+
+    This should use the same connection parameters as Django does.
+    However, this also allows overrides specified by the setting
+    PG_NOTIFY_DATABASES and then the LISTENER_DATABASES, deprecated AWX setting.
```
john-westcott-iv: `and then` made me thing it would start with DATABASES, lay on top of that PG_NORIFYT_DATABASES and then on top of that LISTERNER_DATABASES. 
```suggestion
    PG_NOTIFY_DATABASES or, if unset, the LISTENER_DATABASES (deprecated AWX setting).
```

ansible_base/lib/utils/db.py
AlanCoding: Valid. Should probably do `.setdefault` instead to assure it's initialized as a dict.

ansible_base/lib/utils/db.py
AlanCoding: For a reference:

https://docs.djangoproject.com/en/5.2/ref/databases/#postgresql-connection-settings

I don't think this can be a nested dictionary. In terms of intended postgres options.

ansible_base/lib/utils/db.py
AlanCoding: Kept the spirit of the comment, but made into bullet-points.

AlanCoding:
My last commit touches stuff this should have no right to touch, but I established the motivation in https://github.com/ansible/django-ansible-base/pull/720

The checks are failing due to a type mismatch calling `super()` itself. This is a bit of nonsense, but clearly it is using an old pattern, so my solution is to just not use that old pattern.

bzwei (approved):

fosterseth (approved):
looks good!

PR #716, opened 2025-04-21
AAP-39645 Allow non-admin admin

bhavenst:
## Description

Fixes issues blocking a user not named "admin" from being the admin user.  The issue in this case was creation of the db authenticator via "aap-gateway-manage authenticators --initialize" was hard coded to the user "admin" but this PR allows it to be created by the system user or by no user if there is no "admin" user or system user defined.

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Start gateway (with PR for AAP-39645 + this PR) after modifying gateway_admin_username in container-startup.yml to something other than admin.
2. See that everything works OK (there were errors in the log and didn't start before).  Log in as the specified user and see they have system administrator role.
3. It's not really possible to unset SYSTEM_USER for gateway or for test_app, both are not happy without one.  This situation could be tested with awx (since apparently it does not use a system user) but that is a challenge to set up.  Unit tests cover this scenario.

### Expected Results
See above.

## Additional Context
N/A

### Required Actions
- [ ] Blocked by PR/MR: #XXX

### Screenshots/Logs
N/A

### DIFF
--- a/ansible_base/authentication/management/commands/authenticators.py
+++ b/ansible_base/authentication/management/commands/authenticators.py
@@ -17,10 +17,10 @@ class Command(BaseCommand):
     help = "Initialize service configuration with an admin user and a local authenticator"
 
     def add_arguments(self, parser):
-        parser.add_argument("--list", action="store_true", help="list the authenticators", required=False)
+        parser.add_argument("--list", action="store_true", help="List the authenticators", required=False)
         parser.add_argument("--initialize", action="store_true", help="Initialize an admin user and local db authenticator", required=False)
-        parser.add_argument("--enable", type=int, help="Initialize an admin user and local db authenticator", required=False)
-        parser.add_argument("--disable", type=int, help="Initialize an admin user and local db authenticator", required=False)
+        parser.add_argument("--enable", type=int, help="Enable the authenticator with provided ID", required=False)
+        parser.add_argument("--disable", type=int, help="Disable the authenticator with provided ID", required=False)
 
     def handle(self, *args, **options):
         took_action = False
@@ -31,16 +31,19 @@ class Command(BaseCommand):
             for id, state in [(options['enable'], True), (options['disable'], False)]:
                 if not id:
                     continue
-                try:
-                    authenticator = Authenticator.objects.get(id=id)
-                except Authenticator.DoesNotExist:
-                    raise CommandError(_("Authenticator %(id)s does not exist") % {"id": id})
-                if authenticator.enabled is not state:
-                    authenticator.enabled = state
-                    authenticator.save()
+                self._update_authenticator(id, state)
             took_action = True
         if options["list"] or not took_action:
             self.list_authenticators()
+
+    def _update_authenticator(self, id: int, state: bool):
+        try:
+            authenticator = Authenticator.objects.get(id=id)
+        except Authenticator.DoesNotExist:
+            raise CommandError(_("Authenticator %(id)s does not exist") % {"id": id})
+        if authenticator.enabled is not state:
+            authenticator.enabled = state
+            authenticator.save()
 
     def list_authenticators(self):
         authenticators = []
@@ -59,6 +62,10 @@ class Command(BaseCommand):
         self.stdout.write('')
 
     def initialize_authenticators(self):
+        if Authenticator.objects.filter(type="ansible_base.authentication.authenticator_plugins.local").exists():
+            self.stdout.write("Local authenticator already exists, skipping")
+            return
+
         # First try to get the system user
         system_user = get_system_user()
         admin_user = None
@@ -74,23 +81,22 @@ class Command(BaseCommand):
         else:
             creator = None
             self.stderr.write("Neither system user nor admin user were defined, local authenticator will be created without created_by set")
-        existing_authenticator = Authenticator.objects.filter(type="ansible_base.authentication.authenticator_plugins.local").first()
-        if not existing_authenticator:
-            existing_authenticator = Authenticator.objects.create(
-                name='Local Database Authenticator',
-                enabled=True,
-                create_objects=True,
-                configuration={},
-                created_by=creator,
-                modified_by=creator,
-                remove_users=False,
-                type='ansible_base.authentication.authenticator_plugins.local',
+
+        new_authenticator = Authenticator.objects.create(
+            name='Local Database Authenticator',
+            enabled=True,
+            create_objects=True,
+            configuration={},
+            created_by=creator,
+            modified_by=creator,
+            remove_users=False,
+            type='ansible_base.authentication.authenticator_plugins.local',
+        )
+        self.stdout.write("Created default local authenticator")
+        if admin_user is not None:
+            self.stdout.write("Binding admin user to the local authenticator")
+            AuthenticatorUser.objects.get_or_create(
+                uid=admin_user.username,
+                user=admin_user,
+                provider=new_authenticator,
             )
-            self.stdout.write("Created default local authenticator")
-            if admin_user is not None:
-                self.stdout.write("Binding admin user to the local authenticator")
-                AuthenticatorUser.objects.get_or_create(
-                    uid=admin_user.username,
-                    user=admin_user,
-                    provider=existing_authenticator,
-                )
--- a/ansible_base/authentication/management/commands/authenticators.py
+++ b/ansible_base/authentication/management/commands/authenticators.py
@@ -9,7 +9,7 @@ from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand, CommandError
 from django.utils.translation import gettext_lazy as _
 
-from ansible_base.authentication.models import Authenticator, AuthenticatorUser
+from ansible_base.authentication.models import Authenticator
 from ansible_base.lib.utils.models import get_system_user
 
 
@@ -82,7 +82,7 @@ class Command(BaseCommand):
             creator = None
             self.stderr.write("Neither system user nor admin user were defined, local authenticator will be created without created_by set")
 
-        new_authenticator = Authenticator.objects.create(
+        Authenticator.objects.create(
             name='Local Database Authenticator',
             enabled=True,
             create_objects=True,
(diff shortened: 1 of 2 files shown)


ansible_base/authentication/management/commands/authenticators.py
```diff
@@ -58,26 +59,38 @@ def list_authenticators(self):
...
-                user=admin_user,
-                provider=existing_authenticator,
-            )
+            if admin_user is not None:
+                self.stdout.write("Binding admin user to the local authenticator")
+                AuthenticatorUser.objects.get_or_create(
+                    uid=admin_user.username,
+                    user=admin_user,
+                    provider=existing_authenticator,
+                )
```
trucdg: 1. I'd suggest only query for `admin_user` or `system_user` when there is no `existing_authenticator` to save some querying performance
2. John commented on the Jira issue saying that AWX wasn't going to support the _SYSTEM user, should we prioritize the `admin_user` to be the creator rather than the `system_user`
3. Should we have a `get_admin_user()` too :-?

So my suggestion is something like this:
```python
existing_authenticator = Authenticator.objects.filter(type="ansible_base.authentication.authenticator_plugins.local").first()

if not existing_authenticator:
    system_user = get_system_user()
    admin_user = get_admin_user()
    creator = admin_user or system_user
    if creator is None:
        self.stderr.write("Neither system user nor admin user were defined, local authenticator will be created without created_by set")

    existing_authenticator = Authenticator.objects.create(
                name='Local Database Authenticator',
                enabled=True,
                create_objects=True,
                configuration={},
                created_by=creator,
                modified_by=creator,
                remove_users=False,
                type='ansible_base.authentication.authenticator_plugins.local',
            )
    self.stdout.write("Created default local authenticator")
    if admin_user is not None:
        self.stdout.write("Binding admin user to the local authenticator")
                AuthenticatorUser.objects.get_or_create(
                    uid=admin_user.username,
                    user=admin_user,
                    provider=existing_authenticator,
                )
```

trucdg (changes requested):

AlanCoding (changes requested):
I only have 1 question here - would it make more sense to bind the local authenticator to _all_ admin users? What exactly are we losing by not associating the local authenticator? Because it seems suspicious to do that for admin user but not for other usernames.

More of a question than a request for change here. it's what raises eyebrows.

bhavenst:
@AlanCoding I was also a bit confused by not binding to any/all admin users if "admin" wasn't present, but implemented it as John suggested.  He's back tomorrow so I'll discuss w/ him.

The AuthenticatorUser does get created(bound) for the authenticator when the user logs in, but I'm not really sure of the implications of relying on that vs pre-creating it or the converse as you say, why we are doing that for "admin".

ansible_base/authentication/management/commands/authenticators.py
bhavenst: Hi @trucdg,
1. Sure, that should be fine, will do.
2. I think it's standard and intentional that system user should own all system created objects unless there is not a system user, in which it falls back to the admin user (or none in this current code if specifically "admin" doesn't exist).  It's an option for each component that extends DAB whether or not they have a system user (AFAIK) so it needs to support both cases.
3. system user is special and get_system_user comes from DAB and interacts with settings etc.  admin users on the other hand are nothing special and there can be many of them.  I'm not sure of the utility of a get_admin_users vs a regular query on is_superuser or where it would go..

ansible_base/authentication/management/commands/authenticators.py
```diff
@@ -30,17 +31,20 @@ def handle(self, *args, **options):
...
                 if not id:
                     continue
-                try:
-                    authenticator = Authenticator.objects.get(id=id)
-                except Authenticator.DoesNotExist:
-                    raise CommandError(_("Authenticator %(id)s does not exist") % {"id": id})
-                if authenticator.enabled is not state:
-                    authenticator.enabled = state
-                    authenticator.save()
+                self._update_authenticator(id, state)
```
bhavenst, line 34: Fixes Sonar cognitive complexity complaint..

john-westcott-iv:
There are two things here. First, we need a user to use as the creator of the local authenticator. This is the logic for looking for system user or admin.

As for binding the admin user via an AuthenticatorUser; I don't think we technically need to do this any more. That should automatically happen at login now on the fly. We should test that and, if so, just remove the binding of the admin user to the authenticator.

Also, we technically shouldn't need the first() because we are filtering on username which is unique anyway so there will only ever be 1 at most. But since we are doing a filter and not a get it would have to be there.

bhavenst:
Thanks @john-westcott-iv I have tested that and it works as you say (auto-association).  In the dev environment it happens immediately for the non-"admin" admin due to the logins for creating the services etc.

I'll remove it.

Also updated a test case to check that the AuthenticatorUser is auto-created on login.

trucdg:
Following steps to test: Step 1 and 2 work as expected. However, when commented out the SYSTEM_USERNAME in defaults.py, I got the below error:

```python
aap_gw_1           | Traceback (most recent call last):
aap_gw_1           |   File "<frozen runpy>", line 198, in _run_module_as_main
aap_gw_1           |   File "<frozen runpy>", line 88, in _run_code
aap_gw_1           |   File "/opt/aap_gateway/src/aap_gateway_api/__main__.py", line 4, in <module>
aap_gw_1           |     manage()
aap_gw_1           |   File "/opt/aap_gateway/src/aap_gateway_api/__init__.py", line 22, in manage
aap_gw_1           |     execute_from_command_line(sys.argv)
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
aap_gw_1           |     utility.execute()
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/core/management/__init__.py", line 436, in execute
aap_gw_1           |     self.fetch_command(subcommand).run_from_argv(self.argv)
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/core/management/base.py", line 412, in run_from_argv
aap_gw_1           |     self.execute(*args, **cmd_options)
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/core/management/base.py", line 458, in execute
aap_gw_1           |     output = self.handle(*args, **options)
aap_gw_1           |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/core/management/base.py", line 106, in wrapper
aap_gw_1           |     res = handle_func(*args, **kwargs)
aap_gw_1           |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/core/management/commands/migrate.py", line 356, in handle
aap_gw_1           |     post_migrate_state = executor.migrate(
aap_gw_1           |                          ^^^^^^^^^^^^^^^^^
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/db/migrations/executor.py", line 135, in migrate
aap_gw_1           |     state = self._migrate_all_forwards(
aap_gw_1           |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/db/migrations/executor.py", line 167, in _migrate_all_forwards
aap_gw_1           |     state = self.apply_migration(
aap_gw_1           |             ^^^^^^^^^^^^^^^^^^^^^
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/db/migrations/executor.py", line 252, in apply_migration
aap_gw_1           |     state = migration.apply(state, schema_editor)
aap_gw_1           |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/db/migrations/migration.py", line 132, in apply
aap_gw_1           |     operation.database_forwards(
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/db/migrations/operations/special.py", line 193, in database_forwards
aap_gw_1           |     self.code(from_state.apps, schema_editor)
aap_gw_1           |   File "/opt/aap_gateway/src/aap_gateway_api/migrations/_initialize_tasks.py", line 50, in create_system_user
aap_gw_1           |     system_username = settings.SYSTEM_USERNAME
aap_gw_1           |                       ^^^^^^^^^^^^^^^^^^^^^^^^
aap_gw_1           |   File "/opt/aap_gateway/venv/lib64/python3.11/site-packages/django/conf/__init__.py", line 104, in __getattr__
aap_gw_1           |     val = getattr(_wrapped, name)
aap_gw_1           |           ^^^^^^^^^^^^^^^^^^^^^^^
aap_gw_1           | AttributeError: 'Settings' object has no attribute 'SYSTEM_USERNAME'
aap_gw_1           |   Applying aap_gateway_api.0001_initial...
aap_gw_1 exited with code 1
aap_gw_proxy1      | [2025-04-24 13:16:57.509][1][warning][config] [external/envoy/source/common/config/http_subscription_impl.cc:120] REST update for /v3/discovery:clusters failed

```

bhavenst:
Yeah, OK.  I didn't actually try removing the system user entirely like I suggested :).  It might be that gateway can't deal with that.  There is a unit test that temporarily unsets it, so maybe that's enough.

AlanCoding:
What I tried, on devel, and then on this branch:

```
python manage.py authenticators
python manage.py authenticators --initialize
```

The devel behavior is:

> CommandError: Admin user with username 'admin' not defined.

And on this branch it populates:

```
|   ID | Enabled   | Name                         |   Order |
|------|-----------|------------------------------|---------|
|    1 | True      | Local Database Authenticator |       1 |
```

I do kind of agree that this is generally useful. I was worried about some corner cases, I tried:
 - initializing authentications
 - THEN, creating superuser

For whatever reason, I'm able to login as the superuser (test_app) in this scenario. So there's not a gotcha for ordering in this case. I'm okay with all this. Legit, I remember being annoyed by the ordering that we had to do things with before.

@trucdg may have a point about the SYSTEM_USER. My testing is pretty limited, and I'm sure that's set for test_app.

AlanCoding (approved):

trucdg (approved):

john-westcott-iv:
Yea, gateway requires the system user to be present. You would have to test with an app that doesn't require it.

PR #720, opened 2025-04-24
Fix sqlite3 type error to super

AlanCoding:
this is an attempted CI fix, seen in output from other PR


AlanCoding:
Replaced by https://github.com/ansible/django-ansible-base/pull/722

PR #724, opened 2025-05-06
[AAP-45287] Fix IPv6 handling for Redis

jshimkus-rh:
## Description

- What is being changed?
* Extending the class of addresses supported for Redis clients to include IPv6 values.

- Why is this change needed?
* Current Redis client rejects IPv6 address as invalid.

Accepts IPv6 addresses in the common forms of <addr>:<port> and [<addr>]:<port>.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Deploy AAP environment in an IPv6-only environment *not utilizing host names for Redis client.  Installer operates as such.

### Expected Results
Redis client connections successfully instantiated.

### DIFF
--- a/ansible_base/lib/redis/client.py
+++ b/ansible_base/lib/redis/client.py
@@ -12,7 +12,7 @@ from redis.cluster import ClusterNode, RedisCluster
 from redis.exceptions import NoPermissionError, RedisClusterException
 
 from ansible_base.lib.constants import STATUS_DEGRADED, STATUS_FAILED, STATUS_GOOD
-from ansible_base.lib.utils.address import AddressType, classify_and_split_address_string
+from ansible_base.lib.utils import address
 
 logger = logging.getLogger('ansible_base.lib.redis.client')
 
@@ -161,27 +161,27 @@ class RedisClientGetter:
             had_host_errors = False
             host_ports = self.redis_hosts.split(',')
             for host_port in host_ports:
-                (addr_type, node, port_string) = classify_and_split_address_string(host_port)
-                if addr_type == AddressType.UNKNOWN:
+                response = address.classify_address(host_port)
+                if response.type == address.AddressType.UNKNOWN:
                     logger.error(
                         f"Specified cluster_host {host_port} is not valid; "
                         "it is of an unknown address type and must be one of <hostname>:<port>, <ipv4>:<port> or <ipv6>:<port>"
                     )
                     had_host_errors = True
                     continue
-                if not port_string:
+                if not response.port:
                     logger.error(f"Specified cluster_host {host_port} is not valid; it needs to be in the format <host>:<port>")
                     had_host_errors = True
                     continue
                 # Make sure we have an int for the port
                 try:
-                    port = int(port_string)
+                    port = int(response.port)
                 except ValueError:
                     logger.error(f'Specified port on {host_port} is not an int')
                     had_host_errors = True
                     continue
 
-                self.connection_settings['startup_nodes'].append(ClusterNode(node, port))
+                self.connection_settings['startup_nodes'].append(ClusterNode(response.address, port))
 
             if had_host_errors:
                 raise ValueError()
--- a/ansible_base/lib/redis/client.py
+++ b/ansible_base/lib/redis/client.py
@@ -12,6 +12,7 @@ from redis.cluster import ClusterNode, RedisCluster
 from redis.exceptions import NoPermissionError, RedisClusterException
 
 from ansible_base.lib.constants import STATUS_DEGRADED, STATUS_FAILED, STATUS_GOOD
+from ansible_base.lib.utils.address import AddressType, classify_and_split_address_string
 
 logger = logging.getLogger('ansible_base.lib.redis.client')
 
@@ -160,13 +161,15 @@ class RedisClientGetter:
             had_host_errors = False
             host_ports = self.redis_hosts.split(',')
             for host_port in host_ports:
-                try:
-                    node, port_string = host_port.split(':')
-                except ValueError:
+                (addr_type, node, port_string) = classify_and_split_address_string(host_port)
+                if addr_type == AddressType.UNKNOWN:
+                    logger.error(f"Specified cluster_host {host_port} is not valid; it is of an unknown address type")
+                    had_host_errors = True
+                    continue
+                if not port_string:
                     logger.error(f"Specified cluster_host {host_port} is not valid; it needs to be in the format <host>:<port>")
                     had_host_errors = True
                     continue
-
                 # Make sure we have an int for the port
                 try:
                     port = int(port_string)
--- a/ansible_base/lib/redis/client.py
+++ b/ansible_base/lib/redis/client.py
@@ -165,7 +165,7 @@ class RedisClientGetter:
                 if response.type == address.AddressType.UNKNOWN:
                     logger.error(
                         f"Specified cluster_host {host_port} is not valid; "
-                        "it is of an unknown address type and must be one of <hostname>:<port>, <ipv4>:<port> or <ipv6>:<port>"
+                        "it is of an unknown address type and must be one of <hostname>:<port>, <ipv4>:<port> or [<ipv6>]:<port>"
                     )
                     had_host_errors = True
                     continue
--- a/ansible_base/lib/redis/client.py
+++ b/ansible_base/lib/redis/client.py
@@ -163,7 +163,10 @@ class RedisClientGetter:
             for host_port in host_ports:
                 (addr_type, node, port_string) = classify_and_split_address_string(host_port)
                 if addr_type == AddressType.UNKNOWN:
-                    logger.error(f"Specified cluster_host {host_port} is not valid; it is of an unknown address type")
+                    logger.error(
+                        f"Specified cluster_host {host_port} is not valid; "
+                        "it is of an unknown address type and must be one of <hostname>:<port>, <ipv4>:<port> or <ipv6>:<port>"
+                    )
                     had_host_errors = True
                     continue
                 if not port_string:
--- a/ansible_base/lib/utils/address.py
+++ b/ansible_base/lib/utils/address.py
@@ -1,0 +1,113 @@
+import enum
+import ipaddress
+import re
+
+
+class AddressType(enum.Enum):
+    HOSTNAME = "hostname"
+    IPv4 = "ipv4"
+    IPv6 = "ipv6"
+    UNKNOWN = "unknown"
+
+
+def _classify_address_string(address_string) -> AddressType:
+    """
+    Categorizes a given string as IPv4, IPv6, hostname, or unknown.
+
+    Args:
+        address_string: The string to categorize.
+
+    Returns:
+        A value of AddressType indicating the category.
+    """
+    try:
+        ipaddress.IPv4Address(address_string)
+        return AddressType.IPv4
+    except ipaddress.AddressValueError:
+        pass
+
+    try:
+        ipaddress.IPv6Address(address_string)
+        return AddressType.IPv6
+    except ipaddress.AddressValueError:
+        pass
+
+    # Basic hostname check (can be expanded for more rigorous validation)
+    if re.match(
+        r"^[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?"
+        r"(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$",
+        address_string
+    ):
+        return AddressType.HOSTNAME
+
+    return AddressType.UNKNOWN
+
+
+def classify_address_string(address_string) -> AddressType:
+    """
+    Categorizes a given string as IPv4, IPv6, hostname, or unknown.
+
+    An IPv6 address wrapped in [] is considered a hostname.
+
+    Args:
+        address_string: The string to categorize.
+
+    Returns:
+        A value of AddressType indicating the category.
+    """
+    address_type = _classify_address_string(address_string)
+    if address_type != AddressType.UNKNOWN:
+        return address_type
+
+    # We could be dealing with an IPv6 address wrapped in [].
+    # If that is the case we want to identify it as a host name.
+    if (address_string[0] == "[") and (address_string[-1] == "]"):
+        address_type = _classify_address_string(address_string[1:-1])
+
+        # Only an address type of IPv6 is considered valid here.
+        # We don't want to treat any other specification as valid.
+        if address_type == AddressType.IPv6:
+            return AddressType.HOSTNAME
+
+    return AddressType.UNKNOWN
+
+
+def classify_and_split_address_string(address_string) -> tuple[AddressType, str, str]:
+    """
+    Categorizes a given string with optional ":<port>" suffix as IPv4, IPv6,
+    hostname, or unknown.
+
+    Args:
+        address_string: The string to categorize.
+
+    Returns:
+        A tuple of the form (AddressType, address string, port string) where
+        address_string and port_string are only valid (though port string may
+        be an empty string) when AddressType is not AddressType.UNKNOWN.
+    """
+    address_type = classify_address_string(address_string)
+    if address_type != AddressType.UNKNOWN:
+        # A known type with no port.
+        return (address_type, address_string, "")
+
+    # Split into potential address and port and classify the address.
+    (address_string, _, port) = address_string.rpartition(":")
+    address_type = classify_address_string(address_string)
+    if address_type != AddressType.UNKNOWN:
+        # A known type with a port.
+        return (address_type, address_string, port)
+
+    # An unknown address type.
+    return (AddressType.UNKNOWN, "", "")
+
+
+def is_hostname_address_string(address_string: str) -> bool:
+    return classify_address_string(address_string) == AddressType.HOSTNAME
+
+
+def is_ipv4_address_string(address_string: str) -> bool:
+    return classify_address_string(address_string) == AddressType.IPv4
+
+
+def is_ipv6_address_string(address_string: str) -> bool:
+    return classify_address_string(address_string) == AddressType.IPv6
--- a/ansible_base/lib/utils/address.py
+++ b/ansible_base/lib/utils/address.py
@@ -38,17 +38,26 @@ class AddressTypeResponse(object):
     existing code to facilitate use of the classification functionality
     provided.  An empty string indicates the non-existence of that particular
     attribute as part of the classified address.
-
-    The address and port fields are not valid in any sense if the type field is
-    AddressType.UNKNOWN.
     """
 
     type: AddressType
     address: str = ""
     port: str = ""
 
+    @property
+    def ipv6_bracketed(self):
+        """
+        IPv6 addresses are stored without brackets in the address field.
+        If the type of this instance is AddressType.IPv6 the method will return
+        the address enclosed in brackets.
+        If the type is not AddressType.IPv6 it will return None.
+        """
+        if self.type != AddressType.IPv6:
+            return None
+        return f"[{self.address}]"
 
-def _classify_base_address(address: str) -> AddressType:
+
+def _classify_base_address(address: str) -> AddressTypeResponse:
     """
     Categorizes a given string as IPv4, IPv6, hostname, or unknown.
 
@@ -56,17 +65,17 @@ def _classify_base_address(address: str) -> AddressType:
         address: The string to categorize.
 
     Returns:
-        A value of AddressType indicating the category.
+        A value of AddressTypeResponse indicating the classified address.
     """
     try:
         ipaddress.IPv4Address(address)
-        return AddressType.IPv4
+        return AddressTypeResponse(AddressType.IPv4, address)
     except ipaddress.AddressValueError:
         pass
 
     try:
         ipaddress.IPv6Address(address)
-        return AddressType.IPv6
+        return AddressTypeResponse(AddressType.IPv6, address)
     except ipaddress.AddressValueError:
         pass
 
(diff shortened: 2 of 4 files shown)


test_app/tests/lib/utils/test_address.py
```diff
@@ -0,0 +1,99 @@
...
+    "address,expected",
+    [
+        ("127.0.0.1", AddressType.IPv4),
+        ("[127.0.0.1]", AddressType.UNKNOWN),
+        ("10.0.1.1", AddressType.IPv4),
+        ("[10.0.1.1]", AddressType.UNKNOWN),
+        ("::1", AddressType.IPv6),
+        ("[::1]", AddressType.HOSTNAME),
+        ("2600:1f18:218b:5902:e5d4:54de:fdc1:24b8", AddressType.IPv6),
+        ("[2600:1f18:218b:5902:e5d4:54de:fdc1:24b8]", AddressType.HOSTNAME),
```
bhavenst: Aren't these bracket notation ipv6 addresses?  I thought the [] was just to avoid interpreting the port as part of the address.  Probably wrong, but it seems odd that these resolve to hostname.

test_app/tests/lib/utils/test_address.py
```diff
@@ -0,0 +1,99 @@
...
+        "localhost",
+        "a-host-name",
+        "[2600:1f18:218b:5902:e5d4:54de:fdc1:24b8]",
+    ],
+)
+def test_is_hostname_address(address):
+    assert is_hostname_address_string(address)
+
+
+@pytest.mark.parametrize(
```
bhavenst: Maybe a couple of invalid cases? 127.0.0.257 for instance.

test_app/tests/lib/utils/test_address.py
bhavenst: Oh, noticed the note stating that.  Interesting, is that a general thing or just for purposes of this usage?

ansible_base/lib/redis/client.py
```diff
@@ -160,13 +161,15 @@ def _get_hosts(self) -> None:
...
-                try:
-                    node, port_string = host_port.split(':')
-                except ValueError:
+                (addr_type, node, port_string) = classify_and_split_address_string(host_port)
+                if addr_type == AddressType.UNKNOWN:
+                    logger.error(f"Specified cluster_host {host_port} is not valid; it is of an unknown address type")
+                    had_host_errors = True
+                    continue
+                if not port_string:
                     logger.error(f"Specified cluster_host {host_port} is not valid; it needs to be in the format <host>:<port>")
```
bhavenst, line 173: Need to expand this "needs to be in format ..." list?

bhavenst (approved):

test_app/tests/lib/utils/test_address.py
jshimkus-rh: It's to differentiate between a raw address that must be wrapped in []s and one already so wrapped.  I chose to use hostname as that's the analogous common use case we have: `https://[<ipv6>]:<port>/....`

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,109 @@
+import enum
+import ipaddress
+import re
+
+
+class AddressType(enum.Enum):
```
tznamena, line 8: Some documentation would be nice for this class.

tznamena (changes requested):

ansible_base/lib/utils/address.py
jshimkus-rh: Added.

tznamena: Awesome docs! :+1:

tznamena (approved):

bhavenst (approved):

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # We could be dealing with an IPv6 address wrapped in [].
+    # If that is the case we want to identify it as a host name.
+    if (address_string[0] == "[") and (address_string[-1] == "]"):
+        address_type = _classify_address_string(address_string[1:-1])
+
+        # Only an address type of IPv6 is considered valid here.
+        # We don't want to treat any other specification as valid.
+        if address_type == AddressType.IPv6:
+            return AddressType.HOSTNAME
```
john-westcott-iv: Is HOSTNAME correct here? Wouldn't it be an IPv6 or do we consider [<ip v6 address>] as a hostname?

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,125 @@
...
+    address_type = classify_address_string(address_string)
+    if address_type != AddressType.UNKNOWN:
+        # A known type with a port.
+        return (address_type, address_string, port)
+
+    # An unknown address type.
+    return (AddressType.UNKNOWN, "", "")
+
+
+def is_hostname_address_string(address_string: str) -> bool:
```
john-westcott-iv: This name of this one feels weird to me. Can we call it `is_hostname` or would that conflict with something else?

john-westcott-iv: Also, do we want an `is_hostname_with_port` method?

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Args:
+        address_string: The string to categorize.
+
+    Returns:
+        A tuple of the form (AddressType, address string, port string) where
+        address_string and port_string are only valid (though port string may
+        be an empty string) when AddressType is not AddressType.UNKNOWN.
+    """
+    address_type = classify_address_string(address_string)
```
john-westcott-iv: When I'm looking at this, I'm wondering if we can shrink this logic with a regex like:
```
address_re = re.compile('^\[?([a-zA-Z0-9:]*)\]?(:([^:]+))?')
address_re.matches(address_string)
```
Its not quite working, `localhost:` passes the check. But it feels like something like this might be able to reduce some logic in here.

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,125 @@
...
+        pass
+
+    try:
+        ipaddress.IPv6Address(address_string)
+        return AddressType.IPv6
+    except ipaddress.AddressValueError:
+        pass
+
+    # Basic hostname check (can be expanded for more rigorous validation)
+    if re.match(r"^[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$", address_string):
```
relrod: Can you leave a comment with where this regex came from, assuming you didn't write it yourself? Or alternatively, maybe split it into multiple lines and comment what is going on here?

This is a moderately complex regex, I worry about being able to maintain it going forward.

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,125 @@
...
+        address_string: The string to categorize.
+
+    Returns:
+        A value of AddressType indicating the category.
+    """
+    address_type = _classify_address_string(address_string)
+    if address_type != AddressType.UNKNOWN:
+        return address_type
+
+    # We could be dealing with an IPv6 address wrapped in [].
```
relrod, line 119: Rather than having to parse it twice in this case, would it be better to just strip the `[` and `]` at the beginning, parse the string, then if we get IPv6 back and the original string had `[` and `]` then we characterize it as `HOSTNAME`?

In other words, something like this (untested):


```python
is_bracketed = address_string[0] == '[' and address_string[-1] == ']'
address_type = _classify_address_string(address_string.strip('[]'))
if address_type == AddressType.IPv6 and is_bracketed:
    return AddressType.HOSTNAME
return address_type
```

And probably a check above to make sure we're not given the empty string as input, otherwise we'd crash on indexing into it.

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,125 @@
...
+        pass
+
+    # Basic hostname check (can be expanded for more rigorous validation)
+    if re.match(r"^[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$", address_string):
+        return AddressType.HOSTNAME
+
+    return AddressType.UNKNOWN
+
+
+def classify_address_string(address_string) -> AddressType:
```
relrod:
```suggestion
def classify_address_string(address_string: str) -> AddressType:
```

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,125 @@
...
+    successfully utilize it.
+    """
+
+    HOSTNAME = "hostname"
+    IPv4 = "ipv4"
+    IPv6 = "ipv6"
+    UNKNOWN = "unknown"
+
+
+def _classify_address_string(address_string) -> AddressType:
```
relrod:
```suggestion
def _classify_address_string(address_string: str) -> AddressType:
```

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Only an address type of IPv6 is considered valid here.
+        # We don't want to treat any other specification as valid.
+        if address_type == AddressType.IPv6:
+            return AddressType.HOSTNAME
+
+    return AddressType.UNKNOWN
+
+
+def classify_and_split_address_string(address_string) -> tuple[AddressType, str, str]:
```
relrod:
```suggestion
def classify_and_split_address_string(address_string: str) -> tuple[AddressType, str, str]:
```

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Returns:
+        A tuple of the form (AddressType, address string, port string) where
+        address_string and port_string are only valid (though port string may
+        be an empty string) when AddressType is not AddressType.UNKNOWN.
+    """
+    address_type = classify_address_string(address_string)
+    if address_type != AddressType.UNKNOWN:
+        # A known type with no port.
+        return (address_type, address_string, "")
```
relrod: Is there a good reason for port to be a string, beyond "might be empty"?

If not, I'd prefer `Optional[int]` here.

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Split into potential address and port and classify the address.
+    (address_string, _, port) = address_string.rpartition(":")
+    address_type = classify_address_string(address_string)
+    if address_type != AddressType.UNKNOWN:
+        # A known type with a port.
+        return (address_type, address_string, port)
+
+    # An unknown address type.
+    return (AddressType.UNKNOWN, "", "")
```
relrod: I'd consider making the address part an `Optional[str]` and have it be `None` here rather than empty string.

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,125 @@
...
+
+def is_hostname_address_string(address_string: str) -> bool:
+    return classify_address_string(address_string) == AddressType.HOSTNAME
+
+
+def is_ipv4_address_string(address_string: str) -> bool:
+    return classify_address_string(address_string) == AddressType.IPv4
+
+
+def is_ipv6_address_string(address_string: str) -> bool:
```
relrod: Not a blocker, but: Future optimization for these wrapper functions would be to _only_ check the thing we care about in the cases where we can.

In other words, as it stands right now, for `is_ipv6_address_string()` to work, it will always try to be parsed as v4 first, and we should be able to avoid that.

ansible_base/lib/utils/address.py
relrod: I'd go further and say, can we drop the `_string` everywhere? Both in the function names, and also in the function parameter? In both cases, the type of the input tells us it's a string.

I'd prefer e.g. `classify_address(address: str)`, etc.

ansible_base/lib/utils/address.py
jshimkus-rh: The comment before the preceding if can be more explicit.  The intention here is two-fold:
1. minimize the type space; i.e., not have two kinds of IPv6
2. minimize the logic that a client has to implement 

In the case of 2 the way in which a []ed IPv6 is used is as if it were a hostname.  That is, no special processing is required (beyond convenience this is an additional reason there's a method that does the splitting into "address" and "port"; otherwise the client would still have to special case processing). 

So the only, if it matters to the component, address type it may have to care about is the single IPv6 type.

ansible_base/lib/utils/address.py
jshimkus-rh: > This name of this one feels weird to me. Can we call it `is_hostname` or would that conflict with something else?

@john-westcott-iv Sure, we can shorten the name.  In some sense I'd prefer that based on my perspective that python code should be using (at the least) the terminal namespace to access offerings; primarily for some identification as to origin but also to avoid name collisions.

jshimkus-rh: > Also, do we want an `is_hostname_with_port` method?

Possibly.  I was thinking about how far we want to split the identification breakdown.

I included the is_* methods mostly because that's what I did in the gateway code.  But nothing uses them and we if only provide the full breakdown (type, address, port) method we eliminate the inconsistencies in which is_ipv6 with <addr>:<port> reports "No" but the full breakdown reports IPv6.

So I'm going to eliminate everything other than the full breakdown.

ansible_base/lib/utils/address.py
jshimkus-rh: Good idea re: the comment.  
I tried the splitting originally but our CI linting complained about it.

< aside> The AI generated version of the core categorization provided a simpler regex that just didn't work (not even for names like "a-hostname".  But asking solely for a regex that would identify hostnames gave the one in the code. < /aside>

ansible_base/lib/utils/address.py
jshimkus-rh: Not unreasonable re: parsing.

Definitely yes re: empty string.

ansible_base/lib/utils/address.py
jshimkus-rh: It minimizes modifying existing code which already deals with the port as split from the string.

I'm not saying it's a wrong idea but minimizing changes to existing code being able to use this functionality was the reason.

ansible_base/lib/utils/address.py
jshimkus-rh: Considered that as I'm not a fan of non-explicit comparison of a value as boolean (unless it *is* a boolean).  In this case "" and None would be logically equivalent.

Also, see comment about minimizing changes to existing code.

ansible_base/lib/utils/address.py
jshimkus-rh: Gonna change to just the one categorization method so this won't be applicable any more.

test_app/tests/lib/utils/test_address.py
```diff
@@ -0,0 +1,32 @@
...
+        ("127.0.0.1:1234", AddressTypeResponse(AddressType.IPv4, "127.0.0.1", "1234")),
+        ("10.0.1.1:1234", AddressTypeResponse(AddressType.IPv4, "10.0.1.1", "1234")),
+        ("10.0.1.1", AddressTypeResponse(AddressType.IPv4, "10.0.1.1")),
+        ("[10.0.1.1]", AddressTypeResponse(AddressType.UNKNOWN)),
+        ("::1", AddressTypeResponse(AddressType.IPv6, "::1")),
+        ("[::1]", AddressTypeResponse(AddressType.HOSTNAME, "[::1]")),
+        ("[::1]:1234", AddressTypeResponse(AddressType.HOSTNAME, "[::1]", "1234")),
+        ("2600:1f18:218b:5902:e5d4:54de:fdc1:24b8", AddressTypeResponse(AddressType.IPv6, "2600:1f18:218b:5902:e5d4:54de:fdc1:24b8")),
+        ("[2600:1f18:218b:5902:e5d4:54de:fdc1:24b8]", AddressTypeResponse(AddressType.HOSTNAME, "[2600:1f18:218b:5902:e5d4:54de:fdc1:24b8]")),
+        ("2600:1f18:218b:5902:e5d4:54de:fdc1:24b8:1234", AddressTypeResponse(AddressType.IPv6, "2600:1f18:218b:5902:e5d4:54de:fdc1:24b8", "1234")),
```
PabloHiro, line 20: This should throw an error imho. No unbracketed IPv6+Port combination should be fed to Gateway in the first place. And if it is, throw an error instead of trying to accommodate for that.

test_app/tests/lib/redis/test_client.py
```diff
@@ -209,6 +209,13 @@ def test_redis_cluster_mget_raises_expected_exception():
         ('a:1,b:1', False, 2),
         ('a:b', True, None),
         ('a,b,c', True, None),
+        ('[::1]:1', False, 1),
+        ('2600:1f18:218b:5902:e5d4:54de:fdc1:24b8:1', False, 1),
+        ('[2600:1f18:218b:5902:e5d4:54de:fdc1:24b8]:1', False, 1),
+        ('[::1]:1,2600:1f18:218b:5902:e5d4:54de:fdc1:24b8:1', False, 2),
```
PabloHiro, line 215: Lines 213 and 215 should throw errors too

ansible_base/lib/utils/address.py
jshimkus-rh: Added this discussion of the regex.
```
    # The regex may appear more complicated than it actually is. First you can
    # ignore the "?:" which simply makes the group in which it appears
    # non-capturing.
    #
    # Second you can conceptually collapse the portions delineated by "{" and
    # "}" to "*". The bracked portions simply put a limit on minimum and
    # maximum lengths of the preceding construct.
    #
    # With the two above changes you'll see that there's effectively only one
    # construct in the regex:
    #
    #   [a-zA-Z]([a-zA-Z0-9-]*[a-zA-Z0-9])?
    #
    # It's second usage simply includes the requirement for a leading "." and
    # allows it to be specified zero or more times.
```

test_app/tests/lib/redis/test_client.py
tznamena: Agree with previous comment(address must be between `[` and `]` when followed by port number) and I would suggest to use a different number for port number to make it more obvious. Since segments of the address are hex, using a five digit number for port number would make it very obvious.

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,156 @@
...
+    The address and port fields are not valid in any sense if the type field is
+    AddressType.UNKNOWN.
+    """
+
+    type: AddressType
+    address: str = ""
+    port: str = ""
+
+
+def _classify_base_address(address: str) -> AddressType:
```
tznamena: There are two methods here with very similar name, same parameters, same return type and very similar doc(1 sentence difference).
Can we rename these or consolidate them?

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,156 @@
...
+        return AddressType.HOSTNAME
+
+    return AddressType.UNKNOWN
+
+
+def _classify_address(address: str) -> AddressType:
+    """
+    Categorizes a given string as IPv4, IPv6, hostname, or unknown.
+
+    An IPv6 address wrapped in [] is considered a hostname.
```
tznamena: Does it make sense to support address wrapped in `[` and `]` but not have a port following it?

ansible_base/lib/utils/address.py
jshimkus-rh: They are functionally different in that one (the base version) classifies whatever it is passed.  The non-base version adds the special handling of []-enclosed IPv6.
We can't reduce to one and simply recurse lest we end up interpreting something like this, `[[[[[[[[<ipv6>]:<port>]]]]]]]`, as bracketed IPv6.
We'd have to duplicate code.

ansible_base/lib/utils/address.py
jshimkus-rh: Yes.  E.g., `[::1]` as the IPv6 localhost to use in a URL or any address utilizing the default HTTP port.

ansible_base/lib/utils/address.py
jshimkus-rh: Changed to categorize IPv6, wrapped or not, as IPv6.  Added a property to the AddressTypeResponse which will return the bracketed IPv6 or None if the address type is not IPv6.

test_app/tests/lib/redis/test_client.py
jshimkus-rh: As discussed with @PabloHiro in the IPv6 WG channel unless/until the current mixture of IPv6 specification is standardized and implemented we cannot impose a standard here without potentially breaking deployments.

test_app/tests/lib/utils/test_address.py
jshimkus-rh: As discussed with @PabloHiro in the IPv6 WG channel unless/until the current mixture of IPv6 specification is standardized and implemented we cannot impose a standard here without potentially breaking deployments.

dhaustein (approved):

ansible_base/lib/redis/client.py
```diff
@@ -160,22 +161,27 @@ def _get_hosts(self) -> None:
...
             host_ports = self.redis_hosts.split(',')
             for host_port in host_ports:
-                try:
-                    node, port_string = host_port.split(':')
-                except ValueError:
+                response = address.classify_address(host_port)
+                if response.type == address.AddressType.UNKNOWN:
+                    logger.error(
+                        f"Specified cluster_host {host_port} is not valid; "
+                        "it is of an unknown address type and must be one of <hostname>:<port>, <ipv4>:<port> or <ipv6>:<port>"
```
john-westcott-iv: Would this be more accurate?
```suggestion
                        "it is of an unknown address type and must be one of <hostname>:<port>, <ipv4>:<port> or [<ipv6>]:<port>"
```

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,159 @@
...
+
+    Strings are used for the address and port so as to minimize changes to
+    existing code to facilitate use of the classification functionality
+    provided.  An empty string indicates the non-existence of that particular
+    attribute as part of the classified address.
+    """
+
+    type: AddressType
+    address: str = ""
+    port: str = ""
```
john-westcott-iv: Non-blocking just curious why empty string instead of None?

ansible_base/lib/utils/address.py
```diff
@@ -0,0 +1,159 @@
...
+
+    # Check that the address is at least two characters long.
+    if (len(address) >= 2) and (address[0] == "[") and (address[-1] == "]"):
+        response = _classify_base_address(address[1:-1])
+
+        # Only an address type of IPv6 is considered valid here.
+        # We don't want to treat any other specification as valid.
+        if response.type == AddressType.IPv6:
+            return response
+        return AddressTypeResponse(AddressType.UNKNOWN, address)
```
john-westcott-iv, line 131: Can we add a comment in here? Or, alternatively, should we assume its a hostname if its not an IPv6 address? Would we care if someone did `[<hostname>]` or `[<ip v4 address]`?
```suggestion
        # Returning unknown because we gor a [<address>] type string but the address inside wasn't IPv6 but nothing else should be encapsulated with []
        return AddressTypeResponse(AddressType.UNKNOWN, address)
```

john-westcott-iv (approved):
Approving with minor comments/questions.

ansible_base/lib/redis/client.py
jshimkus-rh: It's probably better that way.  I considered adding both IPv6 but that seemed extreme.

ansible_base/lib/utils/address.py
jshimkus-rh: Originally (when returning a tuple) to try to minimize necessary client code changes to just a call to the functionality.  Now, None probably makes more sense.

ansible_base/lib/utils/address.py
jshimkus-rh: Adding a comment is fine.  We want to reject anything that should not be wrapped in brackets.  I'll add the comment separately 'cause of the misspelling. 😄

bhavenst (approved):

PR #727, opened 2025-05-15
AAP-36738: enhance ldap filter validation

bhavenst:
## Description

Please review/approve as desired, but DO NOT MERGE until https://issues.redhat.com/browse/AAP-46053 is closed (will break PDE builds before that point).

Use ldap-filter package to do LDAP filter validation rather than the regex we had before.

A customer had a problem where we were declaring a valid filter invalid.  Their filter is added to the unit test and now validates OK.

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Create ldap authenticator with filter like customer's filter (or whatever you like).  See bug.
2. Log in and see that it works OK.
3. 

### Expected Results
All valid filters should be declared valid.

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/ldap.py
+++ b/ansible_base/authentication/authenticator_plugins/ldap.py
@@ -5,6 +5,7 @@ from collections import OrderedDict
 from typing import Any
 
 import ldap
+from ldap_filter import Filter, ParseError
 from django.utils.translation import gettext_lazy as _
 from django_auth_ldap import config
 from django_auth_ldap.backend import LDAPBackend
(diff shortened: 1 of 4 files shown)


ansible_base/authentication/authenticator_plugins/ldap.py
```diff
@@ -183,14 +184,14 @@ def validate_ldap_filter(value: Any, with_user: bool = False) -> None:
 
         dn_value = value.replace(user_search_string, 'USER')
 
-    if re.match(r'^\([A-Za-z0-9-]+?=[^()]+?\)$', dn_value):
-        return
-    elif re.match(r'^\([&|!]\(.*?\)\)$', dn_value):
+    if re.match(r'^\([&|!]\(.*?\)\)$', dn_value):
```
john-westcott-iv, line 188: Can you add a comment to this regex for future people looking at this code?

john-westcott-iv (approved):

ansible_base/authentication/authenticator_plugins/ldap.py
```diff
@@ -183,14 +184,15 @@ def validate_ldap_filter(value: Any, with_user: bool = False) -> None:
...
         for sub_filter in dn_value[3:-2].split(')('):
             # We only need to check with_user at the top of the recursion stack
             validate_ldap_filter(f'({sub_filter})', with_user=False)
-        return
-    raise ValidationError(_('Invalid filter: %s') % value)
+    try:
+        Filter.parse(dn_value)
+    except ParseError:
+        raise ValidationError(_('Invalid filter: %s') % value)
```
jshimkus-rh, line 196: What purpose does the sub filter recursion serve?
The totality of the `dn_value` is passed to `Filter.parse` after sub filter recursion which, presumably, would fail if those sub filters were incorrect.

bhavenst: Each piece of the filter is checked independently so that the error message will highlight exactly which part is invalid.  dn_value becomes the subfilter string upon recursion, which then hops over the regex and gets checked.

PR #734, opened 2025-06-02
[AAP-45443] Allow filtering by ansible_id

fosterseth:
### What is being changed?
Allows API query filtering by ansible id
`user_ansible_id`
`team_ansible_id`
`object_ansible_id`

Note, filtering with object_ansible_id accrues an additional database query for the Resource model.

Example API calls

```
/v1/role_user_assignments/?user_ansible_id=<uuid>
/v1/role_user_assignments/?object_ansible_id=<uuid>
/v1/role_team_assignments/?team_ansible_id=<uuid>
```
query params can be combined where appropriate

#### Validation

Adds some validation if the provided ansible_id is not a UUID
```
[
    "Invalid UUID format for user_ansible_id: 80c7e291-b121-48fc-"
]
```

Currently there is no way to filter on these fields, so working with the API becomes cumbersome and inefficient

Adds a django viewset filter backend to alias the query param to the appropriate queryset call.
e.g.
`user_ansible_id` applies a `user__resource__ansible_id` query

## Type of Change

- [ ] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/rbac/api/filter_backends.py
+++ b/ansible_base/rbac/api/filter_backends.py
@@ -1,7 +1,10 @@
+import uuid
+
+from rest_framework.exceptions import ValidationError
 from rest_framework.filters import BaseFilterBackend
+
 from ansible_base.resource_registry.models import Resource
-from django.core.exceptions import ObjectDoesNotExist
-from django.contrib.contenttypes.models import ContentType
+
 
 class AnsibleIdAliasFilterBackend(BaseFilterBackend):
     '''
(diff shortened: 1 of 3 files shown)


AlanCoding:
This is so far superior to the approach I was thinking of taking. I don't know what I was thinking. Totally :100: on the methods here. Just needs checks cleaned up and satisfying the bot.

tznamena:
Is there a Jira item for this work that we could add here?

ansible_base/rbac/api/filter_backends.py
```diff
@@ -0,0 +1,81 @@
...
+
+        return queryset
+
+
+class UserAnsibleIdAliasFilterBackend(AnsibleIdAliasFilterBackend):
+    """
+    Filter backend for user ansible_id.
+
+    Example:
+    /api/v1/role_user_assignments/?user_ansible_id=80c7e291-b121-48fc-8fb1-174aac6f57a6
```
AlanCoding, line 47: Would help to clarify in the docstring that this also enables `object_ansible_id` filtering as per its parent class.

test_app/tests/rbac/test_ansible_id_alias_filter.py
```diff
@@ -0,0 +1,72 @@
...
+    def test_filter_user_ansible_id(self, admin_api_client, org_inv_rd, rando, organization):
+        '''
+        Test filtering RoleUserAssignment by user_ansible_id and object_ansible_id.
+        '''
+        user_resource = Resource.objects.get(object_id=rando.pk, content_type=ContentType.objects.get_for_model(rando).pk)
+        organization_resource = Resource.objects.get(object_id=organization.pk, content_type=ContentType.objects.get_for_model(organization).pk)
+        url = get_relative_url('roleuserassignment-list')
+        data = dict(role_definition=org_inv_rd.id, content_type='shared.organization', user_ansible_id=user_resource.ansible_id, object_id=organization.id)
+        response = admin_api_client.post(url, data=data, format="json")
+        assert response.status_code == 201, response.data
```
AlanCoding, line 22: I think you should check that the total number of user_role_assignments is >1, and if not, assign another permission to any random user for any random object.

Otherwise there's a risk that this test doesn't cover the null case in any way. It's returning the desired object, but it's not clear that it doesn't return undesired objects.

Test is good, I think it's overkill to add null-case testing as a separate test. But doing a two-liner additional assignment would cover the bases fine.

fosterseth: add some logic to ensure we have at least 2 assignments, and that filtering returns the expected number of items

ansible_base/rbac/api/filter_backends.py
fosterseth: updated the doc strings to emphasize object_ansible_id usage

AlanCoding (approved):

PR #735, opened 2025-06-03
AAP-41842: Fix pagination

bhavenst:
## Description
Provide a base ordering via rest filter such that all queries are returned in order of pk (if no other ordering is specified).  This ensures that paginated returns are complete and no records are duplicated.

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Create enough records (users or organizations for example) to exceed one page.  (the page size can also be reduced via setting DEFAULT_PAGE_SIZE if desired), ideally at least 3 pages.
2. Fetch API endpoint (this issue was discovered on service-index/resources/ so that would be a good one to verify)
3. Ensure the content has no missing or duplicated items

### Expected Results
No records are missing and/or duplicated.

### DIFF
--- a/ansible_base/activitystream/views.py
+++ b/ansible_base/activitystream/views.py
@@ -37,9 +37,10 @@ class EntryReadOnlyViewSet(ReadOnlyModelViewSet, AnsibleBaseDjangoAppApiView):
     API endpoint that allows for read-only access to activity stream entries.
     """
 
-    queryset = Entry.objects.order_by('-id')
+    queryset = Entry.objects.all()
     serializer_class = EntrySerializer
     filter_backends = calculate_filter_backends()
+    ordering = ['-id']
 
     def get_permissions(self):
         """
--- a/ansible_base/lib/utils/views/ansible_base.py
+++ b/ansible_base/lib/utils/views/ansible_base.py
@@ -11,6 +11,8 @@ logger = logging.getLogger('ansible_base.lib.utils.views.ansible_base')
 
 class AnsibleBaseView(APIView):
 
+    ordering = ['pk']
+
     # pulp openapi generator compatibility
     endpoint_name = ''
 


bhavenst:
From the looks of the awx and eda CI jobs, it appears that they both have unit tests that rely on the previously unspecified ordering of records returned by a list API.  So not really sure what to do with that.

ansible_base/rest_filters/rest_framework/order_backend.py
```diff
@@ -52,6 +52,8 @@ def get_default_ordering(self, view):
         ordering = getattr(view, 'ordering', None)
         if isinstance(ordering, str):
             return (ordering,)
+        elif ordering is None:
+            return ("pk",)
         return ordering
```
john-westcott-iv: Can ordering be a non-None string here? (i.e. an int) ... after looking at this I'm realizing it might be a tuple. 
Maybe add a comment like:
```suggestion
        if isinstance(ordering, str):
            # If we got just a string turn it into a tuple of one object
            return (ordering,)
        elif ordering is None:
            # If unspecified, than default to the pk (which all objects should have)
            return ("pk",)
        # If we got something else just return it and let Django error if its not valid
        return ordering
```

test_app/tests/activitystream/test_api.py
```diff
@@ -17,18 +17,18 @@ def test_activitystream_api_read(admin_api_client, user):
     response = admin_api_client.get(url)
     assert response.status_code == 200
     assert response.data["count"] > 0
-    assert response.data["results"][0]["operation"] == "create"
+    assert response.data["results"][-1]["operation"] == "create"
```
john-westcott-iv: This is required just because we've reversed the order of the activity stream?
If so, should we consider altering ActivityStream to use a default ordering of created_on to not change behavior?

john-westcott-iv (commented):
I guess just a general question (repeating from another comment). Should we alter the APIs to set an order so that we don't alter how the API appears to be sorted now. We could add a feature flag if we wanted so that people could "turn it off" if they had a script or something relying on the current order (or lack there of).

test_app/tests/activitystream/test_api.py
bhavenst: @john-westcott-iv This struck me as odd since that is indeed the implication, so I am taking another look at this.  It appears that this ordering default is being applied when the view does not contain an order_by but the model does.  It should respect the model's ordering setting too so I'll see how to fix that.  activitystream is sorted in reverse order intentionally, but it's done in the model so it seems like this change overrides it improperly to flip the order.

bhavenst: I tried a couple of things and wound up adding a default ordering to AnsibleBaseView, which is overridden by any view that sets an intentional ordering.  It seems to work well and doesn't require any UT changes.

john-westcott-iv (approved):

PR #736, opened 2025-06-04
[HOLD] AAP-45875 Runtime Feature Flags

zkayyali812:
## Description

https://issues.redhat.com/browse/AAP-45875

- What is being changed?
This change is the foundation for enabling runtime platform feature flags for AAP. This updates the django-ansible-base to be the central location where all platform flags are defined. Components can inherit the `ansible_base.feature_flags` application to inherit all platform feature flag definitions.
- Why is this change needed?
To enable runtime platform feature flags in AAP.
This change addresses the issue by defining a database flag source, which contains all the feature flags along with their associated metadata.
These feature flags are installed into each components database at install-time and kept in sync via resource sync (Gateway is the provider)

Before this can be merged, the following should be done -
1. Confirm each feature flags metadata
    1. Only Dispatcherd and EDA_Analytics feature flags need their metadata confirmed now.

## Type of Change

- [x] Documentation update
- [x] Test update

## Self-Review Checklist

### Steps to Test
Recommend testing this through AAP Dev, but it can be tested directly via the test app as well.

1.  Clone [aap-dev](https://github.com/ansible/aap-dev/)
3. enter aap-dev directory and run `make configure-sources`
4. Use this PR for the DAB source
5. Use this PR for the Gateway source - https://github.com/ansible/aap-gateway/pull/795

### Expected Results
AAP Deploys as expected with database feature flags.

### Notes

This PR is the prerequisite for the following component PRs - 
Gateway - https://github.com/ansible/aap-gateway/pull/795
EDA - https://github.com/ansible/eda-server/pull/1346
Galaxy - https://github.com/ansible/galaxy_ng/pull/2615
AWX - https://github.com/ansible/awx/pull/16030


ansible_base/feature_flags/apps.py
```diff
@@ -1,8 +1,17 @@
...
 
 class FeatureFlagsConfig(AppConfig):
     default_auto_field = 'django.db.models.BigAutoField'
     name = 'ansible_base.feature_flags'
     label = 'dab_feature_flags'
     verbose_name = 'Feature Flags'
+
+    def ready(self):
+        try:
+            create_initial_data()
```
AlanCoding: Creating database objects in the `.ready` call is a very bad pattern. It has happened elsewhere, yes, and it was recognized as a bad thing then, but doing it in DAB would create much more damage.

I see here it connects the same method to the ready call as to the post-migrate signal. Is post-migrate not sufficient? It would probably be _best_ practice to do in actual migration steps, which are also being added here. post-migrate can break reverse migrations and other unexpected things like that.

ansible_base/feature_flags/flag_source.py
```diff
@@ -0,0 +1,25 @@
+from django.apps import apps
+from flags.sources import Condition
+
+
+class DatabaseCondition(Condition):
+    """Condition that includes the AAPFlags database object"""
```
AlanCoding: I'm going to have to admit I can't understand this class and its docstring easily. It would make sense if I could then connect it to what the init method is doing below. But that's very non-obvious.

ansible_base/feature_flags/apps.py
zkayyali812: Hey Alan,
Thank you for the feedback.

Is there a better pattern to use here instead? I'm definitely open to addressing this if so, but in my research so far, this was the best option I had found and was not aware that this could potentially be a bad practice.

The reason I had declared this with/without post migrate is to ensure that if the container is updated after the migration has completed, that it would attempt to update the list of feature flags in the database again.

ansible_base/feature_flags/flag_source.py
zkayyali812: Good note, I will update this docstring to better define it.

zkayyali812: After checking further, this class isn't needed and I removed it.

zkayyali812:
This PR was reviewed in the DAB PR review meeting, before it can be merged, the remaining tasks are outstanding -
Confirm expected flag support levels
Confirm each feature flags metadata

ansible_base/feature_flags/flag_source.py
zkayyali812: I realize now this function was required to ensure that the django-flag functions `enable_flag/disable_flag` can work as intended.
I added this back, clarified the doc string, and added a test to cover this case.

AlanCoding:
Downstream failures, starting to hit the database for references like this in AWX:

```
  File "/awx_devel/awx/main/tasks/jobs.py", line 976, in build_env
    if flag_enabled("FEATURE_INDIRECT_NODE_COUNTING_ENABLED"):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

These are probably expected. This is in the job execution path, so that's not too big of a deal.

docs/apps/feature_flags.md
```diff
@@ -5,55 +5,32 @@ Additional library documentation can be found at https://cfpb.github.io/django-f
 
 ## Settings
 
-Add `ansible_base.feature_flags` to your installed apps:
+Add `ansible_base.feature_flags` to your installed apps and ensure `ansible_base.resource_registry` as added to enable flag state to sync across the platform:
```
AlanCoding, line 8: Don't you also need to register the flag model in the resource registry config?

Just trying to complete the circle for ability for someone to understand. That should be mentioned here if I understand. Or mention briefly how you're doing that automatically, if that's the case.

ansible_base/resource_registry/shared_types.py
```diff
@@ -75,3 +75,23 @@ class TeamType(SharedResourceTypeSerializer):
         default="",
         allow_blank=True,
     )
+
+
+class FeatureFlagType(SharedResourceTypeSerializer):
+    """Serialize list of feature flags"""
```
AlanCoding: Not really for a list.

ansible_base/feature_flags/urls.py
```diff
@@ -1,12 +1,16 @@
...
+from ansible_base.lib.routers import AssociationResourceRouter
 
 app_name = FeatureFlagsConfig.label
 
-api_version_urls = [
-    path('feature_flags_state/', views.FeatureFlagsStateListView.as_view(), name='feature-flags-state-list'),
-]
+router = AssociationResourceRouter()
+
+router.register(r'feature_flags/states', views.FeatureFlagsStatesView, basename='aap_flags_states')
```
AlanCoding, line 11: The serializer for this has an awfully small number of fields. The model has many more fields. Even the management command gives more information.

Were you intending to add some other endpoint which would give the UI the information necessary to render choices and stuff?

ansible_base/feature_flags/utils.py
```diff
@@ -1,5 +1,97 @@
...
+            if e.messages[0] == 'Aap flag with this Name and Condition already exists.':
+                logger.info(f"Feature flag: {flag['name']} already exists")
+            else:
+                error_msg = f"Invalid feature flag: {flag['name']}. Error: {e}"
+                logger.error(error_msg)
+
+
+def delete_feature_flags():
+    """
+    If a feature flag has been removed from the platform flags list, delete it from the database.
```
AlanCoding: Strange between docstring & method name. Maybe it would flow more smoothly if named `purge_feature_flags`. "delete" makes me expect a flag name as argument.

ansible_base/feature_flags/views.py
```diff
@@ -1,25 +1,41 @@
...
+from ansible_base.lib.utils.views.django_app_api import AnsibleBaseDjangoAppApiView
+from ansible_base.lib.utils.views.permissions import IsSuperuserOrAuditor, try_add_oauth2_scope_permission
 
 from .utils import get_django_flags
 
 
-class FeatureFlagsStateListView(AnsibleBaseView):
+class FeatureFlagsStatesView(AnsibleBaseDjangoAppApiView, ModelViewSet):
+    """
+    A view class for displaying feature flags states
```
AlanCoding: Might be help to explain where they should go to change them?

Makes sense that this is read-only. But it would be really really helpful to direct users to where they need to be.

AlanCoding (changes requested):
I'm good with the decisions made here, totally ready to :+1: this

Almost all my comments are to improve documentation, very simple things.

I'm just unclear if the API endpoints (views, serializers) are finished. I'm confused about that, in terms of what the UI needs to be delivered with this, what the response looks like, and if that's planned for later or anything.

docs/apps/feature_flags.md
zkayyali812: Hey Alan,
This is mentioned directly below - 
https://github.com/ansible/django-ansible-base/pull/736/files#diff-c7bfc38afdd2f4d54a88f9d3035edcf2ed284bcd517ea086f38f34aff403f14eR27-R33

ansible_base/feature_flags/urls.py
zkayyali812: Hey Alan,
This is intentional for this endpoint.

The endpoint which the UI would use, which contains the remainder of fields, is only exposed via gateway. 
Please see this [view](https://github.com/ansible/aap-gateway/pull/795/files#diff-c636e8a18790306b0cce5de71da895bd085ea6ab342f95f1510f0370c46dbf41R14)

And its corresponding [serializers](https://github.com/ansible/aap-gateway/pull/795/files#diff-c5a86b58f9e7b848abdfdd3dd012793fd9da1b9d62d1900f8c3caaae21161f1fR7)

zkayyali812: This DAB endpoints is intentionally minimal currently. Only gateway exposes the full set of metadata, as it is the only service which needs it

ansible_base/feature_flags/utils.py
zkayyali812: Great note, I will update

ansible_base/feature_flags/views.py
zkayyali812: Great note, I will put a note here to guide folks to the readme, which contains the instructions about this.

zkayyali812:
Hey @AlanCoding,

Thanks for the feedback. I just updated to address all of these notes.

Regarding your comments -
1. The AWX and EDA tests here in DAB should be fixed by their respective PRs - 
    1. [AWX PR](https://github.com/ansible/awx/pull/16030)
    2. [EDA PR](https://github.com/ansible/eda-server/pull/1346)
2. The API endpoints are finished here. The remainder of the metadata for them is currently intentionally exposed only via gateway, please see [Gateway PR](https://github.com/ansible/aap-gateway/pull/795/files#diff-c636e8a18790306b0cce5de71da895bd085ea6ab342f95f1510f0370c46dbf41R14)

AlanCoding (approved):

fao89:
I can't push to this PR, so I started this: https://github.com/ansible/django-ansible-base/pull/875

AlanCoding:
Is this PR defunct?

zkayyali812:
@AlanCoding , Yes this PR is no longer needed. I will close it out.

PR #743, opened 2025-06-06
AAP-47292 feature flags response description

bhavenst:
## Description
Just a small addition to fix openapi docs for feature flags (describe the response type).

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test (test w/ gateway)
1. Load docs without this change, see no response documented for feature flags endpoint
2. Include this change, load again and see the response type documented
3. 

### Expected Results
Response type/example shown in openapi docs

### DIFF
--- a/ansible_base/feature_flags/views.py
+++ b/ansible_base/feature_flags/views.py
@@ -1,4 +1,5 @@
 from django.utils.translation import gettext_lazy as _
+from drf_spectacular.extensions import OpenApiViewExtension
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiExample, extend_schema
 from rest_framework.response import Response
@@ -9,7 +10,6 @@ from ansible_base.lib.utils.views.ansible_base import AnsibleBaseView
 from .utils import get_django_flags
 
 
-@extend_schema(request=None, responses=OpenApiTypes.OBJECT, examples=[OpenApiExample(name="featureflags", value={"FLAG1": True, "FLAG2": False})])
 class FeatureFlagsStateListView(AnsibleBaseView):
     """
     A view class for displaying feature flags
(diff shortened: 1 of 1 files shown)


BrennanPaciorek (approved):

ansible_base/feature_flags/views.py
```diff
@@ -7,6 +9,7 @@
 from .utils import get_django_flags
 
 
+@extend_schema(request=None, responses=OpenApiTypes.OBJECT, examples=[OpenApiExample(name="featureflags", value={"FLAG1": True, "FLAG2": False})])
```
BrennanPaciorek: Can we be sure that drf_spectacular is installed at all times? that particular dependency is behind the docs dependencies.

BrennanPaciorek: ie: we may want to have our own decorator util in DAB to check `INSTALLED_APPS` for `ansible_base.api_documentation` before importing drf_spectacular and trying to extend the schema.

bhavenst: Yeah, that's a good idea.  It's meant to be optional to use api_documentation so anything that would be loaded if that app isn't loaded should probably have a check.  We have a couple of other places like that too, but I'll see if I can find a nice way to handle that.

BrennanPaciorek: https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/api_documentation/customizations.py has a different approach, where I'm guessing we can extend the view schemas as well. 

I'm guessing we'd want to keep the schema extensions where the view is defined though.

bhavenst: Those are normally intended to add documentation to things defined outside the code (libraries, dependencies etc.) but seems reasonable to use for the purpose of loading only if docs are enabled.  It doesn't solve the import issue, but I switched to that format to match what was done in https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/jwt_consumer/eda/auth.py and so that the view is only modified when docs are enabled.

bhavenst: Which didn't work for exactly the reason that it doesn't prevent imports..  AWX consumer test failed since it couldn't import drf_spectacular.  The other one I pointed out isn't loaded by AWX.  So will need a different solution.

ansible_base/feature_flags/views.py
```diff
@@ -17,9 +18,23 @@ class FeatureFlagsStateListView(AnsibleBaseView):
...
-    def get(self, request, format=None):
+    def _get(self, request, format=None):
         self.serializer = FeatureFlagSerializer()
         return Response(self.serializer.to_representation())
 
     def get_queryset(self):
         return get_django_flags()
+
+    # Conditionally add openapi documentation for feature flags
+    if 'ansible_base.api_documentation' in settings.INSTALLED_APPS:
```
john-westcott-iv, line 29: I was torn between this and `get_setting('INSTALLED_APPS', [])` but it looks like we overwhelming just pull this from settings.

john-westcott-iv (approved):

ansible_base/feature_flags/views.py
bhavenst: Your option is definitely better for questionable settings but it's super unlikely we'd ever encounter no INSTALLED_APPS (and all hell would be breaking loose in that case anyway :) ).

PR #746, opened 2025-06-25
AAP-44925 Make auth map attr and group comparison case insensitive

bhavenst:
## Description

Does case insensitive comparison for authenticator map attributes, values, and group names:

```
"attributes": {
    "InSenSiTiVe_Attr_nAme": {
        "contains": "value"
    }
}

"groups": {
  "has_or": [
    "cN=auDitors,Ou=GroupS,DC=example,dc=ORG"
  ]
}
```

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites
Start gateway w/ LDAP sidecar.

### Steps to Test
1. Create an auth map using random case for a known user attribute and with a noticeable result (adds user to org for instance).  givenName for instance in the LDAP server arrives as "givenname", but create a map for attribute "GIVENnAME"
2. Log in with user matching attribute value and ensure user is added to org (or whichever effect you configured).
3. Repeat the above, but with another authenticator map targeting an LDAP group with mixed case.

### Expected Results
Maps use case insensitive comparison for attribute names and group DNs

### DIFF
--- a/ansible_base/authentication/utils/claims.py
+++ b/ansible_base/authentication/utils/claims.py
@@ -181,24 +181,28 @@ def _add_rbac_role_mapping(has_permission, role_mapping, role, organization=None
 def process_groups(trigger_condition: dict, groups: list, authenticator_id: int) -> TriggerResult:
     """
     Looks at a maps trigger for a group and users groups and determines if the trigger is defined for this user.
+    Group DNs are compared case-insensitively.
     """
 
     invalid_conditions = set(trigger_condition.keys()) - set(TRIGGER_DEFINITION['groups']['keys'].keys())
     if invalid_conditions:
         logger.warning(f"The conditions {', '.join(invalid_conditions)} for groups in mapping {authenticator_id} are invalid and won't be processed")
 
-    set_of_user_groups = set(groups)
+    set_of_user_groups = set([group.casefold() for group in groups])
 
     if "has_or" in trigger_condition:
-        if set_of_user_groups.intersection(set(trigger_condition["has_or"])):
+        trigger_groups = set([group.casefold() for group in trigger_condition["has_or"]])
+        if set_of_user_groups.intersection(trigger_groups):
             return TriggerResult.ALLOW
 
     elif "has_and" in trigger_condition:
-        if set(trigger_condition["has_and"]).issubset(set_of_user_groups):
+        trigger_groups = set([group.casefold() for group in trigger_condition["has_and"]])
+        if trigger_groups.issubset(set_of_user_groups):
             return TriggerResult.ALLOW
 
     elif "has_not" in trigger_condition:
-        if not set(trigger_condition["has_not"]).intersection(set_of_user_groups):
+        trigger_groups = set([group.casefold() for group in trigger_condition["has_not"]])
+        if not trigger_groups.intersection(set_of_user_groups):
             return TriggerResult.ALLOW
 
     return TriggerResult.SKIP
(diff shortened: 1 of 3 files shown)


bhavenst:
Will add UT

bhavenst:
I can't explain the coverage report.  The lines it is claiming aren't covered are well covered by the unit tests.  I ran it locally and looked at the report and it is strange and inconsistent, like claiming that a statement always results in false, but then showing coverage that would only happen if it were true, etc.


Coverage was in the low 60%s but it seems to have resolved itself.

john-westcott-iv:
So for clarity, this is doing lowercase on attribute names (not values) and lowercase on group names? Or am I reading this wrong?
Looks like maybe [this line](https://github.com/ansible/django-ansible-base/pull/746/files#diff-7f6b8f4df1be270aa850255184d42657f290642f5ac3751fb7adadbd03a7a421R259) is doing the lowercase of the attribute values.

bhavenst:
Updated the description.  It's names/keys, values and group names, so it kind of expanded from the original ticket after talking it all through w/ Ron..

ansible_base/authentication/utils/claims.py
```diff
@@ -178,10 +179,24 @@ def _add_rbac_role_mapping(has_permission, role_mapping, role, organization=None
             logger.warning(f"Role mapping is not possible, organization for team '{team}' is missing")
 
 
+def _lowercase_group_names(trigger_condition: dict) -> dict:
```
jshimkus-rh: Maybe include `trigger` in the name of the function?  Other code (e.g., line 198) does inline group case folding and this name is pretty generic which might confuse folk?

ansible_base/authentication/utils/claims.py
```diff
@@ -218,10 +233,35 @@ def has_access_with_join(current_access: Optional[bool], new_access: bool, condi
...
+    return ci_trigger_condition
+
+
 def process_user_attributes(trigger_condition: dict, attributes: dict, authenticator_id: int) -> TriggerResult:
     """
     Looks at a maps trigger for an attribute and the users attributes and determines if the trigger is defined for this user.
+    Attribute names are compared case-insensitively.
     """
+    if flag_enabled("FEATURE_CASE_INSENSITIVE_AUTH_MAPS"):
+        attributes = {f"{k}".casefold(): v for k, v in attributes.items()}
```
jshimkus-rh, line 266: The key can be something other than a string?  I.e., we have to coerce it via an f-string rather than directly k.casefold()?

ansible_base/authentication/utils/claims.py
```diff
@@ -265,7 +305,7 @@ def process_user_attributes(trigger_condition: dict, attributes: dict, authentic
         for a_user_value in user_value:
             # We are going to do mostly string comparisons, so convert the attribute to a
             #  string just in case it came back as an int or something funky
-            a_user_value = f"{a_user_value}"
+            a_user_value = f"{a_user_value}".casefold() if flag_enabled("FEATURE_CASE_INSENSITIVE_AUTH_MAPS") else f"{a_user_value}"
```
jshimkus-rh: Given the more than one use of `flag_enabled("FEATURE_CASE_INSENSITIVE_AUTH_MAPS")` I'd like to see a separate boolean function that encapsulates this.

ansible_base/authentication/utils/claims.py
bhavenst: sure, will do

ansible_base/authentication/utils/claims.py
bhavenst: Sure

ansible_base/authentication/utils/claims.py
bhavenst: Belt and suspenders I guess.  That was already being done elsewhere so I added it here too.

jshimkus-rh (approved):

john-westcott-iv (approved):
One nit here... I might have had a function that was like:
```
def lowercase_string_if_needed(input: str) -> str:
    if _is_case_insensitivity_enabled():
        return input.casefold()
    return input
```

Then you don't need the if's and casefold all over the place. But I won't hold this PR up on this.

PR #749, opened 2025-06-30
[AAP-48392] Models and APIs for tracking remote permissions in DAB RBAC

### LINKED ISSUE #80: [RBAC] Feature to include permissions of external resources, with no corresponding local model
The RBAC proposal https://github.com/ansible/django-ansible-base/pull/45 creates several tables for the purpose of the app, and has a fairly loose connection to the models that it applies for. Relationships are typically GenericForeignKey type (in its _most_ bound form) which manage things like cascade deletion at the application level (if at all).

Because of this, there should be no fundamental blocker to managing RBAC for a model that doesn't exist on the local server, and instead exists in a remote server. The only problem is that we have to run some querysets in order to support features like resource permission inheritance. Solutions could be:

1. sync a representational entry for every _object_ in the remote system to the local system so that the `RoleEvaluation` entries can be filled in, and object-level access can be shown
2. don't bother with object-level access of remote objects, and just skip all re-computation logic for external objects

Right now I think I prefer (2), which still has some implementation challenges because of source-of-truth issues, but I looked at that before and believed that an acceptable solution would be possible.

Of course the big missing piece is a replacement for the **content type** model which can include entries for non-local models. The RBAC system should be flexible enough to customize the content type... DAB probably needs to provide at least an abstract model since certain extra fields will be needed, like an indicator that a type is external, and probably a representation of _which_ external system it exists in.

AlanCoding:
## Description
This modifies the RBAC app so that the models can store permissions for remote objects - objects that don't actually exist in the local server. To know which are which, a `service` field is added to our type-tracking model, which is also new as of this work. Importantly, permission _evaluations_ can be done for both local items and remote items.

Why? Just as we have synchronization to a "resource server" via the resource registry app, this allows you to appoint a single service to be the gatekeeper for RBAC. This still requires synchronization, making it different from other approaches. Several new endpoints under `/service-index/` are introduced to help facilitate that synchronization.

EDITing some snapshots of the progress state

 - As of opening, the core code is not  finished being written, just want to get CI output continuously.
 - As of July 14 - relatively stable with current AWX, but integration work for rest of components is still WIP

Fixes https://github.com/ansible/django-ansible-base/issues/80

Also request review from @dleehr @TheRealHaoLiu 

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
Some tests show how you can use this in isolation for types & permissions. That means, you can call `DABContentType.objects.load_remote_objects` to import some remote types, and then make roles using these types, and assign permissions.

But that doesn't do much unless you set up a resource server and another service for it to track, and synchronize permissions between them. For this, you pretty much need aap-dev. Watching the demos also might be good to see test cases. Post-install those are mainly:

 - modify a role definition
 - assign a permission
 - repeat with the API from the resource server & the other server (AWX probably)

### Expected Results
Role definitions & assignments synchronized using the endpoint system added here.

## Additional Context
See backlinks for PRs that adopt this PR

### Required Actions

- [ ] Blocked by PR/MR: #XXX


AlanCoding:
Documenting the eda-server test failure I'm troubleshooting, log:

```
2025-07-02 19:16:07,056 WARNING  ansible_base.rbac.models.content_type Could not find content type for ('eda', 'core', 'awxtoken'), so creating new, out of:
dict_keys([('eda', 'core')])
```

failure:

```
FAILED tests/integration/api/test_activation_with_event_stream.py::test_create_activation_with_event_stream - django.db.utils.IntegrityError: duplicate key value violates unique constraint "dab_rbac_dabcontenttype_pkey"
DETAIL:  Key (id)=(1) already exists.
```

It tried to create at:

```
  File "/home/runner/work/django-ansible-base/django-ansible-base/ansible_base/rbac/models/role.py", line 82, in give_creator_permissions
    cts = DABContentType.objects.get_for_models(*model_and_children).values()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/django-ansible-base/django-ansible-base/ansible_base/rbac/models/content_type.py", line 145, in get_for_models
    ct = self.create(
         ^^^^^^^^^^^^
```

I might have ran into the indexing bug, because I set a pk value in migrations for the types, but the current pk index for new objects does not increment in some databases. This might be a good lead.

Apparently this is called "sequence desynchronization"

AlanCoding:
The higher-level error is that AwxToken is not a registered permission model

https://github.com/ansible/eda-server/blob/b6c03d41936fcce4a842760cf3f101c9903ca9df/src/aap_eda/core/models/__init__.py#L69

So that suggests that `give_creator_permissions` was, in some way, wrong to determine that it is a child model of something else. Or else the tests added it unexpectedly.

AlanCoding:
Update on downstream tests:

```
FAILED awx/main/tests/functional/dab_rbac/test_dab_rbac_api.py::test_custom_read_role
```

with

> ansible_base.resource_registry.models.resource.Resource.DoesNotExist: Resource matching query does not exist.

This is a novel failure, due to adding `resource` on to the RoleDefinition model. This needs a change in AWX which is not there yet. It is also extremely challenging to add that to AWX in a smooth way to keep tests passing. I am considering a new hack for that.

AlanCoding:
this explains some test failures:

```
permission_registry.register(
    Activation,
    EdaCredential,
    CredentialInputSource,
    DecisionEnvironment,
    Project,
    Organization,
    Team,
    EventStream,
    parent_field_name="organization",
)
```

A new model was added `CredentialInputSource` that did not exist in prior model states. Naturally, the reverse migration isn't working, because the model is in the registry, but does not exist in that prior model state.

What a crazy situation.

djyasin (approved):

ansible_base/rbac/service_api/router.py
```diff
@@ -0,0 +1,11 @@
+from ansible_base.lib.routers import AssociationResourceRouter
+
+from . import views
+
+service_router = AssociationResourceRouter()
+
+service_router.register(r'role-types', views.RoleContentTypeViewSet)
+service_router.register(r'role-permissions', views.RolePermissionTypeViewSet)
+# Different basenames used to distinguish between the duplicate, public, endpoints
+service_router.register(r'role-user-assignments', views.ServiceRoleUserAssignmentViewSet, basename='serviceuserassignment')
```
dleehr, line 10: Well I like this 👍

ansible_base/rbac/api/queries.py
```diff
@@ -0,0 +1,35 @@
+from typing import Union
+
+from django.db.models import Model
+
+from ..models import DABContentType, get_evaluation_model
+from ..remote import RemoteObject
+
+
+def assignment_qs_user_to_obj(actor: Model, obj: Union[Model, RemoteObject]):
+    """Querset of assignments (team or user) that grants the actor any form of permission to obj"""
```
zkayyali812:
```suggestion
    """Queryset of assignments (team or user) that grants the actor any form of permission to obj"""
```

ansible_base/rbac/management/__init__.py
```diff
@@ -84,8 +123,19 @@ def create_dab_permissions(app_config, verbosity=2, interactive=True, using=DEFA
...
+    perms_by_model_name = defaultdict(list)
     for perm in perms:
-        logger.debug("Adding permission '%s'" % perm)
+        perms_by_model_name[perm.content_type.model].append(perm.codename)
+    for model_name, codenames in perms_by_model_name.items():
+        codename_prnt = ', '.join(codenames)
+        logger.debug(f"Added DAB permissions for model {model_name}: {codename_prnt}")
+
+    # Reset the sequence to avoid PK collision later
+    if connection.vendor == 'postgresql' and new_cts:
```
zkayyali812, line 141: Is this potentially an area we would need to revisit if we ever decide to support other DBs?
Wondering if we would even need this logic if we could use UUIDs instead of integers for primary keys, that would be a bigger/later refactor though potentially

zkayyali812 (approved):
Approving with minor notes. I was able to deploy this successfully with aap-dev, along with the other component PRs, and the system appears to function as expected. I was not able to drill too deeply into that yet.
I believe it would be good to get this in sooner than later to let this bake, and if any further issues are identified we can address them as they come up.

ansible_base/rbac/management/__init__.py
AlanCoding: I've got to go through a couple of layers to explain why the sequence reset is necessary:

 - AWX still has a _lot_ of logic around the _old_ RBAC API that has existed since Ansible Tower 3.0. Absurd as this is, we still support a `/api/v2/job_templates/42/access_list/` endpoint, which this gives a replacement for. That, and filtering of what roles a user can see, are the most complex still-active logic for that old RBAC API.
   - It's just kind of not realistic to refactor that. Like, it would be beyond me to migrate the old Ansible Tower role models to use the new DABContentType model introduced here. So the only way to avoid rewriting that is to set the new model's pk values to the canonical `ContentType` model, which works.
     - setting PK values introduces a new problem, that _only in postgres_, you will "outrun" your current sequence pointer. This introduces a conflict when creating a new item that the id clashes with an existing id.

So this is like a problem due to another problem's fix due to a temporary measure. If you introduced a new database type, it's not at all obvious that this would be a problem, just as it's not a problem in sqlite.

Also, what makes this even less of a problem, is that only _this app_ should be creating DABContenType objects. When that's done, we can just again set the pk value. Only people manually mucking with the database or writing tests would ever even hit the problem, I'm seriously split on whether this SQL should even be included for this reason.

sean-m-sullivan (approved):

BrennanPaciorek (approved):

PR #754, opened 2025-07-07
[AAP-48878] Add email field to AuthenticatorUser

zkayyali812:
## Description
- What is being changed?
The AuthenticatorUser model is extended with the `email` field. This PR is only to add the field, which will be populated by a later ticket- https://issues.redhat.com/browse/AAP-48066.
- Why is this change needed?
This change is needed to associate AuthenticatorUsers directly with their email address. This will make it easier to debug issues regarding how users authenticate into the platform, since we will be able to tell directly which email address their authenticatoruser is configured with. When authenticating via multiple different authenticators, we will sign the user into the AAP account associated with their email.
This change addresses the issue by adding the email field to the AuthenticatorUser table, which will be later populated by AAP-48066.

## Type of Change

## Self-Review Checklist

### Steps to Test
1. 
2. 
3. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX


ansible_base/authentication/models/authenticator_user.py
```diff
@@ -42,6 +42,7 @@ class AuthenticatorUser(AbstractUserSocialAuth, AbstractCommonModel):
         on_delete=models.CASCADE,
         help_text=_("The local DB user related to this authenticator user."),
     )
+    email = models.EmailField(default=None, null=True, blank=True, help_text=_("The e-mail associated with this authenticator user."), max_length=250)
```
john-westcott-iv: I was initially concerned with the 250 length here but I think its likely fine.

john-westcott-iv (approved):

ansible_base/authentication/models/authenticator_user.py
zkayyali812: I can check if we can remove this, I had initially added this, as I had initially used a CharField, and max_length was required for that field

zkayyali812: I can remove this, by default an emailfield has a max length of 254 - 
https://docs.djangoproject.com/en/5.2/ref/models/fields/#emailfield

john-westcott-iv (approved):

chrismeyersfsu (approved):

PR #757, opened 2025-07-10
AAP-21301 Make serializer accept jinja like syntax

john-westcott-iv:
## Description

- What is being changed?
Allowing the authenticator_map serializer to store `{% for_attr_value(something) %}`

- Why is this change needed?
This stages authenticator_maps to support a missing SAML option.

This will allow the authenticator maps to have templates in them.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Log into your instance and create an authenticator_map with different formats of {% in it.
2. 
3. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/authentication/migrations/0018_authenticatoruser_email.py
+++ b/ansible_base/authentication/migrations/0018_authenticatoruser_email.py
@@ -15,4 +15,19 @@ class Migration(migrations.Migration):
             name='email',
             field=models.EmailField(blank=True, default=None, help_text='The e-mail associated with this authenticator user.', max_length=254, null=True),
         ),
+        migrations.AlterField(
+            model_name='authenticatormap',
+            name='organization',
+            field=models.CharField(blank=True, default=None, help_text='An organization name this rule works on. Will expand {% for_attr_value(user_orgs) %} syntax', max_length=512, null=True),
+        ),
+        migrations.AlterField(
+            model_name='authenticatormap',
+            name='role',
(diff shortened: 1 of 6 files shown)


john-westcott-iv:
I'm not sure why code overage doesn't think I'm at 100%. If you look in Sonar it thinks a couple if statements aren't covered but the lines in the if/else are green so it got to both sides. 🤷

zkayyali812 (approved):

BrennanPaciorek (commented):
Have we considered just actually using Jinja2 for this??
- https://jinja.palletsprojects.com/en/stable/sandbox/

This seems like we can heavily restrict the operations allowed to just be able to expand the list of attrs passed in.

Downsides to just using Jinja2 would be that we'd have to introduce a dependency on it only to use a very small subset of the features, and that we'd have to be very deliberate setting up the sandbox to make sure we do not allow access to any potentially dangerous filters.

I'm only half-heatedly suggesting that one, since we really just want to pass in variables to the input, then allow the authenticator_maps to expand those variables.

BrennanPaciorek (commented):
Should we also at least be able to return a list of expandable expressions in some way that can be consumed by our variable-expansion logic?
- https://docs.python.org/3/library/re.html#re.findall
- https://docs.python.org/3/library/re.html#re.Match.start

A slight modification to `has_expansion` to return None if no matches, return list of all matches with `re.findall`, and name the `.*` with `.*<expression-value>`, would get us in a better shape to implement the rest of the logic; we'd just need a note to prevent someone from attempting any sort of recursive expansions.

BrennanPaciorek (approved):
I'm fine with these changes, and the rest of the changes are in the follow-up, so I'll approve as well.

PR #758, opened 2025-07-10
AAP-48726: Fallback to controller auth

zkayyali812:
## Description
- What is being changed?
Updates the local authenticator to fallback to controller authentication when logging in via gateway local authenticator.
- Why is this change needed?
This change is needed to support users logging in in the following case -
1. A users AAP account was migrated over from 2.4->2.6 (this causes the account to exist in both AAP and gateway, but gateway is not aware of the user password, since we can't migrate over the password).
2. When a user attempts to login via gateway the initial auth fails, we check -
    a. If the user actually exists in gateway as a local user
    b. If the user is a valid user in controller, and we check if we can authenticate successfully via the `/api/controller/v2/me/` endpoint
3. If checks pass, we update the gateway user account with the updated password.

This change addresses the issue by adding the necessary fallback logic.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. 
4. 
5. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/local.py
+++ b/ansible_base/authentication/authenticator_plugins/local.py
@@ -1,16 +1,24 @@
 import logging
+from urllib.parse import urljoin
 
+import requests
+from django.contrib.auth import get_user_model
 from django.contrib.auth.backends import ModelBackend
 from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy as _
+from requests.auth import HTTPBasicAuth
 
 from ansible_base.authentication.authenticator_plugins.base import AbstractAuthenticatorPlugin, BaseAuthenticatorConfiguration
 from ansible_base.authentication.utils.authentication import determine_username_from_uid, get_or_create_authenticator_user
 from ansible_base.authentication.utils.claims import update_user_claims
+from ansible_base.lib.utils.settings import get_setting
 
 logger = logging.getLogger('ansible_base.authentication.authenticator_plugins.local')
 
+
 # TODO: Change the validator to not allow it to be deleted or a second one added
+
+UserModel = get_user_model()
 
 
 class LocalConfiguration(BaseAuthenticatorConfiguration):
@@ -44,7 +52,7 @@ class AuthenticatorPlugin(ModelBackend, AbstractAuthenticatorPlugin):
 
         # Determine the user name for this authenticator, we have to call this so that we can "attach" to a pre-created user
         new_username = determine_username_from_uid(username, self.database_instance)
-        # However we can't really accept a different username because we are the local authenticator imageine if:
+        # However we can't really accept a different username because we are the local authenticator imagine if:
         #    User "a" is from another authenticator and has an AuthenticatorUser
         #    User "a" tried to login from local authenticator
         #    The above function will return a username of "a<hash>"
@@ -53,6 +61,24 @@ class AuthenticatorPlugin(ModelBackend, AbstractAuthenticatorPlugin):
             return None
 
         user = super().authenticate(request, username, password, **kwargs)
+        controller_login_results = None
+        if (
+            not user
+            and request
+            and request.path.startswith('/api/gateway/v1/login/')
+            and (controller_login_results := self._can_authenticate_from_controller(username, password))
+        ):
+            logger.warning("User has been validated by controller, updating gateway user.")
+            self.update_gateway_user(username, password)
+            user = super().authenticate(request, username, password, **kwargs)
+        elif not user:
+            logger.info(
+                "Fallback authentication condition not met: "
+                f"username={username}, "
+                f"request={'set' if request else 'None'}, "
+                f"login_path={'True' if request and request.path.startswith('/api/gateway/v1/login/') else 'False'}, "
+                f"controller_login_results={controller_login_results}"
+            )
 
         # This auth class doesn't create any new local users, but we still need to make sure
         # it has an AuthenticatorUser associated with it.
@@ -69,5 +95,126 @@ class AuthenticatorPlugin(ModelBackend, AbstractAuthenticatorPlugin):
                     "is_superuser": user.is_superuser,
                 },
             )
-
         return update_user_claims(user, self.database_instance, [])
+
+    def _can_authenticate_from_controller(self, username, password):
+        """
+        Check if a user exists in the AuthenticatorUser table with the local authenticator provider.
+        If the user is valid, update the gateway users credentials with the controller credentials.
+        """
+        try:
+            UserModel._default_manager.get_by_natural_key(username)
+        except UserModel.DoesNotExist:
+            logger.warning(f"User '{username}' does not exist in the database.")
+            return False
+
+        if controller_user := self._get_controller_user(username, password):
+            # Validate controller_user has a ldap_dn field, if it is not None, then the user is a local user
+            ldap_dn = controller_user.get("ldap_dn")
+            if ldap_dn is None or ldap_dn != "":
+                logger.warning(f"User '{username}' is an ldap user and can not be authenticated.")
+                return False
+            if controller_user.get('password', None) != "$encrypted$":
+                logger.warning(f"User '{username}' is an enterprise user and can not be authenticated.")
+                return False
+            return True
+        else:
+            return False
+
+    def _get_controller_user(self, username: str, password: str):
+        """
+        Get the user from the controller by making a request to the controller API /me/ endpoint.
+        If the user is not found, return None.
+        If the user is found, return the user.
+        """
+
+        controller_base_domain = get_setting('gateway_proxy_url')
+        if not controller_base_domain:
+            logger.warning("Controller authentication failed, unable to get controller base domain")
+            return None
+        controller_url = urljoin(controller_base_domain, "/api/controller/v2/me/")
+
+        timeout = get_setting('GRPC_SERVER_AUTH_SERVICE_TIMEOUT')
+        timeout = self._convert_to_seconds(timeout)
+
+        try:
+            response = requests.get(controller_url, auth=HTTPBasicAuth(username, password), timeout=int(timeout))
+            response.raise_for_status()
+            user_data = response.json()
+
+            # Check if count exists and equals 1
+            count = user_data.get("count")
+            if count != 1:
+                logger.warning(f"Unable to authenticate user '{username}' with controller.")
+                return None
+
+            # Check if results exists and is a non-empty list
+            results = user_data.get("results")
+            if not results or not isinstance(results, list) or len(results) == 0:
+                logger.info(f"Unable to authenticate user '{username}' with controller. Invalid or empty results.")
+                return None
+            if not isinstance(results[0], dict):
+                logger.warning(f"Unable to authenticate user '{username}' with controller. user was not a dictionary.")
+                return False
+
+            return results[0]
+        except requests.exceptions.HTTPError as http_err:
+            logger.warning(f"HTTP error occurred: {http_err}")
+            return None
+        except requests.exceptions.ConnectionError as conn_err:
+            logger.warning(f"Connection error occurred: {conn_err}")
+            return None
+        except requests.exceptions.Timeout as timeout_err:
+            logger.warning(f"Timeout error occurred: {timeout_err}")
+            return None
+        except requests.exceptions.RequestException as err:
+            logger.warning(f"An unexpected error occurred: {err}")
+            return None
+        except ValueError as json_err:
+            logger.warning(f"JSON decode error occurred: {json_err}")
+            return None
+        except Exception as err:
+            logger.warning(f"An unexpected error occurred: {err}")
+            return None
+
+    def update_gateway_user(self, username, password):
+        """
+        Update the gateway user with the controller credentials and set is_partially_migrated to False.
+        """
+        user = UserModel._default_manager.get_by_natural_key(username)
+        user.set_password(password)
+        user.save(update_fields=['password'])
+        logger.info(f"Updated user {username} gateway account")
+
+    def _convert_to_seconds(self, s):
+        """
+        Converts a time string like '15s', '5m', '1h', '2d', '3w' to seconds.
+        """
+        default = 10
+        try:
+            unit = s[-1].lower()
+            value = int(s[:-1])
+
+            ret_val = 0
+            # Check units
+            if unit == '-':
+                ret_val = default
+            elif unit == 's':
+                ret_val = value
+            elif unit == 'm':
+                ret_val = value * 60
+            elif unit == 'h':
+                ret_val = value * 3600  # 60 * 60
+            elif unit == 'd':
+                ret_val = value * 86400  # 60 * 60 * 24
+            elif unit == 'w':
+                ret_val = value * 604800  # 60 * 60 * 24 * 7
+            else:
+                ret_val = int(s)
+            # If less than or equal to 0, return default
+            if ret_val <= 0:
+                ret_val = default
+            return ret_val
+        except Exception:
+            logger.warning(f"Invalid duration format: '{s}'")
+            return default
--- a/test_app/tests/authentication/authenticator_plugins/test_local.py
+++ b/test_app/tests/authentication/authenticator_plugins/test_local.py
@@ -1,6 +1,8 @@
 from unittest import mock
 
 import pytest
+import requests
+from django.contrib.auth import get_user_model
 from django.test.client import RequestFactory
 
 from ansible_base.authentication.authenticator_plugins.local import AuthenticatorPlugin
@@ -8,6 +10,16 @@ from ansible_base.authentication.session import SessionAuthentication
 from ansible_base.lib.utils.response import get_relative_url
 
 authenticated_test_page = "authenticator-list"
+
+
+def mock_get_setting(setting_name):
+    """Helper function to mock get_setting with appropriate values for different settings."""
+    if setting_name == 'gateway_proxy_url':
+        return 'http://controller.example.com'
+    elif setting_name == 'GRPC_SERVER_AUTH_SERVICE_TIMEOUT':
+        return '30s'  # Return a valid duration format
+    else:
+        return None
 
 
 @mock.patch("rest_framework.views.APIView.authentication_classes", [SessionAuthentication])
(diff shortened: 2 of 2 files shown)


ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +82,100 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+        1. The user exists in the gateway database
+        2. The user is_partially_migrated=True
+        """
+        try:
+            user = UserModel._default_manager.get_by_natural_key(username)
+        except UserModel.DoesNotExist:
+            logger.warning(f"User '{username}' does not exist in the database.")
+            return False
+
+        if not user.resource.is_partially_migrated:
```
john-westcott-iv: I think `is_partially_migrated` is going away so we might not want to reference it here. If we remove this I'm not sure how I feel about this function. I kind of like that we are checking if they are in the DB but also, if they weren't we would just fail on the password update anyway.

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +82,100 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+        if controller_user := self._get_controller_user(username, password):
+            # Validate controller_user is a dictionary
+            if not isinstance(controller_user, dict):
+                logger.warning(f"Unable to authenticate user '{username}' with controller. Invalid result format.")
+                return False
+
+            # Validate controller_user has a ldap_dn field, if it is not None, then the user is a local user
+            ldap_dn = controller_user.get("ldap_dn")
+            if ldap_dn is None or ldap_dn != "":
+                logger.warning(f"User '{username}' is unable to be confirmed as a local user.")
```
john-westcott-iv: I think we also have to check for `password == "$encrypted$"` too to ensure they are local. The social auth provides will make the user serializer remove the password field.

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -54,6 +62,11 @@ def authenticate(self, request, username=None, password=None, **kwargs):
 
         user = super().authenticate(request, username, password, **kwargs)
 
+        if not user and not (request and request.path.startswith('/api/controller/')) and self.is_a_partially_migrated_controller_user(username, password):
```
john-westcott-iv: Not sure about this. I'm assuming you are doing this so you can make the request to `/api/controller/v2/me/`. But if you go to `https://<host>/api/controller/v2/instances/` and click login you will be taken to `https://34.207.163.197/api/controller/login` which actually looks like the gateway page and you can log in with gateway creds. So I think having this that open might prevent that from working? Its kind of a nice case that you are logging in for the first time to the API via a controller endpoint.

john-westcott-iv: Also, this portion of DAB is made to be used generically and this would prevent that from happening (making it very AAP specific). It might be better to have some kind of variable that would kick this logic off.

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Updated to remove this check

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Updated to include this check

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: I updated to instead ensure the path starts with /api/controller/v2/me/ to consider this

zkayyali812: I updated further to only check on `/api/gateway/v1/login/`

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -53,6 +61,10 @@ def authenticate(self, request, username=None, password=None, **kwargs):
             return None
 
         user = super().authenticate(request, username, password, **kwargs)
+        if not user and request and request.path.startswith('/api/gateway/v1/login/') and self._can_authenticate_from_controller(username, password):
```
chrismeyersfsu: `request.path.startswith('/api/gateway/v1/login/')` feels wrong, I wonder if we can make it less so. Could you try the below?

```
from django.urls import reverse
login_url = reverse('login')
```

Ideally we would somehow scope it to the Gateway app.

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Updated even further to be - 
```
and request.path.startswith(reverse('login', current_app='aap_gateway_api'))
```
Based on @chrismeyersfsu suggestion

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Hey Chris,
Great note. I updated this logic to be - 
```
and request.path.startswith(reverse('login', current_app='aap_gateway_api'))
```
to encapsulate that

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +87,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+        """
+        Check if a user exists in the AuthenticatorUser table with the local authenticator provider.
+        If the user is valid, update the gateway users credentials with the controller credentials.
+        """
+        try:
+            user = UserModel._default_manager.get_by_natural_key(username)
+        except UserModel.DoesNotExist:
+            logger.warning(f"User '{username}' does not exist in the database.")
+            return False
+
```
chrismeyersfsu, line 110:
```suggestion
if user.role.is_partially_migrated is False:
  return False
```

I expected to find a short-circuit before calling to the Controller API.

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -53,6 +62,15 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
         user = super().authenticate(request, username, password, **kwargs)
+        if (
+            not user
+            and request
+            and request.path.startswith(reverse('login', current_app='aap_gateway_api'))
+            and self._can_authenticate_from_controller(username, password)
+        ):
+            logger.warning("User has been validated by controller, updating gateway user.")
+            self.update_gateway_user(username, password)
+            user = super().authenticate(request, username, password, **kwargs)
```
chrismeyersfsu, line 73:
```suggestion
            user = super().authenticate(request, username, password, **kwargs)
        else:
            logger.info(
                "Fallback authentication condition not met: "
                f"user={user}, "
                f"request={'set' if request else 'None'}, "
                f"login_path={'True' if request and request.path.startswith(reverse('login', current_app='aap_gateway_api')) else 'False'}, "
                f"can_auth_controller={'True' if self._can_authenticate_from_controller(username, password) else 'False'}"
            )
```

* So we, and an aap admin, can answer the quesiton: "why did my login fail?"
* info level (vs. debug level) so that, as an aap admin logging data to splunk, when there is an incident I can provide the security team with data about failed login attempts, by default.

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +87,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+        """
+
+        controller_base_domain = get_setting('gateway_proxy_url')
+        if not controller_base_domain:
+            logger.warning("Controller authentication failed, unable to get controller base domain")
+            return None
+        controller_url = urljoin(controller_base_domain, "/api/controller/v2/me/")
+
+        try:
+            response = requests.get(controller_url, auth=HTTPBasicAuth(username, password), timeout=10)
```
chrismeyersfsu: TL;DR set the timeout to the underlying component request/response cycle timeout.

What settings are available to us related to request timeouts?
@kdelee 

You had to pick something, but I'd rather the timeout be calculated based on the remaining time left in the request/response cycle.

```
envoy --> grpc --> gateway --> controller
```
@kdelee has added a parameter to the installer, I can't remember the name of it, but it's the total request response time. That is the time envoy gets set to. Every other service to the right of envoy gets a timeout set that is < envoy, and that number gets subtracted from the request response time total. So you get `total request response time (envoy) = grpc + gateway + component`.

I'm not sure how we get a handle on that installer setting.

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +87,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+            # Check if count exists and equals 1
+            count = user_data.get("count")
+            if count != 1:
+                logger.warning(f"Unable to authenticate user '{username}' with controller.")
+                return None
+
+            # Check if results exists and is a non-empty list
+            results = user_data.get("results")
+            if not results or not isinstance(results, list) or len(results) == 0:
+                logger.warning(f"Unable to authenticate user '{username}' with controller. Invalid or empty results.")
```
chrismeyersfsu:
```suggestion
                logger.info(f"Unable to authenticate user '{username}' with controller. Invalid or empty results.")
```

I'm thinking info log for this one. This one can happen if someone types the wrong password. It's not necessarily a system issue

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +87,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+                return None
+            return results[0]
+        except requests.exceptions.HTTPError as http_err:
+            logger.info(f"HTTP error occurred: {http_err}")
```
chrismeyersfsu:
```suggestion
            logger.warning(f"HTTP error occurred: {http_err}")
```

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +87,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+            return None
+        except requests.exceptions.ConnectionError as conn_err:
+            logger.info(f"Connection error occurred: {conn_err}")
```
chrismeyersfsu:
```suggestion
            logger.warning(f"Connection error occurred: {conn_err}")
```

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +87,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+            return None
+        except requests.exceptions.Timeout as timeout_err:
+            logger.info(f"Timeout error occurred: {timeout_err}")
```
chrismeyersfsu:
```suggestion
            logger.warning(f"Timeout error occurred: {timeout_err}")
```

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +87,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+            return None
+        except requests.exceptions.RequestException as err:
+            logger.info(f"An unexpected error occurred: {err}")
```
chrismeyersfsu:
```suggestion
            logger.warning(f"An unexpected error occurred: {err}")
```

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +87,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+            return None
+        except ValueError as json_err:
+            logger.info(f"JSON decode error occurred: {json_err}")
```
chrismeyersfsu:
```suggestion
            logger.warning(f"JSON decode error occurred: {json_err}")
```

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +87,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+            return None
+
+    def update_gateway_user(self, username, password):
+        """
+        Update the gateway user with the controller credentials and set is_partially_migrated to False.
+        """
+        user = UserModel._default_manager.get_by_natural_key(username)
+        user.set_password(password)
```
chrismeyersfsu, line 185:
```suggestion
        user.set_password(password)
        user.is_partially_migrated = False
```

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +87,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+        user.save()
```
chrismeyersfsu:
```suggestion
        user.save(update_fields=['password', 'is_partially_migrated'])
```

What I imagine the sql of the two looks like:

save sql: `UPDATE "auth_user" SET "username" = 'newusername', "email" = 'existing_email', "password" = 'existing_password' WHERE "auth_user"."id" = 1;`

`update_fields` sql: `UPDATE "auth_user" SET  "password" = 'password-hash-here', "is_partially_migrated" = False WHERE "auth_user"."id" = 1;`

ansible_base/authentication/authenticator_plugins/local.py
chrismeyersfsu: Should we "blindly" include error messages from another service in our logs, especially when it's the authentication path?

Part of me likes it. The downstream service is ours, it shouldn't be outputting sensitive data in the response.

But then part of me thinks about layers of security/checks.

I vote for debuggability and to continue forwarding the message directly to the logs. No change.

**Story time:**
Controller recently had a security issue for doing this sort of thing with Cyberark responses or it was the GET url issued to Cyberark. Because Cyberark did a bad best practice (putting sensitive data in GET params) we had to account for it.

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: This was removed due to @john-westcott-iv's comment here - https://github.com/ansible/django-ansible-base/pull/758#discussion_r2205103814

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: is_partially_migrated was removed from this PR as this is likely going away

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Nice!! Did not know you could do this

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: I think this should be acceptable. I don't think any sensitive data is returned from this API directly. Even if a password is returned, it would be its encrypted value.

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: I think Ill change the case here to `elif not user` so it does not trigger on every login

zkayyali812: Also Ill change `user` to `username`

zkayyali812: I also removed `_can_authenticate_from_controller` from this info log. If this is included, it adds risk to somehow lead to infinite loops if all the potential authentication paths are not excluded.

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -53,6 +62,22 @@ def authenticate(self, request, username=None, password=None, **kwargs):
             return None
 
         user = super().authenticate(request, username, password, **kwargs)
+        if (
+            not user
+            and request
+            and request.path.startswith(reverse('login', current_app='aap_gateway_api'))
```
john-westcott-iv: If this was ever run under a service w/o this endpoint this would raise an exception:
```
In [3]: reverse('login2', current_app='aap_gateway_api')
NoReverseMatch                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 reverse('login2', current_app='aap_gateway_api')

File ~opt/aap_gateway/venv/lib64/python3.11/site-packages/django/urls/base.py:88, in reverse(viewname, urlconf, args, kwargs, current_app)
     83     if ns_pattern:
     84         resolver = get_ns_resolver(
     85             ns_pattern, resolver, tuple(ns_converters.items())
     86         )
---> 88 return resolver._reverse_with_prefix(view, prefix, *args, **kwargs)

File ~opt/aap_gateway/venv/lib64/python3.11/site-packages/django/urls/resolvers.py:828, in URLResolver._reverse_with_prefix(self, lookup_view, _prefix, *args, **kwargs)
    823 else:
    824     msg = (
    825         "Reverse for '%(view)s' not found. '%(view)s' is not "
    826         "a valid view function or pattern name." % {"view": lookup_view_s}
    827     )
--> 828 raise NoReverseMatch(msg)

NoReverseMatch: Reverse for 'login2' not found. 'login2' is not a valid view function or pattern name.

```

ansible_base/authentication/authenticator_plugins/local.py
john-westcott-iv: I've never tried it in an and condition but could you do:
```
and controller_login_results := self._can_authenticate_from_controller(username, password)
```

And then in the print statement just dump the variable?

ansible_base/authentication/authenticator_plugins/local.py
john-westcott-iv: Yea, we can double check with the upgrade patch but I'm pretty sure removing this flag was part of that. @AlanCoding might be able to confirm off the top of his head.

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +94,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+                logger.warning(f"Unable to authenticate user '{username}' with controller. Invalid result format.")
+                return False
+            # Validate controller_user has a ldap_dn field, if it is not None, then the user is a local user
+            ldap_dn = controller_user.get("ldap_dn")
+            if (ldap_dn is None or ldap_dn != "") and user.password == "$encrypted$":
+                logger.warning(f"User '{username}' is unable to be confirmed as a local user.")
+                return False
+            return True
+        else:
+            return False
```
john-westcott-iv:
```suggestion
            # We can only validate local controller users so we need to check two things
            # First lets see if they were an ldap user
            ldap_dn = controller_user.get("ldap_dn", None)
            if ldap_dn is None or ldap_dn != "":
                logger.warning(f"User '{username}' is an ldap user and can not be authenticated.")
                return False
             # Second, lets see if the were an enterprise user
             if controller_user.get('password, None) != "$encrypted$":
                logger.warning(f"User '{username}' is an enterprise user and can not be authenticated.")
                return False
            return True
        else:
            return False
```

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +94,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+            return None
+        except requests.exceptions.Timeout as timeout_err:
+            logger.warning(f"Timeout error occurred: {timeout_err}")
+            return None
+        except requests.exceptions.RequestException as err:
+            logger.warning(f"An unexpected error occurred: {err}")
+            return None
+        except ValueError as json_err:
+            logger.warning(f"JSON decode error occurred: {json_err}")
+            return None
```
john-westcott-iv, line 175: Do we also want to catch a generic Exception here incase something else is raised up?

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +94,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+            if count != 1:
+                logger.warning(f"Unable to authenticate user '{username}' with controller.")
+                return None
+
+            # Check if results exists and is a non-empty list
+            results = user_data.get("results")
+            if not results or not isinstance(results, list) or len(results) == 0:
+                logger.info(f"Unable to authenticate user '{username}' with controller. Invalid or empty results.")
+                return None
+            return results[0]
```
john-westcott-iv, line 160:
```suggestion
            if not isinstance(results[0], dict):
                logger.warning(f"Unable to authenticate user '{username}' with controller. user was not a dictionary.")
                return False
                
            return results[0]
```

ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -69,5 +94,84 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
+            user = UserModel._default_manager.get_by_natural_key(username)
+        except UserModel.DoesNotExist:
+            logger.warning(f"User '{username}' does not exist in the database.")
+            return False
+
+        if controller_user := self._get_controller_user(username, password):
+            # Validate controller_user is a dictionary
+            if not isinstance(controller_user, dict):
+                logger.warning(f"Unable to authenticate user '{username}' with controller. Invalid result format.")
+                return False
```
john-westcott-iv: I almost want to move this into `_get_controller_user` so that you are dealing with entirety of the response in that function. (i.e. because you already check the results count and what not in there). But this is a nit. 

```suggestion
```

ansible_base/authentication/authenticator_plugins/local.py
chrismeyersfsu: Asked if `is_partially_migrated` is going away slack https://redhat-internal.slack.com/archives/C08NZHY16Q1/p1752586713478919

ansible_base/authentication/authenticator_plugins/local.py
kdelee: Do you have access to the gateway Django settings? 

```
{
    "gateway_token_name": "X-DAB-JW-TOKEN",
    "gateway_access_token_expiration": 600,
    "jwt_expiration_buffer_in_seconds": 2,
    "gateway_basic_auth_enabled": true,
    "gateway_proxy_url": "http://elijah-minimal-25.aap-25.svc.cluster.local/",
    "gateway_proxy_url_ignore_cert": false,
    "jwt_private_key": "$encrypted$",
    "jwt_public_key": "-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA4AP9e+qu6JRGJYRm2YxN\nXqJmzpCxrffiInidINfcLwbQ787y4S+xI7IH5iVsjfiGSS6Q8weY8WIIh/vvJ8fz\n2mH+9mfFBpmR8/CYaFtzQvaLnYSGt8EmlDzDLnJ0rQTugVi6V7gt0OzU+6cAv1ej\nwbBfYM/tSBp39i5O3JV4aFkIz4G0HPM9ghLwuXM4I66+uBs131iTCKuZtIlUK5Xl\n54K4j7ubrW5b47hGCsvp/HO4mjS6aYH04gqvPdsfiEjdtmAAVC2yNR3dt4DuKBDi\n+vjZSPXXSpyhndTrIoWhegS9lP9AU4WT6NBEFvqz63nb6Ge6Pic+tB5FRs+1atnD\nZwIDAQAB\n-----END PUBLIC KEY-----\n",
    "status_endpoint_backend_timeout_seconds": 5,
    "status_endpoint_backend_verify": true,
    "resource_client_request_timeout": 2.0,
    "request_timeout": 30,
    "password_min_length": 0,
    "password_min_digits": 0,
    "password_min_upper": 0,
    "password_min_special": 0,
    "allow_admins_to_set_insecure": false,
    "LOGIN_REDIRECT_OVERRIDE": "",
    "custom_login_info": "",
    "custom_logo": "",
    "SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL": false,
    "SESSION_COOKIE_AGE": 900,
    "DEFAULT_PAGE_SIZE": 50,
    "MAX_PAGE_SIZE": 200,
    "ALLOW_OAUTH2_FOR_EXTERNAL_USERS": false,
    "CONTROLLER_SSO_URL": "",
    "AUTOMATION_HUB_SSO_URL": "",
    "INSIGHTS_TRACKING_STATE": true,
    "RED_HAT_CONSOLE_URL": "https://console.redhat.com/",
    "REDHAT_USERNAME": "",
    "REDHAT_PASSWORD": "$encrypted$",
    "SUBSCRIPTIONS_USERNAME": "",
    "SUBSCRIPTIONS_PASSWORD": "$encrypted$",
    "AUTOMATION_ANALYTICS_GATHER_INTERVAL": 14400
}
```
Gateway's  setting `request_timeout` is the timeout for the components (their nginx + wsgi) because grpc timeout is counted separately 
```
settings.GRPC_SERVER_AUTH_SERVICE_TIMEOUT
'30s'
```
This is the grpc service timeout.

The top level `client_request_timeout` is `request_timeout + settings.GRPC_SERVER_AUTH_SERVICE_TIMEOUT` but is not set anywhere in gateway settings

Controller/EDA/Hub don't seem to have any setting INSIDE the application that makes it aware of the nginx/uwsgi/gunicorn timeouts

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: In this case I am going to remove the `reverse` and go back to a hardcoded url. 
CC: @chrismeyersfsu . The other alternative is to do a try/catch to try to get this value, but for simplicity, probably easier to just hardcode the expected path.

Open to other options/opinions here though

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Adding a note here for posterity - it was confirmed by @AlanCoding that `is_partially_migrated` is going away

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Noting for posterity, that it was confirmed in slack that this is going away

ansible_base/authentication/authenticator_plugins/local.py
AlanCoding: yep ping @TheRealHaoLiu

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Done

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Added timeout based on - `GRPC_SERVER_AUTH_SERVICE_TIMEOUT`. Will fallback to `30` seconds if not provided

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Added

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Done

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Done

ansible_base/authentication/authenticator_plugins/local.py
zkayyali812: Done

PR #759, opened 2025-07-11
[AAP-48713] Alter authenticator maps to work with attribute expansion

john-westcott-iv:
## Description

- What is being changed?
If you add the syntax `{% for_attr_value(<attribute name>) %}` to an authenticator map we will expand that value based on the values returned from an attribute on login.

- Why is this change needed?
This will make gateway backwards compatible with an obscure SAML setting from AWX.

This allows authenticator maps to behave like the SAML authenticator can.

NOTES:
* I still need to setup the gateway LDAP side care to allow for multi-attributes to be passed to gateway to do user testing of this feature.
* This is built on top of PR #757 

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [X] Refactoring (no functional changes)
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. 
2. 
3. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/authentication/utils/authenticator_map.py
+++ b/ansible_base/authentication/utils/authenticator_map.py
@@ -1,9 +1,21 @@
+import logging
 import re
-from typing import Any, Optional
+from typing import Dict, Optional
 
+from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist
 from django.utils.translation import gettext_lazy as _
 
+from ansible_base.authentication.models import AuthenticatorMap
+from ansible_base.lib.utils.auth import get_organization_model, get_team_model
+from ansible_base.lib.utils.collection import dict_cartesian_product
+from ansible_base.lib.utils.string import is_empty
+from ansible_base.lib.utils.typing import TranslatedString
+
+logger = logging.getLogger('ansible_base.authentication.utils.authenticator_map')
+
 _EXPANSION_FIELDS = ['organization', 'role', 'team']
+_EXPANSION_RE = re.compile(r'{%\s*for_attr_value\(([^(%})]+)\)\s*%}')
 
 
 def has_expansion(value: Optional[str]) -> bool:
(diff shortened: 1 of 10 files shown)


test_app/tests/authentication/utils/test_claims.py
```diff
@@ -232,10 +243,13 @@ def test_create_claims_single_map_acl(
     assert res["access_allowed"] == exp_access_allowed
     assert res["is_superuser"] == exp_is_superuser
     assert res["claims"] == exp_claims
-    if connection.vendor != 'postgresql':
-        assert res["last_login_map_results"] == exp_last_login_map_results
-    else:
-        assert list(res["last_login_map_results"][0].values())[0] == list(exp_last_login_map_results[0].values())[0]
+
+    if connection.vendor == 'postgresql' and local_authenticator_map.id != 1:
+        # Since we are running in postgres we need to "massage" the exp_last_login_map_results as the saved auth map might have a different id
```
tznamena: Let's replace the "massage" with specifics that explain why and what's changed.

tznamena (commented):
Is this build on top of the PR #757 ? It seems to include the same commit as that PR, should the other PR be closed then?

ansible_base/authentication/utils/authenticator_map.py
```diff
@@ -18,12 +30,166 @@ def has_expansion(value: Optional[str]) -> bool:
         return False
 
 
-def check_expansion_syntax(value: Optional[str]) -> Optional[Any]:
+def check_expansion_syntax(value: Optional[str]) -> Optional[TranslatedString]:
     """
     Check a given field to see if it contains the proper syntax for {% for_attr_value(user_orgs) %}
 
     Raises a ValidationError if its incorrect
```
jshimkus-rh, line 37: Says it raises ValidationError but it does not.

ansible_base/authentication/utils/authenticator_map.py
```diff
@@ -18,12 +30,166 @@ def has_expansion(value: Optional[str]) -> bool:
...
+
+            if len(normalized_attributes) == 0:
+                logger.info(
+                    f"Authenticator Map {map.name} on {map.authenticator.name} tried to expand attribute {attr_name} "
+                    f"but there were not values in that attribute, ignoring"
+                )
+                had_issues = True
+
+            # If any of the values are not a string then we can't expand it
+            for value in normalized_attributes[attr_name]:
```
jshimkus-rh, line 98: I'd suggest something like this instead:

`non_strings = [value for value in normalized_atttributes[attr_name] if not isinstance(value, str)]`
`for non_string in non_strings:`
    `...do the logging...`

to log all invalid attributes rather than just the first that the `break` would result in.

ansible_base/authentication/utils/authenticator_map.py
john-westcott-iv: Good catch, that was an older implementation, comment removed.

ansible_base/authentication/utils/authenticator_map.py
john-westcott-iv: Good call on not breaking out so that every logs. But if I do the `non_strings` approach you mentioned the logging becomes header so I think I'll just remove the break to make it log everything. Are you good with that?

PR #762, opened 2025-07-15
AAP-49479  Fix RoleTeamAssignmentSerializer team_ansible_id allow_null

PabloHiro:
## Description

- What is being changed? `allow_null field being added to `team_ansible_id` for RoleTeamAssignmentSerializer
- Why is this change needed? Align `RoleTeamAssignmentSerializer` with `RoleUserAssignmentSerializer` for proper API schema generation
- How does this change address the issue? Changing the serializer corrects schema generation by indicating to the introspection tools that None is a potential value for the field coming from the API

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites

test_app up and running

### Steps to Test

1. Generate openAPI spec file /api/v1/docs/schema
2. Read the spec file to ensure expected results
 

### Expected Results
RoleTeamAssignment:

```yaml
        team_ansible_id:
          type: string
          format: uuid
          nullable: true
          description: The resource ID of the team who will receive permissions from
            this assignment. An alternative to team field.
```

### DIFF
--- a/ansible_base/rbac/api/serializers.py
+++ b/ansible_base/rbac/api/serializers.py
@@ -323,6 +323,7 @@ class RoleTeamAssignmentSerializer(BaseAssignmentSerializer):
     team_ansible_id = serializers.UUIDField(
         required=False,
         help_text=_('The resource ID of the team who will receive permissions from this assignment. An alternative to team field.'),
+        allow_null=True,
     )
 
     class Meta:


AlanCoding (changes requested):
This is not correct. The `RoleTeamAssignment.team` field is non-null

```
In [3]: RoleTeamAssignment._meta.get_field('team').null
Out[3]: False
```

You might have a bug where the field is null, either because the resource registry failed to set up a resource for that team (thus not having the ansible_id), or the serializer failed to get the ansible_id. In either case, this is not the correct response.

AlanCoding (approved):
I see that `user_ansible_id` has allow_null, so we should just take this as a minor tweak for short-term

zkayyali812 (approved):

PR #751, opened 2025-07-01
[PoC] [AAP-48066] Make users join by emai

john-westcott-iv:
## Description

- What is being changed?
PoC allowing for multiple authenticators merging into the same user

- Why is this change needed?
This allows users from multiple sources to be merged into a single django user.

- How does this change address the issue?
This change makes it work.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test update
- [ ] Refactoring (no functional changes)
- [ ] Development environment change
- [ ] Configuration change

## Self-Review Checklist

- [X] I have performed a self-review of my code
- [X] I have added relevant comments to complex code sections
- [X] I have updated documentation where needed
- [X] I have considered the security impact of these changes
- [X] I have considered performance implications
- [X] I have thought about error handling and edge cases
- [X] I have tested the changes in my local environment

### Steps to Test
1. Log in through different authenticators with different username/emails and see what user you are logged in as.
2. 
3. 

### Required Actions

- [ ] Requires documentation updates
  
- [ ] Requires downstream repository changes
  
- [ ] Requires infrastructure/deployment changes
  
- [ ] Requires coordination with other teams
  
- [ ] Blocked by PR/MR: #XXX

PR #764, opened 2025-07-15
Feature ANSTRAT-1459

zkayyali812:
## Description
- What is being changed? We are planning to merge our feature branch into devel to leverage quicker testing and feedback from the nightlies.
- Why is this change needed? To continuously deliver on 1459, and to get more iterative feedback from the nightlies.
- How does this change address the issue? This change addresses the issue by merging our in progress feature branch into devel to incorporate our already validated changes.

@AlanCoding attempted summary:

This includes 2 merges to the feature branch of 1459

 - https://github.com/ansible/django-ansible-base/pull/757
 - https://github.com/ansible/django-ansible-base/pull/754

The change descriptions are there.

 - Allowing the `authenticator_map` (_field_ @john-westcott-iv ) serializer to store `{% for_attr_value(something) %}`
   - This will allow the authenticator maps to have templates in them. (so basically, adding templating to the authenticator_maps field?)
 - `email` field is added
   - This will make it easier to debug issues regarding how users authenticate into the platform, since we will be able to tell directly which email address their authenticatoruser is configured with.
   - the PoC for this was https://github.com/ansible/django-ansible-base/pull/751
   - the PoC used the email from the User model, which is likely improved by moving the emtil to the AuthenticatorUser

## Type of Change

- [x] Documentation update
- [x] Test update

## Self-Review Checklist

### Steps to Test
1. 
2. 
3. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX


john-westcott-iv (approved):

ansible_base/authentication/models/authenticator_user.py
```diff
@@ -42,6 +42,7 @@ class AuthenticatorUser(AbstractUserSocialAuth, AbstractCommonModel):
         on_delete=models.CASCADE,
         help_text=_("The local DB user related to this authenticator user."),
     )
+    email = models.EmailField(default=None, null=True, blank=True, help_text=_("The e-mail associated with this authenticator user."))
```
AlanCoding, line 45: If I'm reading the help text, the question I'm trying to answer to myself is "why does this duplicate the email field from the User model?" That leads into a whole myriad of concerns about the email getting changed, going out of sync between user & authenticator user.

Perhaps this is used as a lookup when provided by SSO of some kind. But why not look up by the User model's email? These may have good answers... but this help text is practically an insult.

zkayyali812: Hey @AlanCoding ,
As part of 1459, we are aiming to store the email returned from the authenticator directly, to help in debugging and provide information about how the authentication is working. This is purposefully intended to be separate from the users email field. 
This may involve cases where an authenticator does not return an email (field would be blank),or an authenticator returns an email which is different from the users email, but a user has manually linked it via the API.

The help text is generic here, but truthful that this is just intended to be the email associated with this specific AuthenticatorUser, rather than the AAP user.

AlanCoding (approved):

PR #765, opened 2025-07-15
AAP-49223 Fix missing query parameters openapi spec

PabloHiro:
## Description

- What is being changed? OpenAPI schema now accurately displays the possible query parameters of an endpoint.
- Why is this change needed? To accurately build API clients out of the OpenAPI schema.
- How does this change address the issue? Create several classes extending `OpenApiFilterExtension` part of drf-spectacular(https://drf-spectacular.readthedocs.io/en/latest/drf_spectacular.html#drf_spectacular.extensions.OpenApiFilterExtension) to have granular control of how the schema gets populated.

## Type of Change

- [x] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

## Additional Context
Assited by AI. Cursor on claude-4-sonnet model.

### Screenshots/Logs
Snippet of the swagger view of the API schema:
image

### DIFF
--- a/ansible_base/api_documentation/apps.py
+++ b/ansible_base/api_documentation/apps.py
@@ -13,3 +13,8 @@ class ApiDocumentationConfig(AppConfig):
 
         if 'ansible_base.authentication' in settings.INSTALLED_APPS:
             apply_authentication_customizations()
+
+        # Import filter extensions to register them with drf-spectacular
+        if 'ansible_base.rest_filters' in settings.INSTALLED_APPS and 'ansible_base.api_documentation' in settings.INSTALLED_APPS:
+            # If this service is using DAB rest filters and api documentation, load our filter extensions for OpenAPI
+            from ansible_base.api_documentation import filter_extensions  # noqa: F401
--- a/ansible_base/api_documentation/filter_extensions.py
+++ b/ansible_base/api_documentation/filter_extensions.py
@@ -1,0 +1,160 @@
+from drf_spectacular.extensions import OpenApiFilterExtension
+from drf_spectacular.plumbing import build_parameter_type
+from drf_spectacular.utils import OpenApiParameter
+
+from ansible_base.rest_filters.rest_framework.field_lookup_backend import FieldLookupBackend
+from ansible_base.rest_filters.rest_framework.order_backend import OrderByBackend
+from ansible_base.rest_filters.rest_framework.type_filter_backend import TypeFilterBackend
+
+
+class FieldLookupBackendExtension(OpenApiFilterExtension):
+    target_class = FieldLookupBackend
+
+    def get_schema_operation_parameters(self, auto_schema, *args, **kwargs):
+        """
+        Generate OpenAPI parameters for FieldLookupBackend.
+
+        This filter backend supports field lookups on any model field using Django's
+        field lookup syntax (e.g., field__exact, field__contains, field__gt, etc.).
+        Since the actual fields depend on the model, we provide generic examples.
+        """
+        parameters = []
+
+        # Add model-based parameters if model is available
+        if self._has_model_queryset(auto_schema):
+            model = auto_schema.view.queryset.model
+            model_fields = self._get_relevant_model_fields(model)
+            parameters.extend(self._create_model_field_parameters(model, model_fields))
+
+        # Add RBAC parameter
+        parameters.append(self._create_role_level_parameter())
+
+        return parameters
+
+    def _has_model_queryset(self, auto_schema):
+        """Check if the view has a model queryset."""
+        return hasattr(auto_schema.view, 'queryset') and auto_schema.view.queryset is not None
+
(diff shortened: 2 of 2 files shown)


ansible_base/api_documentation/apps.py
```diff
@@ -13,3 +13,10 @@ def ready(self):
 
         if 'ansible_base.authentication' in settings.INSTALLED_APPS:
             apply_authentication_customizations()
+
+        # Import filter extensions to register them with drf-spectacular
+        try:
+            from ansible_base.api_documentation import filter_extensions  # noqa: F401
+        except ImportError:
+            # drf_spectacular may not be available
+            pass
```
john-westcott-iv: `ansible_base.rest_filters` is also an optional django application that a service /could/ use. We might want to wrap this in something like:

```suggestion
        if 'ansible_base.rest_filters' in settings.INSTALLED_APPS and 'ansible_base.api_documentation' in settings.INSTALLED_APPS:
            # If this service is using DAB rest filters and api documentation, load our filter extensions for OpenAPI
            from ansible_base.api_documentation import filter_extensions  # noqa: F401
```

zkayyali812 (approved):

ansible_base/api_documentation/filter_extensions.py
```diff
@@ -0,0 +1,160 @@
...
+        """Check if the view has a model queryset."""
+        return hasattr(auto_schema.view, 'queryset') and auto_schema.view.queryset is not None
+
+    def _get_relevant_model_fields(self, model):
+        """Get relevant model fields, excluding complex relationships."""
+        model_fields = []
+        for field in model._meta.get_fields():
+            if self._is_simple_field(field):
+                model_fields.append(field.name)
+        return model_fields[:10]  # Limit to avoid overwhelming the schema
```
bhavenst: Is this really an issue?  It seems like we should include all the fields.

ansible_base/api_documentation/filter_extensions.py
```diff
@@ -0,0 +1,160 @@
...
+        return parameters
+
+    def _create_field_parameters(self, model, field_name):
+        """Create all parameter variations for a single field."""
+        parameters = []
+
+        # Basic exact match parameter
+        parameters.append(self._create_parameter(field_name, f'Filter by {field_name} (exact match)'))
+
+        # Add field-type specific parameters
```
bhavenst, line 64: Does the presence of each of these filter types enable/disable the use of it from ATF?  Or can they be used anyway (including things like isnull, regex, etc. that aren't included at all)?

I am wondering if it's worth filtering some things like created, modified since those are everywhere and seem less likely to be filtered than the model specific fields.  And it would probably substantially reduce the schema size.

ansible_base/api_documentation/filter_extensions.py
```diff
@@ -0,0 +1,160 @@
...
+    def get_schema_operation_parameters(self, auto_schema, *args, **kwargs):
+        """
+        Generate OpenAPI parameters for OrderByBackend.
+
+        This filter backend supports ordering results by field names.
+        """
+        parameters = []
+
+        # Add the ordering parameters
+        for param_name in ['order', 'order_by']:
```
bhavenst, line 149: I'm thinking order isn't appropriate here, but could be wrong.  In gateway at least, the models that do have an order field use it for other purposes (like hierarchical ordering, application/usage order, etc.).  order_by on the other hand is for ordering the query results.

bhavenst: Ope, nevermind.  I see it isn't a field but a param and the order by backend supports both.  Ignore!

ansible_base/api_documentation/filter_extensions.py
PabloHiro: > Does the presence of each of these filter types enable/disable the use of it from ATF?

Yes. If a query_parameter is not present in the spec file, the auto-generated client with [openapi-python-client/](https://github.com/openapi-generators/openapi-python-client/tree/main), won't let us perform query calls using that query_parameter.

I have not considered isnull, regex or others due to my lack of knowledge of what is available. We don't have to aim for perfection for now, the problem I am trying to fix is that currently the spec file only supports "search" as a query_parameter

ansible_base/api_documentation/filter_extensions.py
PabloHiro: Honestly, this chunk was generated by AI. When I was reviewing 10 seemed like a sensible choice, but I guess this is putting a band aid before the wound, so removing the limit.

AlanCoding (approved):

ansible_base/api_documentation/filter_extensions.py
bhavenst: Yeah that's cool.  We can add them as desired since you created the base template, so that is totally fine.

bhavenst (approved):

PR #768, opened 2025-07-15
[AAP-48713] Alter authenticator maps to work with attribute expansion

john-westcott-iv:
## Description

- What is being changed?
If you add the syntax `{% for_attr_value(<attribute name>) %}` to an authenticator map we will expand that value based on the values returned from an attribute on login.

- Why is this change needed?
This will make gateway backwards compatible with an obscure SAML setting from AWX.

This allows authenticator maps to behave like the SAML authenticator can.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [X] Refactoring (no functional changes)
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Setup an authenticator which return multiple attribute values
2. Create authenticator maps that expand values
3. Login through the authenticator
4. Assert that the values were expanded as expected

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/authentication/utils/authenticator_map.py
+++ b/ansible_base/authentication/utils/authenticator_map.py
@@ -1,6 +1,6 @@
 import logging
 import re
-from typing import Dict, Optional
+from typing import Any, Dict, Optional
 
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
@@ -33,15 +33,13 @@ def has_expansion(value: Optional[str]) -> bool:
 def check_expansion_syntax(value: Optional[str]) -> Optional[TranslatedString]:
     """
     Check a given field to see if it contains the proper syntax for {% for_attr_value(user_orgs) %}
-
-    Raises a ValidationError if its incorrect
     """
 
     if has_expansion(value) and not _EXPANSION_RE.search(value):
         return _("Expansion only supports the format {% for_attr_value(attribute) %}")
 
 
-def expand_syntax(attributes: dict, map: AuthenticatorMap) -> list[Dict[str, str]]:
+def expand_syntax(attributes: dict, auth_map: AuthenticatorMap) -> list[Dict[str, str]]:
     """
     Given attributes and a map, look for the fields that can be expanded and do the expansion
     If no fields require expansion, do nothing
@@ -51,7 +49,7 @@ def expand_syntax(attributes: dict, map: AuthenticatorMap) -> list[Dict[str, str
     """
     expanded_strings = {}
     for field in _EXPANSION_FIELDS:
-        field_value = getattr(map, field, None)
+        field_value = getattr(auth_map, field, None)
 
         if field_value is None:
             # If this particular field is None we can move on.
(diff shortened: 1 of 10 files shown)


AlanCoding (approved):
Doesn't really seem to change a lot long along the interface that matters with other systems, and fairly clear what this is doing and why it is needed.

ansible_base/authentication/utils/authenticator_map.py
```diff
@@ -18,12 +30,201 @@ def has_expansion(value: Optional[str]) -> bool:
...
+            if type(value) is not str:
+                attr_errors.append(
+                    f"Authenticator Map {auth_map.name} on {auth_map.authenticator.name} tried to expand attribute {attr_name} "
+                    f"but that was not a list of string, instead got {value}, ignoring"
+                )
+
+        if attr_errors == []:
+            normalized_attributes[attr_name] = attr_value
+        else:
+            # If we got errors bubble them up and log them
```
jshimkus-rh: Doesn't actually bubble them up (which I read as 'make them available outside this method') as if there are errors an unparameterized ValueError is raised.  Though it does raise them up a scope in order to raise that ValueError.

jshimkus-rh (approved):

zkayyali812 (approved):

ansible_base/authentication/utils/authenticator_map.py
```diff
@@ -18,12 +30,201 @@ def has_expansion(value: Optional[str]) -> bool:
...
+    If no fields require expansion, do nothing
+
+    returns:
+    list of named tuples of expanded value
+    """
+    expanded_strings = {}
+    for field in _EXPANSION_FIELDS:
+        field_value = getattr(auth_map, field, None)
+
+        if field_value is None:
```
mkanoor, line 54: @john-westcott-iv Is this the code change that would  allow organization to be None to support SOCIAL_AUTH_SAML_ORGANIZATION_ATTR SAML where the organization name comes at runtime?

ansible_base/authentication/utils/authenticator_map.py
john-westcott-iv: Updated to be clearer.

ansible_base/authentication/utils/authenticator_map.py
john-westcott-iv: This who code change is for the runtime values. Organization won't be None though, it will have a "template" syntax inside of it. For example lets say your attribute had the values `["database", "it operations"]`, you could set the organization field to be `{% for_attr_value(attribute) %} Organization`. This could would run the rule for 2 organizations: `["database Organization", "it operations Organization"]`. So we still need to know how to template out the organization name.

This code can also do this with team or role values on the authenticator map as well.

It can even do multiple expansions per string. So say you had organization as `{% for_attr_value(aap-orgs) %}` and team was set to `{% for_attr_value(aap-teams) %}` and role was set to `{% for_attr_value(aap-roles) %}`. Now lets say your login had the attributes:
```
{
    "aap-orgs": ["Org 1", "Org 2"],
    "aap-teams": ["Team 1", "Team 2"],
    "aap-roles": ["Team Member"],
}
```
This one rule would be expanded into 4 rules:
* Team Member of Team 1 in Org 1
* Team Member of Team 2 in Org 1
* Team Member of Team 1 in Org 2
* Team Member of Team 2 in Org 2

All of these rules would be fired off the same trigger.

PR #771, opened 2025-07-16
AAP-45580 Add callback url to generic OIDC authenticator

jshimkus-rh:
## Description

- What is being changed?
Adding CALLBACK_URL to generic OIDC authenticator plug-in.

- Why is this change needed?
To provide the callback url via the UI.
 
The UI automatically presents the callback URL when present.

## Type of Change

- [ ] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

### Required Actions

- [x] Requires documentation updates
  
The Configuring authentication in the Ansible Automation Platform section on configuring generic OIDC authentication requires updating similarly to that of configuring SAML authentication.  The update needed is the inclusion of a suitably worded version of the following from SAML....

> The SAML Assertion Consumer Service (ACS) URL field registers the service as a service provider (SP) with each identity provider (IdP) you have configured. Leave this field blank. After you save this authentication method, it is auto generated. This field must match the Reply URL setting in your IdP.

Something like the following...

> The OIDC Callback URL field registers the service as a service provider (SP) with each OIDC provider (IdP) you have configured. Leave this field blank. After you save this authentication method, it is auto generated. This field must match the Redirect Endpoint URL setting in your IdP.


ansible_base/authentication/authenticator_plugins/oidc.py
```diff
@@ -90,6 +90,15 @@ class OpenIdConnectConfiguration(BaseAuthenticatorConfiguration):
         ui_field_label=_("Authorization URL"),
     )
 
+    CALLBACK_URL = URLField(
+        help_text=_(
+            "Provide this URL as the callback URL for your application as part of your registration process. Refer to the documentation for more detail."
```
tznamena, line 95: Here I would suggest to simply describe the field without telling user(s) what to do.

test_app/tests/conftest.py
```diff
@@ -305,10 +305,11 @@ def google_oauth2_authenticator(google_oauth2_configuration):
 def oidc_configuration():
     return {
         "OIDC_ENDPOINT": "https://localhost/api/gateway/callback/oidc_test/",
-        "OIDC_VERIFY_SSL": True,
+        "VERIFY_SSL": True,
```
tznamena, line 308: Why this change, was the name `OIDC_VERIFY_SSL` wrong?

zkayyali812, line 308: Nice catch

zkayyali812 (approved):

test_app/tests/conftest.py
zkayyali812: I assumed this was wrong initially since OIDC_VERIFY_SSL does not exist in the class, but VERIFY_SSL does.

jshimkus-rh: All the authenticator plugins use VERIFY_SSL (there was a migration that cleaned stuff up; not certain what happened here) and having OIDC_VERIFY_SSL resulted in validation failures.

ansible_base/authentication/authenticator_plugins/oidc.py
jshimkus-rh: This was a copy/paste of CALLBACK_URL from the Azure AD authenticator plugin.  
I used that as the JIRA is about a customer using Generic OIDC in an Azure AD environment.
Other plugins also provide instructions (alternatively worded for their specifics).

test_app/tests/conftest.py
jshimkus-rh: Apparently the first time someone actually tried to use this. 😄

bhavenst:
Looks good to me.  I was just testing this though and wonder if it should go into the keycloak type as well.

Dostonbek1 (approved):

BrennanPaciorek (approved):

BrennanPaciorek:
> Looks good to me. I was just testing this though and wonder if it should go into the keycloak type as well.

This is a good idea, will probably save us future problems if we do it here, but it would also be optional, so I'm approving.

jshimkus-rh:
> > Looks good to me. I was just testing this though and wonder if it should go into the keycloak type as well.
> 
> This is a good idea, will probably save us future problems if we do it here, but it would also be optional, so I'm approving.

It is a reasonable idea.  However, it's a bit of mission creep and, more importantly in the short-term, it would require knock-on changes at a minimum affecting documentation (a la my recent edit of the description of this PR).  

Also, We're proposing (see the associated JIRA) that the UI unconditionally suppress (rather than its current conditional suppression) `CALLBACK_URL` on create of an authenticator.  If we can get that it'd be better to not do this and have to change doc to just then change doc again when it's no longer an issue; i.e., let's see if we can get a good solution and avoid churn.

AlanCoding (approved):

PR #775, opened 2025-07-21
[AAP-48723] Update user.last_login_from when authenticating

fao89:
Issue: AAP-48723
Related: https://github.com/ansible-automation-platform/aap-gateway/pull/908
Signed-off-by: Fabricio Aguiar <fabricio.aguiar@gmail.com>

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

## Description

- What is being changed?
The `AnsibleBaseAuth` backend is modified to update the `last_login_from` attribute of the user object upon successful authentication.
- Why is this change needed?
To track which authenticator a user last used to log in. This is related to issue AAP-48723.
By explicitly setting and saving the `last_login_from` field on the user model after a successful login.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Configure an authenticator.
2. Log in as a user using the configured authenticator.
3. Check the user's `last_login_from` field to see if it is set to the authenticator used.

### Expected Results

The user's `last_login_from` field should be populated with the authenticator they used to log in.

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/authentication/backend.py
+++ b/ansible_base/authentication/backend.py
@@ -56,5 +56,8 @@ class AnsibleBaseAuth(ModelBackend):
                     return None
 
                 logger.info(f'User {user.username} logged in from authenticator with ID "{authenticator_id}"')
+                if hasattr(user, "last_login_from"):
+                    user.last_login_from = authenticator_object.database_instance
+                    user.save(update_fields=['last_login_from'])
                 return user
         return None
--- a/test_app/tests/authentication/test_backend.py
+++ b/test_app/tests/authentication/test_backend.py
@@ -172,3 +172,168 @@ def test_authentication_exception(expected_log):
         # Expect the log we emit
         with expected_log('ansible_base.authentication.backend.logger', "exception", "Exception raised while trying to authenticate with"):
             backend.AnsibleBaseAuth().authenticate(None)
+
+
+@pytest.mark.django_db
+def test_last_login_from_with_attribute(local_authenticator, random_user, expected_log):
+    """Test that last_login_from is set when user has the attribute."""
+    # Add last_login_from attribute to user
+    random_user.last_login_from = None
+
+    # Create a mock authenticator plugin object with proper structure
+    mock_authenticator_plugin = mock.MagicMock()
+    mock_authenticator_plugin.authenticate.return_value = random_user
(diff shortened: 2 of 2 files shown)


zkayyali812 (approved):

ansible_base/authentication/backend.py
```diff
@@ -55,6 +55,9 @@ def authenticate(self, request, *args, **kwargs):
                     )
                     return None
 
-                logger.info(f'User {user.username} logged in from authenticator with ID "{authenticator_id}"')
+                if hasattr(user, "last_login_from"):
+                    logger.info(f'User {user.username} logged in from authenticator with ID "{authenticator_id}"')
```
bhavenst: I'm not sure we want the info msg to appear only if hasattr is true, maybe move it out of the if block?  (though I'm not sure when/why last_login_from wouldn't be a user attribute)

test_app/tests/authentication/test_backend.py
```diff
@@ -172,3 +172,163 @@ def authenticate(*args, **kwars):
...
+            assert random_user.last_login_from == local_authenticator
+
+            # Verify save was called with the correct update_fields
+            mock_save.assert_called_once_with(update_fields=['last_login_from'])
+
+
+@pytest.mark.django_db
+def test_last_login_from_without_attribute(local_authenticator, random_user):
+    """Test that authentication works normally when user doesn't have last_login_from attribute."""
+    # Ensure user doesn't have last_login_from attribute
```
bhavenst, line 215: Yeah in this case I think we should get the user logged in message as well since we always did before.

bhavenst (approved):
Minor comments..

ansible_base/authentication/backend.py
fao89: right now last_login_from isn't an attribute yet, it will be introduced by https://github.com/ansible-automation-platform/aap-gateway/pull/908

AlanCoding:
These sorts of updates are notorious performance killers, if a client is making multiple requests at the same time. Because the settings often make the requests atomic, it has the effect of basically forcing requests to be processed in parallel, as one request waits for the commit of the change from the other requests.

But this is affected by the question of "what constitutes a login", and the UI is using, etc.

PR #778, opened 2025-07-24
AAP-50130: Fallback authentication only on migrated users

zkayyali812:
## Description

- What is being changed? Modifies the controller fallback authentication so that it only is triggered if the gateway user has flag `use_controller_password=true`. Once the controller fallback authentication is completed, this flag is set to false to disallow users from then signing in again with their controller password.
- Why is this change needed? This change is needed to ensure passwords can only be set from their controller passwords once. This provides more security benefits so that the controller password cannot always be used as a "fallback" password, limiting exposure if this password had ever been leaked after it has been changed.
- How does this change address the issue? This change addresses the issue by adding another safeguard check in the fallback controller authentication logic to validate that this flag is set on the user, and if it is, it will update it to false, after the password has been modified.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test

Testing instructions use aap-dev for validation

1. Deploy AAP Dev, with this branch, and also this [gateway PR](https://github.com/ansible-automation-platform/aap-gateway/pull/918)
2. Check `/gateway/v1/users/`, validate the administrator account has `use_controller_password=false`
3. Attempt to sign in via gateway with administrator controller password. This should fail. Only gateway password will work
4. Exec into controller-task container and run the following command to create a user on controller only - 
```bash
ANSIBLE_REVERSE_RESOURCE_SYNC=false awx-manage shell_plus --quiet-load -c "User.objects.create_user(username='dummy', email='dummy@example.com', password='test', first_name='dummy', last_name='user')"
```
5. Exec into gateway-api pod and run the migrate_service_data command -
```bash
aap-gateway-manage migrate_service_data --username=admin
```
6. Check `/gateway/v1/users`, validate dummy account exists, with `use_controller_password=true`
7. Sign into gateway with the dummy user account, using the password created in controller, sign in should succeed.
8. Check `/gateway/v1/users`, see that, with `use_controller_password=false`
9. Change dummy user password in gateway
10. Attempt to sign in with dummy user again using original controller password. Signin should fail. Only gateway password can be used now.

### Expected Results

Controller fallback authentication can only happen once for any account, and only on those accounts which were migrated via migrate_service_data

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/local.py
+++ b/ansible_base/authentication/authenticator_plugins/local.py
@@ -94,9 +94,15 @@ class AuthenticatorPlugin(ModelBackend, AbstractAuthenticatorPlugin):
         If the user is valid, update the gateway users credentials with the controller credentials.
         """
         try:
-            UserModel._default_manager.get_by_natural_key(username)
+            user = UserModel._default_manager.get_by_natural_key(username)
         except UserModel.DoesNotExist:
             logger.warning(f"User '{username}' does not exist in the database.")
+            return False
+
+        # Skip controller authentication if user has use_controller_password field set to False
+        # Default to False when field doesn't exist (test environments)
+        if not getattr(user, 'use_controller_password', False):
+            logger.warning(f"User '{username}' password not in Controller.")
             return False
 
         if controller_user := self._get_controller_user(username, password):
@@ -174,7 +180,14 @@ class AuthenticatorPlugin(ModelBackend, AbstractAuthenticatorPlugin):
         """
         user = UserModel._default_manager.get_by_natural_key(username)
         user.set_password(password)
-        user.save(update_fields=['password'])
+
+        # Set use_controller_password to False if the field exists
+        update_fields = ['password']
+        if hasattr(user, 'use_controller_password'):
+            user.use_controller_password = False
+            update_fields.append('use_controller_password')
+
+        user.save(update_fields=update_fields)
         logger.info(f"Updated user {username} gateway account")
 
     def _convert_to_seconds(self, s):
--- a/ansible_base/authentication/utils/authentication.py
+++ b/ansible_base/authentication/utils/authentication.py
@@ -145,11 +145,7 @@ def _handle_no_merge_strategy(uid: str, authenticator: Authenticator) -> str:
     if AuthenticatorUser.objects.filter(Q(uid=uid) | Q(user__username=uid)).exists():
         # Some other provider is providing this username so we need to create our own username
         new_username = get_local_username({'username': uid})
-        logger.info(
-            f'Authenticator {authenticator.name} wants to authenticate {uid} but that'
-            f' username is already in use by another authenticator,'
-            f' the user from this authenticator will be {new_username}'
-        )
+        logger.warning(f"User {uid} is already associated with an existing user, creating a new user with username {new_username}")
         return new_username
     else:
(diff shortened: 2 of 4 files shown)


ansible_base/authentication/authenticator_plugins/local.py
```diff
@@ -94,11 +94,17 @@ def _can_authenticate_from_controller(self, username, password):
...
-            UserModel._default_manager.get_by_natural_key(username)
+            user = UserModel._default_manager.get_by_natural_key(username)
         except UserModel.DoesNotExist:
             logger.warning(f"User '{username}' does not exist in the database.")
             return False
 
+        # Skip controller authentication if user has use_controller_password field set to False
+        # Default to False when field doesn't exist (test environments)
+        if not getattr(user, 'use_controller_password', False):
+            logger.warning(f"User '{username}' has a usable password already.")
```
chrismeyersfsu:
```suggestion
            logger.warning(f"User '{username}' password not in Controller.")
```

There will be users with `use_controller_password=False` and unsuable password (LDAP users in Gateway).

zkayyali812: Great suggestion, updated

chrismeyersfsu (approved):

jshimkus-rh:
Looking at these step in the testing procedure and  the code...

> 6. Check /gateway/v1/users, validate dummy account exists, with use_controller_password=true
> 7. Sign into gateway with the dummy user account, using the password created in controller, sign in should succeed.
> 8. Check /gateway/v1/users, see that, with use_controller_password=false 
> 9. Change dummy user password in gateway

it appears that if the user logs out before changing the password that they will be prevented from logging in again.

zkayyali812:
@jshimkus-rh In this case, the gateway password will be updated to be the controller password, so the user will be able to login as expected again using their gateway password.

jshimkus-rh:
Ah, missed that as it was pre-existing and just outside the changes.

jshimkus-rh (approved):

ansible_base/authentication/utils/authentication.py
```diff
@@ -145,11 +145,7 @@ def _handle_no_merge_strategy(uid: str, authenticator: Authenticator) -> str:
     if AuthenticatorUser.objects.filter(Q(uid=uid) | Q(user__username=uid)).exists():
         # Some other provider is providing this username so we need to create our own username
         new_username = get_local_username({'username': uid})
-        logger.info(
-            f'Authenticator {authenticator.name} wants to authenticate {uid} but that'
-            f' username is already in use by another authenticator,'
-            f' the user from this authenticator will be {new_username}'
-        )
+        logger.warning(f"User {uid} is not found in the database, creating a new user with username {new_username}")
```
bhavenst: This warning doesn't make sense to me.  It appears that this branch is for uid found in db (but w/ a different authenticator) not that it's missing from the db.

bhavenst (approved):

ansible_base/authentication/utils/authentication.py
zkayyali812: Great call,
I will update to this log - 
```
logger.warning(f"User {uid} is already associated with an existing user, creating a new user with username {new_username}")
```

Issue #534, opened 2024-07-30
The role assignment serializers fail to give fields like `user_ansible_id`


AlanCoding:
See test in:

https://github.com/ansible/django-ansible-base/pull/527

Assignment serializers show ansible_id fields but they are returned blank.

We need a near-term solution for this.

AlanCoding:
expected fixed with backlinked patch

PR #783, opened 2025-07-30
AAP-50626 AAP-50627 Use ansible_id references in access lists

AlanCoding:
This is needed to complete the UI connection of the access lists.

Background:

Due to technical issues, we have asked the UI to use the component-based access list for resources owned by a certain component. So if there's an AWX resource, get the access list served from AWX for that item, like an inventory or job template.

However, our endpoints were still using the local team or user id. This is a problem, because all other RBAC management is centralized around the resource server. So a translation is needed, and right now there's no easy way to do that translation. So the solution (this) is to formulate all the access list related endpoints around the ansible_id of the actor, that being the user or the team.

Corresponding to this, we need a fix for the bug https://github.com/ansible/django-ansible-base/issues/534, so that the client has the ansible_id as a starting point when navigating to the component endpoints

### DIFF
--- a/ansible_base/rbac/api/serializers.py
+++ b/ansible_base/rbac/api/serializers.py
@@ -132,15 +132,6 @@ class BaseAssignmentSerializer(CommonModelSerializer):
 
         return super().validate(attrs)
 
-    def get_actor_from_data(self, validated_data, requesting_user):
-        actor_aid_field = f'{self.actor_field}_ansible_id'
-        if validated_data.get(self.actor_field):
-            actor = validated_data[self.actor_field]
-        else:
-            # Actor is already resolved by ActorAnsibleIdField and validated
-            actor = validated_data[self.actor_field]
-        return actor
-
     def get_object_from_data(self, validated_data, role_definition, requesting_user):
         obj = None
         if validated_data.get('object_id') and validated_data.get('object_ansible_id'):
@@ -175,7 +166,7 @@ class BaseAssignmentSerializer(CommonModelSerializer):
         requesting_user = self.context['view'].request.user
 
         # Resolve actor - team or user
-        actor = self.get_actor_from_data(validated_data, requesting_user)
+        actor = validated_data[self.actor_field]
 
         # Resolve object
         obj = self.get_object_from_data(validated_data, rd, requesting_user)
@@ -285,7 +276,6 @@ class AccessListMixin:
         except ObjectDoesNotExist:
             # Resource doesn't exist, stick with pk
             logger.error(f"Resource does not exist for {self.Meta.model} {obj.pk}")
-            pass
 
         related_fields['details'] = get_relative_url(
             f'role-{actor_cls._meta.model_name}-access-assignments',
--- a/ansible_base/rbac/api/serializers.py
+++ b/ansible_base/rbac/api/serializers.py
@@ -125,9 +125,7 @@ class BaseAssignmentSerializer(CommonModelSerializer):
         has_actor_in_request = self.actor_field in self.initial_data
         has_actor_aid_in_request = actor_aid_field in self.initial_data
 
-        if has_actor_in_request and has_actor_aid_in_request:
-            self.raise_id_fields_error(self.actor_field, actor_aid_field)
-        elif not has_actor_in_request and not has_actor_aid_in_request:
+        if (has_actor_in_request and has_actor_aid_in_request) or (not has_actor_in_request and not has_actor_aid_in_request):
             self.raise_id_fields_error(self.actor_field, actor_aid_field)
 
         return super().validate(attrs)
@@ -153,10 +151,11 @@ class BaseAssignmentSerializer(CommonModelSerializer):
         elif validated_data.get('object_ansible_id'):
             obj = self.get_by_ansible_id(validated_data.get('object_ansible_id'), requesting_user, for_field='object_ansible_id')
             if permission_registry.content_type_model.objects.get_for_model(obj) != role_definition.content_type:
+                model_name = getattr(role_definition.content_type, 'model', 'global')
                 raise ValidationError(
                     {
                         'object_ansible_id': _('Object type of %(model_name)s does not match role type of %(role_definition)s')
-                        % {'model_name': obj._meta.model_name, 'role_definition': role_definition.content_type.model}
+                        % {'model_name': obj._meta.model_name, 'role_definition': model_name}
                     }
                 )
         return obj
--- a/ansible_base/rbac/api/serializers.py
+++ b/ansible_base/rbac/api/serializers.py
@@ -20,6 +20,7 @@ from ansible_base.rbac.validators import check_locally_managed, validate_permiss
 
 from ..models import DABContentType, DABPermission
 from ..remote import RemoteObject
+from .fields import ActorAnsibleIdField
 from .queries import assignment_qs_user_to_obj, assignment_qs_user_to_obj_perm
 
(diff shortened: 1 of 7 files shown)


AlanCoding:
Demo from test_app

http://localhost:8000/api/v1/role_user_access/shared.organization/1/

```json
{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 2,
            "url": "/api/v1/users/2/",
            "related": {
                "details": "/api/v1/role_user_access/shared.organization/1/b6283e9e-f04f-4a50-a978-28b51af9de9e/"
            },
            "username": "admin",
            "is_superuser": true,
            "object_role_assignments": []
        },
        {
            "id": 6,
            "url": "/api/v1/users/6/",
            "related": {
                "details": "/api/v1/role_user_access/shared.organization/1/b62c1b42-2dc4-4d0f-9216-d51409013809/"
            },
            "username": "org_admin",
            "is_superuser": false,
            "object_role_assignments": [
                {
                    "type": "direct",
                    "role_definition": {
                        "name": "Organization Admin",
                        "url": "/api/v1/role_definitions/2/"
                    }
                }
            ]
        }
    ]
}
```

Follow through the link, this works

http://localhost:8000/api/v1/role_user_access/shared.organization/1/b62c1b42-2dc4-4d0f-9216-d51409013809/

```json
{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 1,
            "related": {
                "role_definition": "/api/v1/role_definitions/2/",
                "user": "/api/v1/users/6/",
            },
            "summary_fields": {
                "created_by": {
                    "id": 1,
                    "username": "_system",
                    "first_name": "",
                    "last_name": ""
                },
                "role_definition": {
                    "id": 2,
                    "name": "Organization Admin",
                    "description": "Has all permissions to a single organization and all objects inside of it",
                    "managed": true
                },
                "user": {
                    "id": 6,
                    "username": "org_admin",
                    "first_name": "",
                    "last_name": ""
                },
                "content_object": {
                    "id": 1,
                    "name": "AWX_community"
                }
            },
            "created": "2025-07-30T19:35:22.833517Z",
            "created_by": 1,
            "object_id": "1",
            "object_ansible_id": null,
            "role_definition": 2,
            "user": 6,
            "user_ansible_id": "b62c1b42-2dc4-4d0f-9216-d51409013809",
            "intermediary_roles": []
        }
    ]
}
```

So I'm happy with this.

AlanCoding:
This gives team access

http://localhost:8000/api/v1/role_team_access/aap.inventory/2/

And the link from there works

http://localhost:8000/api/v1/role_team_access/aap.inventory/2/dc661d94-4e3d-410c-b342-ddf4593f6aa5/

ansible_base/rbac/api/serializers.py
```diff
@@ -112,17 +117,18 @@ def get_by_ansible_id(self, ansible_id, requesting_user, for_field):
...
+    def validate(self, attrs):
+        """Validate that exactly one of actor or actor_ansible_id is provided"""
         actor_aid_field = f'{self.actor_field}_ansible_id'
-        if validated_data.get(self.actor_field) and validated_data.get(actor_aid_field):
+
+        # Check what was actually provided in the request
+        has_actor_in_request = self.actor_field in self.initial_data
+        has_actor_aid_in_request = actor_aid_field in self.initial_data
+
+        if (has_actor_in_request and has_actor_aid_in_request) or (not has_actor_in_request and not has_actor_aid_in_request):
```
john-westcott-iv: Can this be simplified to:
```suggestion
        # If both actor and actor_ansible_id are present or both not present than we error out
        if has_actor_in_request == has_actor_aid_in_request:
```

ansible_base/rbac/api/serializers.py
```diff
@@ -257,9 +266,19 @@ def _get_related(self, obj) -> dict[str, str]:
...
         actor_cls = self.Meta.model
+
+        # Use ansible_id if available, otherwise fall back to pk
+        actor_identifier = obj.pk
+        try:
+            if hasattr(obj, 'resource') and obj.resource:
+                actor_identifier = str(obj.resource.ansible_id)
+        except ObjectDoesNotExist:
+            # Resource doesn't exist, stick with pk
+            logger.error(f"Resource does not exist for {self.Meta.model} {obj.pk}")
```
john-westcott-iv: Should this actually be an error or just a warning since we are going to continue on?
```suggestion
            logger.warn(f"Resource does not exist for {self.Meta.model} {obj.pk}, unable to get ansible_id, will attempt to use PK instead")
```

ansible_base/rbac/api/views.py
```diff
@@ -379,6 +382,28 @@ class UserAccessAssignmentViewSet(
     def get_url_actor(self):
         actor_pk = self.kwargs.get("actor_pk")
         actor_cls = self.get_actor_model()
+
+        # First, try to parse as UUID for ansible_id lookup
+        try:
+            parsed_uuid = uuid.UUID(actor_pk)
```
john-westcott-iv, line 388: Not for this PR but my brain is wondering if we want to make an AnsibleID object that would "abstract" uuid so that if we ever wanted to change it we only have to do it in one place.

john-westcott-iv (approved):

john-westcott-iv:
Approving with minor comments.

ansible_base/rbac/api/serializers.py
AlanCoding: Yes, I agree it can be a warn. We are only here when `hasattr(obj, 'resource')` is true. In that case the model _is_ registered with the resource registry, but the signals failed to do their expected job of creating an associated resource object.

The related link will still kind of work, but on the higher-level of UI-API interaction, the user experience will break.

I'll reword the message, but will diverge somewhat from your suggestion.

ansible_base/rbac/api/serializers.py
AlanCoding: so smart

AlanCoding:
Super surprising got a sqlite3 specific failure

```
FAILED test_app/tests/rbac/api/test_access_assignment_uuid_lookup.py::TestAccessListRelatedLinksWithAnsibleID::test_team_access_list_related_links_use_ansible_id - AssertionError: assert not True
 +  where True = <built-in method endswith of str object at 0x7fe5f4ef32b0>('1/')
 +    where <built-in method endswith of str object at 0x7fe5f4ef32b0> = '/api/v1/role_team_access/aap.inventory/1/5ca295bc-9064-47da-8983-7e1ed5592d31/'.endswith
= 1 failed, 2103 passed, 25 skipped, 13 xfailed, 42 warnings in 323.82s (0:05:23) =
py311sqlite: exit 1 (328.71 seconds) /home/runner/work/django-ansible-base/django-ansible-base> pytest -n auto --cov=. --cov-report=xml:coverage.xml --cov-report=html --cov-report=json --cov-branch --junit-xml=django-ansible-base-test-results.xml test_app/tests pid=3955
py311sqlite: docker> remove '113dcd32ace8' (from 'db')
```


Ooooooh! I see it! This is numerology flake. The UUID string _just happens to_ end with the same decimal number as the primary key.

```
E        +    where <built-in method endswith of str object at 0x7fe5f4ef32b0> = '/api/v1/role_team_access/aap.inventory/1/5ca295bc-9064-47da-8983-7e1ed5592d31/'.endswith
```

Because we asserted that it does not end with `"1/"`, and it does! Because the uuid4 ends with 1 for this one single run. That will fail 1 out of every 10 times.

PR #787, opened 2025-08-05
AAP-50614 Add a methods for user claims serialization and hashing

AlanCoding:
## Description
Up until this point, the DAB app jwt_consumer has housed logic for _saving_ user claims to the RBAC models in the local database.

We have planned a change which will modify this contract. A service will still get serialized claims from the resource server and save them by existing jwt_consumer logic, but when this happens we will store a hash of those claims (producible from the claims data, no DB interaction needed) in the local cache. Then, we will cross-check against this hash for future requests and do nothing if it matches. This will allow us to move the transmission of the cache data to a new mechanism which avoids the header size limitations that plague JWT usage right now.

This just does this first part - move _generation_ of claims data into DAB. Before this we _consumed_ the claims data, and this remains unchanged. Additionally, we add some simple logic to _hash_ the claims data and test that the hash is consistent.

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Expected Results
Generated hash should match claims data that is currently a subset of the JWT.

### Required Actions
- [ ] Blocked by PR/MR: #XXX


ansible_base/rbac/claims.py
```diff
@@ -0,0 +1,317 @@
...
+    Example usage:
+        assignments = get_user_object_roles(user)
+        for assignment in assignments:
+            print(assignment.rd_name, assignment.aid, assignment.resource_name, assignment.content_type_id)
+    """
+    # Create subqueries for resource data
+    resource_cls = apps.get_model('dab_resource_registry', 'Resource')
+    ansible_id_subquery = resource_cls.objects.filter(object_id=OuterRef('object_id'), content_type=OuterRef('content_type')).values('ansible_id')
+
+    resource_name_subquery = resource_cls.objects.filter(object_id=OuterRef('object_id'), content_type=OuterRef('content_type')).values('name')
```
AlanCoding, line 46: :point_up: this is a novel change added by me.

Before, we were doing a query of `Organization.objects.filter` or something like that, and that would get the name so that we can fill in the rest of the needed object data. But here I am, saying, the `Resource` model already caches the name, so no need to. We just add it to our fancy query here, and that avoids the need for a specific Organization-based query later on to fill in the object data.

This does _not_ work for the `Team` model, and that reveals a true "special case" for teams. I am trying to think ahead for the scenario where we communicate _all_ user permissions in the claims data. When that happens, those will be remote objects. For remote objects, there will be no need to summarize the object, because it doesn't live in Gateway, and we will not communication instructions for creating the object. We _only_ need to include a non-name field for the Team model, and that is the reference to its organization. So there is some additional purpose-built logic just for handing the team-org reference, which starts to become logical.

test_app/tests/rbac/test_claims.py
```diff
@@ -0,0 +1,472 @@
...
+        print("Query details:")
+        for i, query in enumerate(connection.queries[queries_before:], 1):
+            print(f"  {i}. {query['sql'][:100]}{'...' if len(query['sql']) > 100 else ''}")
+            print(f"     Time: {query['time']}s")
+        print("=" * 45)
+
+        # Assert we maintain our improved performance baseline
+        # Baseline improved after N+1 fix: 4 queries for the mixed_large scenario
+        # (Even more efficient: team organization resolution optimized with single join query)
+        assert total_queries == 4, f"Claims generation used {total_queries} queries, expected 4 (improved baseline)"
```
AlanCoding, line 410: Because of that 1 novel change I called out, I do believe this is better than before.

ansible_base/rbac/claims.py
```diff
@@ -0,0 +1,306 @@
...
+    for team_data in object_arrays['team']:
+        team_ansible_id = team_data['ansible_id']
+        org_info = team_org_mapping.get(team_ansible_id)
+
+        if org_info:
+            org_ansible_id = org_info['ansible_id']
+            org_name = org_info['name']
+
+            # Ensure the organization is in our arrays
+            if org_ansible_id not in org_ansible_id_to_index:
```
fosterseth, line 102: trying to determine if this block of code is needed. are there situations where the org id isn't in the provided array? i guess it doesn't hurt to play defense

AlanCoding: Yes, the test case "teams_no_orgs" was created specifically for this purpose.

This happens for the incredibly common case where a user is a member of a team, but has no role to the team's associated organization. Up until this point, we only fill out the object data for the case where the user has a direct role assignment to a particular object.

The consequence really helps to better clarify what it is we're doing in the first place. You might make a request, and you are a member of a team. Within the course of that request, not only will the app have to create the _team_ specified by the claims data, but will also have to _create the organization the team goes in_ because both of these entities came from SSO claims data ultimately.

So all this JWT claims stuff is a data serializer/creator in the general. It just so happens that we only solve the problem for only 2 models - organizations and teams. The logic to create the objects an apply assignments already exists in the jwt_consumer app.

To close the loop, we need to better formalize the _saver_ logic and test that locally to DAB. So create user1, give that user a lot of permissions. Then serializer user1's permissions, and apply those permissions to user2. Ok, now delete _everything_ the users, teams, orgs, create a new user, and apply the claims again, and see that everything in the chain is re-created.

AlanCoding: The saver logic is functionally this:

https://github.com/ansible/django-ansible-base/blob/44e4ae8b3ca5c5ac54676cfa69c5e0298938b960/ansible_base/jwt_consumer/common/auth.py#L259

I would prefer that be moved here, and probably refactored and worked into the testing.

ansible_base/rbac/claims.py
```diff
@@ -0,0 +1,306 @@
...
+    object_arrays, object_roles = _build_objects_and_roles(user)
+
+    # Get global roles
+    global_roles = _get_user_global_roles(user)
+
+    # Build final claims structure
+    return {'objects': object_arrays, 'object_roles': object_roles, 'global_roles': global_roles}
+
+
+def get_user_claims_hashable_form(claims: dict) -> dict[str, Union[list[str], dict[str, list[str]]]]:
```
fosterseth, line 236: is there a reason we can't just call this in get_claims_hash?

I expected the flow to be

```
claims = get_user_claims()
hashed = get_claims_hash(claims)
```

instead we have a call sandwiched in there

fosterseth: this is code nitpick, we don't need to make a change necessarily, but just wondering why it's split up

fosterseth (approved):

ansible_base/rbac/claims.py
AlanCoding: What I think would be better-aligned with our eventual use is a method that gets the claims hash from a `user` object passed in. So that would be 3 internal calls, get claims, hashable form, then hash. I don't think we should combine pre-maturely, but adding another method that combines them would be totally valid.

AlanCoding:
Linking for later

https://github.com/ansible/django-ansible-base/blob/44e4ae8b3ca5c5ac54676cfa69c5e0298938b960/ansible_base/jwt_consumer/common/auth.py#L259

PR #788, opened 2025-08-06
[AAP-50642] memoize pem private key to avoid openssl slowdown

kdelee:
For 2.6, we've moved to RHEL9 for the base container, which has openssl 3.2 in it. Apparently , its is a known issue that the RSA key validation is more secure and more slow in this version vs. Openssl 1 that RHEL8 uses. We were loading and validating the same jwt private key on every request, at a cost of ~ 38 ms, vs previously this took ~ 3 ms. We noticed this slowdown in the benchmarks and after deep dive/profiling we found this was the difference and memoizing the validated private key saves us from this broad slowdown. See https://issues.redhat.com/browse/AAP-50642 for more details

https://github.com/pyca/cryptography/issues/7236#issuecomment-1131908472 <--- report that details Openssl 3 slowdown

## Description

- What is being changed?
- Why is this change needed?

## Type of Change

## Self-Review Checklist

### Prerequisites

The speedup is visible running hte benchmark tests in the test-suite
```
pytest tests/perf/benchmark/test_api_benchmark.py -k 'test_benchmark_controller_endpoint_list_response[notifications]' --tsd=../tsd.json
```

### Expected Results

~ 30ms speedup, which brings us back in line with 2.5 results on containerized and OCP
Note: we didn't see the slow down comparing AAP 2.5 and AAP 2.6 on RHEL9, because they BOTH use same openssl version. The difference between containerized and OCP AAP 2.5 and AAP 2.6 is the base image (we moved from RHEL8 to RHEL9)

## Additional Context

For the return value type of the function, I did try importing the class that this returns `from cryptography.hazmat.bindings._rust.openssl.rsa import RSAPrivateKey` but that caused the gRPC service to not be able to start 

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/jwt_consumer/common/util.py
+++ b/ansible_base/jwt_consumer/common/util.py
@@ -22,7 +22,7 @@ def _load_pem_private_key(key: str):
     return serialization.load_pem_private_key(bytes(key, 'utf-8'), password=None)
 
 def generate_x_trusted_proxy_header(key: str) -> str:
-    private_key = _load_private_key(key)
+    private_key = _load_pem_private_key(key)
     timestamp = time.time_ns()
     message = f'{_SHARED_SECRET}-{timestamp}'
     signature = private_key.sign(bytes(message, 'utf-8'), padding.PSS(mgf=padding.MGF1(hashes.SHA256()), salt_length=padding.PSS.MAX_LENGTH), hashes.SHA256())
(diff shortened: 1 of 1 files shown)


chrismeyersfsu:
Closing in favor of https://github.com/ansible/django-ansible-base/pull/791

PR #789, opened 2025-08-06
[AAP-50407] Refactor JWT claims handling to use gateway endpoint

TheRealHaoLiu:
# [AAP-50407](https://issues.redhat.com/browse/AAP-50407) JWT Claims Gateway Integration - Complete Implementation

## Description

This PR implements a complete migration from JWT token-embedded claims to gateway-sourced claims via the `service-index/jwt_claims/<user_ansible_id>` endpoint.

### What is being changed?
- **JWT Authentication**: Modified to fetch permission claims from gateway service-index API instead of using embedded JWT token fields
- **RBAC Processing**: Updated to exclusively use gateway claims data with no fallback to deprecated JWT token fields
- **JWT Validation**: Removed deprecated fields (`objects`, `object_roles`, `global_roles`) from required JWT token validation
- **Resource Creation**: Updated to use gateway claims exclusively for building organization and team resources

- **Deprecation**: JWT token fields `objects`, `object_roles`, and `global_roles` are being deprecated across the platform
- **Centralization**: Move permission data management to the gateway service for better consistency and real-time updates
- **Performance**: Reduce JWT token size by removing large permission data structures
- **Architecture**: Clean separation between authentication (JWT) and authorization (gateway)

- Implements `get_jwt_claims()` method in `ResourceAPIClient` to call the gateway endpoint
- Updates `JWTCommonAuth` to fetch and store gateway claims in `self.gateway_claims`
- Modifies `process_rbac_permissions()` and `get_or_create_resource()` to use gateway data exclusively
- Removes all fallback logic to deprecated JWT token fields
- Provides clear error handling when gateway claims are unavailable

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Self-Review Checklist

### Prerequisites
- Gateway service running with functional `service-index/jwt_claims/<user_ansible_id>` endpoint
- Valid JWT tokens for authentication
- Test users with various permission configurations

### Steps to Test
1. **Basic Authentication Test**:
   - Send JWT token via `X-DAB-JW-TOKEN` header
   - Verify user authentication succeeds
   - Check that `self.gateway_claims` is populated

2. **RBAC Processing Test**:
   - Authenticate user with various role assignments
   - Call `process_rbac_permissions()`
   - Verify roles are correctly assigned from gateway claims data
   - Confirm no usage of deprecated JWT token fields

3. **Resource Creation Test**:
   - Test with team and organization creation scenarios
   - Verify resources are created using gateway claims data
   - Test error handling when gateway claims are unavailable

4. **Error Handling Test**:
   - Simulate gateway endpoint unavailability
   - Verify authentication succeeds but RBAC processing fails gracefully
   - Check appropriate error messages are logged

### Expected Results
- JWT authentication works without deprecated token fields
- Permission processing exclusively uses gateway claims
- Clear error messages when gateway is unavailable
- No fallback to deprecated JWT token fields under any circumstance

### Required Actions

### Breaking Changes

This is a **breaking change** that affects systems relying on JWT token embedded claims:

1. **Gateway Dependency**: RBAC functionality now requires the gateway `service-index/jwt_claims/` endpoint
2. **No Fallback**: Systems cannot process permissions without gateway connectivity
3. **JWT Token Changes**: Deprecated fields are no longer required in JWT tokens

### Migration Impact

**Before**: JWT tokens contained full permission data
```json
{
  "sub": "user-id",
  "objects": {"organization": [...], "team": [...]},
  "object_roles": {"Org Admin": {...}},
  "global_roles": ["Platform Auditor"]
}
```

**After**: JWT tokens contain minimal auth data, permissions from gateway
```json
{
  "sub": "user-id",
  "user_data": {...}
}
```

#### Files Modified
- `ansible_base/jwt_consumer/common/auth.py` - Main JWT processing logic
- `ansible_base/jwt_consumer/common/cache.py` - Cache methods (cleanup)
- `ansible_base/resource_registry/rest_client.py` - Gateway API client
- `test_app/tests/conftest.py` - Test fixtures updated

#### Key Methods Updated
- `JWTCommonAuth.parse_jwt_token()` - Now fetches from gateway
- `JWTCommonAuth.process_rbac_permissions()` - Uses gateway claims only
- `JWTCommonAuth.get_or_create_resource()` - Uses gateway claims only
- `ResourceAPIClient.get_jwt_claims()` - New gateway endpoint method

#### Error Handling Strategy
- **Authentication**: Succeeds even if gateway unavailable (basic user creation)
- **Authorization**: Fails gracefully with clear error messages
- **Logging**: Detailed debug/error logs for troubleshooting

### Performance Considerations

- **Network Calls**: Now makes HTTP request to gateway for each JWT authentication
- **Caching**: Removed complex claims-hash optimization (can be re-added later)
- **Token Size**: JWT tokens will be significantly smaller without embedded claims

### Security Impact

- **Separation of Concerns**: Authentication (JWT) and authorization (gateway) are cleanly separated
- **Real-time Permissions**: Gateway provides up-to-date permission data
- **Reduced Attack Surface**: Smaller JWT tokens, less embedded data to validate

This implementation provides a clean, no-fallback solution that exclusively uses the gateway service for permission claims while maintaining robust error handling for operational scenarios.

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -181,7 +181,7 @@ class JWTCommonAuth:
             # Recalculate hash from local database to verify the mismatch
             # It is possible that the cached hash is stale, but the local data is synced to the resource server.
             # This is an optimization to avoid fetching claims from the resource server if the local data is synced.
-            logger.debug(f"Claims hash mismatch for user {user_ansible_id}: cached={cached_hash}, current={current_claims_hash}")
+            logger.debug(f"Claims hash mismatch for user {user_ansible_id}: cached={cached_hash}, from token={current_claims_hash}")
             logger.debug(f"Recalculating hash from local database for user {user_ansible_id}")
 
             try:
@@ -190,13 +190,14 @@ class JWTCommonAuth:
                 hashable_claims = get_user_claims_hashable_form(user_claims)
                 recalculated_hash = get_claims_hash(hashable_claims)
 
-                logger.debug(f"Recalculated hash for user {user_ansible_id}: {recalculated_hash}")
                 # Compare recalculated hash with current hash from token
                 if recalculated_hash != current_claims_hash:
-                    logger.debug(f"Claims hash still differs after recalculation for user {user_ansible_id}: local={recalculated_hash}, current={current_claims_hash}")
+                    logger.debug(f"Claims hash still differs after recalculation for user {user_ansible_id}: local={recalculated_hash}, from token={current_claims_hash}")
                     return True
                 else:
-                    logger.debug(f"Recalculated hash matches current hash for user {user_ansible_id}")
+                    logger.debug(f"Recalculated local hash matches token hash for user {user_ansible_id}")
+                    logger.debug(f"Caching claims hash for user {user_ansible_id}: {recalculated_hash}")
+                    self._cache_claims_hash(user_ansible_id, recalculated_hash)
                     return False
 
             except Exception as e:
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -19,6 +19,7 @@ from ansible_base.lib.logging.runtime import log_excess_runtime
 from ansible_base.lib.utils.auth import get_user_by_ansible_id
 from ansible_base.lib.utils.translations import translatableConditionally as _
 from ansible_base.resource_registry.models import Resource, ResourceType
+from ansible_base.resource_registry.rest_client import get_resource_server_client
 from ansible_base.resource_registry.signals.handlers import no_reverse_sync
 
 logger = logging.getLogger("ansible_base.jwt_consumer.common.auth")
@@ -52,6 +53,7 @@ class JWTCommonAuth:
         self.cache = JWTCache()
         self.user = None
         self.token = None
+        self.gateway_claims = None  # Store claims from gateway
 
     @log_excess_runtime(logger, debug_cutoff=0.01)
     def parse_jwt_token(self, request):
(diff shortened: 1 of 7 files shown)


TheRealHaoLiu:
no matter how much test i add... sonar just saying i have 0% coverage for the new code! LIES!

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -142,10 +144,94 @@ def parse_jwt_token(self, request):
...
+        if not current_claims_hash:
+            logger.debug(f"No claims_hash in token for user {user_ansible_id}, will fetch claims")
+            return True
+
+        cached_hash = self.cache.get_claims_hash(user_ansible_id)
+        if cached_hash != current_claims_hash:
+            logger.debug(f"Claims hash changed for user {user_ansible_id}: cached={cached_hash}, current={current_claims_hash}")
+            return True
+
+        # Hash matches cached value, try to get cached claims
```
AlanCoding, line 208: No, don't do that.

If hashes match, do nothing. Don't do approximately nothing, do _exactly_ nothing. Don't save the claims for later. If the hashes match that means the claims _have already been saved_.

TheRealHaoLiu: will update to do approximately nothing 🤣

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -52,6 +53,7 @@ def __init__(self, user_fields=default_mapped_user_fields) -> None:
         self.cache = JWTCache()
         self.user = None
         self.token = None
+        self.gateway_claims = None  # Store claims from gateway
```
AlanCoding, line 57: Very strongly convinced this shouldn't be on `self`. Let the de-referenced be garbage collected.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -259,16 +346,23 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
             return
 
         from ansible_base.rbac.models import RoleUserAssignment
 
         role_diff = RoleUserAssignment.objects.filter(user=self.user, role_definition__name__in=settings.ANSIBLE_BASE_JWT_MANAGED_ROLES)
 
-        for system_role_name in self.token.get("global_roles", []):
+        # Process global roles from gateway claims
+        global_roles = self.gateway_claims.get("global_roles", [])
+        for system_role_name in global_roles:
```
AlanCoding, line 390: You're trying too hard to not change the existing code. And in this case it's not good code.

Make a new method, like `save_claims(user, claims)`. See other stuff in `ansible_base/rbac/claims.py`, it should fit in with that crowd.

It should take the claims, and make them true for that user. You don't _have_ to refactor the logic itself here (although you should, you don't _have_ to), but you do really need to refactor the interface.

This method should be callable from unit tests, and it should be called from unit tests. We shouldn't need to see the JWT auth class within a mile of that logic.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -142,10 +144,94 @@ def parse_jwt_token(self, request):
...
+            logger.debug(f"Claims hash matches but no cached claims found for user {user_ansible_id}")
+            return True
+
+    def _cache_claims_hash(self, user_ansible_id, claims_hash):
+        """Cache the claims hash and gateway claims for future comparisons."""
+        if claims_hash and self.gateway_claims:
+            self.cache.set_claims_hash(user_ansible_id, claims_hash)
+            self.cache.set_cached_claims(user_ansible_id, self.gateway_claims)
+
+    def _fetch_jwt_claims_from_gateway(self, user_ansible_id):
```
AlanCoding, line 223:
```suggestion
    def fetch_jwt_claims_from_gateway(self, user_ansible_id) -> dict[str,dict[dict[dict[str,list],list],dict]:
```

ansible_base/jwt_consumer/hub/auth.py
AlanCoding, line 1: Just do what https://github.com/ansible/django-ansible-base/pull/583 does. I'm telling you right now, the reason this is a duplicated mess is because:

> # TODO - galaxy does not have an org admin roledef yet

and that this should no longer apply. If galaxy_ng errors because it doesn't have an Organization Admin role, add it there.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -142,10 +144,94 @@ def parse_jwt_token(self, request):
...
+        Returns True if claims need to be fetched (hash changed or not cached).
+        """
+        if not current_claims_hash:
+            logger.debug(f"No claims_hash in token for user {user_ansible_id}, will fetch claims")
+            return True
+
+        cached_hash = self.cache.get_claims_hash(user_ansible_id)
+        if cached_hash != current_claims_hash:
+            logger.debug(f"Claims hash changed for user {user_ansible_id}: cached={cached_hash}, current={current_claims_hash}")
+            return True
```
AlanCoding: As a modifier to this, from another tangent conversation - if hashes mismatch, you _can_ re-generate the local hash, and if that matches, replace the local cache value with it, and return `True`. But if you re-generate the local hash and it still doesn't match, then request has to be made.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -177,8 +178,31 @@ def _should_fetch_claims_from_gateway(self, user_ansible_id, current_claims_hash
...
+                # Get user claims from local database
+                user_claims = get_user_claims(self.user)
+                hashable_claims = get_user_claims_hashable_form(user_claims)
+                recalculated_hash = get_claims_hash(hashable_claims)
+
+                logger.debug(f"Recalculated hash for user {user_ansible_id}: {recalculated_hash}")
+                # Compare recalculated hash with current hash from token
+                if recalculated_hash != current_claims_hash:
+                    logger.debug(f"Claims hash still differs after recalculation for user {user_ansible_id}: local={recalculated_hash}, current={current_claims_hash}")
+                    return True
```
AlanCoding, line 196: Before you return True, for this to be valid, you would also need to place the value in the local cache. Or else you'll run this logic every time after the event.

fosterseth: I think you mean the `return False` case, which is valid. I updated the code

ansible_base/jwt_consumer/hub/auth.py
fosterseth: good idea, we'll do this in a separate PR

john-westcott-iv:
This has had no activity since Aug 11th and the linked JIRA is closed. Closing this for now, please reopen if needed.

PR #796, opened 2025-08-12
[AAP-47811] Update jwt_consumer to load user claims from gateway, if needed

TheRealHaoLiu:
#  [AAP-47811](https://issues.redhat.com/browse/AAP-47811) Update jwt_consumer to load user claims from gateway, if needed
## Description

- **What is being changed?** Refactored JWT RBAC permission processing to use claims hash validation and caching
- **Why is this change needed?** To improve performance by avoiding unnecessary gateway calls and ensure RBAC permissions are properly validated
- **How does this change address the issue?** Implements a claims hash mechanism that caches permission data and only fetches from the gateway when the hash changes, reducing redundant API calls while maintaining data consistency

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites
- Use aap-dev build from sources for all components
- Deploy with the corresponding gateway changes from https://github.com/ansible-automation-platform/aap-gateway/pull/930
- Full AAP environment with JWT authentication enabled

### Steps to Test
1. Run the existing JWT authentication test suite: `tox -e py311`
2. Verify all RBAC permission tests pass: `pytest test_app/tests/jwt_consumer/ -v`
3. Deploy aap-dev with all components built from source (including gateway PR changes)
4. Login to the platform as various users with different permission levels (admin, org admin, regular users)
5. Perform actions across different services (Controller, EDA, Hub) as these users
6. Validate that RBAC permissions are correctly enforced and nothing breaks
7. Verify that users can successfully authenticate and perform authorized actions

### Expected Results
- All existing JWT authentication functionality continues to work
- RBAC permissions are correctly applied based on claims hash validation
- Users can successfully authenticate and perform authorized actions
- Performance improvement due to reduced gateway calls when claims haven't changed
- Test suite passes with improved coverage of edge cases

### Changes Made:
- **ansible_base/jwt_consumer/common/auth.py**: Added claims hash validation logic, cache management, and gateway claims fetching
- **ansible_base/jwt_consumer/common/cache.py**: Added claims hash caching utilities
- **ansible_base/jwt_consumer/hub/auth.py**: Refactored to use `_apply_rbac_permissions` with explicit parameters
- **Test files**: Updated to match new method signatures and added edge case coverage

### Required Actions

- [ ] Blocked by PR/MR: https://github.com/ansible-automation-platform/aap-gateway/pull/930
  

### Performance Impact:
This change reduces unnecessary gateway API calls by caching claims hashes per user and only fetching fresh claims when the hash indicates changes have occurred.

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -1,4 +1,5 @@
 import logging
+import uuid
 from datetime import datetime
 from typing import Optional, Tuple
 
@@ -18,7 +19,9 @@ from ansible_base.jwt_consumer.common.exceptions import HTTP_498_INVALID_TOKEN, 
 from ansible_base.lib.logging.runtime import log_excess_runtime
 from ansible_base.lib.utils.auth import get_user_by_ansible_id
 from ansible_base.lib.utils.translations import translatableConditionally as _
+from ansible_base.rbac.claims import get_claims_hash, get_user_claims, get_user_claims_hashable_form
 from ansible_base.resource_registry.models import Resource, ResourceType
+from ansible_base.resource_registry.rest_client import get_resource_server_client
 from ansible_base.resource_registry.signals.handlers import no_reverse_sync
 
 logger = logging.getLogger("ansible_base.jwt_consumer.common.auth")
@@ -226,7 +229,7 @@ class JWTCommonAuth:
         return validated_body
 
     def decode_jwt_token(self, unencrypted_token, decryption_key, additional_options={}):
-        local_required_field = ["sub", "user_data", "exp", "objects", "object_roles", "global_roles", "version"]
+        local_required_field = ["sub", "user_data", "exp", "claims_hash", "version"]
         options = {"require": local_required_field}
         options.update(additional_options)
         return jwt.decode(
@@ -258,17 +261,98 @@ class JWTCommonAuth:
 
     def process_rbac_permissions(self):
         """
-        This is a default process_permissions which should be usable if you are using RBAC from DAB
+        Process RBAC permissions using claims hash logic
         """
         if self.token is None or self.user is None:
-            logger.error("Unable to process rbac permissions because user or token is not defined, please call authenticate first")
+            logger.error("Unable to process rbac permissions because user or token is not defined")
             return
 
+        jwt_claims_hash = self.token.get("claims_hash")
+        if not jwt_claims_hash:
+            logger.error("No claims_hash found in JWT token")
+            return
+
+        user_ansible_id = self.token.get("sub")
+        if not user_ansible_id:
+            logger.error("No subject (sub) found in JWT token")
+            return
+
+        # Validate UUID format (consistent with rest of codebase)
+        try:
+            uuid.UUID(user_ansible_id)
+        except (ValueError, TypeError):
+            logger.error(f"Invalid UUID format for user_ansible_id: {user_ansible_id}")
+            return
+
+        # Check cached claims hash
+        cached_claims_hash = self.cache.get_cached_claims_hash(user_ansible_id)
+
+        if cached_claims_hash == jwt_claims_hash:
+            logger.debug(f"Claims hash matches cached value for user {user_ansible_id}")
+            return
+
+        # Calculate local claims hash
+        local_claims = get_user_claims(self.user)
+        local_hashable_claims = get_user_claims_hashable_form(local_claims)
+        local_claims_hash = get_claims_hash(local_hashable_claims)
+
+        if local_claims_hash == jwt_claims_hash:
+            logger.debug(f"Claims hash matches local calculation for user {user_ansible_id}")
+            # Update cache with the correct hash
+            self.cache.cache_claims_hash(user_ansible_id, jwt_claims_hash)
+            return
+
+        # Claims hash mismatch - fetch from gateway
+        logger.info(f"Claims hash mismatch for user {user_ansible_id}. JWT: {jwt_claims_hash}, Local: {local_claims_hash}. Fetching from gateway.")
+        gateway_claims = self._fetch_jwt_claims_from_gateway(user_ansible_id)
+
+        if gateway_claims:
+            # Extract claims structure from gateway response
+            objects = gateway_claims.get('objects', {})
+            object_roles = gateway_claims.get('object_roles', {})
+            global_roles = gateway_claims.get('global_roles', [])
+
+            # Process the RBAC permissions with the gateway claims
+            self._apply_rbac_permissions(objects, object_roles, global_roles)
+
+            # Update cache with the new hash
+            self.cache.cache_claims_hash(user_ansible_id, jwt_claims_hash)
+        else:
+            self.log_and_raise(
+                _("Unable to validate user permissions - gateway claims fetch failed for user %(user_ansible_id)s"), {"user_ansible_id": user_ansible_id}
+            )
+
+    def _fetch_jwt_claims_from_gateway(self, user_ansible_id: str) -> Optional[dict]:
+        """
+        Fetch JWT claims from the gateway endpoint using resource server client
+        """
+        try:
+            # Use the resource server client to make the request
+            client = get_resource_server_client(service_path="api/gateway/v1")
+
+            logger.debug(f"Fetching claims from gateway for user {user_ansible_id}")
+            response = client._make_request("GET", f"jwt_claims/{user_ansible_id}/")
+
+            if response.status_code == 200:
+                claims_data = response.json()
+                return claims_data
+            else:
+                logger.error(f"Gateway request failed with status {response.status_code}")
+                return None
+
+        except Exception as e:
+            logger.error(f"Error fetching claims from gateway: {e}")
+            return None
+
+    def _apply_rbac_permissions(self, objects, object_roles, global_roles):
+        """
+        Apply RBAC permissions from claims data
+        """
         from ansible_base.rbac.models import RoleUserAssignment
 
         role_diff = RoleUserAssignment.objects.filter(user=self.user, role_definition__name__in=settings.ANSIBLE_BASE_JWT_MANAGED_ROLES)
 
-        for system_role_name in self.token.get("global_roles", []):
+        for system_role_name in global_roles:
             logger.debug(f"Processing system role {system_role_name} for {self.user.username}")
             rd = self.get_role_definition(system_role_name)
             if rd:
@@ -282,7 +366,7 @@ class JWTCommonAuth:
                 logger.error(f"Unable to grant {self.user.username} system level role {system_role_name} because it does not exist")
                 continue
 
-        for object_role_name in self.token.get('object_roles', {}).keys():
+        for object_role_name in object_roles.keys():
             rd = self.get_role_definition(object_role_name)
             if rd is None:
                 logger.error(f"Unable to grant {self.user.username} object role {object_role_name} because it does not exist")
@@ -291,11 +375,11 @@ class JWTCommonAuth:
                 logger.error(f"Unable to grant {self.user.username} object role {object_role_name} because it is not a JWT managed role")
                 continue
 
-            object_type = self.token['object_roles'][object_role_name]['content_type']
-            object_indexes = self.token['object_roles'][object_role_name]['objects']
+            object_type = object_roles[object_role_name]['content_type']
+            object_indexes = object_roles[object_role_name]['objects']
 
             for index in object_indexes:
-                object_data = self.token['objects'][object_type][index]
+                object_data = objects[object_type][index]
                 try:
                     resource, obj = self.get_or_create_resource(object_type, object_data)
                 except IntegrityError as e:
--- a/ansible_base/jwt_consumer/common/cache.py
+++ b/ansible_base/jwt_consumer/common/cache.py
@@ -38,6 +38,19 @@ class JWTCache:
         cache.set(validated_body["sub"], expected_cache_value, timeout=self.get_cache_timeout())
         return False, expected_cache_value
 
+    def cache_claims_hash(self, user_ansible_id: str, claims_hash: str) -> None:
+        """Cache the claims hash for a user"""
+        cache_key = f"claims_hash_{user_ansible_id}"
+        cache.set(cache_key, claims_hash, timeout=self.get_cache_timeout())
+        logger.debug(f"Cached claims hash for user {user_ansible_id}: {claims_hash}")
+
+    def get_cached_claims_hash(self, user_ansible_id: str) -> Optional[str]:
+        """Get cached claims hash for a user"""
+        cache_key = f"claims_hash_{user_ansible_id}"
+        cached_hash = cache.get(cache_key, None)
+        logger.debug(f"Retrieved cached claims hash for user {user_ansible_id}: {cached_hash}")
+        return cached_hash
+
     def get_key_from_cache(self) -> Optional[str]:
         # If we are not ignoring the cache (forcing a reload of the key), check it
         key = cache.get(cache_key, None)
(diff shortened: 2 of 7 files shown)


ansible_base/jwt_consumer/common/auth.py
```diff
@@ -258,17 +260,90 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
+            logger.error("No claims_hash found in JWT token")
+            return
+
+        user_ansible_id = self.token.get("sub")
+        if not user_ansible_id:
+            logger.error("No subject (sub) found in JWT token")
+            return
+
+        # Check cached claims hash
+        cached_claims_hash = self.cache.get_cached_claims_hash(user_ansible_id)
```
BrennanPaciorek, line 280: Is there a reason we don't just try to `get_user_claims`, which checks cache and computes it if miss, then compare?

Have not fully thought this one through, so I might just delete this comment later.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -258,17 +260,90 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
+        try:
+            # Use the resource server client to make the request
+            client = get_resource_server_client(service_path="api/gateway/v1")
+
+            logger.debug(f"Fetching claims from gateway for user {user_ansible_id}")
+            response = client._make_request("GET", f"jwt_claims/{user_ansible_id}")
+
+            if response.status_code == 200:
+                claims_data = response.json()
+                logger.info(f"Successfully fetched claims from gateway for user {user_ansible_id}")
```
chrismeyersfsu:
```suggestion
```

Too much logging.

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -258,17 +260,90 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
+        if not user_ansible_id:
+            logger.error("No subject (sub) found in JWT token")
+            return
+
+        # Check cached claims hash
+        cached_claims_hash = self.cache.get_cached_claims_hash(user_ansible_id)
+
+        if cached_claims_hash == jwt_claims_hash:
+            logger.debug(f"Claims hash matches cached value for user {user_ansible_id}")
+            return
```
BrennanPaciorek, line 292: So if we hit this code block, we are assuming we already applied the RBAC we need to locally?

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -258,17 +260,90 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
+
+            if response.status_code == 200:
+                claims_data = response.json()
+                logger.info(f"Successfully fetched claims from gateway for user {user_ansible_id}")
+                return claims_data
+            else:
+                logger.error(f"Gateway request failed with status {response.status_code}: {response.text}")
+                return None
+
+        except Exception as e:
```
chrismeyersfsu, line 343: Pretty broad exception catching. Things I _think_ can fail:

`response = client._make_request("GET", f"jwt_claims/{user_ansible_id}")`

`claims_data = response.json()`

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -258,17 +260,90 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
+
+            logger.debug(f"Fetching claims from gateway for user {user_ansible_id}")
+            response = client._make_request("GET", f"jwt_claims/{user_ansible_id}")
+
+            if response.status_code == 200:
+                claims_data = response.json()
+                logger.info(f"Successfully fetched claims from gateway for user {user_ansible_id}")
+                return claims_data
+            else:
+                logger.error(f"Gateway request failed with status {response.status_code}: {response.text}")
```
chrismeyersfsu: We don't want to include the response.text here. This could be all the objects a user has access to. Kind of sensitive, kind of not.

ansible_base/jwt_consumer/common/auth.py
TheRealHaoLiu: yes

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -258,17 +260,90 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
+    def _fetch_jwt_claims_from_gateway(self, user_ansible_id: str) -> Optional[dict]:
+        """
+        Fetch JWT claims from the gateway endpoint using resource server client
+        """
+        try:
+            # Use the resource server client to make the request
+            client = get_resource_server_client(service_path="api/gateway/v1")
+
+            logger.debug(f"Fetching claims from gateway for user {user_ansible_id}")
+            response = client._make_request("GET", f"jwt_claims/{user_ansible_id}")
```
chrismeyersfsu: I was expecting a trailing `/`. You should change your jwt_claims url to require a `/` to follow the established pattern.

ansible_base/jwt_consumer/common/auth.py
chrismeyersfsu: You can dismiss this or address it, don't feel strongly.

chrismeyersfsu (approved):

BrennanPaciorek (commented):
Question about this flow:

Someone logs in -> JWT producer and caches -> permissions update, JWT producer will not quite know this happened, since it is not a part of DAB (so we cannot call some hook into that functionality without signals or a similar mechanism we use to call a dynamic settings function) -> user logs in and request is propagating using cached token. 

Are we okay without updating this and with the wait before roles propagate?

fosterseth (approved):

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -258,17 +260,89 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
+            logger.error("No claims_hash found in JWT token")
+            return
+
+        user_ansible_id = self.token.get("sub")
+        if not user_ansible_id:
+            logger.error("No subject (sub) found in JWT token")
+            return
+
+        # Check cached claims hash
+        cached_claims_hash = self.cache.get_cached_claims_hash(user_ansible_id)
```
john-westcott-iv, line 288: I was just did a quick look and I don't think we ever validate that sub is a valid ansible_id anywhere. We likely don't care here but we should probablly do that somewhere to prevent getting a common value like "None" (as a string), which would make us care here since everyones permissions would effectivly end up as whomever last logged in. Thoughts?

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -258,17 +260,89 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
+
+        if cached_claims_hash == jwt_claims_hash:
+            logger.debug(f"Claims hash matches cached value for user {user_ansible_id}")
```
john-westcott-iv, line 291:
```suggestion
            logger.debug(f"Claims hash matches cached value for user {user_ansible_id}, not updating user claims")
```

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -258,17 +260,89 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
+
+        # Calculate local claims hash
+        local_claims = get_user_claims(self.user)
+        local_hashable_claims = get_user_claims_hashable_form(local_claims)
+        local_claims_hash = get_claims_hash(local_hashable_claims)
+
+        if local_claims_hash == jwt_claims_hash:
+            logger.debug(f"Claims hash matches local calculation for user {user_ansible_id}")
+            # Update cache with the correct hash
+            self.cache.cache_claims_hash(user_ansible_id, jwt_claims_hash)
```
john-westcott-iv, line 302: Why do we have to cache_claims_hash here since we just did `cached_claims_hash == jwt_claims_hash` shouldn't the cache already be correct?

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -258,17 +260,89 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
+            client = get_resource_server_client(service_path="api/gateway/v1")
+
+            logger.debug(f"Fetching claims from gateway for user {user_ansible_id}")
+            response = client._make_request("GET", f"jwt_claims/{user_ansible_id}/")
+
+            if response.status_code == 200:
+                claims_data = response.json()
+                return claims_data
+            else:
+                logger.error(f"Gateway request failed with status {response.status_code}")
```
john-westcott-iv, line 340: Would it make sense to also try and dump out the response body here (incase there is an embedded error)?

ansible_base/jwt_consumer/common/auth.py
```diff
@@ -258,17 +260,89 @@ def get_role_definition(self, name: str) -> Optional[Model]:
...
+            object_roles = gateway_claims.get('object_roles', {})
+            global_roles = gateway_claims.get('global_roles', [])
+
+            # Process the RBAC permissions with the gateway claims
+            self._apply_rbac_permissions(objects, object_roles, global_roles)
+
+            # Update cache with the new hash
+            self.cache.cache_claims_hash(user_ansible_id, jwt_claims_hash)
+        else:
+            logger.error(f"Failed to fetch claims from gateway for user {user_ansible_id}")
```
john-westcott-iv: Hmmm... so we know the users claims are wrong but if we can't call the gateway for some reason we just leave them incorrect and allow access? Should we raise here so that a request is not processed with the incorrect permissions?

ansible_base/jwt_consumer/common/cache.py
```diff
@@ -38,6 +38,19 @@ def check_user_in_cache(self, validated_body: dict) -> Tuple[bool, dict]:
         cache.set(validated_body["sub"], expected_cache_value, timeout=self.get_cache_timeout())
         return False, expected_cache_value
 
+    def cache_claims_hash(self, user_ansible_id: str, claims_hash: str) -> None:
+        """Cache the claims hash for a user"""
+        cache_key = f"claims_hash_{user_ansible_id}"
```
john-westcott-iv, line 43: Its kind of meh but maybe make the "claims_hash" a variable and us it in the put/get cache methods so if we ever needed to change it its only in one place.

ansible_base/jwt_consumer/common/cache.py
```diff
@@ -38,6 +38,19 @@ def check_user_in_cache(self, validated_body: dict) -> Tuple[bool, dict]:
...
+        cache.set(cache_key, claims_hash, timeout=self.get_cache_timeout())
+        logger.debug(f"Cached claims hash for user {user_ansible_id}: {claims_hash}")
+
+    def get_cached_claims_hash(self, user_ansible_id: str) -> Optional[str]:
+        """Get cached claims hash for a user"""
+        cache_key = f"claims_hash_{user_ansible_id}"
+        cached_hash = cache.get(cache_key, None)
+        logger.debug(f"Retrieved cached claims hash for user {user_ansible_id}: {cached_hash}")
```
john-westcott-iv, line 51:
```suggestion
        if cached_hash is not None:
            logger.debug(f"Retrieved cached claims hash for user {user_ansible_id}: {cached_hash}")
        else:
            logger.debug(f"There was not a cache of the user has for user {user_ansible_id}, permissions will have to be processed")
```

ansible_base/jwt_consumer/common/cache.py
```diff
@@ -38,6 +38,19 @@ def check_user_in_cache(self, validated_body: dict) -> Tuple[bool, dict]:
         cache.set(validated_body["sub"], expected_cache_value, timeout=self.get_cache_timeout())
         return False, expected_cache_value
 
+    def cache_claims_hash(self, user_ansible_id: str, claims_hash: str) -> None:
+        """Cache the claims hash for a user"""
+        cache_key = f"claims_hash_{user_ansible_id}"
+        cache.set(cache_key, claims_hash, timeout=self.get_cache_timeout())
```
john-westcott-iv, line 44: The get_cache_timeout is the value for the JWT itself which is a generous 7 days if ANSIBLE_BASE_JWT_CACHE_TIMEOUT_SECONDS is not set. But do we even care about timing this cache out? i.e. if my permissions are stable do we even want to reprocess them in 7 days. I kind of think this should be non-expiring. The JWT would get recreated but the claim hash could just live forever. Thoughts?

john-westcott-iv (commented):
Comments left, I'd like to get someone from the hub team to review their specific jwt_consumer.

ansible_base/jwt_consumer/common/auth.py
TheRealHaoLiu: this is `if local_claims_hash == jwt_claims_hash` which means that `cached_claims != jwt_claims_hash` and we recalculated the hash based on local claims info so no the cache is not correct at this point

ansible_base/jwt_consumer/common/auth.py
TheRealHaoLiu: Yep that's what the first iteration of my code did, but @chrismeyersfsu tell me to remove it because it may contain sensitive information

TheRealHaoLiu: @john-westcott-iv vs @chrismeyersfsu ... ready, set, GO!

ansible_base/jwt_consumer/common/auth.py
TheRealHaoLiu: good point

ansible_base/jwt_consumer/common/cache.py
TheRealHaoLiu: nah

TheRealHaoLiu: find and replace exist

ansible_base/jwt_consumer/common/cache.py
TheRealHaoLiu: we discussed this it will be ok

ansible_base/jwt_consumer/common/auth.py
TheRealHaoLiu: stop critiquing my debug messages!!!!

ansible_base/jwt_consumer/common/cache.py
TheRealHaoLiu: nope `permissions will have to be processed` messages like that happens in the code that handles the permission not the cache 

cache just care about the cache not how the data will be used once retrieved from the cache

ansible_base/jwt_consumer/common/auth.py
TheRealHaoLiu:
```
        user_ansible_id = self.token.get("sub")
        if not user_ansible_id:
            logger.error("No subject (sub) found in JWT token")
            return
```

This would prevent cache keys like claims_hash_None

we should probably add some empty string check... looking around for any other places that would validate the user_ansible_id

TheRealHaoLiu: adding

```python
        try:
            uuid.UUID(user_ansible_id)
        except (ValueError, TypeError):
            logger.error(f"Invalid UUID format for user_ansible_id: {user_ansible_id}")
            return
```

ansible_base/jwt_consumer/common/auth.py
TheRealHaoLiu: i think ur right... this is a new failure case that does not exist in previous code

fao89 (approved):
👻

BrennanPaciorek:
> Someone logs in -> JWT producer and caches -> permissions update, JWT producer will not quite know this happened, since it is not a part of DAB (so we cannot call some hook into that functionality without signals or a similar mechanism we use to call a dynamic settings function) -> user logs in and request is propagating using cached token.

Ran this is aap-dev. human usage does not lead to any delays in role_user_assignment propagation.

BrennanPaciorek (approved):
My primary concern is not an issue. Approving.

zkayyali812 (approved):

ansible_base/jwt_consumer/common/auth.py
john-westcott-iv: I would have said do a debug level log with the info but its all good.

PR #809, opened 2025-08-18
AAP-51654 Enforce that references to objects are strictly done in claims data, not token data

AlanCoding:
## Description
The prior solution partially migrate us from passing claims data inside of the JWT data to a new user "claims" data structure.

However, it was still referencing "objects" inside of the JWT data itself.

This refactors the methods out of the auth class, which makes this kind of bug likely. Stuff is just passed as method arguments in this new approach. This is what I originally expected out of this work.

## Type of Change
- [ ] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

## Testing Instructions
Should be covered well in the DAB tests here.

### Expected Results
claims processing should work

### Required Actions
- [ ] Blocked by PR/MR: #XXX


test_app/tests/rbac/test_claims.py
```diff
@@ -356,6 +311,73 @@ def test_identical_permissions_same_hash(self, claims_scenario, scenario_name):
...
+        # Save the claims to reapply later
+        user_claims = get_user_claims(user)
+        claims_hash = get_claims_hash(get_user_claims_hashable_form(user_claims))
+
+        # Rage delete everything
+        for team in Team.objects.all():
+            team.delete()
+
+        for org in Organization.objects.all():
+            org.delete()
```
AlanCoding, line 364: @TheRealHaoLiu this test is the actual regression testing here.

This is relying heavily on some test infrastructure that already exists. I don't know what's in the scenarios, but it looks like a lot. We run the scenario and then delete all the objects, and then assert that applying the claims data will re-create objects with the same names and links.

zkayyali812 (approved):

fosterseth (approved):

PR #818, opened 2025-08-21
[donotmerge] Changes to get some test performance data

AlanCoding:
## Description
This PR should print more data about why our tests are taking as long as they are.


AlanCoding:
Dependencies:

```
Fetched 5817 kB in 1min 37s (59.8 kB/s)
```

Oh, and that's all from the `apt-get install`.

And it seems like the debugging for the actual tests isn't really working.

PR #819, opened 2025-08-21
AAP-51502 Make authenticator.configuration reqd

bhavenst:
* The model lists the field as blank=True but it is required by the validator (blank is OK for some auth types)
 * Required for all operations except PATCH
 * Fixes openapi spec/ATF
 * Pins drf-spectacular in dev to the same version used by PDE (aligns spec file output)

## Description
- What is being changed?  "configuration" field made required in serializer
- Why is this change needed? to fix openapi spec for ATF purposes
- How does this change address the issue? DRF spectacular respects the serializer configuration, showing configuration as required

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Start gateway, load docs from /api/gateway/v1/docs/
2. See that configuration field is shown as required for all operations except PATCH
3. 

### Expected Results
Openapi spec is correct.

### DIFF
--- a/ansible_base/authentication/serializers/authenticator.py
+++ b/ansible_base/authentication/serializers/authenticator.py
@@ -35,7 +35,10 @@ class AuthenticatorSerializer(NamedCommonModelSerializer, ImmutableFieldsMixin):
             "slug": {
                 "allow_blank": True,
                 "required": False,
-            }
+            },
+            "configuration": {
+                "required": True,
+            },
         }
 
     # TODO: Do we need/want to delve into dicts and search their keys?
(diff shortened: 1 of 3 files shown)


ansible_base/authentication/serializers/authenticator.py
```diff
@@ -35,7 +35,10 @@ class Meta:
             "slug": {
                 "allow_blank": True,
                 "required": False,
-            }
+            },
+            "configuration": {
```
Dostonbek1, line 39: Looking at line 110, it says `# Not having configuration is only valid for a PATCH`; seems like we are not requiring the field for PATCH method. Would this change break that logic?

bhavenst: No, DRF spectacular knows that fields are not required for PATCH even if they are for other actions.

Dostonbek1 (approved):

john-westcott-iv (approved):

PR #824, opened 2025-08-27
[AAP-52133] Improve RoleUserAssignmentsCache.cache_existing method

john-westcott-iv:
**Changes to claims.py:**
- Fix object_id type handling in cache_existing method to support UUID, int, and string types
- Add proper error handling and logging for invalid object_id types and conversion failures
- Replace simple int conversion with robust type checking and validation
- Add UUID import and improve type safety for object_id processing
- Replace direct role_assignments query with get_user_object_roles for better data handling

**New comprehensive test suite:**
- Create new TestRoleUserAssignmentsCache class with full test coverage
- Add parameterized tests for different object_id types (None, int, UUID, string)
- Test error handling for invalid object_id types and string conversion failures
- Verify cache structure, status assignment, and role definition caching behavior
- Test edge cases: empty lists, multiple assignments, existing cache preservation
- Refactor similar tests into parameterized tests for better maintainability
- All tests follow pytest best practices with descriptive IDs and proper fixtures

## Description

- What is being changed?
Instead of using our own filter for getting user.role_assignments we now use the RBAC `get_user_object_roles` method. This should filter out any UUID permissions which were causing 500s before.
Additionally we updated the `cache_existing` method so that it would now properly accept a UUID incase one ever made it through to avoid 500s in the future.

- Why is this change needed?
Login can fail if the user logging in has permissions to non-gateway items (which return a UUID instead of an int).

This change should: 1) make the UUID objects not cached to start with 2) alter the cache logic to better tolerate invalid values.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Configure gateway with a service
2. Login as a user from a source with authenticator maps
3. Grant the user permissions to objects in the service
4. Log out and login as the user, login should fail

### Required Actions

- [ ] Blocked by PR/MR: #XXX


ansible_base/authentication/utils/claims.py
```diff
@@ -905,9 +904,28 @@ def cache_existing(self, role_assignments):
             # Cache Role User Assignment
             self._init_cache_key(role_definition.name, content_type_id=role_assignment.content_type_id)
 
-            # object_id is TEXT db type
-            object_id = int(role_assignment.object_id) if role_assignment.object_id is not None else None
-            obj = role_assignment.content_object if object_id else None
+            has_valid_object_id = False
+            object_id = role_assignment.object_id  # Initialize with the original value
+
+            if role_assignment.object_id is None:
```
AlanCoding: this is a text field, so I'm not sure if this even works, since it might be `""`

ansible_base/authentication/utils/claims.py
```diff
@@ -905,9 +904,28 @@ def cache_existing(self, role_assignments):
...
+                # If we have None we will cache with a None object
+                pass
+            elif isinstance(role_assignment.object_id, UUID) or isinstance(role_assignment.object_id, int):
```
AlanCoding: Here again, skeptical that the criteria works.

Instead of this, what if you checked `role_assignment.content_type is None or role_assignment.content_type.service == "gateway"`. Instead of hard-coding gateway, there's a method you can call to get the local service ID.

If this criteria is not met, `continue`. I do not see any universe where, in the current design, claims grant objects to things in other services. Yes, this is where we would like to go, but if we do, we should better consolidate with the claims-processing logic that already exists in the RBAC app. So firstly, make the user claims presented there include all claims, not just JWT-designed-roles. This is a very light lift. With that in place, we should try to see how those methods can be used here. Instead of having a JWT claims system and a separate authentication claims system.

john-westcott-iv: My understanding of the intent of this cache was that we want to load all of the users existing permissions into a cache (which is just a dict) marking them as "EXISTING". We then go through the authenticator maps and mark items as ADD/REMOVE based on the the maps. After that we have to reconcile any non-existing items and then there is a toggle on the authenticator if it should remove a users permissions which are not explicitly granted from the authenticator. If set to true, we want to remove all of the "EXISTING" items. If gateway can now see roles in the services on objects I would think we would want to be able to remove them if the authenticator said to. Do you agree with that or should we still only try to "cache" the gateway roles (which would be orgs and teams).

Also, I think we can simplify this logic somewhat by just doing something like `str(role_assignment.object_id)` but I'm now wondering if all object_ids are going to come back as UUIDs we may need to alter our caching logic because right now we are "add/removing" orgs/eams based off their `.id` and we should likely change to `org.resource.ansible_id` so that the new permissions and existing permissions line up.

zkayyali812:
I tried testing this out using the instructions above with aap-dev,  and I got stuck at step 2 - `Login as a user from a source with authenticator maps`.

I am seeing error - 
```
│ api 2025-09-02 13:51:29,479 ERROR    ade6bfa8-69d0-41a1-a2d8-8a955282388f ansible_base.authentication.utils.claims Failed to reconcile user claims: [ErrorDetail(string='Role definition content type must be null to assign globally', code='invalid')]                                │
│ api Traceback (most recent call last):                                                                                                                                                                                                                                                  │
│ api   File "/app/django-ansible-base/ansible_base/authentication/utils/claims.py", line 629, in update_user_claims                                                                                                                                                                      │
│ api     reconcile_user_class.reconcile_user_claims(user, authenticator_user)                                                                                                                                                                                                            │
│ api   File "/app/django-ansible-base/ansible_base/authentication/utils/claims.py", line 718, in reconcile_user_claims                                                                                                                                                                   │
│ api     cls(claims, user, authenticator_user).manage_permissions()                                                                                                                                                                                                                      │
│ api   File "/app/django-ansible-base/ansible_base/authentication/utils/claims.py", line 750, in manage_permissions                                                                                                                                                                      │
│ api     self.apply_permissions()                                                                                                                                                                                                                                                        │
│ api   File "/app/django-ansible-base/ansible_base/authentication/utils/claims.py", line 799, in apply_permissions                                                                                                                                                                       │
│ api     self._apply_permission(object_with_status, role_name)                                                                                                                                                                                                                           │
│ api   File "/app/django-ansible-base/ansible_base/authentication/utils/claims.py", line 806, in _apply_permission                                                                                                                                                                       │
│ api     self._give_permission(self.permissions_cache.rd_by_name(role_name), obj)                                                                                                                                                                                                        │
│ api   File "/app/django-ansible-base/ansible_base/authentication/utils/claims.py", line 839, in _give_permission                                                                                                                                                                        │
│ api     role_definition.give_global_permission(self.user)                                                                                                                                                                                                                               │
│ api   File "/app/django-ansible-base/ansible_base/rbac/models/role.py", line 182, in give_global_permission                                                                                                                                                                             │
│ api     return self.give_or_remove_global_permission(actor, giving=True)                                                                                                                                                                                                                │
│ api            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                │
│ api   File "/app/django-ansible-base/ansible_base/rbac/models/role.py", line 189, in give_or_remove_global_permission                                                                                                                                                                   │
│ api     raise ValidationError('Role definition content type must be null to assign globally')                                                                                                                                                                                           │
│ api rest_framework.exceptions.ValidationError: [ErrorDetail(string='Role definition content type must be null to assign globally', code='invalid')]    
```

This is using the following mapping - 
<img width="894" height="506" alt="image" src="https://github.com/user-attachments/assets/c0127b87-1a99-4313-b8ac-d9fe12765cc0" />

CC: @john-westcott-iv

zkayyali812:
I think the above bug might be a separate issue?
If I remove the role mapping and just use a team mapping that works..

However it is noted that the login should fail, but in my test the login still succeeds after I've manually granted roles to the user which already had roles mapped.
Should it be expected that the login succeeds?

ansible_base/authentication/utils/claims.py
AlanCoding: > but I'm now wondering if all object_ids are going to come back as UUIDs

They won't. As the person causing this chaos, I can say for sure.

The meaning of `object_id` will not change. That is the primary key. There is a related `object_ansible_id` that may presented by the serializer, but that is a prefetched/related field. It does not actually exist on the assignment. The assignment may be for a local or a remote object, but in both cases `object_id` is the primary key of the object. In both cases, we _have_ `object_id`, if the assignment is for an object, at all.

On the other hand, the JWT claims data is shared between services. So that must use the ansible_id for shared objects.

There's nothing wrong about having a claims blob using `object_id` for local references, as long as that is not shared between different services. This can exist independently of the JWT claims stuff, justified by solving a different problem, and `object_id` references are more performant.

However, my prior assertion still stands

> I do not see any universe where, in the current design, claims grant objects to things in other services.

Maybe I should be more specific to say "SSO claims". Until we have the feature for SSO claims to give users arbitrary roles, we're storing a lot of data that will never be used. I'm worried about that.

john-westcott-iv: Hmmm... so were we just in a bad code spot when we hit an error trying to convert an object_id into an int (when we got a UUID)? 

If so, maybe we just revert this change? Or maybe simplify it to:
```
obj = None
if role_assignment.object_ansible_id:
    obj = role_assignment.content_object

self.cache[role_definition.name][role_assignment.content_type_id][role_assignment.object_ansible_id] = {'object': obj, 'status': self.STATUS_EXISTING}
```
And then I make the other areas which use the cache key off ansible_id instead of object_id? That feels like it would better prep us to support object based roles in the future but let me know if that is incorrect.

AlanCoding: Oh I think I get it. Here's where I'm unintentionally trolling you.

> The meaning of `object_id` will not change. That is the primary key.

I said "primary key". Primary keys can be UUIDs. That is _highly_ specific to EDA & galaxy, because devs early in those projects just liked primary key UUIDs. Now, I still think it should be text, so I'm not claiming to fully understand the error.

But back to my recommendation, if you filter out non-Gateway resources, then you should not ever see any UUID primary keys, and it will become a problem for another day.

> And then I make the other areas which use the cache key off ansible_id instead of object_id? That feels like it would better prep us to support object based roles in the future but let me know if that is incorrect.

I don't know right now. But `role_assignment.object_ansible_id` is not "free", and I think that's a relevant immediate consideration.

BrennanPaciorek:
Sonar appears to think the entire code file + another unedited file is uncovered code.

ansible_base/authentication/utils/claims.py
```diff
@@ -905,11 +906,17 @@ def cache_existing(self, role_assignments):
...
-            # object_id is TEXT db type
-            object_id = int(role_assignment.object_id) if role_assignment.object_id is not None else None
-            obj = role_assignment.content_object if object_id else None
-
-            self.cache[role_definition.name][role_assignment.content_type_id][object_id] = {'object': obj, 'status': self.STATUS_EXISTING}
+            # Only should cache local role assignments
+            if role_assignment.content_type is None or role_assignment.content_type.service in [get_local_resource_prefix(), "shared"]:
+                try:
+                    # object_id should be TEXT db type
+                    object_id = int(role_assignment.object_id) if role_assignment.object_id is not None else None
```
BrennanPaciorek: The object_id can be 0 right? I updated this to not make it cache as globally applied if the object it applies to has the id of 0.

BrennanPaciorek (commented):
Other changes I made: https://github.com/ansible/django-ansible-base/pull/824/files#diff-7f6b8f4df1be270aa850255184d42657f290642f5ac3751fb7adadbd03a7a421L738 changes seemed to directly impact existing test results, so I reverted them.

BrennanPaciorek:
The result here is that I cannot get the error, but I also cannot assign roles controller roles to users via authenticator maps, which I don't think we would want.

They show in the UI, the maps pass, but do not provide the role.

ansible_base/authentication/utils/claims.py
```diff
@@ -905,11 +906,17 @@ def cache_existing(self, role_assignments):
             # Cache Role User Assignment
             self._init_cache_key(role_definition.name, content_type_id=role_assignment.content_type_id)
 
-            # object_id is TEXT db type
-            object_id = int(role_assignment.object_id) if role_assignment.object_id is not None else None
-            obj = role_assignment.content_object if object_id else None
-
-            self.cache[role_definition.name][role_assignment.content_type_id][object_id] = {'object': obj, 'status': self.STATUS_EXISTING}
+            # Only should cache local role assignments
+            if role_assignment.content_type is None or role_assignment.content_type.service in [get_local_resource_prefix(), "shared"]:
```
AlanCoding: If the `role_assignment.content_type is None` case is hit, then it's expected that object_id will be blank, which will lead to erroneous `logger.exception` calls. It looks like the caller method `manage_permissions` handles those separately, so I would say change this

```suggestion
            if (role_assignment.content_type is not None) and role_assignment.content_type.service in [get_local_resource_prefix(), "shared"]:
```

to avoid log spam

ansible_base/authentication/utils/claims.py
john-westcott-iv: I believe all IDs in Django start at 1.

AlanCoding (approved):

AlanCoding (approved):

zkayyali812 (approved):

PR #830, opened 2025-09-09
[AAP-52121] Fix attribute handling with 'and' condition

tznamena:
## Description

- What is being changed?
  - When authentication map is configured with "and" condition and the user who is logging does not have the attribute used in the map, the map can not be applied
- Why is this change needed?
  - The authentication map can be setup to grant access on condition of users having certain attributes, however, without this fix, the mapping was being used even for users who had subset of the attributes (were missing some of them).
  - This change ensures that if any attributes used in the mapping are not present for the user the map will not be applied/used for the user

## Type of Change

- [ ] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites

Since the issue was reported with LDAP, the testing described here will be for LDAP but it could be tested with different authentication source as well.

Your LDAP configuration will need to have two users, one containing extra attribute. A simple change is to add attribute "employeeType" with value "manager" to one of the existing users:

user 1:
```yaml
  cn: ldap_admin
  sn: LdapAdmin
  mail: ldap-admin@example.org
```
user 2:
```yaml
  cn: ldap_org_admin
  sn: LdapOrgAdmin
  mail: ldap-org-admin@example.org
  employeeType: manager
```
Then setup an authentication map that adds a user to a team, with following condition using "AND" operation:
- cn ends with `_admin`
- employee type is `manager`

Example of such authentication map triggers:
```json
            "triggers": {
                "attributes": {
                    "cn": {
                        "ends_with": "_admin"
                    },
                    "employeeType": {
                        "equals": "manager"
                    },
                    "join_condition": "and"
                }
            }
```

### Steps to Test
1. Setup your users in LDAP as described above and add the mapping
2. Login with the user that does not have the attribute used in the map
3. Login with the user that hase the attribute used in the map

### Expected Results
User with the attribute "employeeType" set to "manager" will be added to a team but the other user will not.

### DIFF
--- a/ansible_base/authentication/utils/claims.py
+++ b/ansible_base/authentication/utils/claims.py
@@ -459,7 +459,15 @@ def process_user_attributes(trigger_condition: dict, attributes: dict, map_id: i
         # Check if user has the attribute
         user_value = attributes.get(attribute, None)
         if user_value is None:
-            _prefixed_debug(map_id, tracking_id, f"Attr [{attribute}] is not present in user attributes, skipping")
+            # if condition is not "and", the attribute value is not required, just move on
+            if join_condition != 'and':
+                _prefixed_debug(map_id, tracking_id, f"Attr [{attribute}] is not present in user attributes, skipping")
+            # else, condition is "and" which means the attribute value IS required, set access to False
+            else:
+                _prefixed_debug(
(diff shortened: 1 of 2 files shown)


bhavenst (approved):

jay-steurer (approved):

jay-steurer:
@bhavenst  I believe you are the one that can merge this PR

PR #834, opened 2025-09-10
[AAP-51985] Implements cross-service RBAC cleanup for object deletion 

arrestle:
## Description

**What is being changed?**
Implements cross-service RBAC cleanup for object deletion by enhancing DAB deletion signals and adding Gateway service-index endpoints.

Resolves AAP-51985 where deleting Controller resources (inventories, projects, job templates) leaves orphaned role assignments in Gateway database, causing 404 errors in UI.

- Enhanced rbac_post_delete_remove_object_roles signal to detect object-level assignments and trigger cross-service cleanup
- Added object_delete endpoints to service API for bulk assignment cleanup
- Integrated with existing reverse-sync client infrastructure for Controller → Gateway communication

## Type of Change

## Self-Review Checklist

### Prerequisites
- AAP deployment with Controller and Gateway services
- Test data: inventory with object-level role assignments

### Steps to Test
1. Create inventory and assign user-specific role (e.g., Inventory Admin to user for specific inventory)
2. Delete the inventory via UI or API
3. Verify Gateway database cleanup via manual endpoint test:
   ```bash
   curl -X POST -H "Authorization: Bearer <user-token>" \
     -d '{"resource_type": "main.inventory", "resource_pk": "<inventory-id>"}' \
     {url}/api/gateway/v1/service-index/role-user-assignments/object_delete/
   ```

### Expected Results
- Controller database cleaned automatically during deletion
- Gateway database cleaned via cross-service sync call
- UI no longer shows orphaned role assignments

## Additional Context

Implements Corner Case 1 from "0082-Addendum: RBAC Synchronization Hardening" proposal.

### Required Actions
- [x] Blocked by Jira: AAP-53406


ansible_base/rbac/service_api/views.py
```diff
@@ -127,6 +173,10 @@ def assign(self, request):
     def unassign(self, request):
         return self._unassign(request)
 
+    @action(detail=False, methods=['post'], url_path='object_delete')
+    def object_delete(self, request):
```
AlanCoding: oooh ok ok, I'm worried that by having a custom action name (totally serious here) that the authentication class for service tokens will cause you to get a 401. Because it hard-codes a list of allowed actions, see "resource_api_actions", and the concern is that `object_delete` isn't in that list.

It wasn't obvious to me it would go this way. Depending on how the URL was included I figured it could be a new viewset that has the `create` method and no other CRUD methods, which should get past the bouncer.

ansible_base/rbac/service_api/views.py
```diff
@@ -100,6 +100,52 @@ def _unassign(self, request):
...
+            # Get content type for the resource
+            app_label, model = data['resource_type'].split('.')
+            content_type = DABContentType.objects.get(app_label=app_label, model=model)
+            
+            # Delete ALL assignments for this object (bulk operation)
+            with transaction.atomic():
+                deleted_count = self.get_queryset().filter(
+                    content_type=content_type,
+                    object_id=data['resource_pk']
+                ).delete()[0]
```
AlanCoding: I think the other signals didn't do this, taking faith in the cascade deletion working

https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/rbac/triggers.py#L260

But whether you do or don't, it may be equally efficient.

AlanCoding:
Looking good, just need to verify requests work and write tests

AlanCoding:
Formally confirming via aap-dev, and using the UI:
 - Created user
 - created inventory
 - In "User Access" gave the user admin to the inventory
 - deleted the inventory

through these steps, watched

http://localhost:44926/api/gateway/v1/role_user_assignments/

End result

```json
{
    "count": 0,
    "next": null,
    "previous": null,
    "results": []
}
```

whereas before deletion, the assignment from giving the user access persisted. It was 1 entry before the inventory deletion.

ansible_base/rbac/triggers.py
```diff
@@ -257,13 +257,29 @@ def rbac_post_delete_remove_object_roles(instance, *args, **kwargs):
         ObjectRole.objects.filter(users__isnull=True, teams__isnull=True).delete()
 
     ct = permission_registry.content_type_model.objects.get_for_model(instance)
+    
+    # Check if there were object-level assignments before deletion
+    had_object_assignments = ObjectRole.objects.filter(content_type=ct, object_id=instance.pk).exists()
+    
     ObjectRole.objects.filter(content_type=ct, object_id=instance.pk).delete()
```
AlanCoding: The variable `had_object_assignments` will fail in cases related to team deletion. See above.

This is also an inefficient solution to call `.exists()` and incur another query. We want to be as miserly as possible for performance, because this is deeply integrated into app signals, where those are not "our" apps.

Instead, I would like to use the return value of the bulk deletion, as you are already doing to render the response data from a request. Do that for both `ObjectRole` bulk deletion cases, and if either deletes 1 or more object roles, then you need to do the reverse-sync.

ansible_base/rbac/triggers.py
```diff
@@ -257,13 +257,29 @@ def rbac_post_delete_remove_object_roles(instance, *args, **kwargs):
...
+    # NEW: Conditionally trigger cross-service cleanup for orphaned assignments in Gateway
+    # Only sync when object-level assignments existed - this is the key performance optimization
+    if had_object_assignments:
+        try:
+            from ansible_base.rbac.sync import maybe_reverse_sync_object_deletion
+            maybe_reverse_sync_object_deletion(instance)
+        except Exception:
+            # Continue with local deletion even if cross-service sync fails
+            # This ensures we don't break local operations due to network/auth issues
+            logger.exception(f"Failed to sync object deletion for {instance}")
```
AlanCoding, line 283: Non-blocking comment - this means that we will not block deletion of an object when the internal requests have problems. We don't have a good recovery option for this case. But I generally agree we should front-load errors for _creation_ cases but probably not for _deletion_ cases.

AlanCoding (changes requested):
I have one issue related to team deletions, in comment. This is because of the case where there are no members to a team, but the team has permissions to things. Uncommon, but we need to take care of it.

ansible_base/rbac/triggers.py
```diff
@@ -257,13 +257,28 @@ def rbac_post_delete_remove_object_roles(instance, *args, **kwargs):
         ObjectRole.objects.filter(users__isnull=True, teams__isnull=True).delete()
```
AlanCoding, line 257:
```suggestion
        ObjectRole.objects.filter(users__isnull=True, teams__isnull=True).delete()
        deleted_count, _ = ObjectRole.objects.filter(users__isnull=True, teams__isnull=True).delete()
        if deleted_count:
            had_object_assignments = True
```

AlanCoding (approved):

TheRealHaoLiu:
manually tested the scenario that this is PR is suppose to fix and confirm that the PR did indeed fix the problem

ansible_base/rbac/triggers.py
```diff
@@ -255,15 +255,33 @@ def rbac_post_delete_remove_object_roles(instance, *args, **kwargs):
...
+        deleted_count, _ = ObjectRole.objects.filter(users__isnull=True, teams__isnull=True).delete()
+        if deleted_count:
+            had_object_assignments = True
 
     ct = permission_registry.content_type_model.objects.get_for_model(instance)
-    ObjectRole.objects.filter(content_type=ct, object_id=instance.pk).delete()
+
+    # Use bulk delete return value to determine if object-level assignments existed
+    # This avoids the inefficient .exists() query and works correctly for team deletion cases
+    deleted_count, _ = ObjectRole.objects.filter(content_type=ct, object_id=instance.pk).delete()
```
AlanCoding, line 266: This `deleted_count` would shadow the name of the variable above. We would want something like `had_object_assignments = deleted_count or deleted_count_team`, otherwise there's kind of some dead code here

chrismeyersfsu (approved):
LGTM. The reason for this PR is sound.

PR #850, opened 2025-09-23
[AAP-52229] Add AuditableModel inheritance to RBAC assignment models

arrestle:
## Description

- [AAP-52229] Platform RBAC actions not being logged in AAP-UI Activity Stream
- **What is being changed?** Enhanced `RoleUserAssignment` and `RoleTeamAssignment` models in django-ansible-base to inherit from `AuditableModel`
- **Why is this change needed?** Platform RBAC actions (role assignments) are not being logged in AAP-UI Activity Stream, breaking audit trails for compliance and security oversight
- **How does this change address the issue?** Adding `AuditableModel` inheritance enables automatic activity stream logging for all role assignments, providing complete audit trails showing "who assigned what role to whom when"

## Type of Change

- [ ] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

## Testing Instructions
- Create role assignment in test_app and check activity stream entries
-  Run: `tox -e py -- test_app/tests/rbac/test_rbac_activity_stream.py`

### Expected Results
- Both CREATE and DELETE user and team role assignment operations are properly logged

## Additional Context

[aap-gateway test pr#1011](https://github.com/ansible-automation-platform/aap-gateway/pull/1011l)


AlanCoding:
I don't know what's up with checks, `docker image pull mirror.gcr.io/library/postgres:15` works for me locally.

ansible_base/activitystream/models/entry.py
```diff
@@ -63,6 +63,13 @@ class Meta:
     )
 
     def __str__(self):
+        # Enhanced display for RBAC role assignments
+        if self.content_type and self.content_type.model.lower() in ['roleuserassignment', 'roleteamassignment']:
+            operation_text = self.get_operation_display()
+            created_by_text = str(self.created_by) if self.created_by else "Unknown"
+            return f'[{self.created}] Role assignment {operation_text.lower()} by {created_by_text}'
```
AlanCoding, line 70: I agree a custom string would make sense because "roleuserassignment" is verbose. But I would question whether we could solve this generally by using `_meta.verbose_name.title()` instead? Because "Role assignment" drops the user/team designation, which could be useful. This also appears to drop the object_id. I don't think content_type (as a string) was ever useful but it dropped that too.

AlanCoding: You could drop the object_id, but you would want to replace it by like `self.content_object.object_id`, which yeah, is confusing. That's the target object _of the assignment_.

arrestle: I thought about this and decided that if there were any additional assignments implemented in the future they might not necessarily want to be tracked, that decision should be up the PM and the Developer. This code is here to prevent unwanted side-effects. 

*note* There are lots of ways to handle this, such as an attribute on the class

ansible_base/rbac/models/dummy_models.py
AlanCoding, line 1: My suggestion here is that we try to make `AuditableModel` just a normal python class, and not a Django abstract model at all. That way, it can be "seen" and imported by thing not even using Django. You could even move to `ansible_base.lib` and have the activitystream app import it.

arrestle: I agree, but since I was asked to move the ticket back into the backlog by John let's take this up on the Thursday [Ansible Staff Engineering Weekly Late](https://drive.google.com/open?id=1cc4GsIf124Tsm8nOnoSI9X0Xq9KwjPMXATwbS__4-eQ&authuser=0) -- unless you have a better meeting?

PR #863, opened 2025-10-07
[AAP-55305] Short circuiting claims processing if user is a super user

john-westcott-iv:
## Description

- What is being changed?
When running migrate service data gateway is put into a state that prevents querying the gateway for user/object info. During JWT authentication if the claims hash in the token doesn't match the claims hash in the service a call from the service to the gateway to get the claims (which then fails). With this patch, if a user has is_super_user we will just completely skip processing the claims data which will prevent this scenario and optimize authenticator for superusers in general. 

- Why is this change needed?
See above.

Short circuits the claims processing if user is superuser.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. 
2. 
3. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -23,6 +23,15 @@ from ansible_base.resource_registry.signals.handlers import no_reverse_sync
 logger = logging.getLogger("ansible_base.jwt_consumer.common.auth")
 
 
+class GatewayLockedException(Exception):
+    """
+    Exception raised when the gateway is locked (returns 423 status).
+    This typically happens when migrations are in progress.
+    """
+
+    pass
+
+
 # These fields are used to both map the user as well as to validate the JWT token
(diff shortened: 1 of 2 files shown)


ansible_base/jwt_consumer/common/auth.py
```diff
@@ -235,6 +235,10 @@ def process_rbac_permissions(self):
             logger.error("Unable to process rbac permissions because user or token is not defined")
             return
 
+        if self.token.get('user_data', {}).get("is_super_user", None) is True:
```
fosterseth:
```suggestion
        if self.token.get('user_data', {}).get("is_superuser", None) is True:
```

AlanCoding (approved):

fosterseth (approved):

PR #874, opened 2025-10-24
AAP 51882 - shared jwt authenticator class

pb82:
## Description

Currently, there are three component specific JWTConsumer classes in DAB:

AWX: https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/jwt_consumer/awx/auth.py#L9
Hub: https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/jwt_consumer/hub/auth.py#L8
EDA: https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/jwt_consumer/eda/auth.py#L10

They all do the same, the goal of this PR is to provide one common RbacEnabledJWTAuthentication class that all components use.

After this is merged, EDA, AWX and Hub need to be updated to use the common class. Then we can remove the component specific classes in another PR.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -7,7 +7,6 @@ import jwt
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.utils import IntegrityError
-from drf_spectacular.extensions import OpenApiAuthenticationExtension
 from rest_framework.authentication import BaseAuthentication
 from rest_framework.exceptions import AuthenticationFailed
 
@@ -365,9 +364,16 @@ class RbacAwareJWTAuthentication(JWTAuthentication):
         self.use_rbac_permissions = is_rbac_installed()
 
 
-class RbacAwareJWTAuthScheme(OpenApiAuthenticationExtension):
-    target_class = RbacAwareJWTAuthentication
-    name = "RbacAwareJWTAuthentication"
+try:
+    from drf_spectacular.extensions import OpenApiAuthenticationExtension
 
-    def get_security_definition(self, auto_schema):
-        return {"type": "apiKey", "name": "X-DAB-JW-TOKEN", "in": "header"}
+    class RbacAwareJWTAuthScheme(OpenApiAuthenticationExtension):
+        target_class = RbacAwareJWTAuthentication
+        name = "RbacAwareJWTAuthentication"
+
+        def get_security_definition(self, auto_schema):
+            return {"type": "apiKey", "name": "X-DAB-JW-TOKEN", "in": "header"}
+
+except ImportError:
+    # drf_spectacular is not available, skip the schema definition
+    pass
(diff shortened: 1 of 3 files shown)


ansible_base/jwt_consumer/common/auth.py
```diff
@@ -353,3 +354,11 @@ def process_permissions(self):
...
         else:
             logger.info("process_permissions was not overridden for JWTAuthentication")
+
+
+class RbacAwareJWTAuthentication(JWTAuthentication):
+    use_rbac_permissions = False
+
+    def __init__(self):
+        super().__init__()
+        self.use_rbac_permissions = is_rbac_installed()
```
AlanCoding, line 364: ok, this is novel. Didn't expect it, but it's probably the smartest way to do it 👍

ansible_base/jwt_consumer/eda/auth.py
```diff
@@ -12,8 +12,8 @@ class EDAJWTAuthentication(JWTAuthentication):
 
 
 class EDAJWTAuthScheme(OpenApiAuthenticationExtension):
-    target_class = EDAJWTAuthentication
-    name = "EDAJWTAuthentication"
+    target_class = RbacAwareJWTAuthentication
+    name = "RbacAwareJWTAuthentication"
```
AlanCoding: I'm not sure if this is going to work as expected. This is used for the OpenAPI generation (cc @PabloHiro) and yes it's poorly documented what it's doing.

Making the "scheme" class (only used by drf-spectacular machinery presumably) point to a different class is likely to be unexpected, and result unclear. Most likely will cause the spec to miss these details and issue a warning, but could mess up.

But backing up, what should we be doing in the first place? This PR introduces `RbacAwareJWTAuthentication` class. It would be reasonable to along-side of it, introduce a "scheme" class for the new class. But you would put that in `ansible_base/jwt_consumer/common/auth.py` directly next to the new class. _this_ `EDAJWTAuthScheme` might properly register for `RbacAwareJWTAuthentication`, but even if it did, that would be confusing.

Re-summarizing - I would strive to keep this the same, introduce new scheme class for `RbacAwareJWTAuthentication`.

pb82: thanks, I've added a new schema class now for RbacAwareJWTAuthentication

AlanCoding (approved):
LGTM

AlanCoding:
On the AWX failure, we may have a real coordination problem:

```
/var/lib/awx/venv/awx/lib64/python3.11/site-packages/rest_framework/settings.py:182: in import_from_string
    raise ImportError(msg)
E   ImportError: Could not import 'ansible_base.jwt_consumer.awx.auth.AwxJWTAuthentication' for API setting 'DEFAULT_AUTHENTICATION_CLASSES'. ModuleNotFoundError: No module named 'drf_spectacular'.
```

https://github.com/ansible/awx/pull/16154 will add that library. But for now, it might be easiest to conditionally include the scheme class.

AlanCoding (approved):

PR #875, opened 2025-10-24
AAP-45875 Runtime Feature Flags

fao89:
## Description
https://issues.redhat.com/browse/AAP-45875

- What is being changed?
  This change is the foundation for enabling runtime platform feature flags for AAP. This updates the django-ansible-base to be the central location where all platform flags are defined. Components can inherit the `ansible_base.feature_flags` application to inherit all platform feature flag definitions.
- Why is this change needed?
  To enable runtime platform feature flags in AAP.
  This change addresses the issue by defining a database flag source, which contains all the feature flags along with their associated metadata. These feature flags are installed into each components database at install-time and kept in sync via resource sync (Gateway is the provider)

Before this can be merged, the following should be done:
1. Confirm each feature flags metadata
   1. Only Dispatcher and EDA_Analytics feature flags need their metadata confirmed now.

## Type of Change

- [x] Documentation update
- [x] Test update

## Self-Review Checklist

### Steps to Test

Recommend testing this through AAP Dev, but it can be tested directly via the test app as well.

1. Clone https://github.com/ansible/aap-dev/
2. Enter aap-dev directory and run `make configure-sources`
3. Use this PR for the DAB source
4. Use this PR for the Gateway source - https://github.com/ansible-automation-platform/aap-gateway/pull/1056

### Expected Results

AAP Deploys as expected with database feature flags.

### Required Actions

  - Gateway - https://github.com/ansible-automation-platform/aap-gateway/pull/1056
  - EDA - https://github.com/ansible/eda-server/pull/1407
  - Galaxy - https://github.com/ansible-automation-platform/galaxy_ng/pull/127
  - AWX - https://github.com/ansible/awx/pull/16148


ansible_base/lib/dynamic_config/dynamic_urls.py
```diff
@@ -13,8 +13,6 @@
 installed_apps = getattr(settings, 'INSTALLED_APPS', [])
 for app in installed_apps:
     if app.startswith('ansible_base.'):
-        if app in getattr(settings, 'ANSIBLE_BASE_APPS_EXCLUDE_VIEW_LIST', []):
-            continue
```
AlanCoding, line 17: :open_mouth: 

This is being used by aap-gateway, and I suspect it may have been removed by mistake in the rebase.

fao89: Thank you!
I'll add it back

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -360,6 +360,16 @@ def get_orphan_resources(
...
 
+    # Exclude system users from orphan cleanup for shared.user resource type
+    if resource_type_name == "shared.user":
+        from ansible_base.lib.utils.models import get_system_user
+
+        system_user = get_system_user()
+        if system_user:
+            queryset = queryset.exclude(object_id=system_user.id)
+
+    return queryset
```
AlanCoding, line 387: This seems extremely out-of-place for this patch. On the surface it seems reasonable, but I want to ask that we treat it separately so we can have a paper trail (new Jira, etc.). In what case is it known that we had a user sync problem with the _system user?

I want that for basic accountability surrounding the change. I believe there is risk in the change, and I don't want that risk to be attributed to the feature flags feature, which it seems to have nothing to do with it.

ansible_base/resource_registry/shared_types.py
AlanCoding, line 1: Going to ask for @chrismeyersfsu review, as he has some patches to code here, philosophically non-conflicting. The sync changes are something I want to advertise more widely.

AlanCoding:
> Only Dispatcher and EDA_Analytics feature flags need their metadata confirmed now.

I'm not sure if we're going to keep the dispatcher feature flag, as it was for a transition period. So it's likely better to never port it to the updated flags system.

ansible_base/feature_flags/views.py
```diff
@@ -1,25 +1,43 @@
...
-class FeatureFlagsStateListView(AnsibleBaseView):
+class FeatureFlagsStatesView(AnsibleBaseDjangoAppApiView, ModelViewSet):
+    """
+    A view class for displaying feature flags states.
+    To add/update/remove a feature flag, see the instructions in
+    `docs/apps/feature_flags.md`
+    """
+
+    queryset = AAPFlag.objects.order_by('id')
+    permission_classes = try_add_oauth2_scope_permission([IsSuperuserOrAuditor])
```
AlanCoding, line 23: @john-westcott-iv remind me, is the rule for using `try_add_oauth2_scope_permission` that we need it for any view that is surfaced by aap-gateway? That would explain why it is added here, I just want to have it written down.

ansible_base/feature_flags/views.py
```diff
@@ -1,25 +1,43 @@
...
+from ansible_base.lib.utils.views.permissions import IsSuperuserOrAuditor, try_add_oauth2_scope_permission
 
 from .utils import get_django_flags
 
 
-class FeatureFlagsStateListView(AnsibleBaseView):
+class FeatureFlagsStatesView(AnsibleBaseDjangoAppApiView, ModelViewSet):
+    """
+    A view class for displaying feature flags states.
+    To add/update/remove a feature flag, see the instructions in
```
AlanCoding, line 18: Here comes the big questions - we have a new endpoint being added, follow that.

Under what conditions will this be added? What specific actions are needed downstream of this change? Trying to fill in those blanks:
 - Each service (AWX, eda-server, etc.) already have this added to INSTALLED_APPS, but the new migration means that installers need to be sure that migration is run
 - the UI will need to make use of these new endpoints.

But even from these 2 above points, I have a point of confusion. _All_ services will expose these new endpoints. So then what endpoints should the UI make use of? Which endpoints will _work_?

If a flag is modified via the Gateway API, will that be actively synchronized to the services, or passively?

If the UI is only going to use the endpoints from the Gateway API, should we **disable** the new API in the other services or make them **read-only**?

ansible_base/feature_flags/serializers.py
```diff
@@ -1,16 +1,42 @@
...
+
+    def get_state(self, instance):
+        return flag_state(instance.name)
+
+    class Meta:
+        model = AAPFlag
+        fields = ["name", "state"]
+
+    def to_representation(self, instance=None) -> dict:
+        instance.state = True
```
AlanCoding: This line is surprising to me. What is going on here?

ansible_base/feature_flags/views.py
AlanCoding: > All services will expose these new endpoints. So then what endpoints should the UI make use of? Which endpoints will work?

Answered:

There is an additional viewset included in aap-gateway that adds `partial_update`, which is not present here. So this exposes everywhere as read-only. The read-only, GET, view is still in gateway (the resource server) but not in the other components, which are fully read-only.

> If a flag is modified via the Gateway API, will that be actively synchronized to the services, or passively?

Actively.

The `partial_update` method definition _does_ synchronize to components, actively.

> If the UI is only going to use the endpoints from the Gateway API, should we disable the new API in the other services or make them read-only?

Yes, already done. :+1:

ansible_base/feature_flags/models/aap_flag.py
```diff
@@ -0,0 +1,67 @@
...
+        default=False,
+        help_text=_("The visibility of the feature flag. If false, flag is hidden."),
+    )
+    toggle_type = models.CharField(
+        max_length=20,
+        null=False,
+        choices=[('install-time', 'install-time'), ('run-time', 'run-time')],
+        default='run-time',
+        help_text=_("Details whether a flag is toggle-able at run-time or install-time. (Default: 'run-time')."),
+    )
```
AlanCoding, line 81: I looked in the proposal and I don't see any mention of this. I would like to get better clarity on how this will be used. The help text isn't actually doing it for me.

The default is incomparable with most currently-declared flags in `feature_flags.yml`.

By this, I mean, let's get together with UI developers who may work on it, like @nixocio, @PabloHiro on API design.

I just really want to get clarity on the meaning of this field, potential invalid states, and things like that. If we talk about this in advance, I do believe that is likely to prevent unnecessary migration work or bugs later on.

ansible_base/feature_flags/models/aap_flag.py
```diff
@@ -0,0 +1,67 @@
...
+    description = models.CharField(max_length=500, null=False, default="", help_text=_("A detailed description giving an overview of the feature flag."))
+    support_url = models.CharField(max_length=250, null=False, default="", blank=True, help_text="A link to the documentation support URL for the feature")
+    labels = models.JSONField(null=True, default=list, help_text=_("A list of labels for the feature flag."), blank=True)
```
AlanCoding: Can you do a field-level validator here to enforce that it's a list of strings? I don't see any validation, and I don't want to deal with validating at any layer higher than this because it's new and this is the easy way.

ansible_base/feature_flags/models/aap_flag.py
```diff
@@ -0,0 +1,67 @@
...
+    support_level = models.CharField(
+        max_length=25,
+        null=False,
+        help_text=_("The support criteria for the feature flag. Must be one of (DEVELOPER_PREVIEW or TECHNOLOGY_PREVIEW)."),
+        choices=(('DEVELOPER_PREVIEW', 'Developer Preview'), ('TECHNOLOGY_PREVIEW', 'Technology Preview')),
+        blank=False,
+    )
+    visibility = models.BooleanField(
+        default=False,
+        help_text=_("The visibility of the feature flag. If false, flag is hidden."),
```
AlanCoding: hidden to who?

ansible_base/feature_flags/models/aap_flag.py
```diff
@@ -0,0 +1,67 @@
...
+    required = models.BooleanField(
+        default=False,
+        help_text=_("If multiple conditions are required to be met to enable a feature flag, 'required' can be used to specify the necessary conditionals."),
+    )
+    support_level = models.CharField(
+        max_length=25,
+        null=False,
+        help_text=_("The support criteria for the feature flag. Must be one of (DEVELOPER_PREVIEW or TECHNOLOGY_PREVIEW)."),
+        choices=(('DEVELOPER_PREVIEW', 'Developer Preview'), ('TECHNOLOGY_PREVIEW', 'Technology Preview')),
+        blank=False,
```
AlanCoding, line 68: `editable=False`?

ansible_base/feature_flags/models/aap_flag.py
```diff
@@ -0,0 +1,67 @@
...
+
+    name = models.CharField(
+        max_length=64,
+        null=False,
+        help_text=_("The name of the feature flag. Must follow the format of FEATURE_<flag-name>_ENABLED."),
+        validators=[validate_feature_flag_name],
+        blank=False,
+    )
+    ui_name = models.CharField(max_length=64, null=False, blank=False, help_text=_("The pretty name to display in the application User Interface"))
+    condition = models.CharField(max_length=64, default="boolean", help_text=_("Used to specify a condition, which if met, will enable the feature flag."))
```
AlanCoding, line 50: I want an example. Also, it seems likely we need to set `editable=False`.

Ok, maybe _almost all_ of these should set `editable=False`.

ansible_base/feature_flags/serializers.py
fao89: I also don't get it too, @zkayyali812 could you please take a look?

ansible_base/feature_flags/models/aap_flag.py
fao89: Ideally the feature flags would all be runtime, I believe this is just a way to make it possible to have feature flags that can only be toggled at install time

ansible_base/feature_flags/serializers.py
zkayyali812: I dont exactly recall why this is here, but I think this is a great call out. Im guessing this line is a remnant, and could likely be removed.

ansible_base/resource_registry/tasks/sync.py
fao89: I was having issues with test_app/tests/resource_registry/test_resource_sync.py::test_resource_sync it seems to be flaky

ansible_base/feature_flags/models/aap_flag.py
fao89: @zkayyali812 could you help with that?

ansible_base/feature_flags/views.py
BrennanPaciorek: This is correct. Without this, any views that don't have the permission class will get un-scoped permissions when using PATs (`read` tokens effectively become `read write` tokens).

ansible_base/feature_flags/models/aap_flag.py
zkayyali812: If True here, the flag would be visible in the UI (phase 3 of feature flags). False would mean the flag is not to be shown in the UI (unless it is enabled?).

ansible_base/feature_flags/models/aap_flag.py
zkayyali812: This has now been set

ansible_base/feature_flags/models/aap_flag.py
zkayyali812: We could likely set editable=False here too. I expect we will only ever have boolean conditions.

ansible_base/feature_flags/migrations/example_migration
```diff
@@ -0,0 +1,21 @@
+### INSTRUCTIONS ###
+# If updating the feature_flags.yaml, create a new migration file by copying this one.
+# 1. Name the file XXXX_manual_YYYYMMDD.py. For example 0002_manual_20250808.py
+# 1. Uncomment the migration below, by uncommenting everything below the FileHash
+# 2. Update the dependency to point to the last dependency
+# 3. Set the FileHash
+###
```
zkayyali812, line 7: Upon thinking about this further, we should probably update our feature flag handbook proposal with these steps too.

docs/apps/feature_flags.md
```diff
@@ -70,3 +47,37 @@ urlpatterns = [
     ...
 ]
 ```
+
+## Adding/updating/removing feature flags
```
zkayyali812, line 51: This should be added to the handbook too I expect. We can create a separate ticket for this if we'd like.

zkayyali812 (commented):
Approving, recommending addressing the latest comments too from others before merge.

zkayyali812 (approved):

AlanCoding (approved):

ansible_base/feature_flags/definitions/feature_flags.yaml
```diff
@@ -0,0 +1,64 @@
...
+  visibility: False
+  condition: boolean
+  value: 'False'
+  support_level: TECHNOLOGY_PREVIEW
+  description: A service to run python tasks in subprocesses, designed specifically to work well with pg_notify, but intended to be extensible to other message delivery means.
+  support_url: ''
+  toggle_type: install-time
+  labels:
+    - eda
+    - controller
```
AlanCoding, line 54: We have planned for this flag to be removed, so I'll have to file another issue for that.

AlanCoding: Aside - do you have a strategy for removing a flag?

AlanCoding: AAP-58540

I do want some confirmation that this won't lead to problems, will ping you there.

ansible_base/feature_flags/definitions/feature_flags.yaml
fao89, line 1: feature flag defaults

ansible_base/feature_flags/definitions/feature_flags.yaml
```diff
@@ -0,0 +1,64 @@
...
+  description: Submit Event-Driven Ansible usage analytics to console.redhat.com.
+  support_url: https://access.redhat.com/solutions/7112810
+  toggle_type: install-time
+  labels:
+    - eda
+- name: FEATURE_GATEWAY_IPV6_USAGE_ENABLED
+  ui_name: Gateway IPv6 Enablement
+  visibility: False
+  condition: boolean
+  value: 'False'
```
bhavenst: Please make the default True for FEATURE_GATEWAY_IPV6_USAGE_ENABLED (for 2.7 only.  I assume this is not being backported)

ansible_base/feature_flags/definitions/feature_flags.yaml
```diff
@@ -0,0 +1,64 @@
...
+  support_url: ''
+  toggle_type: install-time
+  labels:
+    - eda
+    - controller
+- name: FEATURE_CASE_INSENSITIVE_AUTH_MAPS_ENABLED
+  ui_name: Case Insensitive Authentication Maps
+  visibility: False
+  condition: boolean
+  value: 'False'
```
bhavenst: This is defaulted True for 2.6+

ansible_base/feature_flags/definitions/feature_flags.yaml
fao89: according to @Spredzy this will be backported to 2.6

ansible_base/feature_flags/models/aap_flag.py
```diff
@@ -0,0 +1,84 @@
...
+
+    resource = AnsibleResourceField(primary_key_field="id")
+
+    name = models.CharField(
+        max_length=64,
+        null=False,
+        help_text=_("The name of the feature flag. Must follow the format of FEATURE_<flag-name>_ENABLED."),
+        validators=[validate_feature_flag_name],
+        blank=False,
+    )
```
fao89, line 48: feature flag pattern

ansible_base/feature_flags/definitions/feature_flags.yaml
BrennanPaciorek: ~In which case, we'd want to tweak the backport to default this to False, and devel to default it to true, right?~ Probably cleaner to flip the flag default in separate PR, especially since the issue that flipped the flag for 2.7 is still in sprint.

ansible_base/feature_flags/definitions/feature_flags.yaml
fao89: done

ansible_base/feature_flags/definitions/feature_flags.yaml
fao89: I think we can merge and backport on this PR, and then update this on devel

BrennanPaciorek (approved):

ansible_base/feature_flags/definitions/feature_flags.yaml
bhavenst: The setting of True for this flag is already [backported](https://github.com/ansible-automation-platform/aap-gateway/pull/1130) to 2.6.  Please make it True here as well.  (I wasn't sure that was going to 2.6, but it is)

bhavenst (approved):

PR #879, opened 2025-11-04
AAP-56282: Include `x-ai-description` in generated spec

huffmanca:
## Description
There is a hard requirement being added, to support our MCP server creation, that every endpoint in the generated spec include a `x-ai-description` field that will be used to generate the description for MCP tools. This description should be concise (ideally less than 200 characters, hard limit of 300) and convey to AI/LLM/MCP the purpose of the identified tool. It is **not** designed for a human operator (who should still use the default `description` field).

This PR introduces a series of processing hooks that will ensure this `x-ai-description` is present. It attempts to do this by performing the following in order:

1. Respect any explicitly defined `x-ai-description` values on the endpoint.
2. Use a new `resource_purpose` field that can be defined on ViewSets.
3. Use contextual details from the operation and resource.

## Type of Change
- [x] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
The following instructions were used with Gateway, and will be from that perspective.
1. Launch a local instance of an AAP component that exposes an OpenAPI spec.
2. Download the generated spec:
```
$ curl --insecure -o /tmp/openapi.json https://$AAP_INSTANCE/api/gateway/v1/docs/schema/
```
3. Confirm that the operations and endpoints have the `x-ai-description` field included.
4. (Optionally) Confirm that human modifications take precedence by manually defining an `x-ai-description` on an endpoint and confirming that the new value is shown on subsequent versions of spec.
5. (Optionally) Confirm that `resource_purpose` is effective, by adding the `resource_purpose` field to a ViewSet and confirming that the new value is shown on subsequent versions of spec.

### Expected Results
The generated spec will have `x-ai-description` on every listed operation, unless it has been explicitly excluded.

## Additional Context
Note that a single endpoint may generate multiple MCP tools. For instance, all of the following tools were created from the AuthenticatorViewSet, and each tool must have a unique description that is relevant to the specific operation. I've been informed that it's insufficient to have the MCP server "guess" at the function by a combination of the description + operation.

- gateway.authenticators_create
- gateway.authenticators_destroy
- gateway.authenticators_list
- gateway.authenticators_retrieve
- gateway.authenticators_update

### Required Actions
- [x] Requires coordination with other teams - we'll likely need to mention this functionality to other teams and coordinate its usage. The hope is that it lessens the effort on their part.

### DIFF
--- a/.github/workflows/dab-consumers.yml
+++ b/.github/workflows/dab-consumers.yml
@@ -27,7 +27,7 @@ jobs:
           cat <<EOF > awx-dockerfile/Dockerfile
           FROM ghcr.io/ansible/awx_devel:devel
           COPY . /opt/dab/
-          RUN pip install -e /opt/dab/
+          RUN pip install -e /opt/dab/[api_documentation]
           RUN pip install jq
           EOF
           docker build -t awx-with-dab:latest -f awx-dockerfile/Dockerfile .
@@ -98,6 +98,6 @@ jobs:
           cd /tmp/eda-server
           pipx install poetry
           poetry install -E all --only main,test
-          poetry run pip install -e "$DAB_PATH"
+          poetry run pip install -e "$DAB_PATH[api_documentation]"
           poetry run pip show django-ansible-base
           poetry run pytest
--- a/.github/workflows/dab-consumers.yml
+++ b/.github/workflows/dab-consumers.yml
@@ -27,7 +27,7 @@ jobs:
           cat <<EOF > awx-dockerfile/Dockerfile
           FROM ghcr.io/ansible/awx_devel:devel
           COPY . /opt/dab/
-          RUN pip install -e /opt/dab/[api_documentation]
+          RUN pip install -e /opt/dab/
           RUN pip install jq
           EOF
           docker build -t awx-with-dab:latest -f awx-dockerfile/Dockerfile .
@@ -98,6 +98,6 @@ jobs:
           cd /tmp/eda-server
           pipx install poetry
           poetry install -E all --only main,test
-          poetry run pip install -e "$DAB_PATH[api_documentation]"
+          poetry run pip install -e "$DAB_PATH"
           poetry run pip show django-ansible-base
           poetry run pytest
--- a/ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
+++ b/ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
@@ -153,7 +153,8 @@ class AuthenticatorViewSet(ModelViewSet):
 
 ## Opting Out
 
-If you need to disable automatic x-ai-description generation for a specific ViewSet/APIView, set:
+If you need to disable automatic x-ai-description generation for a specific ViewSet/APIView, set the
+`skip_ai_description` field on the ViewSet, as seen in the following example:
 
 ```python
 class MyCustomViewSet(ModelViewSet):
@@ -196,13 +197,36 @@ This separation ensures:
 
 After adding or updating `resource_purpose` fields:
 
-1. Regenerate the OpenAPI schema
-2. Check the generated x-ai-description fields
-3. Validate using the validation script:
-   ```bash
-   python tools/validate_ai_descriptions.py /path/to/schema.json
-   ```
-4. Ensure descriptions:
+1. **Generate the OpenAPI schema** using your application's schema generation method
+2. **Inspect the x-ai-description fields** in the generated schema
+
+Example using `jq` to extract all x-ai-description fields:
+
+```bash
+# If schema is available as JSON file
+jq '.. | ."x-ai-description"? // empty' schema.json
+
+# Or extract specific operations
+jq '.paths[][] | select(."x-ai-description") | ."x-ai-description"' schema.json
+```
+
+Example using Python:
+
+```python
+import json
+
+with open('schema.json') as f:
+    schema = json.load(f)
+
+# Find all x-ai-description fields
+for path, operations in schema.get('paths', {}).items():
+    for method, operation in operations.items():
+        if isinstance(operation, dict) and 'x-ai-description' in operation:
+            print(f"{method.upper()} {path}")
+            print(f"  {operation['x-ai-description']}\n")
+```
+
+3. **Verify descriptions:**
    - Are clear and explain WHAT and WHY
    - Stay under 200 characters (preferred) or 300 characters (maximum)
    - Provide useful context for MCP tool selection
@@ -227,9 +251,3 @@ SPECTACULAR_SETTINGS = {
     ],
 }
 ```
-
-## Implementation Details
-
-See the hook implementations:
-- Preprocessing: `ansible_base/api_documentation/preprocessing_hooks.py`
-- Postprocessing: `ansible_base/api_documentation/postprocessing_hooks.py`
--- a/ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
+++ b/ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
@@ -6,80 +6,150 @@ The `x-ai-description` field is automatically generated for all API endpoints an
 
 ## How It Works
 
-The postprocessing hook automatically:
-1. Extracts the operation type (list, create, update, delete, etc.)
-2. Prepends an action verb to your endpoint's description
-3. Generates concise, AI-friendly descriptions
+The system uses a **two-tier approach** to generate descriptions:
+
+### Priority 1: Explicit x-ai-description (Highest Priority)
+If you define `x-ai-description` explicitly using `@extend_schema`, it will be used as-is.
+
+### Priority 2: resource_purpose Field (Recommended)
+If you define a `resource_purpose` field on your ViewSet, the hook will generate contextual descriptions for standard CRUD operations.
+
+### Fallback: Auto-generation
+If neither of the above are present, the hook generates basic descriptions from resource names and operation types.
 
 ### Example Transformations
 
-| Your Description | Operation | Generated x-ai-description |
+With `resource_purpose`:
+
+| resource_purpose | Operation | Generated x-ai-description |
 |------------------|-----------|---------------------------|
-| "groups to be viewed or edited" | GET (list) | "List all groups" |
-| "groups to be viewed or edited" | POST (create) | "Create new group" |
-| "groups to be viewed or edited" | DELETE (destroy) | "Delete existing group" |
-| "status of platform services" | GET (retrieve) | "Retrieve single status of platform services" |
+| "audit trail entries for tracking system changes" | GET (list) | "List audit trail entries for tracking system changes" |
+| "audit trail entries for tracking system changes" | POST (create) | "Create an audit trail entry for tracking system changes" |
+| "audit trail entries for tracking system changes" | GET (retrieve) | "Retrieve an audit trail entry for tracking system changes" |
+| "audit trail entries for tracking system changes" | DELETE (destroy) | "Delete an audit trail entry for tracking system changes" |
 
 ## Writing Good Descriptions
 
-### Guidelines
+### Using resource_purpose (Recommended)
 
-**DO write descriptions that explain WHAT and WHY:**
-- ✅ "teams for organizing users and managing group permissions"
-- ✅ "activity stream entries for auditing system changes"
-- ✅ "HTTP port configurations for routing traffic to backend services"
+The `resource_purpose` field should describe:
+- **WHAT** the resource is (plural noun phrase)
+- **WHY** it exists or when an MCP tool would use it
 
-**DON'T include the action verb (it's added automatically):**
-- ❌ "List all teams"
-- ❌ "Create a new organization"
-- ❌ "Retrieve user details"
-
-**DON'T use generic boilerplate:**
-- ❌ "API endpoint that allows teams to be viewed or edited"
-- ❌ "API endpoint for managing users"
-
-### Sentence Structure
-
-Use this pattern in your ViewSet/APIView docstrings:
-
+**Pattern:**
 ```
-"<resource> for <purpose/why it's used>"
+"<resource_plural> for <purpose/use_case>"
 ```
 
 **Examples:**
-```python
-class TeamViewSet(ModelViewSet):
-    """
-    teams for organizing users and managing group permissions
-    """
-    queryset = Team.objects.all()
-    serializer_class = TeamSerializer
-```
 
 ```python
 class ActivityStreamViewSet(ReadOnlyModelViewSet):
-    """
-    activity stream entries for auditing system changes and compliance tracking
-    """
+    """API endpoint for activity stream entries."""
+
+    resource_purpose = "audit trail entries for tracking system changes and user actions"
+
     queryset = ActivityStream.objects.all()
     serializer_class = ActivityStreamSerializer
 ```
 
 ```python
-class StatusView(APIView):
-    """
-    platform service status for monitoring system health
-    """
-    def get(self, request):
-        ...
+class AuthenticatorViewSet(ModelViewSet):
+    """API endpoint for authenticators."""
+
+    resource_purpose = "authentication providers for configuring user login methods (LDAP, SAML, OAuth)"
+
+    queryset = Authenticator.objects.all()
+    serializer_class = AuthenticatorSerializer
 ```
+
+```python
+class RoleDefinitionViewSet(ModelViewSet):
+    """API endpoint for role definitions."""
+
+    resource_purpose = "RBAC role templates defining permissions that can be assigned to users and teams"
+
+    queryset = RoleDefinition.objects.all()
+    serializer_class = RoleDefinitionSerializer
+```
+
+### Guidelines for resource_purpose
+
+**DO explain WHAT and WHY:**
+- ✅ "authentication providers for configuring user login methods (LDAP, SAML, OAuth)"
+- ✅ "audit trail entries for tracking system changes and user actions"
+- ✅ "HTTP listener ports for routing incoming traffic to backend services"
+
+**DON'T include action verbs (they're added automatically):**
+- ❌ "List authentication providers"
+- ❌ "Create new audit trail entries"
+- ❌ "Manage HTTP ports"
+
+**DON'T use generic boilerplate:**
+- ❌ "resources that can be viewed or edited"
+- ❌ "objects for managing system configuration"
+
+**DO use plural nouns** (the hook will singularize for create/update/delete/retrieve):
+- ✅ "audit trail entries" → becomes "audit trail entry" for create/retrieve
+- ✅ "authentication providers" → becomes "authentication provider" for update
+- ✅ "HTTP listener ports" → becomes "HTTP listener port" for delete
 
 ## Character Limits
 
 - **Preferred:** < 200 characters
 - **Maximum:** < 300 characters (enforced with truncation)
 
-The hook will log warnings if descriptions exceed these limits.
+The hook will automatically truncate descriptions that exceed 300 characters.
+
+**Tips for staying under 200 chars:**
+- Focus on the essential purpose
+- Use parentheses for examples: "(LDAP, SAML, OAuth)"
+- Avoid redundant words: "for" instead of "used for"
+
+## When to Use Each Approach
+
+### Use `resource_purpose` when:
+- ✅ The resource is **domain-specific** or uses technical jargon (e.g., "RBAC role templates", "JWT signing keys", "authentication providers")
+- ✅ The resource purpose isn't immediately obvious from its name (e.g., "authenticator maps" → "attribute mapping rules")
+- ✅ The resource needs contextual explanation for MCP tool selection (e.g., "audit trail entries for tracking system changes")
+- ✅ You have standard CRUD operations that all share the same core purpose
+
+### Use `@extend_schema` with explicit `x-ai-description` when:
+- ✅ You have custom actions beyond CRUD
+- ✅ An operation needs unique context different from the resource purpose
+- ✅ You need fine-grained control over the description
+
+### Prefer auto-generation (no resource_purpose) when:
+- ✅ The resource is **self-explanatory** (e.g., "teams", "users", "organizations")
+- ✅ The resource name clearly conveys its purpose
+- ✅ No domain-specific context is needed for MCP tool selection
+
+**Philosophy:** `resource_purpose` should be used sparingly. Docstrings serve human developers; `resource_purpose` serves AI tool selection. Only add `resource_purpose` when it provides meaningful context that auto-generation cannot capture.
+
+### Example combining both:
+
+```python
+from drf_spectacular.utils import extend_schema
+from rest_framework.decorators import action
+
+class AuthenticatorViewSet(ModelViewSet):
+    """API endpoint for authenticators."""
+
+    # Handles list, create, retrieve, update, delete automatically
+    resource_purpose = "authentication providers for configuring user login methods (LDAP, SAML, OAuth)"
+
+    queryset = Authenticator.objects.all()
+    serializer_class = AuthenticatorSerializer
+
+    # Custom action needs explicit description
+    @extend_schema(
+        extensions={'x-ai-description': 'Test authenticator connection and validate configuration'}
+    )
+    @action(detail=True, methods=['post'])
+    def test(self, request, pk=None):
+        """Test authenticator connectivity"""
+        ...
+```
 
 ## Opting Out
 
@@ -92,82 +162,53 @@ class MyCustomViewSet(ModelViewSet):
     serializer_class = MySerializer
 ```
 
-## Explicit Override
+## Separation of Concerns
 
-To provide a custom x-ai-description for a specific operation, use `@extend_schema`:
-
+**Docstrings are for human developers:**
 ```python
-from drf_spectacular.utils import extend_schema
-
 class TeamViewSet(ModelViewSet):
-    """teams for organizing users"""
+    """
+    API endpoint for managing teams.
 
-    @extend_schema(
-        operation_id='teams_special_action',
-        responses={200: TeamSerializer},
-        **{'x-ai-description': 'Perform special team synchronization operation'}
-    )
-    @action(detail=False, methods=['post'])
-    def sync(self, request):
-        ...
+    Teams organize users and control group-based permissions.
+    """
+    queryset = Team.objects.all()
+    serializer_class = TeamSerializer
 ```
 
-## Current State Analysis
-
-Based on existing Gateway endpoints, most descriptions follow this pattern:
-- ❌ "API endpoint that allows X to be viewed or edited"
-
-**Recommended migration:**
-- ✅ "X for <purpose>"
-
-### Migration Examples
-
-**Before:**
+**resource_purpose is for AI/MCP tool selection:**
 ```python
-"""API endpoint that allows teams to be viewed or edited."""
+class AuthenticatorMapViewSet(ModelViewSet):
+    """API endpoint for authenticator maps."""
+
+    resource_purpose = "attribute mapping rules for mapping external user attributes to AAP user fields"
+
+    queryset = AuthenticatorMap.objects.all()
+    serializer_class = AuthenticatorMapSerializer
 ```
 
-**After:**
-```python
-"""teams for organizing users and managing group permissions"""
-```
-
----
-
-**Before:**
-```python
-"""API endpoint that allows authenticators to be viewed or edited."""
-```
-
-**After:**
-```python
-"""authenticators for configuring user authentication methods (LDAP, SAML, OAuth)"""
-```
-
----
-
-**Before:**
-```python
-"""API endpoint that shows status of platform services."""
-```
-
-**After:**
-```python
-"""platform service status for monitoring system health and availability"""
-```
+This separation ensures:
+- ✅ Docstrings can be detailed and conversational for developers
+- ✅ resource_purpose stays concise and optimized for AI tool selection
+- ✅ Each serves its intended audience without compromise
 
 ## Testing Your Descriptions
 
-After updating descriptions, verify the generated x-ai-description by:
+After adding or updating `resource_purpose` fields:
 
-1. Regenerating the OpenAPI schema
-2. Checking the x-ai-description field for your endpoints
-3. Ensuring descriptions are:
-   - Clear and concise
-   - Under 200 characters (preferred) or 300 characters (maximum)
-   - Describe WHAT and WHY, not HOW
+1. Regenerate the OpenAPI schema
+2. Check the generated x-ai-description fields
+3. Validate using the validation script:
+   ```bash
+   python tools/validate_ai_descriptions.py /path/to/schema.json
+   ```
+4. Ensure descriptions:
+   - Are clear and explain WHAT and WHY
+   - Stay under 200 characters (preferred) or 300 characters (maximum)
+   - Provide useful context for MCP tool selection
 
-## Questions?
+## Implementation Details
 
-See the postprocessing hook implementation at:
-`ansible_base/api_documentation/postprocessing_hooks.py`
+See the hook implementations:
+- Preprocessing: `ansible_base/api_documentation/preprocessing_hooks.py`
+- Postprocessing: `ansible_base/api_documentation/postprocessing_hooks.py`
--- a/ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
+++ b/ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
@@ -207,6 +207,27 @@ After adding or updating `resource_purpose` fields:
    - Stay under 200 characters (preferred) or 300 characters (maximum)
    - Provide useful context for MCP tool selection
 
+## Enabling Automatic x-ai-description Generation
+
+To enable automatic `x-ai-description` generation for your OpenAPI schema, register the preprocessing and postprocessing hooks in your Django settings.
+
+### Configuration
+
+Add or update the `SPECTACULAR_SETTINGS` dictionary in your Django settings file:
+
+```python
+SPECTACULAR_SETTINGS = {
+    # ... your existing settings ...
+
+    'PREPROCESSING_HOOKS': [
+        'ansible_base.api_documentation.preprocessing_hooks.collect_ai_description_metadata',
+    ],
+    'POSTPROCESSING_HOOKS': [
+        'ansible_base.api_documentation.postprocessing_hooks.add_x_ai_description',
+    ],
+}
+```
+
 ## Implementation Details
 
 See the hook implementations:
--- a/ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
+++ b/ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
@@ -1,0 +1,173 @@
+# x-ai-description Generation Guide
+
+## Overview
+
+The `x-ai-description` field is automatically generated for all API endpoints and is used by MCP (Model Context Protocol) servers to provide better context to AI tools about API operations.
+
+## How It Works
+
+The postprocessing hook automatically:
+1. Extracts the operation type (list, create, update, delete, etc.)
+2. Prepends an action verb to your endpoint's description
+3. Generates concise, AI-friendly descriptions
+
+### Example Transformations
+
+| Your Description | Operation | Generated x-ai-description |
+|------------------|-----------|---------------------------|
+| "groups to be viewed or edited" | GET (list) | "List all groups" |
+| "groups to be viewed or edited" | POST (create) | "Create new group" |
+| "groups to be viewed or edited" | DELETE (destroy) | "Delete existing group" |
+| "status of platform services" | GET (retrieve) | "Retrieve single status of platform services" |
+
+## Writing Good Descriptions
+
+### Guidelines
+
+**DO write descriptions that explain WHAT and WHY:**
+- ✅ "teams for organizing users and managing group permissions"
+- ✅ "activity stream entries for auditing system changes"
+- ✅ "HTTP port configurations for routing traffic to backend services"
+
+**DON'T include the action verb (it's added automatically):**
+- ❌ "List all teams"
+- ❌ "Create a new organization"
+- ❌ "Retrieve user details"
+
+**DON'T use generic boilerplate:**
+- ❌ "API endpoint that allows teams to be viewed or edited"
+- ❌ "API endpoint for managing users"
+
+### Sentence Structure
+
+Use this pattern in your ViewSet/APIView docstrings:
+
+```
+"<resource> for <purpose/why it's used>"
+```
+
+**Examples:**
+```python
+class TeamViewSet(ModelViewSet):
+    """
+    teams for organizing users and managing group permissions
+    """
+    queryset = Team.objects.all()
+    serializer_class = TeamSerializer
+```
+
+```python
+class ActivityStreamViewSet(ReadOnlyModelViewSet):
+    """
+    activity stream entries for auditing system changes and compliance tracking
+    """
+    queryset = ActivityStream.objects.all()
+    serializer_class = ActivityStreamSerializer
+```
+
+```python
+class StatusView(APIView):
+    """
+    platform service status for monitoring system health
+    """
+    def get(self, request):
+        ...
+```
+
+## Character Limits
+
+- **Preferred:** < 200 characters
+- **Maximum:** < 300 characters (enforced with truncation)
+
+The hook will log warnings if descriptions exceed these limits.
+
+## Opting Out
+
+If you need to disable automatic x-ai-description generation for a specific ViewSet/APIView, set:
+
+```python
+class MyCustomViewSet(ModelViewSet):
+    skip_ai_description = True
+    queryset = MyModel.objects.all()
+    serializer_class = MySerializer
+```
+
+## Explicit Override
+
+To provide a custom x-ai-description for a specific operation, use `@extend_schema`:
+
+```python
+from drf_spectacular.utils import extend_schema
+
+class TeamViewSet(ModelViewSet):
+    """teams for organizing users"""
+
+    @extend_schema(
+        operation_id='teams_special_action',
+        responses={200: TeamSerializer},
+        **{'x-ai-description': 'Perform special team synchronization operation'}
+    )
+    @action(detail=False, methods=['post'])
+    def sync(self, request):
+        ...
+```
+
+## Current State Analysis
+
+Based on existing Gateway endpoints, most descriptions follow this pattern:
+- ❌ "API endpoint that allows X to be viewed or edited"
+
+**Recommended migration:**
+- ✅ "X for <purpose>"
+
+### Migration Examples
+
+**Before:**
+```python
+"""API endpoint that allows teams to be viewed or edited."""
+```
+
+**After:**
+```python
+"""teams for organizing users and managing group permissions"""
+```
+
+---
+
+**Before:**
+```python
+"""API endpoint that allows authenticators to be viewed or edited."""
+```
+
+**After:**
+```python
+"""authenticators for configuring user authentication methods (LDAP, SAML, OAuth)"""
+```
+
+---
+
+**Before:**
+```python
+"""API endpoint that shows status of platform services."""
+```
+
+**After:**
+```python
+"""platform service status for monitoring system health and availability"""
+```
+
+## Testing Your Descriptions
+
+After updating descriptions, verify the generated x-ai-description by:
+
+1. Regenerating the OpenAPI schema
+2. Checking the x-ai-description field for your endpoints
+3. Ensuring descriptions are:
+   - Clear and concise
+   - Under 200 characters (preferred) or 300 characters (maximum)
+   - Describe WHAT and WHY, not HOW
+
+## Questions?
+
+See the postprocessing hook implementation at:
+`ansible_base/api_documentation/postprocessing_hooks.py`
--- a/ansible_base/api_documentation/path_utils.py
+++ b/ansible_base/api_documentation/path_utils.py
@@ -1,0 +1,97 @@
+"""
+Utility functions for parsing API paths and operation IDs.
+
+This module provides shared utilities used by both preprocessing and
+postprocessing hooks to ensure consistent path parsing logic.
+"""
+
+
+def parse_path_segments(path):
+    """
+    Parse URL path into clean segments, excluding parameter placeholders.
+
+    This handles both Django URL patterns (<pk>, <int:pk>) and OpenAPI
+    format ({id}, {pk}) consistently.
+
+    Args:
+        path: URL path string (e.g., '/api/gateway/v1/teams/{id}/users/')
+
+    Returns:
+        List of path segments excluding parameters
+        (e.g., ['api', 'gateway', 'v1', 'teams', 'users'])
+
+    Examples:
+        >>> parse_path_segments('/api/gateway/v1/teams/')
+        ['api', 'gateway', 'v1', 'teams']
+
+        >>> parse_path_segments('/api/gateway/v1/teams/{id}/users/')
+        ['api', 'gateway', 'v1', 'teams', 'users']
+
+        >>> parse_path_segments('/api/gateway/v1/teams/<pk>/users/')
+        ['api', 'gateway', 'v1', 'teams', 'users']
+    """
+    # Filter out empty strings and path parameters (both {id} and <pk> formats)
+    return [p for p in path.split('/') if p and not p.startswith(('<', '{'))]
+
+
+def extract_operation_prefix(operation_id):
+    """
+    Extract the resource prefix from an operation_id.
+
+    The prefix is everything before the final action (e.g., 'list', 'create').
+    Handles special case of 'partial_update' action.
+
+    Args:
+        operation_id: The operation ID (e.g., 'teams_users_list')
+
+    Returns:
+        The prefix string (e.g., 'teams_users')
+
+    Examples:
+        >>> extract_operation_prefix('teams_list')
+        'teams'
+
+        >>> extract_operation_prefix('teams_users_list')
+        'teams_users'
+
+        >>> extract_operation_prefix('teams_partial_update')
+        'teams'
+
+        >>> extract_operation_prefix('status_retrieve')
+        'status'
+    """
+    if '_partial_update' in operation_id:
+        return operation_id.rsplit('_partial_update', 1)[0]
+    elif '_' in operation_id:
+        return operation_id.rsplit('_', 1)[0]
+    else:
+        return operation_id
+
+
+def extract_operation_action(operation_id):
+    """
+    Extract the action from an operation_id.
+
+    Args:
+        operation_id: The operation ID (e.g., 'teams_users_list')
+
+    Returns:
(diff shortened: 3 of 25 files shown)


huffmanca:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=ansible_django-ansible-base&pullRequest=879) **Quality Gate failed**
> 
> Failed conditions  [2 New issues](https://sonarcloud.io/project/issues?id=ansible_django-ansible-base&pullRequest=879&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=ansible_django-ansible-base&pullRequest=879)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

I believe the two failures mentioned here are because we're using the same format as the other pre- and post- processing hook, and these two errors are due to us following the expected format.

ansible_base/rbac/service_api/views.py
```diff
@@ -1,5 +1,6 @@
 from django.db import transaction
 from django.db.models import OuterRef, Subquery
+from drf_spectacular.utils import extend_schema
```
BrennanPaciorek: We cannot be sure that drf_spectacular will be present on all things calling this, since that is an `ansible_base[api_documentation]` dependency.

BrennanPaciorek: so we'd want to wrap this kind of import in a custom ansible_base.lib decorator that does nothing if the import fails, does `extend_schema` if the import succeeds.

.github/workflows/dab-consumers.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           cat <<EOF > awx-dockerfile/Dockerfile
           FROM ghcr.io/ansible/awx_devel:devel
           COPY . /opt/dab/
-          RUN pip install -e /opt/dab/
```
BrennanPaciorek, line 30: This will not work.

Installing more dependencies than the actual app will install is just masking a problem which we'll see in nightlies.

.github/workflows/dab-consumers.yml
```diff
@@ -98,6 +98,6 @@ jobs:
           cd /tmp/eda-server
           pipx install poetry
           poetry install -E all --only main,test
-          poetry run pip install -e "$DAB_PATH"
```
BrennanPaciorek, line 101: This will not work.

Installing more dependencies than the actual app will install is just masking a problem which we'll see in nightlies.

BrennanPaciorek (changes requested):
Importing packages that we do not depend on in a given DAB app is a no-go.

We need to either have a wrapper for the import , localize the import in sections of code where the import will always have the necessary dependencies, or undertake the effort to make the package needed a dependency to the individual DAB application.

ansible_base/api_documentation/preprocessing_hooks.py
```diff
@@ -0,0 +1,159 @@
...
+
+            # Register ViewSet attributes for AI description generation
+            _register_skip_ai_description(view_class, class_name, prefix)
+            _register_resource_purpose(view_class, class_name, prefix)
+
+        except Exception as e:
+            logger.debug(f"Error checking view metadata for {path} {method}: {e}")
+            continue
+
+    return endpoints
```
BrennanPaciorek:
```suggestion
    if endpoints:
        for path, path_regex, method, view in endpoints:
            try:
                # Extract ViewSet class from the view
                view_class = _get_view_class(view)
                class_name = view_class.__name__

                # Extract operation_id prefix from path
                prefix, path_parts, path_parts_count = _extract_prefix_from_path(path)
                if prefix is None:
                   continue

                # Store or handle collision for this prefix
                if prefix not in OPERATION_CLASS_MAP:
                    # First time seeing this prefix - store it
                    OPERATION_CLASS_MAP[prefix] = (class_name, path_parts_count, path_parts)
                else:
                    # Handle collision (different ViewSet with same prefix)
                    prefix = _handle_prefix_collision(prefix, class_name, path_parts_count, path_parts, OPERATION_CLASS_MAP)

                # Register ViewSet attributes for AI description generation
                _register_skip_ai_description(view_class, class_name, prefix)
                _register_resource_purpose(view_class, class_name, prefix)

            except Exception as e:
                logger.debug(f"Error checking view metadata for {path} {method}: {e}")
                continue

    return endpoints
```

To appease SonarCloud, we can do this.

BrennanPaciorek (commented):
SonarCloud-related review.

.github/workflows/dab-consumers.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           cat <<EOF > awx-dockerfile/Dockerfile
           FROM ghcr.io/ansible/awx_devel:devel
           COPY . /opt/dab/
-          RUN pip install -e /opt/dab/
+          RUN pip install -e /opt/dab/[api_documentation]
```
john-westcott-iv: I think AWX is using more than just api_documentation: https://github.com/ansible/awx/blob/devel/requirements/requirements_git.txt#L4

.github/workflows/dab-consumers.yml
```diff
@@ -98,6 +98,6 @@ jobs:
           cd /tmp/eda-server
           pipx install poetry
           poetry install -E all --only main,test
-          poetry run pip install -e "$DAB_PATH"
+          poetry run pip install -e "$DAB_PATH[api_documentation]"
```
john-westcott-iv: EDA is also using more than just api_documentation, I think this is their import but 100% not sure since they user poetry. https://github.com/ansible/eda-server/blob/bafe321db5187a1358d5d661505b8b7464aceb00/pyproject.toml#L50

ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
```diff
@@ -0,0 +1,253 @@
...
+## Overview
+
+The `x-ai-description` field is automatically generated for all API endpoints and is used by MCP (Model Context Protocol) servers to provide better context to AI tools about API operations.
+
+## How It Works
+
+The system uses a **two-tier approach** to generate descriptions:
+
+### Priority 1: Explicit x-ai-description (Highest Priority)
+If you define `x-ai-description` explicitly using `@extend_schema`, it will be used as-is.
```
john-westcott-iv: Maybe add an example here?

ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
```diff
@@ -0,0 +1,253 @@
...
+### Example Transformations
+
+With `resource_purpose`:
+
+| resource_purpose | Operation | Generated x-ai-description |
+|------------------|-----------|---------------------------|
+| "audit trail entries for tracking system changes" | GET (list) | "List audit trail entries for tracking system changes" |
+| "audit trail entries for tracking system changes" | POST (create) | "Create an audit trail entry for tracking system changes" |
+| "audit trail entries for tracking system changes" | GET (retrieve) | "Retrieve an audit trail entry for tracking system changes" |
+| "audit trail entries for tracking system changes" | DELETE (destroy) | "Delete an audit trail entry for tracking system changes" |
```
john-westcott-iv: Should we add examples for the other options?

docs/apps/ai_documentation/x_ai_description_guide.md
```diff
@@ -0,0 +1,253 @@
...
+```python
+class RoleDefinitionViewSet(ModelViewSet):
+    """API endpoint for role definitions."""
+
+    resource_purpose = "RBAC role templates defining permissions that can be assigned to users and teams"
+
+    queryset = RoleDefinition.objects.all()
+    serializer_class = RoleDefinitionSerializer
+```
+
```
john-westcott-iv, line 138: Can we add some "bad" examples?

docs/apps/ai_documentation/x_ai_description_guide.md
```diff
@@ -0,0 +1,253 @@
...
+    # ... your existing settings ...
+
+    'PREPROCESSING_HOOKS': [
+        'ansible_base.api_documentation.preprocessing_hooks.collect_ai_description_metadata',
+    ],
+    'POSTPROCESSING_HOOKS': [
+        'ansible_base.api_documentation.postprocessing_hooks.add_x_ai_description',
+    ],
+}
+```
```
john-westcott-iv: Do we do this in dynamic_settings for people or does this def have to be manually done?

ansible_base/api_documentation/path_utils.py
```diff
@@ -0,0 +1,55 @@
...
+Utility functions for parsing API paths and operation IDs.
+
+This module provides shared utilities used by both preprocessing and
+postprocessing hooks to ensure consistent path parsing logic.
+"""
+
+import re
+
+
+def parse_path_segments(path):
```
john-westcott-iv: Can you add mypy context to these functions?

ansible_base/lib/utils/api_path_utils.py
```diff
@@ -0,0 +1,55 @@
+"""
+Utility functions for parsing API paths and operation IDs.
+
+This module provides shared utilities used by both preprocessing and
+postprocessing hooks to ensure consistent path parsing logic.
+"""
```
john-westcott-iv, line 6: Do any of these utilities have value outside of api_documentation? i.e. should we put this into ansible_base/lib/ somewhere so others can reuse/

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,466 @@
...
+        return resource_name[:-3] + 'y'
+    elif resource_name.endswith('ses'):
+        # e.g., 'addresses' -> 'address'
+        return resource_name[:-2]
+    elif resource_name.endswith('s') and len(resource_name) > 1:
+        # Standard plural: just remove 's'
+        return resource_name[:-1]
+
+    # Already singular or unknown pattern
+    return resource_name
```
john-westcott-iv: Is there any kind of library that might do this for us? I also feel like we might do this somewhere already for the routers or urls or something.

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,466 @@
...
+    action_templates = {
+        'list': lambda p: f"List {p}",
+        'retrieve': lambda p: f"Retrieve a {singularize_resource_purpose(p)}",
+        'read': lambda p: f"Retrieve a {singularize_resource_purpose(p)}",
+        'create': lambda p: f"Create a {singularize_resource_purpose(p)}",
+        'update': lambda p: f"Update a {singularize_resource_purpose(p)}",
+        'partial_update': lambda p: f"Update a {singularize_resource_purpose(p)}",
+        'destroy': lambda p: f"Delete a {singularize_resource_purpose(p)}",
+        'delete': lambda p: f"Delete a {singularize_resource_purpose(p)}",
+    }
```
john-westcott-iv: Should this be a global like `OPERATION_DESCRIPTIVE_NAMES` or even merged with it?

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,466 @@
...
+        return f"{operation_name} {resource_name}"
+
+    # For partial_update, use special formatting
+    if action == 'partial_update':
+        if parent_resource:
+            return f"Partially update existing {resource_name}"
+        return f"Partially update existing {singularize_resource(resource_name)}"
+
+    # For other CRUD operations (retrieve, read, create, update, destroy, delete)
+    if action in ['retrieve', 'read', 'create', 'update', 'destroy', 'delete']:
```
john-westcott-iv: Could this be something more like `OPERATION_DESCRIPTIVE_NAMES.keys()` or somehow otherwise be auto-generated instead of a separate maintained list (applies to other locations).

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,466 @@
...
+    # Remove common prefixes
+    prefixes_to_remove = [
+        'API endpoint that allows',
+        'API endpoint that',
+        'API endpoint for',
+        'A view class for managing and displaying',
+        'Endpoint:',
+    ]
+
+    for prefix in prefixes_to_remove:
```
john-westcott-iv, line 299: Should we do a lower here os anything incase the case is off?

ansible_base/authentication/views/authenticator_map.py
```diff
@@ -10,6 +10,8 @@ class AuthenticatorMapViewSet(AnsibleBaseDjangoAppApiView, ModelViewSet):
     API endpoint that allows authenticator maps to be viewed or edited.
     """
 
+    resource_purpose = "attribute mapping rules for mapping external user attributes to AAP user fields"
+
```
john-westcott-iv: It can include roles as well, not just user fields. i.e. organization admin, team member, etc.

ansible_base/rbac/api/views.py
```diff
@@ -116,6 +116,8 @@ class RoleDefinitionViewSet(AnsibleBaseDjangoAppApiView, ModelViewSet):
     but can be assigned to users.
     """
 
+    resource_purpose = "RBAC role templates defining permissions that can be assigned to users and teams"
```
john-westcott-iv: Not sure how I feel about `templates` being in here.

docs/apps/ai_documentation/x_ai_description_guide.md
```diff
@@ -0,0 +1,253 @@
+# x-ai-description Generation Guide
```
john-westcott-iv, line 1: Most docs don't live under the applications, they live in the top level docs folder. There are folders for the apps in there. If ai_documentation dosen't have an folder you can create one.

huffmanca: This has been moved to the `docs/` directory, and the existing `api_documentation.md` file has been updated to reference it.

ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
huffmanca: Examples have been reworked, consolidated, and linked where appropriate.

ansible_base/api_documentation/X_AI_DESCRIPTION_GUIDE.md
huffmanca: The remaining operations should have examples now.

docs/apps/ai_documentation/x_ai_description_guide.md
huffmanca: These have been included.

ansible_base/rbac/service_api/views.py
huffmanca: I've added an `extend_schema_if_available` decorator, and updated this section to reference it.

ansible_base/lib/utils/api_path_utils.py
huffmanca: Moved this to `ansible_base/lib/utils/api_path_utils/`.

ansible_base/api_documentation/path_utils.py
huffmanca: These have been updated.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: You're right about there being a library to do this - inflection exists that can do this for us, and I've updated to use it.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: I've modified this to be a global.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: I've gone through and updated the `for $VAR in $LIST` logic to try and simplify/generalize this where it made sense. This and the other action loop are now pulling from the global list, and I've moved the supported HTTP methods to the top of the file as well.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: The next line has been updated to lower both the `prefix` and the `clean_desc`. I put `prefix.lower()` instead of overriding the list in case people add additional values to the list later on, and didn't want to introduce a human requirement.

ansible_base/api_documentation/preprocessing_hooks.py
huffmanca: I've updated this per the suggestion.

ansible_base/authentication/views/authenticator_map.py
huffmanca: I've updated the resource_purpose here from looking at the Ansible 2.6 docs and how AuthenticatorMap is used. Let me know if the current description is still insufficient.

ansible_base/rbac/api/views.py
huffmanca: Removed.

ansible_base/authentication/views/authenticator.py
```diff
@@ -16,6 +16,8 @@ class AuthenticatorViewSet(AnsibleBaseDjangoAppApiView, ModelViewSet):
     API endpoint that allows authenticators to be viewed or edited.
     """
 
+    resource_purpose = "authentication providers for configuring user login methods (LDAP, SAML, OAuth)"
```
tznamena: Can we shorten this to "authentication types (...)" to better align it with what we call this in the UI?

ansible_base/authentication/views/authenticator_map.py
```diff
@@ -10,6 +10,8 @@ class AuthenticatorMapViewSet(AnsibleBaseDjangoAppApiView, ModelViewSet):
     API endpoint that allows authenticator maps to be viewed or edited.
     """
 
+    resource_purpose = "conditional rules for granting permissions and assigning users to organizations or teams based on external authentication attributes"
```
tznamena: How about: "conditional rule for granting access, membership or roles based on user attributes or groups"
Also, are we using plural for views and singular for actual resource?

tznamena (commented):
I started reviewing the individual descriptions of the resources and realized that in some views the description mentions the resources and their intended use, however, 

I wonder if the views should simply state "list of xyz" and the description of that resource should explain what it is for?

docs/apps/ai_documentation/x_ai_description_guide.md
BrennanPaciorek, line 1: Do we want to document `extend_schema_if_available` in `docs/lib`, or add a clause to the `x_ai_description_guide` to use that decorator if developing for one of the non-`api-documentation` django apps?

ansible_base/rbac/service_api/views.py
BrennanPaciorek: This is good. feel free to resolve this, I don't have the button.

ansible_base/authentication/views/authenticator.py
huffmanca: Updated verbiage to use the suggestion here.

ansible_base/authentication/views/authenticator_map.py
huffmanca: Updated verbiage to use the suggestion here.

docs/apps/ai_documentation/x_ai_description_guide.md
huffmanca: I've added this to the dynamic settings, and updated the docs to be clear regarding when they'll need to explicitly add these hooks.

tznamena (approved):

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+    'read': lambda p: f"Retrieve a {singularize_resource_purpose(p)}",
+    'create': lambda p: f"Create a {singularize_resource_purpose(p)}",
+    'update': lambda p: f"Update a {singularize_resource_purpose(p)}",
+    'partial_update': lambda p: f"Update a {singularize_resource_purpose(p)}",
+    'destroy': lambda p: f"Delete a {singularize_resource_purpose(p)}",
+    'delete': lambda p: f"Delete a {singularize_resource_purpose(p)}",
+}
+
+
+def _extract_resource_parts_from_operation_id(operation_id, action):
```
john-westcott-iv: I would /strongly/ suggest mypy references everywhere if possible.

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+    """Extract resource parts from operation_id by removing the action suffix."""
+    # Handle special case of "partial_update" which has two underscores
+    if action == 'partial_update' and '_partial_update' in operation_id:
+        return operation_id.split('_')[:-2] if operation_id.count('_') >= 2 else []
+
+    # Standard case: remove the last segment (the action)
+    if '_' in operation_id:
+        return operation_id.split('_')[:-1]
```
john-westcott-iv: Not knowing what the other operation_ids are could this just be:
```python
return [x for x in operation_id.split('_') if x]
```

Or as claude points out, the filter command can take a None parameter to remove falsy items so you could do:
```python
return list(filter(None, operation_id.split('_')))
```

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+
+    return []
+
+
+def _has_segments_after_placeholder(parts, placeholder_index):
+    """Check if there are non-placeholder resource segments after a placeholder."""
```
john-westcott-iv: This could use a much better doc string indicating what should be coming into thiese variables and what you are returning for future maintenance.

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+    return None
+
+
+def _find_parent_resource_from_path(path, placeholders=None):
+    """
+    Find parent resource in nested REST path by detecting placeholders with segments after.
+    Returns singular parent resource name or None if not nested.
+    """
+    if placeholders is None:
+        placeholders = ['{id}', '{pk}']
```
john-westcott-iv: If someone used something else besides one of these will there be something "obvious" that they forgot to set placeholders? Like a log warning or something or we just wouldn't know until someone was reviewing the open api spec?

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+    if path_parts and path_parts[-1] in ['associate', 'disassociate']:
+        path_parts = path_parts[:-1]
+
+    if len(path_parts) >= 2:
+        child = path_parts[-1].replace('_', ' ')
+        parent_singular = singularize(path_parts[-2])
+        parent_readable = parent_singular.replace('_', ' ')
+        article = 'an' if parent_readable[0].lower() in 'aeiou' else 'a'
+
+        verb = "Associate" if is_associate else "Disassociate"
```
john-westcott-iv: You could move this line outside of the if to reduce one dupe line. Non blocking.

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+    # For partial_update, use special formatting
+    if action == 'partial_update':
+        if parent_resource:
+            return f"Partially update existing {resource_name}"
+        return f"Partially update existing {singularize(resource_name)}"
+
+    # For other CRUD operations supported by OPERATION_DESCRIPTION_TEMPLATES
+    if action in OPERATION_DESCRIPTION_TEMPLATES.keys():
+        if parent_resource:
+            # Compound resources already formatted
```
john-westcott-iv: I might be good to provide examples of compound s simple resources for future people.

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+        'Endpoint:',
+    ]
+
+    for prefix in prefixes_to_remove:
+        if clean_desc.lower().startswith(prefix.lower()):
+            clean_desc = clean_desc[len(prefix) :].strip()
+            break
+
+    # Remove trailing period
+    if clean_desc.endswith('.'):
```
john-westcott-iv: Would we care about anything other than `.`? Not that someone should really have 'API endpoint that allows authenticators!' or anything who knows.

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+
+    # If description is too long or contains multiple sentences, truncate
+    # This handles cases like settings_getter where entire docstring was used
+    if len(clean_desc) > 100 or '\n' in clean_desc:
+        # Take first sentence or first 100 chars
+        first_line = clean_desc.split('\n')[0]
+        if '.' in first_line[:100]:
+            clean_desc = first_line.split('.')[0]
+        else:
+            clean_desc = first_line[:100]
```
john-westcott-iv: Do you want to do something like `clean_desc = f'{first_line[:97]...}` to indicate the truncation?

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+            clean_desc = clean_desc[len(prefix) :].strip()
+            break
+
+    # Remove trailing period
+    if clean_desc.endswith('.'):
+        clean_desc = clean_desc[:-1]
+
+    # If description is too long or contains multiple sentences, truncate
+    # This handles cases like settings_getter where entire docstring was used
+    if len(clean_desc) > 100 or '\n' in clean_desc:
```
john-westcott-iv: 100 is in 3 places. I might say make this a variable so that if someone was like lets go to 75 or 200 its a simpler change in the future.

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+        # Take first sentence or first 100 chars
+        first_line = clean_desc.split('\n')[0]
```
john-westcott-iv: Is there any chance that a description might like:
```
This is an API
that allows us to 
create authenticators
```
And here we ned up just getting `This is an API`. Would it be better to just do something like `join(' ', clean_desc.split('\n'))` to make it a single line and then do the 100 character checking?

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+
+def _generate_from_resource_purpose(resource_purpose, action):
+    """Generate description from resource_purpose if action is supported."""
+    if not resource_purpose:
+        return None
+
+    if action in OPERATION_DESCRIPTION_TEMPLATES.keys():
+        return generate_description_from_purpose(resource_purpose, action)
+
+    return None
```
john-westcott-iv: `generate_description_from_purpose` returns resource_purpose if it doesn't get a template based on the action so it seems like this could just be something like:
```python
return generate_description_from_purpose(resource_purpose, action) if action in OPERATION_DESCRIPTION_TEMPLATES.keys() else None
```
Unless the reource_purpose could be None or '' but even then you could consense this to be 
```python
    if resource_purpose and action in OPERATION_DESCRIPTION_TEMPLATES.keys():
        return generate_description_from_purpose(resource_purpose, action)

    return None
```

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+
+def _should_skip_ai_description(operation_id):
+    """Check if an operation should skip AI description generation."""
+    if not operation_id:
+        return False
+
+    prefix = operation_id.split('_')[0] if '_' in operation_id else operation_id
+    if prefix in SKIP_AI_DESCRIPTION_PREFIXES:
+        return True
+    return False
```
john-westcott-iv: Instead of split in here I might use a regex and make this a 1 liner:
```python
# Remove a leading _ from operation_id and see if the resulting value is in SKIP_AI_DESCRIPTION_PREFIXES
return re.sub(r'^_', '', '' if operation_id is None else operation_id) in SKIP_AI_DESCRIPTION_PREFIXES
```

Also, if you do something like this don't actually make a 1 line non-reused function, just drop this in place of the function call.

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+        ai_description = generate_custom_action_description(operation_name, resource_name, operation)
+
+    return ai_description
+
+
+def _enforce_character_limit(description, max_length=300):
+    """Enforce character limit on description, truncating if necessary."""
+    if len(description) > max_length:
+        return description[: max_length - 3] + "..."
+    return description
```
john-westcott-iv: Isn't this essentially what we are doing up above with the 100 character limit? Can we just call this from up there?

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+    if 'x-ai-description' in operation:
+        logger.debug(f"x-ai-description already defined for {path} {method.upper()}. Respecting existing value.")
+        return
+
+    # Get operation ID
+    operation_id = operation.get('operationId', '')
+
+    # Check if the ViewSet has opted out of AI description generation
+    if _should_skip_ai_description(operation_id):
+        return
```
john-westcott-iv, line 389: Add a debug here as to why you are just returning to save yourself some trouble shooting headache in the future.

ansible_base/api_documentation/postprocessing_hooks.py
```diff
@@ -0,0 +1,435 @@
...
+
+    for path, path_item in paths.items():
+        for method, operation in path_item.items():
+            # Skip non-operation keys (like 'parameters')
+            if method not in HTTP_METHODS:
+                continue
+
+            # Process this operation to add x-ai-description
+            _process_operation(operation, method, path)
+
```
john-westcott-iv: Save some bits ....

```suggestion
            if method in HTTP_METHODS:
                # Process this operation to add x-ai-description
                _process_operation(operation, method, path)
```

ansible_base/api_documentation/preprocessing_hooks.py
```diff
@@ -0,0 +1,157 @@
...
+# We store the count to resolve collisions: fewer path parts = main resource, gets simple prefix
+# We store path_parts to generate correct compound prefixes when resolving collisions
+OPERATION_CLASS_MAP = {}
+
+
+def _get_view_class(view):
+    """Extract the ViewSet class from a view, handling DRF's view wrapping."""
+    if hasattr(view, 'cls'):
+        return view.cls
+    return view.__class__
```
john-westcott-iv: Shorter version?
```suggestion
    return getattr(view, 'cls', view.__class__)
```

ansible_base/api_documentation/preprocessing_hooks.py
```diff
@@ -0,0 +1,157 @@
...
+
+# Global storage for ViewSet class names by operation_id prefix
+# Maps operation_id prefix (e.g. "authenticators") -> (ViewSet class name, path_parts_count, path_parts)
+# This is shared with postprocessing_hooks.py
+# We store the count to resolve collisions: fewer path parts = main resource, gets simple prefix
+# We store path_parts to generate correct compound prefixes when resolving collisions
+OPERATION_CLASS_MAP = {}
+
+
+def _get_view_class(view):
```
john-westcott-iv: Again, use mypy type hints everywehere we can.

ansible_base/api_documentation/preprocessing_hooks.py
```diff
@@ -0,0 +1,157 @@
...
+        SKIP_AI_DESCRIPTION_PREFIXES.add(prefix)
+        logger.info(f"View class {class_name} (prefix: {prefix}) has skip_ai_description=True")
+
+
+def _register_resource_purpose(view_class, class_name, prefix):
+    """Register a ViewSet's resource_purpose for description generation."""
+    resource_purpose = getattr(view_class, 'resource_purpose', None)
+    if resource_purpose:
+        RESOURCE_PURPOSE_MAP[class_name] = resource_purpose
+        logger.debug(f"View class {class_name} (prefix: {prefix}) has resource_purpose: {resource_purpose[:50]}...")
```
john-westcott-iv: This would print out the ... even if the resoure_purpose is < 50 characters. I don't really care, just pointing it out.

ansible_base/lib/utils/api_path_utils.py
```diff
@@ -0,0 +1,56 @@
...
+import re
+from typing import List
+
+
+def parse_path_segments(path: str) -> List[str]:
+    """
+    Parse URL path into segments, excluding placeholders.
+    Handles both Django (<pk>) and OpenAPI ({id}) formats.
+    """
+    return [p for p in path.split('/') if p and not p.startswith(('<', '{'))]
```
john-westcott-iv, line 22: I love this for reusability, wondering if we would want / if it would make sense to make an object with methods? Thinking like maybe this parse_path_segments is the __init__ and then you could have objects.get_operation_prefix() or whatever. If that is going to be too much rework no worries.

ansible_base/lib/utils/api_path_utils.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Extract resource prefix from operation_id (everything before final action).
+    Handles special case of 'partial_update'.
+    """
+    if '_partial_update' in operation_id:
+        return operation_id.rsplit('_partial_update', 1)[0]
+    elif '_' in operation_id:
+        return operation_id.rsplit('_', 1)[0]
+    else:
+        return operation_id
```
john-westcott-iv, line 35: This feels similar to what we did above.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: Every method should now have mypy types. Note that for `add_x_ai_description` I'm using `Any` to prevent importing libraries.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: This method ended up being largely redundant, so I've removed it and combined this logic into `extract_action_and_resource`.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: Doc strings have been updated to indicate what placeholders are along with some examples.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: You're right. Since this is the postprocessing hook, everything should be in the openapi format. I updated it to search for any placeholder instead of just defining these defaults.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: Done.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: Updated the comment here, and also referenced the method that does the initial formatting for these compounds.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: I'd be surprised, but it's a valid point. The entire `clean_base_description` method has been tweaked a bit, but we're now removing punctuation at the end instead of just '.'

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: You mention later that we already have _enforce_character_limit, and that runs as the final step of this entire hook. I removed the character limit sections in clean_base_description entirely.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: This entire section was removed due to the hard limit being performed in _enforce_character_limit .

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: Updated to use the second version. It seems cleaner to me.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: This has been reworked slightly, and we're now pulling the operation and determining if it should be skipped. The function where this line was referenced has been fully removed.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: Yeah - you're right to identify the duplication. I removed the hard character limit from clean_base_description entirely since we're doing a final truncation at the end.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: Good call. Done.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: Done.

ansible_base/api_documentation/preprocessing_hooks.py
huffmanca: Tweaked to use this one-liner.

ansible_base/lib/utils/api_path_utils.py
huffmanca: Yeah - I removed a couple places in the postprocessing_hooks where we had similar or identical logic, and had them reference this function instead. We *should* be good now.

ansible_base/api_documentation/preprocessing_hooks.py
huffmanca: Done.

ansible_base/api_documentation/postprocessing_hooks.py
huffmanca: Yeah - it's a valid case. I modified this bit to join the entire description into a single string for processing. So the flow now is:
1. Combine everything into a single string.
2. Grab the first sentence for the common punctuation types (I'm intentionally leaving it on `;` or `:` just in case.)
3. Remove final punctuation

ansible_base/api_documentation/preprocessing_hooks.py
huffmanca: Valid. And updated.

ansible_base/lib/utils/api_path_utils.py
huffmanca: This is valid, and I've waffled a bit on it, but considering the size of this PR and the issues that have already been updated, I'm hesitant on changing these bits in the event that it results in another round of reviews.

I'm leaving this open, and we can file a Jira to change the behavior in a subsequent sprint.

huffmanca:
Comments from the most recent round of review should now be addressed. The SonarCloud error is still reported, but I'd prefer to keep the method signature identical to the drf_spectacular's postprocessing_hooks. I can be convinced to change it, since we're not currently doing anything with these arguments, but it seems clearer (to me) to have it be identical to the original hook.

BrennanPaciorek:
>Comments from the most recent round of review should now be addressed. The SonarCloud error is still reported, but I'd prefer to keep the method signature identical to the drf_spectacular's postprocessing_hooks. I can be convinced to change it, since we're not currently doing anything with these arguments, but it seems clearer (to me) to have it be identical to the original hook.

Agreed on this approach. @john-westcott-iv @bhavenst, can either of you manually close the sonar issue? I don't have permissions to do so on this repo.

docs/apps/ai_documentation/x_ai_description_guide.md
huffmanca: I updated the `x_ai_description_guide` with references to this decorator and updated the examples to use it as well. There's also note in the `docs/apps/api_documentation.md` that points to this file now. 

I also went ahead and a very basic `docs/lib/schema.md` file that's rather simple. It effectively acts as a pointer to the decorator, but I'm uncertain if this is going to be enough for users to find it. It seems that if they're using `api_documentation`, then they'd see the docs and potentially not need the custom decorator. If they're *not* using `api_documentation`, then we're hoping they see the docs and use the decorator. 

I'm not certain how to fix that issue, though.

john-westcott-iv (commented):
## Code Review: AAP-56282 - x-ai-description Generation

This PR adds automatic `x-ai-description` generation for OpenAPI specs to support MCP server tools. The implementation is **well-designed with comprehensive tests and documentation**. The three-tier priority system (explicit → resource_purpose → auto-generated) provides good flexibility.

### Strengths
- ✅ **Excellent test coverage** - 1000+ lines of tests across unit, integration, and schema tests
- ✅ **Comprehensive documentation** - 346-line guide with clear examples and best practices
- ✅ **Graceful dependency handling** - `extend_schema_if_available` wrapper prevents hard drf_spectacular dependency
- ✅ **Smart collision resolution** - Handles nested resources elegantly (e.g., `/teams/` vs `/orgs/{id}/teams/`)
- ✅ **Edge case handling** - Properly handles `partial_update` special case

### Issues

**ansible_base/api_documentation/postprocessing_hooks.py:84-92**
```python
def _truncate_description(description: str, max_length: int = 300) -> str:
    """Truncate description to max_length with ellipsis if needed."""
    if len(description) <= max_length:
        return description
    return description[: max_length - 3] + '...'
```
**Issue:** Truncation may cut mid-word, creating awkward descriptions for MCP tools.

```suggestion
def _truncate_description(description: str, max_length: int = 300) -> str:
    """Truncate description to max_length with ellipsis, breaking at word boundary."""
    if len(description) <= max_length:
        return description
    
    # Try to break at last space before max_length
    truncated = description[:max_length - 3]
    last_space = truncated.rfind(' ')
    if last_space > max_length * 0.7:  # Only use word boundary if not too far back
        return truncated[:last_space] + '...'
    return truncated + '...'
```

**ansible_base/api_documentation/preprocessing_hooks.py:14-23**
Global variables for sharing state between hooks raise thread-safety concerns.

```python
SKIP_AI_DESCRIPTION_PREFIXES = set()
RESOURCE_PURPOSE_MAP = {}
OPERATION_CLASS_MAP = {}
```
**Question:** Is schema generation single-threaded? If OpenAPI spec can be generated concurrently (e.g., different API versions), these globals could cause race conditions. Consider documenting thread-safety assumptions or using threading.local().

**ansible_base/api_documentation/postprocessing_hooks.py:156-169**
The `singularize` function uses simple heuristics that may fail for irregular plurals.

```python
def singularize(text: str) -> str:
    # ... handles common cases like "ies" -> "y", "sses" -> "ss", "s" -> ""
```
**Minor Issue:** Will fail for irregular plurals (e.g., "people" → "people", "children" → "children", "data" → "data"). Consider adding a lookup table for common irregular forms or using inflect library if this becomes an issue.

### Suggestions

**ansible_base/lib/utils/api_path_utils.py:14-16**
Missing docstring examples and edge case documentation.

```suggestion
def parse_path_segments(path: str) -> List[str]:
    """
    Parse URL path into segments, excluding placeholders.
    Handles both Django (<pk>) and OpenAPI ({id}) formats.
    
    Examples:
        >>> parse_path_segments('/api/v1/teams/')
        ['api', 'v1', 'teams']
        >>> parse_path_segments('/api/v1/teams/{id}/users/')
        ['api', 'v1', 'teams', 'users']
    """
    return [p for p in path.split('/') if p and not p.startswith(('<', '{'))]
```

**docs/apps/ai_documentation/x_ai_description_guide.md:129**
Documentation states "resource_purpose should be used sparingly" but doesn't provide metrics. Consider adding guidance like "~20-30% of ViewSets typically need resource_purpose."

### Notes
- **SonarCloud unused parameter warning**: Intentionally keeping signature identical to drf_spectacular hooks for consistency - this is the right call.
- **Preprocessing/postprocessing hook pattern**: Clean separation between metadata collection and description generation.
- **Character limits enforced**: 300 char hard limit with 200 preferred - good for MCP tool descriptions.

### Verdict
Strong implementation with minor improvements needed. The word-boundary truncation fix is recommended but not blocking. Thread-safety clarification would be valuable documentation.

huffmanca:
Feedback from the PR review bot should be addressed. Here are my notes on these:
1. Address hard character truncation - using the recommended feedback (although this is in `_enforce_character_limit` instead of `_truncate_description`). Should now truncate based on the last word found instead of a hard limit at the last character.
2. Thread safety status - from research + claude verification, spec generation appears to be single threaded, which includes the pre- and post- processing hooks. I've added a comment before the global variables to indicate this.
3. Inflection's `.singularize` appears to cover irregular plurals from what I can tell. No changes made here.
4. I've updated the docstring with examples.
5. I've updated the verbiage surrounding when resource_purpose should be used, and I've confirmed with Claude locally that it is clear. Personally I want to avoid a hard percentage recommendation, as I don't want to steer teams incorrectly, but I've added guidelines on when it's appropriate.

AlanCoding:
Here's the command I ran to produce this for test_app:

```
python manage.py spectacular --format openapi-json --file openapi.json
```

on your branch here...

and here's an example:

```json
        "/api/v1/activitystream/": {
            "get": {
                "operationId": "activitystream_list",
                "description": "API endpoint that allows for read-only access to activity stream entries.",
                "parameters": [
                    {
                        "in": "query",
                        "name": "order",
                        "schema": {
                            "type": "string"
                        },
                        "description": "Order results by field name. Prefix with '-' for descending order. Supports comma-separated values for multiple fields."
                    },
                    {
                        "in": "query",
                        "name": "order_by",
                        "schema": {
                            "type": "string"
                        },
                        "description": "Order results by field name. Prefix with '-' for descending order. Supports comma-separated values for multiple fields."
                    },
                    {
                        "name": "page",
                        "required": false,
                        "in": "query",
                        "description": "A page number within the paginated result set.",
                        "schema": {
                            "type": "integer"
                        }
                    },
                    {
                        "name": "page_size",
                        "required": false,
                        "in": "query",
                        "description": "Number of results to return per page.",
                        "schema": {
                            "type": "integer"
                        }
                    },
                    {
                        "name": "search",
                        "required": false,
                        "in": "query",
                        "description": "A search term.",
                        "schema": {
                            "type": "string"
                        }
                    },
                    {
                        "in": "query",
                        "name": "type",
                        "schema": {
                            "type": "string"
                        },
                        "description": "Filter by object type. Supports comma-separated values for multiple types."
                    }
                ],
                "tags": [
                    "activitystream"
                ],
                "security": [
                    {
                        "SessionAuthentication": []
                    },
                    {
                        "LoggedBasicAuthentication": []
                    }
                ],
                "responses": {
                    "200": {
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/PaginatedEntryList"
                                }
                            }
                        },
                        "description": ""
                    }
                },
                "x-ai-description": "List audit trail entries for tracking system changes and user actions"
            }
        },
```

Shortening those up:

```
            "get": {
                "operationId": "activitystream_list",
                "description": "API endpoint that allows for read-only access to activity stream entries.",
                "x-ai-description": "List audit trail entries for tracking system changes and user actions"
```

Firstly, in the `docs/` content, I would like it to be a little more clear about what this produces. I mean, specifically, give a an example snippet like this, and clarifying how it adds the `"x-ai-description"` key in the specific location where it does. As that is very non-obvious.

Secondly, I admit to really not getting it. This seems to co-exist on the same level as the `"description"` field. And the code here appears to do a lot of reinventing the wheel with the pre & post processors, which I imagine is very close to how the "description" field is generated. Why would we not use the "description" field?

Even if this direction is intended, the current patch doesn't _remove_ the description field. It has duplicated content.

Ping @mabashian ref AAP-56226

huffmanca:
@AlanCoding , let me do my best to answer the questions based on my current understanding. If I'm misunderstanding, then I'll go ahead and apologize for the wasted time.

> Firstly, in the docs/ content, I would like it to be a little more clear about what this produces. I mean, specifically, give a an example snippet like this, and clarifying how it adds the "x-ai-description" key in the specific location where it does. As that is very non-obvious.

~~I can include this.~~ Edit: This has been included from the latest commit.

> Secondly, I admit to really not getting it. This seems to co-exist on the same level as the "description" field. And the code here appears to do a lot of reinventing the wheel with the pre & post processors, which I imagine is very close to how the "description" field is generated. Why would we not use the "description" field?

I'll confess that I don't like duplicating the description; it creates two places to update and maintain. The reason for it is because I was told there was an explicit requirement to have the `x-ai-description` populated for all of our endpoints for tool descriptions. 

When I ran the default descriptions, from the docstrings, through Claude for the highlighted MCP tools, it repeatedly indicated that the tool descriptions needed improvement. Not just a docstring improvement, but minor variations for each operation. For instance, I believe the TeamViewSet creates the following tools:
- gateway.teams_list
- gateway.teams_retrieve
- gateway.teams_create
- gateway.teams_delete
etc. So the purpose of this is to create a unique string, specifically targeted at the MCP tool description, for every operation. 

Given the above requirement, my focus was attempting to create a decent tool description that could be utilized for the obvious CRUD operations, and then override it where needed. At the end of this, I'm expecting we'll have two separate strings (that may come from the same docstring if `resource_purpose` isn't defined):
1. The docstring which is still visible in the generated and web spec, which would be focused on human users.
2. The `x-ai-description` in the generated spec which should get pulled into the MCP tool description, with a focus on being used by AI tooling.

And with all of that said - I'm fully open to try and find a better way to accomplish this goal; I was working to find a generic way to generate these unique descriptions instead of explicitly overriding every operation with large `extend_schema_view` decorators that also duplicated the docstring. :/

john-westcott-iv (approved):

BrennanPaciorek (approved):

mabashian:
@AlanCoding re: `description` v. `x-ai-description` - I think I'm following you and I agree that in many cases they will be very similar if not identical.  In the cases where we truly want something custom for the AI use case or where the description is just too long and will eat up too much of the LLM's context then we need a way to provide a more concise string that can be used instead of the description.  I think we definitely need the ability to provide a `x-ai-description` in those scenarios. 

Now, if we want to debate whether `x-ai-description` is needed on EVERY endpoint then I think that's valid and we can definitely have that argument.  In instances where the description is sufficient perhaps we don't include `x-ai-description` but we definitely need the ability to add it quickly should we need to in order to improve the UX of an agentic workflow.  On the mcp server side of things we can take a hierarchical approach (ie use `x-ai-description` as long as it's there and if it's not there then use the `description`).  We're getting a bit late in the game though so whatever is decided needs to be decided quickly.

cc @jameswnl in case you have additional thoughts

jay-steurer (approved):

AlanCoding:
Testing application to a service at

https://github.com/ansible/awx/pull/16174

AlanCoding:
> Now, if we want to debate whether x-ai-description is needed on EVERY endpoint 

Operations (such as "create an inventory"), not endpoint.

It's a very bad plan to have every app write their own code to auto-generate the x-ai-description on all operations. But having it be auto-applied for all actions (important, that it's only the action level) using the common framework is a reasonable policy. I think that's what we've gotten here. It's close to what you were saying, but there are some really important points that were getting interpreted wrong.

AlanCoding (approved):
There's a lot of lines of code for producing the schema, and I don't claim to review that actual logic. As as am concerned / involved, the AWX patch experimentally confirms that this _only_ adds the new field at the _action_ level. I'm less concerned about the specific content, and more about the maintenance strategy of it, which this gets clear.

For manual writing, I got:
 - `resource_purpose` property on view / viewset
 - `@extend_schema_if_available(extensions={'x-ai-description'` for action

So this effectively establishes the _contract_ for adding manual overrides for AI-specific text.

PR #885, opened 2025-11-20
AAP-56795 feat: Fix AuthenticatorMap OpenAPI schema by making triggers field required

tyraziel:
# AAP-56795

## Description
  Update base AuthenticatorMapSerializer to explicitly define triggers as a
  required JSONField, which generates accurate OpenAPI schemas automatically
  while preserving existing PATCH behavior through validate_trigger_data().

  Changes:
  - Add explicit triggers = JSONField(required=True) to AuthenticatorMapSerializer
  - Remove @extend_schema_if_available decorator (no longer needed)
  - Existing validate_trigger_data() method still allows optional triggers for PATCH

  This approach is simpler than specialized serializers and follows DRF patterns
  where field-level requirements control schema generation but custom validation
  can override behavior for specific HTTP methods.

  Co-Authored-By: Claude <noreply@anthropic.com>
  Vibe-Coder: Andrew Potozniak <potozniak@redhat.com>
  AIA: Entirely AI, Human-initiated, Reviewed, Claude Code [Sonnet 4] v1.0
  https://aiattribution.github.io/statements/AIA-EAI-Hin-R-?model=Claude%20Code%20%5BSonnet%204%5D-v1.0

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

## Testing Instructions
TBD

### Steps to Test
1. 
2. 
3. 

### Required Actions

- [X] Requires backport to 2.6

### DIFF
--- a/ansible_base/authentication/serializers/authenticator_map.py
+++ b/ansible_base/authentication/serializers/authenticator_map.py
@@ -8,6 +8,7 @@ from ansible_base.authentication.models import AuthenticatorMap
 from ansible_base.authentication.utils.authenticator_map import _EXPANSION_FIELDS, check_expansion_syntax, check_role_type, has_expansion
 from ansible_base.authentication.utils.trigger_definition import TRIGGER_DEFINITION
 from ansible_base.lib.serializers.common import NamedCommonModelSerializer
+from ansible_base.lib.serializers.fields import JSONField
 from ansible_base.lib.utils.string import is_empty
 from ansible_base.lib.utils.typing import TranslatedString
 
@@ -15,6 +16,14 @@ logger = logging.getLogger('ansible_base.authentication.serializers.authenticato
 
 
 class AuthenticatorMapSerializer(NamedCommonModelSerializer):
+    triggers = JSONField(
+        required=True,
+        allow_null=False,
+        error_messages={'required': 'Triggers must be a valid dict'},
+        help_text="Required. Trigger conditions dictionary that determines when this map applies. "
+        "See /trigger_definition/ for structure details. Only one top-level key per request.",
+    )
+
     class Meta:
         model = AuthenticatorMap
         fields = NamedCommonModelSerializer.Meta.fields + ['authenticator', 'map_type', 'role', 'organization', 'team', 'revoke', 'triggers', 'order']


huffmanca (approved):
Assuming that tests pass, this looks good to me (even though my approval in DAB means nothing).

AlanCoding (changes requested):
Delete `validate_trigger_data` and then this should be complete.

I expect the test case should update its expectation as well

https://github.com/ansible/django-ansible-base/blob/43982875957bf65ab9df99798387892d049a9450/test_app/tests/authentication/views/test_authenticator_map.py#L190

bhavenst:
> Delete `validate_trigger_data` and then this should be complete.
> 
> I expect the test case should update its expectation as well
> 
> https://github.com/ansible/django-ansible-base/blob/43982875957bf65ab9df99798387892d049a9450/test_app/tests/authentication/views/test_authenticator_map.py#L190

I don't think I agree with that.  Won't this just validate the presence of a dict, but not all the syntactical content covered by _validate_trigger_data(...)?  I would vote for removing the portion checking the presence, but not the portion validating the structure/content.

tyraziel:
> > Delete `validate_trigger_data` and then this should be complete.
> > I expect the test case should update its expectation as well
> > https://github.com/ansible/django-ansible-base/blob/43982875957bf65ab9df99798387892d049a9450/test_app/tests/authentication/views/test_authenticator_map.py#L190
> 
> I don't think I agree with that. Won't this just validate the presence of a dict, but not all the syntactical content covered by _validate_trigger_data(...)? I would vote for removing the portion checking the presence, but not the portion validating the structure/content.

This here suggests that trigger data isn't needed for PATCH, but then we don't go and validate it in _validate_trigger_data in that case.

```
    def validate_trigger_data(self, data):
        errors = {}
        request = self.context.get('request', None)
        if 'triggers' not in data or not data['triggers']:
            if not request or (request.method != 'PATCH'):
                errors["triggers"] = "Triggers must be a valid dict"
        else:
            errors.update(self._validate_trigger_data(data['triggers'], TRIGGER_DEFINITION, 'triggers'))
        return errors
```

Is there any compelling reason to not just leave this as is?

AlanCoding:
Oh no you're right, `_validate_trigger_data` absolutely needs to be called. My comment was more that the PATCH-specific condition, null-validation is duplicated with the `allow_null` added here.

bhavenst (approved):

AlanCoding (approved):

PR #893, opened 2025-12-02
AAP-53741 Log when OAuth2 tokens are created, modified, used

BrennanPaciorek:
This change logs the creation and usage of OAuth2 tokens, logging the OAuth2
application used where applicable.

Co-authored-by: Claude <noreply@anthropic.com>

## Description
- What is being changed? Create logs where OAuth2 tokens are created and used, associating requests with OAuth2 applications where possible.
- Why is this change needed? To make auditability easier for users.
- How does this change address the issue? by adding logs when creating tokens, updating logs when users use tokens.

## Type of Change

- [ ] Documentation update
- [X] Test update
- [ ] Configuration change

## Self-Review Checklist

  - This one will need to be evaluated for log injection issues, I have yet to do that one myself.

### Prerequisites
- Start test_app
- Create OAuth2 application

### Steps to Test
1. Create an OAuth2 token (both PAT and Application tokens)
2. Use the tokens to authenticate to endpoints in the application

### Expected Results

- On Creation of the token a log should be present noting the user the token was created for and the application it was created for.
- On usage of token, a log should be present noting the user the token was used for, the application it was used for, and the uri it was used to access

---

> [!NOTE]
> Adds audit logs for OAuth2 token creation/modification/use and limits activity stream updates to only the fields explicitly updated.
> 
> - **OAuth2**:
>   - **Auth Logging**: Log authenticated API requests with OAuth2, including user, method, path, token, and application id/name in `oauth2_provider/authentication.py`.
>   - **Token Models**: In `OAuth2AccessToken` and `OAuth2RefreshToken`, log on create/modify (skip trivial changes), hash tokens on save, and exclude `last_used/modified/modified_by` from activity stream.
> - **Models/Utils**:
>   - **Change Detection**: Add `ModifiableModel._has_non_trivial_changes()` and `trivial_fields` to detect meaningful updates (supports selective logging).
> - **Activity Stream**:
>   - **Selective Diff**: `_store_activitystream_entry()` accepts `update_fields` and diffs only the intersection with `activity_stream_limit_field_names`; `pre_save` passes `update_fields`.
> - **Tests**:
>   - Add extensive tests for selective activity stream diffs, OAuth2 auth not creating entries on read, and token create/modify logging (including `update_fields` cases).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 62abd6dd48a5773659a13fe6794da58c066de051. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ansible_base/activitystream/models/entry.py
+++ b/ansible_base/activitystream/models/entry.py
@@ -112,48 +112,6 @@ class AuditableModel(models.Model):
     # Adding field names to this list will limit the activity stream changes dictionaries to only include these fields
     activity_stream_limit_field_names = []
 
-    def _has_non_trivial_changes(self, update_fields=None):
-        """Check if any non-timestamp fields have changed.
-
-        Args:
-            update_fields: Optional list/set of field names being updated. If provided,
-                          only these fields will be checked for changes.
-
-        Returns:
-            bool: True if any non-timestamp field has changed, False otherwise.
-        """
-        if not self.pk:
-            return True
-
-        Model = self.__class__
-        try:
-            old_instance = Model.objects.get(pk=self.pk)
-        except Model.DoesNotExist:
-            pass
-            # Fields to exclude from change detection (timestamp/auto-updated fields)
-
-        # If update_fields is specified, only check those fields (minus excluded ones)
-        if update_fields is not None:
-            fields_to_check = set(update_fields) - set(self.activity_stream_excluded_field_names)
-            if not fields_to_check:
-                # Only timestamp fields are being updated
-                return False
-            # Check only the specified fields
-            for field_name in fields_to_check:
-                old_value = getattr(old_instance, field_name, None)
-                new_value = getattr(self, field_name, None)
-                if old_value != new_value:
-                    return True
-        else:
-            # Check all non-timestamp fields
-            for field in self._meta.get_fields():
-                if field.concrete and hasattr(field, 'name') and field.name not in self.activity_stream_excluded_field_names:
-                    old_value = getattr(old_instance, field.name, None)
-                    new_value = getattr(self, field.name, None)
-                    if old_value != new_value:
-                        return True
-        return False
-
     @property
     def activity_stream_entries(self):
         """
--- a/ansible_base/activitystream/models/entry.py
+++ b/ansible_base/activitystream/models/entry.py
@@ -112,6 +112,48 @@ class AuditableModel(models.Model):
     # Adding field names to this list will limit the activity stream changes dictionaries to only include these fields
     activity_stream_limit_field_names = []
 
+    def _has_non_trivial_changes(self, update_fields=None):
+        """Check if any non-timestamp fields have changed.
+
+        Args:
+            update_fields: Optional list/set of field names being updated. If provided,
+                          only these fields will be checked for changes.
+
+        Returns:
+            bool: True if any non-timestamp field has changed, False otherwise.
+        """
+        if not self.pk:
+            return True
+
+        Model = self.__class__
+        try:
+            old_instance = Model.objects.get(pk=self.pk)
+        except Model.DoesNotExist:
+            pass
+            # Fields to exclude from change detection (timestamp/auto-updated fields)
+
+        # If update_fields is specified, only check those fields (minus excluded ones)
+        if update_fields is not None:
+            fields_to_check = set(update_fields) - set(self.activity_stream_excluded_field_names)
+            if not fields_to_check:
+                # Only timestamp fields are being updated
+                return False
+            # Check only the specified fields
+            for field_name in fields_to_check:
+                old_value = getattr(old_instance, field_name, None)
+                new_value = getattr(self, field_name, None)
+                if old_value != new_value:
+                    return True
+        else:
+            # Check all non-timestamp fields
+            for field in self._meta.get_fields():
+                if field.concrete and hasattr(field, 'name') and field.name not in self.activity_stream_excluded_field_names:
+                    old_value = getattr(old_instance, field.name, None)
+                    new_value = getattr(self, field.name, None)
+                    if old_value != new_value:
+                        return True
+        return False
+
     @property
     def activity_stream_entries(self):
         """
--- a/ansible_base/lib/abstract_models/common.py
+++ b/ansible_base/lib/abstract_models/common.py
@@ -41,6 +41,50 @@ class ModifiableModel(models.Model):
     class Meta:
         abstract = True
 
+    trivial_fields = ["modified", "modified_by"]
+
+    def _has_non_trivial_changes(self, update_fields=None):
+        """Check if any non-timestamp fields have changed.
+
+        Args:
+            update_fields: Optional list/set of field names being updated. If provided,
+                          only these fields will be checked for changes.
+
+        Returns:
+            bool: True if any non-timestamp field has changed, False otherwise.
+        """
+        if not self.pk:
+            return True
+
+        Model = self.__class__
+        try:
+            old_instance = Model.objects.get(pk=self.pk)
+        except Model.DoesNotExist:
+            pass
+            # Fields to exclude from change detection (timestamp/auto-updated fields)
+
+        # If update_fields is specified, only check those fields (minus excluded ones)
+        if update_fields is not None:
+            fields_to_check = set(update_fields) - set(self.trivial_fields)
+            if not fields_to_check:
+                # Only timestamp fields are being updated
+                return False
+            # Check only the specified fields
+            for field_name in fields_to_check:
+                old_value = getattr(old_instance, field_name, None)
+                new_value = getattr(self, field_name, None)
+                if old_value != new_value:
+                    return True
+        else:
+            # Check all non-timestamp fields
+            for field in self._meta.get_fields():
+                if field.concrete and hasattr(field, 'name') and field.name not in self.trivial_fields:
+                    old_value = getattr(old_instance, field.name, None)
+                    new_value = getattr(self, field.name, None)
+                    if old_value != new_value:
+                        return True
+        return False
+
     modified = models.DateTimeField(
         editable=False,
         help_text=_("The date/time this resource was created."),
--- a/ansible_base/lib/abstract_models/common.py
+++ b/ansible_base/lib/abstract_models/common.py
@@ -56,33 +56,29 @@ class ModifiableModel(models.Model):
         if not self.pk:
             return True
 
-        Model = self.__class__
+        model_cls = self.__class__
         try:
-            old_instance = Model.objects.get(pk=self.pk)
-        except Model.DoesNotExist:
-            pass
-            # Fields to exclude from change detection (timestamp/auto-updated fields)
+            old_instance = model_cls.objects.get(pk=self.pk)
+        except model_cls.DoesNotExist:
+            return True
 
         # If update_fields is specified, only check those fields (minus excluded ones)
         if update_fields is not None:
-            fields_to_check = set(update_fields) - set(self.trivial_fields)
-            if not fields_to_check:
-                # Only timestamp fields are being updated
-                return False
-            # Check only the specified fields
-            for field_name in fields_to_check:
-                old_value = getattr(old_instance, field_name, None)
-                new_value = getattr(self, field_name, None)
-                if old_value != new_value:
-                    return True
+            fields_we_can_check = set(update_fields)
         else:
-            # Check all non-timestamp fields
-            for field in self._meta.get_fields():
-                if field.concrete and hasattr(field, 'name') and field.name not in self.trivial_fields:
-                    old_value = getattr(old_instance, field.name, None)
-                    new_value = getattr(self, field.name, None)
-                    if old_value != new_value:
-                        return True
+            fields_we_can_check = [field.name for field in self._meta.get_fields() if field.concrete and hasattr(field, "name")]
+
+        fields_to_check = set(fields_we_can_check) - set(self.trivial_fields)
+
+        if not fields_to_check:
+            # Only timestamp fields are being updated
+            return False
+        # Check only the specified fields
+        for field_name in fields_to_check:
+            old_value = getattr(old_instance, field_name, None)
+            new_value = getattr(self, field_name, None)
+            if old_value != new_value:
+                return True
         return False
 
     modified = models.DateTimeField(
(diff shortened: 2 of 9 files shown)


john-westcott-iv (approved):
Looks like you already address the Cursor concerns.

john-westcott-iv (commented):
## 📋 Code Review Summary

This PR adds audit logging for OAuth2 token creation and usage, logging the application context where applicable. The changes enhance auditability of OAuth2 operations.

**Files Reviewed**: 4 files
**Comments Posted**: 3 review comments

### 🔍 Issues Found
- 0 security/safety issues
- 1 correctness/logic issue (parameter mismatch concern)
- 2 code quality suggestions

### Overall Assessment
LGTM with minor suggestions. The implementation is solid, but there's an inconsistency in logging approaches between the authentication module and the model modules that should be considered.

### General Feedback
- Good use of `log_auth_event` for audit logging in the model save methods
- Tests provide good coverage of the new logging functionality
- The authentication.py module uses direct logger calls while models use `log_auth_event` - consider unifying the approach for consistency

ansible_base/oauth2_provider/authentication.py
```diff
@@ -42,8 +42,14 @@ def authenticate(self, request):
         if ret:
             user, token = ret
             username = user.username if user else '<none>'
+            oauth2_application_pk = token.application.pk if token.application else "N/A"
+            oauth2_application_name = token.application.name if token.application else "Personal Access Token"
             logger.info(
-                smart_str(u"User {} performed a {} to {} through the API using OAuth 2 token {}.".format(username, request.method, request.path, token.pk))
+                smart_str(
+                    u"User {} performed a {} to {} through the API using OAuth 2 token {} for OAuth2 application {} ({}).".format(
```
john-westcott-iv, line 50: Consider using `log_auth_event` here for consistency with the access_token.py and refresh_token.py changes. This would ensure all OAuth2-related events are logged to the audit logger.

Note: This would require importing `log_auth_event` from `ansible_base.lib.logging`.

ansible_base/oauth2_provider/models/access_token.py
```diff
@@ -97,3 +101,6 @@ def save(self, *args, **kwargs):
             self.validate_external_users()
             self.token = hash_string(self.token, hasher=hashlib.sha256, algo="sha256")
         super().save(*args, **kwargs)
+        app_name = self.application.name if self.application else "N/A (Personal Access Token)"
+        user_name = self.user.username if self.user else "N/A"
+        log_auth_event(f"Creating OAuth2 access token for user '{user_name}' with application '{app_name}' and scope '{self.scope}'", second_logger=logger)
```
john-westcott-iv: The `log_auth_event` function signature in `ansible_base/lib/logging/__init__.py` is:

```python
def log_auth_event(message: str, level: Optional[int] = logging.INFO, log_to_callers_logger: Optional[bool] = True):
```

It uses `_get_caller_logger()` to automatically find the caller's logger - it doesn't have a `second_logger` parameter. The function should find the module's `logger` variable automatically via stack inspection.

If CI tests pass, this works as expected due to the automatic logger discovery, but the explicit `second_logger=logger` parameter appears unnecessary and would cause a TypeError if the function signature is as I see in the codebase.

test_app/tests/oauth2_provider/test_models.py
```diff
@@ -32,3 +36,79 @@ def test_oauth2_revoke_refresh_token(oauth2_admin_access_token):
...
+    # Create an access token with an application
+    token = OAuth2AccessToken.objects.create(
+        user=admin_user,
+        application=application,
+        token=generate_token(),
+        scope='write',
+        expires=datetime(2088, 1, 1, tzinfo=timezone.utc),
+    )
+
+    # Verify the logger was called with the correct message
```
john-westcott-iv, line 57: The test mocks the module logger and expects `.info()` to be called, but `log_auth_event` uses `.log(level, message)` not `.info(message)`. These are different method signatures.

If this test passes, either:
1. The `log_auth_event` implementation differs from what I see in the codebase, or
2. The test is checking the wrong method

Consider verifying this behavior matches the actual `log_auth_event` implementation.

ansible_base/oauth2_provider/models/access_token.py
john-westcott-iv: This is inaccurate, my AI was looking at a branch.

test_app/tests/oauth2_provider/test_models.py
john-westcott-iv: This is inaccurate, my AI was looking at a branch.

BrennanPaciorek:
Good to know I actually have zero coverage.

ansible_base/oauth2_provider/models/access_token.py
```diff
@@ -83,6 +87,47 @@ def _update_last_used():
...
+
+        Args:
+            update_fields: Optional list/set of field names being updated. If provided,
+                          only these fields will be checked for changes.
+
+        Returns:
+            bool: True if any non-timestamp field has changed, False otherwise.
+        """
+        if not self.pk:
+            return False
```
john-westcott-iv: Couldn't we just make this return True since we are creating and not updating? (See comment further down)

ansible_base/oauth2_provider/models/access_token.py
```diff
@@ -83,6 +87,47 @@ def _update_last_used():
...
+            else:
+                # Check all non-timestamp fields
+                for field in self._meta.get_fields():
+                    if field.concrete and hasattr(field, 'name') and field.name not in exclude_fields:
+                        old_value = getattr(old_instance, field.name, None)
+                        new_value = getattr(self, field.name, None)
+                        if old_value != new_value:
+                            return True
+        except OAuth2AccessToken.DoesNotExist:
+            pass
```
john-westcott-iv: This could only happen if another thread was deleting the token at the same time this token was being saved right?

ansible_base/oauth2_provider/models/access_token.py
```diff
@@ -83,6 +87,47 @@ def _update_last_used():
             connection.on_commit(_update_last_used)
         return valid
 
+    def _has_non_timestamp_changes(self, update_fields=None):
```
john-westcott-iv: What would you think about doing a base change to models with a function like `has_non_trivial_change(self, update_fields := ) -> bool` and instead of having:
```
exclude_fields = {'created', 'created_by', 'modified', 'modified_by', 'last_used'}
```

We could do something like `exclude_fields =  getattr(self, 'trivial_fields', [])`.

This would enable the pattern for all models and then we could make decisions for things like auditing, logging or anything else based on this.

ansible_base/oauth2_provider/models/refresh_token.py
```diff
@@ -25,7 +29,58 @@ class Meta(oauth2_models.AbstractRefreshToken.Meta):
...
+
+        if create_token:
             self.token = hash_string(self.token, hasher=hashlib.sha256, algo="sha256")
         super().save(*args, **kwargs)
+        access_token_id = self.access_token.pk if hasattr(self, 'access_token') and self.access_token else "N/A"
+        user_name = self.user.username if self.user else "N/A"
+        if create_token:
+            log_auth_event(f"Created OAuth2 refresh token {self.pk} for user '{user_name}' linked to access token {access_token_id}", second_logger=logger)
+        elif modifying_token:
+            log_auth_event(f"Modified OAuth2 refresh token {self.pk} for user '{user_name}' linked to access token {access_token_id}", second_logger=logger)
```
john-westcott-iv: If we make `_has_non_timestamp_changes` return true `if not self.pk` can simplify this to:

```suggestion
        update_fields = kwargs.get('update_fields')
        has_non_trivial_fields = self._has_non_timestamp_changes(update_fields)

        logging_verb = 'Modified'
        if not self.pk:
            self.token = hash_string(self.token, hasher=hashlib.sha256, algo="sha256")
            logging_verb = 'Created'
        super().save(*args, **kwargs)
        access_token_id = self.access_token.pk if hasattr(self, 'access_token') and self.access_token else "N/A"
        user_name = self.user.username if self.user else "N/A"
        if has_non_trivial_fields:
            log_auth_event(f"{logging_verb} OAuth2 refresh token {self.pk} for user '{user_name}' linked to access token {access_token_id}", second_logger=logger)
```

ansible_base/oauth2_provider/models/refresh_token.py
```diff
@@ -25,7 +29,58 @@ class Meta(oauth2_models.AbstractRefreshToken.Meta):
     token = prevent_search(models.CharField(max_length=255, help_text=_("The refresh token value.")))
     updated = None  # Tracked in CommonModel with 'modified', no need for this
 
-    def save(self, *args, **kwargs):
+    def _has_non_timestamp_changes(self, update_fields=None):
```
john-westcott-iv: I've not gone line by line but I'm assuming this function is a complete dupe. Before I mentioned making this part of one of the super classes (like CommonModel). But if we don't want to tackle that at this time, can we at least make a new super class for "tokens"? That can inherit from CommonModel and have this method in there. Then both of these classes can inherit from that model instead of CommonModel to at least reduce this duplication in here.

ansible_base/oauth2_provider/models/access_token.py
```diff
@@ -93,7 +138,21 @@ def validate_external_users(self):
                 )
 
     def save(self, *args, **kwargs):
-        if not self.pk:
+        creating_token = not bool(self.pk)
+        update_fields = kwargs.get('update_fields')
+        modifying_token = self._has_non_timestamp_changes(update_fields) if not creating_token else False
```
john-westcott-iv: See ways to optimize this down below.

ansible_base/oauth2_provider/models/access_token.py
john-westcott-iv: I think if we move the except up to right after the try and have it return false it would reduce the cognitive complexity of this function as I think the indent adds a point modifier.

ansible_base/oauth2_provider/models/access_token.py
BrennanPaciorek: Testing a solution as part of AuditableModel, but it is not my final iteration (since AuditableModel is only conditionally present.

I'll move the solution to work with any common model as described, then set `trivial_fields` as equal to `activity_stream_excluded_fields`

ansible_base/oauth2_provider/models/access_token.py
```diff
@@ -93,7 +97,18 @@ def validate_external_users(self):
                 )
 
     def save(self, *args, **kwargs):
+        update_fields = kwargs.get('update_fields')
+        has_non_trivial_fields = self._has_non_trivial_changes(update_fields)
```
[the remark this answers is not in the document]
BrennanPaciorek: (I know)

ansible_base/activitystream/models/entry.py
```diff
@@ -112,6 +112,48 @@ class Meta:
...
+        """
+        if not self.pk:
+            return True
+
+        Model = self.__class__
+        try:
+            old_instance = Model.objects.get(pk=self.pk)
+        except Model.DoesNotExist:
+            pass
+            # Fields to exclude from change detection (timestamp/auto-updated fields)
```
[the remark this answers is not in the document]
BrennanPaciorek: I'll change this to just return True.

ansible_base/oauth2_provider/authentication.py
```diff
@@ -42,8 +43,15 @@ def authenticate(self, request):
...
-                smart_str(u"User {} performed a {} to {} through the API using OAuth 2 token {}.".format(username, request.method, request.path, token.pk))
+            oauth2_application_pk = token.application.pk if token.application else "N/A"
+            oauth2_application_name = token.application.name if token.application else "Personal Access Token"
+            log_auth_event(
+                smart_str(
+                    u"User {} performed a {} to {} through the API using OAuth 2 token {} for OAuth2 application {} ({}).".format(
+                        username, request.method, request.path, token.pk, oauth2_application_pk, oauth2_application_name
+                    )
+                ),
+                logger,
```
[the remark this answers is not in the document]
BrennanPaciorek: Not sure this is an actual issue. I don't think you can change this to do anything more than make it a little harder to read.
```
2025-12-08 17:47:50,614 INFO     id:1f6886da-299d-4724-b4b3-ea6d6f1f1815 ansible_base.oauth2_provider.authentication User 'admin' performed a GET to '/api/v1/authenticators/' through the API using OAuth 2 token 6 for OAuth2 application 4 ()"Test Potential \n Log Injection \n"().
```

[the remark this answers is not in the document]
BrennanPaciorek: request.path should be sanitized long before it reaches this log call as well.

ansible_base/lib/abstract_models/common.py
```diff
@@ -41,6 +41,46 @@ class ModifiableModel(models.Model):
...
+        # If update_fields is specified, only check those fields (minus excluded ones)
+        if update_fields is not None:
+            fields_we_can_check = set(update_fields)
+        else:
+            fields_we_can_check = [field.name for field in self._meta.get_fields() if field.concrete and hasattr(field, "name")]
+
+        fields_to_check = set(fields_we_can_check) - set(self.trivial_fields)
+
+        if not fields_to_check:
+            # Only timestamp fields are being updated
```
john-westcott-iv:
```suggestion
            # Only trivial fields are being updated
```

ansible_base/lib/abstract_models/common.py
```diff
@@ -41,6 +41,46 @@ class ModifiableModel(models.Model):
...
+        try:
+            old_instance = model_cls.objects.get(pk=self.pk)
+        except model_cls.DoesNotExist:
+            return True
+
+        # If update_fields is specified, only check those fields (minus excluded ones)
+        if update_fields is not None:
+            fields_we_can_check = set(update_fields)
+        else:
+            fields_we_can_check = [field.name for field in self._meta.get_fields() if field.concrete and hasattr(field, "name")]
```
john-westcott-iv: if you only care about concrete fields there is `<model>._meta.concrete_fields`

```suggestion
            fields_we_can_check = [field.name for field in self._meta.concrete_fields if hasattr(field, "name")]
```

john-westcott-iv (approved):

PR #903, opened 2025-12-11
[AAP-60068] Enforce OAuth2 token scope on HTTP methods

john-westcott-iv:
## Description
- **What is being changed?** Added scope validation to OAuth2 authentication layer to enforce that tokens with 'read' scope can only perform safe HTTP methods (GET, HEAD, OPTIONS). Tokens with 'write' scope can perform any method.
- **Why is this change needed?** Previously, tokens with 'read' scope could be used to perform write operations (POST, PUT, PATCH, DELETE), which is a security concern and doesn't honor the intended scope restrictions.
- **How does this change address the issue?** The `LoggedOAuth2Authentication.authenticate()` method now checks the token scope against the request method and raises `AuthenticationFailed` (401) if the scope doesn't permit the method.

## Type of Change

## Self-Review Checklist

### Prerequisites
- A running gateway instance with OAuth2 tokens configured

### Steps to Test
1. Create an OAuth2 token with 'read' scope
2. Attempt a GET request with the token - should succeed (200)
3. Attempt a POST request with the token - should fail (401) with message about scope
4. Create an OAuth2 token with 'write' scope
5. Attempt POST/PUT/PATCH/DELETE requests - should all succeed

### Expected Results
- Read tokens: GET/HEAD/OPTIONS succeed, POST/PUT/PATCH/DELETE return 401
- Write tokens: All methods succeed
- Error message clearly explains scope restriction
- Denied requests logged at WARNING level with 'attempted' terminology
- Successful requests logged at INFO level with 'performed' terminology

### Required Actions
  - AAP Gateway may need to verify this works correctly in the proxy flow to backend services

---

### DIFF
--- a/ansible_base/oauth2_provider/authentication.py
+++ b/ansible_base/oauth2_provider/authentication.py
@@ -4,9 +4,10 @@ import logging
 from django.utils.encoding import smart_str
 from oauth2_provider.contrib.rest_framework import OAuth2Authentication
 from oauth2_provider.oauth2_backends import OAuthLibCore as _OAuthLibCore
-from rest_framework.exceptions import UnsupportedMediaType
+from rest_framework.exceptions import PermissionDenied, UnsupportedMediaType
+from rest_framework.permissions import SAFE_METHODS
 
-from ansible_base.lib.logging import log_auth_event
+from ansible_base.lib.logging import log_auth_event, log_auth_warning
 from ansible_base.lib.utils.hashing import hash_string
 
 logger = logging.getLogger('ansible_base.oauth2_provider.authentication')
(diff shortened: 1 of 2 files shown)


ansible_base/oauth2_provider/authentication.py
```diff
@@ -45,6 +46,27 @@ def authenticate(self, request):
             username = user.username if user else '<none>'
             oauth2_application_pk = token.application.pk if token.application else "N/A"
             oauth2_application_name = token.application.name if token.application else "Personal Access Token"
+
+            # TODO: check oauth_scopes when we have RBAC in Gateway
+            setattr(user, 'oauth_scopes', [x for x in token.scope.split() if x])
+
+            # Check if token scope permits the requested method
+            if 'write' not in token.scope and request.method.upper() not in SAFE_METHODS:
+                log_auth_warning(
```
BrennanPaciorek, line 55: If we're trying to backport this right away, we don't want to include `log_auth_<level>", as that won't be in the stable branches by the time this fix gets delivered.

Its able to be worked around, but another reason why working this fix in the authentication class, instead of where the issue presents, is not my favorite solution.

ansible_base/oauth2_provider/authentication.py
```diff
@@ -45,6 +46,27 @@ def authenticate(self, request):
             username = user.username if user else '<none>'
             oauth2_application_pk = token.application.pk if token.application else "N/A"
             oauth2_application_name = token.application.name if token.application else "Personal Access Token"
+
+            # TODO: check oauth_scopes when we have RBAC in Gateway
```
AlanCoding, line 50:
```suggestion
```

AlanCoding (approved):

ansible_base/oauth2_provider/authentication.py
AlanCoding: my feeling is that someone can just deal with the git conflict in the backport

john-westcott-iv:
Closing in favor of a different PR.

PR #907, opened 2025-12-16
[AAP-66801] Fix race condition in OAuth2AccessToken.is_valid() causing HTTP 500

DenisKoleda:
## Summary

- Fix race condition when concurrent requests use the same OAuth2 token
- Replace `save(update_fields=['last_used'])` with `QuerySet.update()` to prevent `DatabaseError`

## Problem

When multiple parallel requests authenticate with the same OAuth2 token, the Gateway crashes with:

```
django.db.utils.DatabaseError: Save with update_fields did not affect any rows.
```

**Root cause:** In `is_valid()` method, there's a race condition between `exists()` check and `save()` call. When concurrent requests try to update `last_used` timestamp on the same token:

1. Request A checks `exists()` → True
2. Request B checks `exists()` → True  
3. Request B calls `save()` → Success, row version changes
4. Request A calls `save()` → **Fails** because Django's `save(update_fields=...)` expects to affect exactly 1 row, but the row was already modified

## Solution

Replace:
```python
if OAuth2AccessToken.objects.filter(pk=self.pk).exists():
    self.save(update_fields=['last_used'])
```

With:
```python
OAuth2AccessToken.objects.filter(pk=self.pk).update(last_used=self.last_used)
```

`QuerySet.update()` is atomic and simply returns 0 if no rows match, without raising an exception. This is the expected behavior for updating `last_used` — if the token was deleted between validation and update, we don't need to error out.

## Test plan

- [x] Verify existing OAuth2 authentication tests pass
- [x] Manual test with concurrent requests using the same token
- [x] Confirm no HTTP 500 errors under parallel load

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token tracking reliability by refactoring how access token usage timestamps are updated, reducing potential race conditions during concurrent requests.


AlanCoding:
Soft approval. This is the type of solution I would expect for this type of problem.

Still want someone to look at it who was more involved with the oauth2 app development to weigh in. The existing save appears to me to be avoid any other model logic, and if assuming that's the case, this should be good.

ansible_base/oauth2_provider/models/access_token.py
```diff
@@ -82,8 +82,7 @@ def is_valid(self, scopes=None):
             self.last_used = now()
 
             def _update_last_used():
-                if OAuth2AccessToken.objects.filter(pk=self.pk).exists():
-                    self.save(update_fields=['last_used'])
+                OAuth2AccessToken.objects.filter(pk=self.pk).update(last_used=self.last_used)
```
john-westcott-iv, line 91: Nice fix! Small suggestion — let's leave a breadcrumb for future readers so nobody "fixes" this back to `save()`:

```suggestion
                # QuerySet.update() is used instead of save() to avoid a race condition
                # where concurrent requests updating last_used can cause DatabaseError.
                # This is safe because last_used is a trivial field with no save() side effects.
                OAuth2AccessToken.objects.filter(pk=self.pk).update(last_used=self.last_used)
```

john-westcott-iv (commented):
Clean fix for a real race condition — `QuerySet.update()` is the right tool here. Left one suggestion to add a comment explaining why `update()` is used instead of `save()`, so future developers don't inadvertently revert this.

BrennanPaciorek (approved):
This should not cause any issues I do not think.

BrennanPaciorek (commented):
We'll want to check how activitystream handles updates to last used in this case. I think it'll resolve cleanly, but I'm uncertain.

john-westcott-iv:
Thanks for this contribution and sorry for this PR getting lost in the shuffle. This should be scheduled to be worked on starting next Thursday and should merge shortly after that.

AlanCoding (approved):

PR #909, opened 2025-12-23
fix: Handle race condition in RoleDefinition.get_or_create

B-Whitt:
## Description
When multiple threads simultaneously call give_creator_permissions, they can race to create the same RoleDefinition, causing IntegrityError.

Solution avoids using transaction.atomic() savepoints or database locks, as requested by DAB RBAC senior engineering.

Jira Bug: [AAP-60238](https://issues.redhat.com/browse/AAP-60238) Error when creating EDA project by multiple users

Assisted by: Claude Code

## Type of Change

- [x] Test update

## Self-Review Checklist

### DIFF
--- a/ansible_base/rbac/models/role.py
+++ b/ansible_base/rbac/models/role.py
@@ -6,6 +6,7 @@ from uuid import UUID
 # Django
 from django.conf import settings
 from django.db import connection, models, transaction
+from django.db.models import Count
 from django.db.models.functions import Cast
 from django.db.models.query import QuerySet
 from django.db.utils import IntegrityError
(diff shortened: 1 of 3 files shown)


ansible_base/rbac/models/role.py
```diff
@@ -113,18 +113,39 @@ def give_creator_permissions(self, user, obj) -> Optional['RoleUserAssignment']:
...
+            existing_set = {perm.codename for perm in existing_rd.permissions.all()}
+            if existing_set == permissions:
+                return existing_rd
+        return None
+
     def get_or_create(self, permissions=(), defaults=None, **kwargs):
-        "Add extra feature on top of existing get_or_create to use permissions list"
+        """Add extra feature on top of existing get_or_create to use permissions list.
+
+        Handles race conditions where multiple threads may try to create the same
```
AlanCoding: I think this tends to be multiple separate processes rather than threads, but minor nit

ansible_base/rbac/models/role.py
AlanCoding, line 1: I follow what this is doing, and this is a reasonable change. In the abstract I agree with this. It might be more clear to have a retry count in a loop, but this will work as-is.

ansible_base/rbac/models/role.py
```diff
@@ -113,18 +113,39 @@ def give_creator_permissions(self, user, obj) -> Optional['RoleUserAssignment']:
...
             if defaults:
                 create_kwargs.update(defaults)
-            return (self.create_from_permissions(permissions=permissions, **create_kwargs), True)
+            try:
+                with transaction.atomic():
+                    return (self.create_from_permissions(permissions=permissions, **create_kwargs), True)
+            except IntegrityError:
+                # Race condition: another thread created the RoleDefinition between
+                # our check and create. Re-query to find the one that was just created.
+                existing = self._find_by_permissions(permissions)
```
AlanCoding: Add a log here for debugging please. This is out of the ordinary, and `logger.info` would be appropriate in my book, and including `self.pk` in the message would be important.

AlanCoding:
Looks good, I only asked for minor changes. The AWX downstream check is probably not from this change.

ansible_base/rbac/models/role.py
B-Whitt: Doc string updated to read 'Handles race conditions where multiple users attempt to create Projects [therefore
        multiple RoleDefinition are created] by catching IntegrityError and retrying the lookup.'

ansible_base/rbac/models/role.py
B-Whitt: added retry loop

ansible_base/rbac/models/role.py
B-Whitt: I added logger.info. I was not able to add 'self.pk', since RoleDefinitionManager did not have the pk attribute. If you can show me a work around, I will gladly add it.

ansible_base/rbac/models/role.py
mkanoor: @B-Whitt Would it make more sense to change this [line](https://github.com/ansible/django-ansible-base/blob/ab013f251dc24926c3b0e71e6aa243e7ec02c941/ansible_base/rbac/models/role.py#L174) to use [update_or_create](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#update-or-create). Then we dont need to have a retry loop since they are creating the same object.

AlanCoding (approved):
Looks good thanks

B-Whitt:
updated with https://github.com/ansible/django-ansible-base/pull/910

mkanoor (approved):

AlanCoding (approved):
I still think this is much improved, even if it's not airtight. I was working on a test, and I can return to that if there's a need to go deeper.

PR #910, opened 2026-01-07
fix: Handle race condition in RoleDefinition.get_or_create

B-Whitt:
Jira Bug: [AAP-60238](https://issues.redhat.com/browse/AAP-60238) Error when creating EDA project by multiple users

## Description
Fixes a race condition in RoleDefinition.get_or_create() that caused duplicate roles with identical names when multiple threads created roles concurrently.
- Add PostgreSQL advisory lock to prevent duplicate roles
- Handle IntegrityError from concurrent name collisions

This is an update to: https://github.com/ansible/django-ansible-base/pull/909

Assisted by: Claude Code

## Type of Change

## Self-Review Checklist

### DIFF
--- a/ansible_base/rbac/models/role.py
+++ b/ansible_base/rbac/models/role.py
@@ -6,6 +6,7 @@ from uuid import UUID
 # Django
 from django.conf import settings
 from django.db import connection, models, transaction
+from django.db.models import Count
 from django.db.models.functions import Cast
 from django.db.models.query import QuerySet
 from django.db.utils import IntegrityError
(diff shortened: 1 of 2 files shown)


ansible_base/rbac/models/role.py
```diff
@@ -114,17 +115,44 @@ def give_creator_permissions(self, user, obj) -> Optional['RoleUserAssignment']:
...
-                if existing_set == permissions:
-                    return (existing_rd, False)
-            create_kwargs = kwargs.copy()
-            if defaults:
-                create_kwargs.update(defaults)
-            return (self.create_from_permissions(permissions=permissions, **create_kwargs), True)
+
+            # Generate deterministic lock ID from permission set
+            # Hash collision probability: ~1 in 2^31 (negligible)
+            lock_id = hash(frozenset(permissions)) % (2**31)
```
AlanCoding, line 128: Can you just use the `advisory_lock` already in DAB which does this exact name-to-lock_id thing?

ansible_base/rbac/models/role.py
```diff
@@ -114,17 +115,44 @@ def give_creator_permissions(self, user, obj) -> Optional['RoleUserAssignment']:
...
+
+            with transaction.atomic():
+                # PostgreSQL advisory lock - serializes concurrent checks for same permission set
+                # Falls back gracefully on non-PostgreSQL databases (lock simply not acquired)
```
AlanCoding: There are a lot of particular rules for the right way to do this in conjunction with `transaction.atomic()`. But for this I think the best strategy is to get rid of `transaction.atomic()`. Whatever is happening should probably already be in an atomic block.

B-Whitt: @AlanCoding I have removed transaction.atomic(). This does not pose a concern for EDA because, as you stated, the call will already be in an atomic block. But, I am not sure if this will cause any issues for other components.

ansible_base/rbac/models/role.py
B-Whitt: @AlanCoding Alan - I was unable to get the dab advisory_lock  to work in this scenario. Here are some of the blockers that I ran into. If you know of a way to work around these I’d love to get on a call with you to show me. Blockers… 

- different Lock Type for This Use Case
  - advisory_lock uses session-scoped locks (pg_advisory_lock)
  - Current code uses transaction-scoped locks (pg_advisory_xact_lock)
  - Session locks persist beyond transaction boundaries
- Lock Window Mismatch 
  - Current: Lock released when transaction commits (tight coupling)
  - advisory_lock: Lock released when context manager exits (could be before or after transaction)

mkanoor (approved):

ansible_base/rbac/models/role.py
```diff
@@ -114,17 +115,43 @@ def give_creator_permissions(self, user, obj) -> Optional['RoleUserAssignment']:
...
+            if connection.vendor == 'postgresql':
+                with connection.cursor() as cursor:
+                    cursor.execute('SELECT pg_advisory_xact_lock(%s)', [lock_id])
+            else:
+                logger.debug(f'Advisory lock skipped - {connection.vendor} does not support pg_advisory_xact_lock')
+
+            # Check for existing role with matching permissions
+            # Filter candidates by permission count only - do NOT filter on permissions__codename__in
+            target_count = len(permissions)
+            candidates = self.annotate(perm_count=Count('permissions')).filter(perm_count=target_count).prefetch_related('permissions')
```
AlanCoding, line 142: this seems like a clever optimization

ansible_base/rbac/models/role.py
```diff
@@ -114,17 +115,43 @@ def give_creator_permissions(self, user, obj) -> Optional['RoleUserAssignment']:
...
+            lock_id = hash(frozenset(permissions)) % (2**31)
+
+            # NOTE: Caller must wrap in transaction.atomic() for advisory lock to work.
+            # pg_advisory_xact_lock is transaction-scoped - releases when transaction ends.
+            # Without outer transaction, lock acquires and releases immediately (no protection).
+            if connection.vendor == 'postgresql':
+                with connection.cursor() as cursor:
+                    cursor.execute('SELECT pg_advisory_xact_lock(%s)', [lock_id])
+            else:
+                logger.debug(f'Advisory lock skipped - {connection.vendor} does not support pg_advisory_xact_lock')
```
AlanCoding, line 137: I don't think this is a good idea. It is a much better idea for the caller to manage transactions, and for the broader issue to be solved that way. However, still catching an IntegrityError and doing a retry is fine, because it's obviously not a problem from the perspective of the application code here.

B-Whitt:
Senior engineering preferred the solution presented in #909

PR #915, opened 2026-01-14
[AAP-43413] FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED feature flag and OIDC endpoints

PabloHiro:
## Description

* Support in DAB for a new feature flag `FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED`, disabled by default.
* View under `/o/` updated to display `/o/.well-known/jwks.json` and `/o/.well-known/openid-configuration` when the feature flag is enabled.

## Type of Change

## Self-Review Checklist

## Testing Instructions
See instructions under https://github.com/ansible-automation-platform/aap-gateway/pull/1200

### DIFF
--- a/ansible_base/feature_flags/definitions/feature_flags.yaml
+++ b/ansible_base/feature_flags/definitions/feature_flags.yaml
@@ -52,3 +52,14 @@
   support_url: ''
   labels:
     - platform
+- name: FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED
+  ui_name: OIDC Workload Identity
+  visibility: False
+  condition: boolean
+  value: 'False'
+  support_level: TECHNOLOGY_PREVIEW
+  description: Enable identity provision of workloads using OIDC
+  support_url: ''
+  toggle_type: install-time
+  labels:
+    - platform
--- a/ansible_base/feature_flags/migrations/0003_manual_20260113.py
+++ b/ansible_base/feature_flags/migrations/0003_manual_20260113.py
@@ -1,0 +1,16 @@
+###
+# Addition of FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED
+###
+
+# FileHash: 53968cb50a7e8250c60abc96733d40c927fefd4ecd6389a2a5d3b121d07216a1
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+     dependencies = [
+         ('dab_feature_flags', '0002_manual_20251222'),
+     ]
+
+     operations = [
+     ]
--- a/ansible_base/oauth2_provider/urls.py
+++ b/ansible_base/oauth2_provider/urls.py
@@ -1,4 +1,5 @@
 from django.urls import include, path, re_path
+from flags.urls import flagged_re_path
 from oauth2_provider import views as oauth_views
 
 from ansible_base.lib.routers import AssociationResourceRouter
@@ -28,9 +29,25 @@ api_version_urls = [
     path('', include(router.urls)),
 ]
 
+FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED = 'FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED'
+
+oauth_urls = [
+    re_path(r'^$', oauth2_provider_views.ApiOAuthAuthorizationRootView.as_view(), name='oauth_authorization_root_view'),
+    re_path(r"^authorize/$", oauth_views.AuthorizationView.as_view(), name="authorize"),
+    re_path(r"^token/$", oauth2_provider_views.TokenView.as_view(), name="token"),
+    re_path(r"^revoke_token/$", oauth_views.RevokeTokenView.as_view(), name="revoke-token"),
+    # OIDC endpoints - flag is checked at request time, returns 404 when disabled
+    flagged_re_path(
+        FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED,
+        r"^\.well-known/openid-configuration$",
+        oauth_views.ConnectDiscoveryInfoView.as_view(),
+        name="oidc-connect-discovery-info",
+    ),
+    flagged_re_path(FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED, r"^\.well-known/jwks\.json$", oauth_views.JwksInfoView.as_view(), name="jwks-info"),
+    flagged_re_path(FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED, r"^userinfo$", oauth_views.UserInfoView.as_view(), name="user-info"),
+]
+
+
 root_urls = [
-    re_path(r'^o/$', oauth2_provider_views.ApiOAuthAuthorizationRootView.as_view(), name='oauth_authorization_root_view'),
-    re_path(r"^o/authorize/$", oauth_views.AuthorizationView.as_view(), name="authorize"),
-    re_path(r"^o/token/$", oauth2_provider_views.TokenView.as_view(), name="token"),
-    re_path(r"^o/revoke_token/$", oauth_views.RevokeTokenView.as_view(), name="revoke-token"),
+    re_path(r"^o/", include((oauth_urls, 'oauth2_provider'))),
 ]
--- a/ansible_base/oauth2_provider/views/authorization_root.py
+++ b/ansible_base/oauth2_provider/views/authorization_root.py
@@ -1,6 +1,8 @@
 from collections import OrderedDict
 
 from django.utils.translation import gettext_lazy as _
+from drf_spectacular.utils import extend_schema
+from flags.state import flag_enabled
 from rest_framework import permissions
 from rest_framework.response import Response
 
@@ -8,6 +10,7 @@ from ansible_base.lib.utils.response import get_relative_url
 from ansible_base.lib.utils.views.django_app_api import AnsibleBaseDjangoAppApiView
 
 
+@extend_schema(exclude=True)
 class ApiOAuthAuthorizationRootView(AnsibleBaseDjangoAppApiView):
     permission_classes = (permissions.AllowAny,)
     name = _("API OAuth 2 Authorization Root")
@@ -16,7 +19,10 @@ class ApiOAuthAuthorizationRootView(AnsibleBaseDjangoAppApiView):
 
     def get(self, request, format=None):
         data = OrderedDict()
-        data['authorize'] = get_relative_url('authorize')
-        data['revoke_token'] = get_relative_url('revoke-token')
-        data['token'] = get_relative_url('token')
+        data['authorize'] = get_relative_url('oauth2_provider:authorize')
+        data['revoke_token'] = get_relative_url('oauth2_provider:revoke-token')
+        data['token'] = get_relative_url('oauth2_provider:token')
+        if flag_enabled("FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED"):
+            data['oidc-connect-discovery-info'] = get_relative_url('oauth2_provider:oidc-connect-discovery-info')
+            data['jwks-info'] = get_relative_url('oauth2_provider:jwks-info')
         return Response(data)
--- a/test_app/tests/oauth2_provider/views/test_authorization_root.py
+++ b/test_app/tests/oauth2_provider/views/test_authorization_root.py
@@ -1,3 +1,6 @@
+import pytest
+from flags.state import disable_flag, enable_flag
+
 from ansible_base.lib.utils.response import get_relative_url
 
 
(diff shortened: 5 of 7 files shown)


ansible_base/feature_flags/definitions/feature_flags.yaml
```diff
@@ -52,3 +52,15 @@
   support_url: ''
   labels:
     - platform
+- name: FEATURE_GATEWAY_OIDC_PROVIDER_ENABLED
```
dleehr: I think this name makes perfect sense for the scope of AAP-43413 and configuring the OIDC endpoints. 

Thinking through the rest of ANSTRAT-1019 (the "feature" 😉) we'll create an API in gateway to issue JWTs, logic in controller to call that API, credential changes, etc. Do you envision using this single feature flag `FEATURE_GATEWAY_OIDC_PROVIDER_ENABLED` to toggle all of that behavior? Or do you envision having multiple feature flags with descriptive names to control each piece?

dleehr (commented):
Excellent work here. This is concise and makes a ton of sense. Also learning a lot about our feature flag architecture in real time 🥇.

Commenting with a question on the feature flag name - not making a decision here since it's in draft.

ansible_base/feature_flags/definitions/feature_flags.yaml
PabloHiro: Good point. I was thinking of using a single feature flag. Using more can lead to installer/configuration complexity and I think we should try to avoid that.

So to better reflect the feature itself what about `FEATURE_OAUTH_WORKLOAD_SIGN_ENABLED` maybe?

dleehr: Totally agree that a single feature flag makes sense here. Now to do the hard part and come up with a name 😿. I think `OAUTH` is ~~misleading~~ imprecise here, but some suggestions:

- `FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED` Speaks to the standards-based piece, aligns with both the work happening here for gateway and the rest of the feature.
- `FEATURE_JWT_WORKLOAD_IDENTITY_ENABLED` A bit more implementation specific on the feature logic, may not be obvious that it enables an endpoint in gateway.
- `FEATURE_WORKLOAD_IDENTITY_ENABLED` Describes in a generic way may not be obvious what it does.
- `FEATURE_WORKLOAD_SIGNING_ENABLED` Generic, but in a different way.

PabloHiro: Changed the flag here and in the other PR to `FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED`

ansible_base/oauth2_provider/views/authorization_root.py
```diff
@@ -16,7 +17,10 @@ class ApiOAuthAuthorizationRootView(AnsibleBaseDjangoAppApiView):
...
         data = OrderedDict()
-        data['authorize'] = get_relative_url('authorize')
-        data['revoke_token'] = get_relative_url('revoke-token')
-        data['token'] = get_relative_url('token')
+        data['authorize'] = get_relative_url('oauth2_provider:authorize')
+        data['revoke_token'] = get_relative_url('oauth2_provider:revoke-token')
+        data['token'] = get_relative_url('oauth2_provider:token')
+        if flag_enabled("FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED"):
+            data['oidc-connect-discovery-info'] = get_relative_url('oauth2_provider:oidc-connect-discovery-info')
+            data['jwks-info'] = get_relative_url('oauth2_provider:jwks-info')
```
dleehr, line 27: Nice👍

dleehr (approved):
Changes here LGTM. 

I see the AWX consumer check is failing, appears to be that test you noted earlier. I don't see any change needed in DAB, but the check on the DAB repo is failing, so I'm raising it here. 

AWX [asserts the number of feature flags must be exactly `5`](https://github.com/ansible/awx/blob/051357e573bb366c73270e01570bef19d01771a5/awx/main/tests/functional/dab_feature_flags/test_feature_flags_api.py#L29). Earlier in our discussion I assumed AWX was counting the flags relevant to AWX (e.g. `if 'controller' in labels: assert count == 5`) but that doesn't seem to be the case. 

Personally, I don't see a good reason for AWX to `assert len(response.data["results"]) == 5`. That's fragile and will fail any time a feature flag is added to DAB, right? What's the value in that (see https://github.com/ansible/awx/pull/16210#pullrequestreview-3623266580.)? The other assertion in that functional test provides value - it toggles `FEATURE_DISPATCHERD_ENABLED` and checks it.

Edit: Any thoughts on how to unblock this?

dleehr:
~~See also https://github.com/ansible/awx/pull/16210#pullrequestreview-3623266580~~ Edited into above comment

ansible_base/oauth2_provider/views/authorization_root.py
PabloHiro: Opened: https://github.com/ansible/awx/pull/16229

ansible_base/feature_flags/migrations/0003_manual_20260113.py
```diff
@@ -0,0 +1,16 @@
+###
+# Addition of FEATURE_GATEWAY_OIDC_PROVIDER_ENABLED
```
matoval: Is this file generated? The original feature flag is still mentioned here, should be the updated feature flag.

matoval (commented):
Looks good, just one comment about the new migrations file.

ansible_base/feature_flags/migrations/0003_manual_20260113.py
PabloHiro: Nice catch!! Updated

ansible_base/oauth2_provider/views/authorization_root.py
```diff
@@ -16,7 +17,10 @@ class ApiOAuthAuthorizationRootView(AnsibleBaseDjangoAppApiView):
 
     def get(self, request, format=None):
         data = OrderedDict()
-        data['authorize'] = get_relative_url('authorize')
-        data['revoke_token'] = get_relative_url('revoke-token')
-        data['token'] = get_relative_url('token')
+        data['authorize'] = get_relative_url('oauth2_provider:authorize')
+        data['revoke_token'] = get_relative_url('oauth2_provider:revoke-token')
+        data['token'] = get_relative_url('oauth2_provider:token')
+        if flag_enabled("FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED"):
```
AlanCoding, line 25: This was kind of an ugly discussion point in AWX...

If the schema is important now (OpenAPI, spectacular), and we have to ship **static** schemas for each version, but feature flags exist, and feature flags will turn on/off some endpoints...

Should `flag_enabled` just return True unconditionally during schema generation?

Not so much your PR here, just the general question, and I feel like this is a good place to pull more people into the question.

AlanCoding (approved):
Ok the actual feature looks pretty clean here actually.

ansible_base/oauth2_provider/views/authorization_root.py
PabloHiro: Good point I don't have an answer for :sweat_smile: 

Ideally we would generate schemas with all feature flags set to enabled?

zkayyali812: I think we can put feature flags in, and my expectation is it should not affect schema generation.

I recommend, we dont add urls in this method, as I think this can be problematic, and not in line with how we should use flags to feature-flag urls.

Check this out for the recommended approach - 
https://cfpb.github.io/django-flags/usage/#using-flags-in-code

EX -
```
urlpatterns = [
    flagged_path('FLAG_WITH_REQUIRED_CONDITIONS', 'a-url/', view_requiring_flag, state=True),
]
```

john-westcott-iv (commented):
Is this something we want to put into DAB or should this just be part of aap gateway? i.e. would any other service ever be configured as an oauth2 provider?

ansible_base/oauth2_provider/views/authorization_root.py
john-westcott-iv, line 27: Out of curosity, does this reuse the existing JWT signing key or does it use a separate key for the oauth2 provider logic?

PabloHiro: Reuse. See:
https://github.com/ansible-automation-platform/aap-gateway/pull/1200/changes#diff-bf6bf9fc44e88410e9b844ce931c9d5ad38ec823331dfde20c4ca95ef09b1f1fR18

PabloHiro: > Is this something we want to put into DAB or should this just be part of aap gateway? i.e. would any other service ever be configured as an oauth2 provider?

It should just be part of Gateway, but the only work here not related to the feature flag is the views. But the views already have stuff that should be exclusive to Gateway as well, so for homogeneity sake, I just chose to extend it here.

ansible_base/oauth2_provider/views/authorization_root.py
PabloHiro: @zkayyali812 , I am not an expert of django. Should I adapt your suggestion on `ansible_base/oauth2_provider/urls.py` instead?

zkayyali812: If the URLs are intended to be feature flagged, I would ask that we do it via the notation I mention above.

PabloHiro: Done :+1:

zkayyali812 (approved):

ansible_base/feature_flags/definitions/feature_flags.yaml
```diff
@@ -52,3 +52,14 @@
   support_url: ''
   labels:
     - platform
+- name: FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED
+  ui_name: Gateway as OIDC Provider
+  visibility: False
+  condition: boolean
+  value: 'False'
+  support_level: TECHNOLOGY_PREVIEW
+  description: Enable Gateway to act as an OIDC Provider.
```
zkayyali812: Just double checking, is the ui_name, description, visibility, and support level all approved by PM?

We don't currently show these anywhere in the UI, but we may do so in later releases. I'm not sure if we want to show the word "gateway" in the UI which is why I Want to double check this too.

dleehr: Excellent feedback @zkayyali812. Yes let's update that description to scrub "gateway". I know we have agreement that support level is Tech Preview but we can confirm the rest. Edit: @PabloHiro already reaching out 💯

ansible_base/oauth2_provider/urls.py
```diff
@@ -28,9 +29,25 @@
...
+
+oauth_urls = [
+    re_path(r'^$', oauth2_provider_views.ApiOAuthAuthorizationRootView.as_view(), name='oauth_authorization_root_view'),
+    re_path(r"^authorize/$", oauth_views.AuthorizationView.as_view(), name="authorize"),
+    re_path(r"^token/$", oauth2_provider_views.TokenView.as_view(), name="token"),
+    re_path(r"^revoke_token/$", oauth_views.RevokeTokenView.as_view(), name="revoke-token"),
+    # OIDC endpoints - flag is checked at request time, returns 404 when disabled
+    flagged_re_path(
+        FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED,
+        r"^\.well-known/openid-configuration$",
```
tznamena, line 42: This URL will only work without a trailing slash, with trailing slash it will get a 404. 
Since the customers might not have direct control over their 3rd party services that would make this API call, I would suggest to add support for optional trailing slash:

```suggestion
        r"^\.well-known/openid-configuration/?$",
```

tznamena (changes requested):

ansible_base/oauth2_provider/urls.py
PabloHiro: This is a copy paste from the implementation of this url pattern in the django-oauth-toolkit library. See:
https://github.com/django-oauth/django-oauth-toolkit/blob/e62c5f5f6a3db7370f7f11ab164de5211df743b7/oauth2_provider/urls.py#L50

The standard does not mention a trailing slash anywhere:
https://openid.net/specs/openid-connect-discovery-1_0.html

I think we shouldn't be supporting non-standard implementations.

tznamena: Actually, the code from django-oauth-tookit does support optional trailing slash, on the next line, line 51, using the same regexp as suggested. There must have been a reason why they have it there. However, the comment just above it says:
> Support for trailing slash shall be removed in a future release.

I agree with strict compliance with standards, however, in this case we can avoid a potential issue if customers uses a 3rd party service that does add a trailing slash.
Anyone else wants to weigh in on this one? @zkayyali812 @john-westcott-iv

PabloHiro: Ah my bad, I wrote that comment late at night and auto-piloted, but my point still stands. The trailing slash is not part of the standard. But I am open to change it.

tznamena: The proposed change makes the trailing slash optional and will make the OIDC provider handle URLs with and without it as well. Let's get one more opinion on this so we can move forward with the set of PRs.

BrennanPaciorek: I'm leaning towards keeping it as-is, since the original authors probably caved to a enterprise user at some point.

We don't need to worry about external users at the moment, so we might as well go with the more constrained (but standard) implementation while we still can.

tznamena: Sounds good, I will mark this as resolved now so we can merge this.

tznamena (approved):

ansible_base/feature_flags/definitions/feature_flags.yaml
PabloHiro: Changed both the `ui_name` and the `description`.

john-westcott-iv (approved):

PR #921, opened 2026-01-16
[AAP-62956] Change datetime to use utc as timestamp instead of local timezone

RudneiBertolJr:
Fixing the issue reported at AAP-62956

## Description

- What is being changed?
   Changing the module datetime to use UTC data to generate the expiration timestamp for the token.
- Why is this change needed?
   When customer has controller node with TIME_ZONE variable configured to a timezone with timezone grater than 1h, generated tokens are generated already expired.
   Now, it does not matter the timezone, the token will respect the expiration time set.

## Type of Change

## Self-Review Checklist

## Testing Instructions
You can use the instructions for the reproduce at https://issues.redhat.com/browse/AAP-62956

### Expected Results
Now does not matter the timezone between client and server, the expiration timestamp should respect the expiration time based on the UTC timezone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated JWT token expiration to use timezone-aware time handling, ensuring the `exp` claim is calculated consistently across server locations and timezone settings. This helps prevent premature expiration or validity window issues related to timezone and daylight saving changes.

### DIFF
--- a/ansible_base/resource_registry/resource_server.py
+++ b/ansible_base/resource_registry/resource_server.py
@@ -1,8 +1,9 @@
-from datetime import datetime, timedelta
+from datetime import timedelta
 from typing import TypedDict
 
 import jwt
 from django.conf import settings
+from django.utils.timezone import now
 
 from ansible_base.resource_registry.models import service_id
 
@@ -31,6 +32,6 @@ def get_service_token(user_id=None, expiration=60, **kwargs):
         payload["sub"] = user_id
 
     if expiration is not None:
-        payload["exp"] = datetime.now() + timedelta(seconds=expiration)
+        payload["exp"] = now() + timedelta(seconds=expiration)
 
     return jwt.encode(payload, config["SECRET_KEY"], config["JWT_ALGORITHM"])


AlanCoding:
I would just like to have some form of testing for this, but I am struggling to see what that would be.

Ultimately, this is compared against some time diff, but on another server. It's this other side of the token expiration implementation that this need to match. And that needs to use UTC. Presumably that could be hunted down, and if in DAB, tested against.

RudneiBertolJr:
Hey @AlanCoding,
I was able to test it with one controller+gateway node on AAP, which uses this library. I do have a reproducer in place. Let me know if you would like to meet and check it.

AlanCoding (approved):

ansible_base/resource_registry/resource_server.py
```diff
@@ -31,6 +31,6 @@ def get_service_token(user_id=None, expiration=60, **kwargs):
         payload["sub"] = user_id
 
     if expiration is not None:
-        payload["exp"] = datetime.now() + timedelta(seconds=expiration)
+        payload["exp"] = datetime.now(timezone.utc) + timedelta(seconds=expiration)
```
AlanCoding: Actually, if you have the ability to re-test, I would kind of prefer

```python
from django.utils import timezone

now = timezone.now()
```

AlanCoding:
```
In [1]: from django.utils.timezone import now

In [2]: now()
Out[2]: datetime.datetime(2026, 1, 20, 21, 44, 41, 966889, tzinfo=datetime.timezone.utc)

In [3]: from datetime import datetime, timedelta, timezone

In [4]: datetime.now(timezone.utc)
Out[4]: datetime.datetime(2026, 1, 20, 21, 44, 54, 837486, tzinfo=datetime.timezone.utc)

In [5]: now()
Out[5]: datetime.datetime(2026, 1, 20, 21, 44, 59, 91693, tzinfo=datetime.timezone.utc)
```

They seem to be the same thing. I had the wrong import, it is `from django.utils.timezone import now`. This is the most standard thing to use `now()`.

AlanCoding (changes requested):
I think it's worth it to put in the `now()` pattern, because this happens so many times in related code we should standardzie

ansible_base/resource_registry/resource_server.py
RudneiBertolJr: Yeap, I have tested and they are the same. I will change my PR to use `now()`.

```
>>> from django.utils.timezone import now
>>> now()
datetime.datetime(2026, 1, 20, 22, 5, 38, 890072, tzinfo=datetime.timezone.utc)

>>> from datetime import datetime, timedelta, timezone
>>> datetime.now()
datetime.datetime(2026, 1, 20, 14, 6, 8, 126278)
>>> datetime.now(timezone.utc)
datetime.datetime(2026, 1, 20, 22, 6, 24, 138297, tzinfo=datetime.timezone.utc)

>>> now()
datetime.datetime(2026, 1, 20, 22, 8, 12, 107243, tzinfo=datetime.timezone.utc)
```

AlanCoding (approved):

PR #923, opened 2026-01-23
[AAP-62105] Implement default JWT scope aap_controller_automation_job for OIDC workload identity

melissalkelly:
AAP-62105

### What is being changed?
  Adds OIDC workload identity scope definitions to django-ansible-base in `workload_identity` under lib:

  - **Base `Scope` class** - Defines the interface for workload identity scopes
    - `list_claims()` - Returns list of claim names for the scope
    - `populate_claims()` - Stub for future implementation (AAP-62657)

  - **`AutomationControllerJobScope` class** - Implements the `aap_controller_automation_job` scope with 21 claims defined as constants (e.g., `CLAIM_JOB_ID`, `CLAIM_JOB_NAME`)

  - **Tests** - Validates base class raises `NotImplementedError` and verifies `AutomationControllerJobScope` returns the correct 21 claims
This scope definition is the foundation for OIDC workload identity in AAP. Gateway and Controller both need to reference this scope:
- Gateway (AAP-62534) validates that provided claims match the scope 
- Controller (AAP-62657) uses the claims list to know what job fields to populate

The scope is defined in django-ansible-base since it's the shared library between services.
- aap_controller_automation_job scope is defined in django-ansible-base - Implemented as AutomationControllerJobScope class
- Scope resolves to the 21 claims from P3 proposal - All claims defined as constants and returned via list_claims()

## Type of Change

- [ ] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
Run the workload identity unit tests:
  ```bash
  tox -e py312 -- test_app/tests/workload_identity/
  ```
### Expected Results
  All 3 tests should pass:
  - test_base.py::test_list_claims - Verifies base Scope class raises NotImplementedError
  - test_base.py::test_populate_claims - Verifies base Scope class raises NotImplementedError
  - test_controller.py::test_list_claims - Verifies AutomationControllerJobScope returns correct 21 claims


ansible_base/workload_identity/base.py
```diff
@@ -0,0 +1,23 @@
+"""
+Base class for OIDC workload identity scopes.
+"""
+
+
+class Scope:
```
matoval: Does the Scope class need a const for scope name?

matoval (commented):
Great work!

ansible_base/workload_identity/base.py
```diff
@@ -0,0 +1,24 @@
...
```
davemulford: Should this have a more descriptive name to indicate its purpose, like `BaseWorkloadIdentityScope`?

ansible_base/workload_identity/controller.py
```diff
@@ -0,0 +1,63 @@
+"""
+OIDC Workload Identity Scope for AAP Controller.
+
+Defines the scope and claims for Controller automation job workload identity.
+"""
+
+from ansible_base.workload_identity.base import Scope
+
+
+class AutomationControllerJobScope(Scope):
```
davemulford: Does this class not need to implement `populate_claims()`?

test_app/tests/lib/workload_identity/test_controller.py
```diff
@@ -0,0 +1,35 @@
+from ansible_base.workload_identity.controller import AutomationControllerJobScope
+
+
```
davemulford, line 3: If is it expected that `AutomationControllerJobScope` does not implement the `populate_claims()` function, should there be a test for that?

davemulford (changes requested):
Great start! Just a couple of questions around naming and implementation of the `populate_claims()` function in the inheriting class.

ansible_base/workload_identity/controller.py
melissalkelly: It does not need to implement it under the current ticket, but it would be implemented under AAP-62657.

test_app/tests/lib/workload_identity/test_controller.py
melissalkelly: It will implement populate_claims() function under AAP-62657. Do you think a test would be needed in the meantime?

PabloHiro (commented):
@melissalkelly LGTM

Question for DAB maintainers though:
* Should this be placed under `ansible_base/oauth2_provider` instead?

ansible_base/lib/workload_identity/controller.py
```diff
@@ -0,0 +1,63 @@
+"""
+OIDC Workload Identity Scope for AAP Controller.
+
+Defines the scope and claims for Controller automation job workload identity.
+"""
+
+from ansible_base.workload_identity.base import BaseWorkloadIdentityScope
+
+
+class AutomationControllerJobScope(BaseWorkloadIdentityScope):
```
fincamd, line 10: Nitpick here!

Nice work with the inheritance :D, I would only suggest a change to the name of the concrete scope class so we stay closer to the name of the scope itself `aap_controller_automation_job`:

```suggestion
class AAPControllerAutomationJobScope(BaseWorkloadIdentityScope):
```

WDYT?

ansible_base/lib/workload_identity/base.py
```diff
@@ -0,0 +1,24 @@
+"""
+Base class for OIDC workload identity scopes.
+"""
+
+
+class BaseWorkloadIdentityScope:
```
fincamd, line 8: Given that this is a base class, I would suggest adding the following annotation to the class definition (https://docs.python.org/3/library/abc.html):

```suggestion
from abc import ABC

class BaseWorkloadIdentityScope(ABC):
```

ansible_base/lib/workload_identity/base.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    Base class for OIDC workload identity scopes.
+    """
+
+    name = ""
+    description = ""
+
+    def list_claims(self) -> list[str]:
```
fincamd: I'd also suggest to annotate these methods so that the interpreter complains if somebody forgets to implement these methods `list_claims` and `populate_claims`. For this, add this import to the top of the file, complementing what I suggested to annotate the base class itself:

```python
from abc import ABC, abstractmethod
```

```suggestion
    @abstractmethod
    def list_claims(self) -> list[str]:
```

ansible_base/workload_identity/base.py
```diff
@@ -0,0 +1,24 @@
...
+    name = ""
+    description = ""
+
+    def list_claims(self) -> list[str]:
+        """
+        Return a list of all claim names defined in this scope.
+        """
+        raise NotImplementedError("Subclasses must implement list_claims()")
+
+    def populate_claims(self, context: dict) -> dict:
```
fincamd:
```suggestion
    @abstractmethod
    def populate_claims(self, context: dict) -> dict:
```

test_app/tests/lib/workload_identity/test_controller.py
davemulford: No, if it's something that will be implemented in an upcoming ticket then it can be added there.

davemulford (approved):
LGTM!

ansible_base/lib/workload_identity/base.py
```diff
@@ -0,0 +1,24 @@
...
+        """
+        Return a list of all claim names defined in this scope.
+        """
+        raise NotImplementedError("Subclasses must implement list_claims()")
+
+    def populate_claims(self, context: dict) -> dict:
+        """
+        Populate claims from the provided context.
+        """
```
fincamd, line 37: I'd also explain what this `context` parameter is used for:

```suggestion
        """
         Populate claims based on the provided context.

        :param context: The context dictionary used to populate claim values. This context identifies the AAP workload that is requesting a scope.
        :type context: dict

        :return: A dictionary mapping claim names (as strings) to their corresponding values generated from the context.
        :rtype: dict
        """
```

You can also change the name of the `context` variable to something more meaningful like `workload_fields` or something of the sorts, if you want.

fincamd (changes requested):
Good work @melissalkelly!! Thanks for taking care of this.

I'm leaving some suggestions

ansible_base/lib/workload_identity/base.py
PabloHiro: I was gonna suggest the same thing because this is what I prefer as well, but the rest of DAB does not use the "newer" python mechanism to tag abstract classes, so I think for now it is better to leave it.

AlanCoding:
> Question for DAB maintainers though:
> 
> Should this be placed under ansible_base/oauth2_provider instead?


Is it also going to be used by the jwt_consumer app?

I don't like having this new folder `ansible_base/xyz` where that doesn't fit into the app system. Although I realize this might be used by more than 1 app (?). But if it's used by 1 app, or an app that uses it is a dependency of the other apps that use it, I would favor having it in the app folder.

PabloHiro:
> Is it also going to be used by the jwt_consumer app?
> 
> I don't like having this new folder `ansible_base/xyz` where that doesn't fit into the app system. Although I realize this might be used by more than 1 app (?). But if it's used by 1 app, or an app that uses it is a dependency of the other apps that use it, I would favor having it in the app folder.

@AlanCoding , no, it is going to be a separate thing. The "applications" we have in Gateway will not interact with this part of the code. However this feature will leverage the same library (https://django-oauth-toolkit.readthedocs.io/en/latest/) as the stuff under `oauth2_provider`, that's why I am asking.

Is DAB organized by features? Or organized by "technologies"?

ansible_base/lib/workload_identity/controller.py
melissalkelly: I am ok with changing it, my only thought is having AAP at the beginning could be too verbose. The package is ansible_base.workload_identity so it is clear that it is AAP infrastructure. Maybe ControllerAutomationJobScope without the AAP would be best? Does anyone else have any thoughts?

ansible_base/workload_identity/controller.py
```diff
@@ -0,0 +1,63 @@
...
+            self.CLAIM_EXECUTION_ENVIRONMENT_ID,
+            self.CLAIM_PROJECT_NAME,
+            self.CLAIM_PROJECT_ID,
+            self.CLAIM_JOB_TEMPLATE_NAME,
+            self.CLAIM_JOB_TEMPLATE_ID,
+            self.CLAIM_UNIFIED_JOB_TEMPLATE_NAME,
+            self.CLAIM_UNIFIED_JOB_TEMPLATE_ID,
+            self.CLAIM_INSTANCE_GROUP_NAME,
+            self.CLAIM_INSTANCE_GROUP_ID,
+        ]
```
john-westcott-iv: Instead of having to add this in two places, can we do something like:
```suggestion
     return [
        getattr(self, attr) for attr in dir(self) if attr.startswith('CLAIM_')
    ]
```

AlanCoding:
> Is DAB organized by features? Or organized by "technologies"?

I would certainly say features and not by technologies. My takeaway, yes, is that this is a mostly an entirely new thing. I just thought I remembered something about it going through the existing OAuth2 provider.

ansible_base/lib/workload_identity/controller.py
BrennanPaciorek, line 10: I think there should be some acknowledgement either via the class name or some comment that this is not a complete implementation of the base class, as it lacks an implementation of `populate_claims`

PabloHiro (approved):

john-westcott-iv (commented):
## 📋 Code Review Summary

This PR adds OIDC workload identity scope definitions to django-ansible-base, implementing a base `BaseWorkloadIdentityScope` class and the `AutomationControllerJobScope` class with 21 claims for Controller automation job workload identity. The implementation is clean and well-documented.

**Files Reviewed**: 6 files

### 🔍 Issues Found
- 1 correctness/quality suggestion (missing ABC inheritance)
- 0 security/safety issues
- 0 blocking issues

### Overall Assessment
LGTM with minor suggestion - The implementation is solid and achieves the stated goals. The one suggestion is about properly inheriting from `ABC` to make the abstract methods truly enforced by Python's ABC mechanism, though the current implementation works due to the explicit `NotImplementedError` raises.

### General Feedback
- Well-structured code with clear separation between base class and implementation
- Good test coverage for the defined functionality
- Clear docstrings explaining the purpose and future plans for `populate_claims()`
- The use of class constants for claim names is a good pattern for maintainability

ansible_base/lib/workload_identity/base.py
```diff
@@ -0,0 +1,37 @@
+"""
+Base class for OIDC workload identity scopes.
+"""
+
+from abc import abstractmethod
+
+
+class BaseWorkloadIdentityScope:
```
john-westcott-iv, line 8: The class uses `@abstractmethod` decorators but doesn't inherit from `ABC`. While the current implementation works because the methods explicitly raise `NotImplementedError`, inheriting from `ABC` would provide proper Python abstract base class enforcement.

```suggestion
from abc import ABC, abstractmethod


class BaseWorkloadIdentityScope(ABC):
```

This would make Python enforce that subclasses implement the abstract methods at instantiation time, rather than at method call time. However, since the explicit `NotImplementedError` raises accomplish the same goal functionally, this is a minor suggestion rather than a required change.

PabloHiro: @john-westcott-iv  we had a discussion about this here:
https://github.com/ansible/django-ansible-base/pull/923#discussion_r2727409500

I suggested not using `abc` because other parts of the code do not use it either.

melissalkelly: Thanks for the review @john-westcott-iv! As mentioned above by @PabloHiro, I went with the current approach to maintain consistency with existing base classes in the codebase. However, I'm happy to change this to inherit from ABC if you think that would be best. Let me know what you think

BrennanPaciorek (approved):
I'm alright with these changes from a functionality perspective, meaning once Johns comment is resolved one way or another it LGTM.

ansible_base/lib/workload_identity/base.py
john-westcott-iv: I'm ok with this for now, we should have an effort to update all of these locations to inherit from ABC in the future.

ansible_base/lib/workload_identity/controller.py
```diff
@@ -0,0 +1,44 @@
...
+    CLAIM_PROJECT_NAME = 'aap_controller_project_name'
+    CLAIM_PROJECT_ID = 'aap_controller_project_id'
+    CLAIM_JOB_TEMPLATE_NAME = 'aap_controller_job_template_name'
+    CLAIM_JOB_TEMPLATE_ID = 'aap_controller_job_template_id'
+    CLAIM_UNIFIED_JOB_TEMPLATE_NAME = 'aap_controller_unified_job_template_name'
+    CLAIM_UNIFIED_JOB_TEMPLATE_ID = 'aap_controller_unified_job_template_id'
+    CLAIM_INSTANCE_GROUP_NAME = 'aap_controller_instance_group_name'
+    CLAIM_INSTANCE_GROUP_ID = 'aap_controller_instance_group_id'
+
+    def list_claims(self) -> list[str]:
```
dleehr: nit: Have you considered implementing this as a [class method](https://docs.python.org/3/library/functions.html#classmethod) rather than an instance method? 

As an instance method, we're suggesting that the return value of `list_claims` depends on the specific instance of `AutomationControllerJobScope`. But every `AutomationControllerJobScope` should return the same list of claims, which depends on the class, right?

An instance method works, so this is a nitpick.

dleehr (approved):
LGTM 🎉 . One nitpicky suggestion.

dleehr (approved):
LGTM

fosterseth (approved):

john-westcott-iv (approved):

PR #924, opened 2026-01-29
Add workload identity API client library

matoval:
## Description

Implements a workload identity API client for requesting workload identity tokens from Gateway using service token authentication. This enables Ansible services (Controller, EDA) to obtain JWT tokens for automation jobs to authenticate with external systems like HashiCorp Vault.

**Location:** `ansible_base/resource_registry/`

## Type of Change
- [x] New feature
- [x] Test update

## Implementation

**New Package** (`ansible_base/resource_registry/`):
- `service_client.py` - `BaseServiceClient` class with shared JWT authentication logic
- `workload_identity_client.py` - `WorkloadIdentityClient` class with auto-refresh service tokens
- `workload_identity_types.py` - `WorkloadIdentityTokenRequest` and `WorkloadIdentityTokenResponse` NamedTuples  
- `workload_identity_exceptions.py` - Custom exception hierarchy
- `__init__.py` - Updated with import documentation

**Modified Files:**
- `rest_client.py` - Refactored `ResourceAPIClient` to inherit from `BaseServiceClient`

**Tests** (`test_app/tests/resource_registry/`):
- `test_workload_identity_client.py` - 22 comprehensive tests

**Pattern:** Inherits from `BaseServiceClient`

## Acceptance Criteria

All 8 criteria from AAP-62691 met:

1. ✅ Package at `ansible_base/resource_registry`
2. ✅ Client class & types
3. ✅ POST to `/api/gateway/v1/workload_identity_tokens`
4. ✅ `X-ANSIBLE-SERVICE-AUTH` header authentication
5. ✅ Automatic service token refresh
6. ✅ Accepts `claims: dict` and `scope: str` parameters
7. ✅ JSON POST body
8. ✅ Returns JWT on success, raises exceptions on failure

## Testing

```bash
# Run tests
pytest test_app/tests/resource_registry/test_workload_identity_client.py -v
```

## Usage Example

```python
from ansible_base.resource_registry.workload_identity_client import get_workload_identity_client

client = get_workload_identity_client(jwt_user_id=user.id)

response = client.request_token(
    claims={
        "sub": f"job_{job.id}",
        "aud": "hashicorp-vault",
        "controller_project_name": job.project.name,
    },
    scope="read"
)

# Use response.access_token with external service
```

## Notes

- Requires server endpoint `/api/gateway/v1/workload_identity_tokens` (likely from AAP-43414) 

**Related:** AAP-43413, AAP-43414, AAP-62105  
**Jira:** [AAP-62691](https://issues.redhat.com/browse/AAP-62691)

## Summary by CodeRabbit

* **New Features**
  * Added a Workload Identity client to request workload identity tokens with configurable verification, timeout, and automatic service-token refresh.

* **Refactor**
  * Introduced a shared service HTTP client to centralize authentication, token management, and request handling for registry interactions.

* **Documentation**
  * Added module-level documentation and usage guidance for the Resource Registry package.

* **Tests**
  * Added comprehensive tests for client behavior, token refresh/expiration, request payloads, and error handling.


ansible_base/workload_identity/exceptions.py
```diff
@@ -0,0 +1,38 @@
...
+    default_detail = "Workload identity operation failed."
+    default_code = "workload_identity_error"
+
+
+class InvalidTokenError(WorkloadIdentityError):
+    """Raised when a token is invalid or expired."""
+
+    status_code = HTTP_498_INVALID_TOKEN
+    default_detail = "Invalid or expired token."
+    default_code = "invalid_token"
```
PabloHiro: This is unused code, I suggest removing it.

ansible_base/resource_registry/workload_identity_types.py
```diff
@@ -0,0 +1,31 @@
...
+class WorkloadIdentityTokenRequest(NamedTuple):
+    """
+    Request body for workload identity token endpoint.
+    """
+
+    claims: dict
+    """Dictionary of claims to include in the workload identity token."""
+
+    scope: str
+    """Token scope string (e.g., 'read', 'write', 'read write')."""
```
PabloHiro: @fincamd you might need to use this(or a similar thing for your workload_identity work). The idea is that these two PRs consume the same serializer.

ansible_base/resource_registry/workload_identity_types.py
```diff
@@ -0,0 +1,31 @@
...
+    """The JWT access token."""
+
+    token_type: Optional[str] = "Bearer"
+    """Token type, typically 'Bearer'."""
+
+    expires_in: Optional[int] = None
+    """Token expiration time in seconds."""
+
+    scope: Optional[str] = None
+    """The scope of the token."""
```
PabloHiro: @fincamd same for this, the idea is that both the request and response are modeled in a common place.

PabloHiro (commented):
Instead of creating a brand new client under `ansible_base/workload_identity`, shouldn't we be expanding the existing one under `django-ansible-base/ansible_base/resource_registry/rest_client.py`?

Thoughts @ansible/django-ansible-base-write ? Maybe it is time to extract a base client class?

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,196 @@
...
+
+    @property
+    def service_auth_header(self) -> dict:
+        """
+        Get the service authentication headers.
+
+        Returns:
+            dict: Dictionary with X-ANSIBLE-SERVICE-AUTH key and JWT token value
+        """
+        return self.requests_auth_kwargs["headers"]
```
PabloHiro: This method is not used in the code, I would remove it altogether.

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,196 @@
...
+        Overrides BaseServiceClient.refresh_jwt() to wrap errors in ServiceAuthenticationError.
+
+        Raises:
+            ServiceAuthenticationError: If token refresh fails
+        """
+        try:
+            super().refresh_jwt()
+        except Exception as e:
+            logger.error(f"Failed to refresh service token: {e}")
+            raise ServiceAuthenticationError(f"Failed to refresh service token: {e}") from e
```
PabloHiro: I am not a big fan of overriding, it makes less intuitive to understand where a method is coming from and can lead to the [diamond problem](https://en.wikipedia.org/wiki/Multiple_inheritance), also see [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance)

I would either move the try/except to the base class or remove it. And remove this method altogether from here. I don't think wrapping in an exception justifies having an override.

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,196 @@
...
+    def service_auth_header(self) -> dict:
+        """
+        Get the service authentication headers.
+
+        Returns:
+            dict: Dictionary with X-ANSIBLE-SERVICE-AUTH key and JWT token value
+        """
+        return self.requests_auth_kwargs["headers"]
+
+    def request_token(
```
PabloHiro: The [proposal](https://handbook.eng.ansible.com/proposals/ANSTRAT-1019-Management-of-JTWs-to-support-AAP-access-to-external-systems) names this:

```suggestion
    def request_workload_jwt(
```

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,196 @@
...
+        # Create response object
+        return WorkloadIdentityTokenResponse(
+            access_token=access_token,
+            token_type=response_data.get("token_type", "Bearer"),
+            expires_in=response_data.get("expires_in"),
+            scope=response_data.get("scope", scope),
+        )
+
+
+def get_workload_identity_client(**kwargs) -> WorkloadIdentityClient:
```
PabloHiro: `rest_client.py` has the equivalent method at the beginning of the file. Can we keep the structure and place this at the beginning too? This way someone reading the files will have an easier time as the structure is aligned.

ansible_base/resource_registry/workload_identity_exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+
+class TokenRequestError(WorkloadIdentityError):
+    """Raised when token request fails."""
+
+    status_code = HTTP_401_UNAUTHORIZED
+    default_detail = "Failed to obtain workload identity token."
+    default_code = "token_request_failed"
+
+
+class ServiceAuthenticationError(WorkloadIdentityError):
```
PabloHiro: This one was used only by an the `refresh_jwt` in `WorkloadIdentityClient` as an override. I think we should get rid of this and leave the original `refresh_jwt` in the base class as is.

This way we ensure that we do not change existing functionality in `rest_client.py` and we do not add bloat to `workload_identity_client.py`

ansible_base/resource_registry/workload_identity_exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+    default_detail = "Workload identity operation failed."
+    default_code = "workload_identity_error"
+
+
+class TokenRequestError(WorkloadIdentityError):
+    """Raised when token request fails."""
+
+    status_code = HTTP_401_UNAUTHORIZED
+    default_detail = "Failed to obtain workload identity token."
+    default_code = "token_request_failed"
```
PabloHiro:
```suggestion
class TokenRequestError(APIException):
    """Raised when token request fails."""

    status_code = HTTP_401_UNAUTHORIZED
    default_detail = "Failed to obtain workload identity token."
    default_code = "token_request_failed"
```

I would simply create a new exception and move it to `ansible_base/resource_registry/workload_identity_client.py` so we have one file less and all the new code for the feature is in a single file.

ansible_base/resource_registry/workload_identity_types.py
PabloHiro, line 1: Following the same pattern as in `rest_client.py`, I think that the contents of this file can be merged directly at the beginning of `workload_identity_client.py`, we would have one file less and everything related to the feature is in a single file. Thoughts?

PabloHiro (changes requested):
Overall it LGTM, I left comments aiming to trim AI verbosity and keep the changes leaner.

ansible_base/resource_registry/workload_identity_types.py
matoval: Makes sense to me

PabloHiro (approved):
LGTM, thank you @matoval , please look for approvals from the code owners. We also might need to wait to merge this one until we get this merged first:
https://github.com/ansible/django-ansible-base/pull/925

`WorkloadIdentityTokenResponse` here might get simplified to a simple `jwt` field and `WorkloadIdentityTokenRequest` to have `scope`, `audience`, `claims` instead.

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+class WorkloadIdentityTokenRequest(NamedTuple):
+    """Request body for workload identity token endpoint."""
+
+    claims: dict
+    """Dictionary of claims to include in the workload identity token."""
+
+    scope: str
+    """Token scope string (e.g., 'read', 'write', 'read write')."""
```
fincamd: Could the comment be misleading talking about `read` and `write` scopes? These are OIDC standard scopes, but our usage is not the standard one.
For now we have only proposed to add the `aap_controller_automation_job` scope

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,193 @@
...
+from rest_framework.exceptions import APIException
+from rest_framework.status import HTTP_401_UNAUTHORIZED
+
+from ansible_base.resource_registry.resource_server import get_resource_server_config
+from ansible_base.resource_registry.service_client import BaseServiceClient
+
+logger = logging.getLogger("ansible_base.resource_registry.workload_identity_client")
+
+
+class WorkloadIdentityTokenRequest(NamedTuple):
```
fincamd, line 20: There's been an addition to the request schema: the `audience` parameter, that will need to come from the integration itself to the WIT API.

As part of the PRs linked in [AAP-43414](https://issues.redhat.com/browse/AAP-43414). See #925.

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,193 @@
...
+    """The JWT access token."""
+
+    token_type: Optional[str] = "Bearer"
+    """Token type, typically 'Bearer'."""
+
+    expires_in: Optional[int] = None
+    """Token expiration time in seconds."""
+
+    scope: Optional[str] = None
+    """The scope of the token."""
```
fincamd: I apologize for the changes made @matoval but I changed implementation a bit from what the Handbook says.

The WIT endpoint will only return the token itself as a string in a field called `jwt`. We will not need the `expires_in`, `scope` or `token_type`, I believe (unless I am misunderstanding the purpose of this `WorkloadIdentityTokenResponse` class.

See #925. 
```suggestion
class WorkloadIdentityTokenResponse(NamedTuple):
    """Response from workload identity token endpoint."""

    jwt: str
    """The JWT access token signed and containing."""
```

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,193 @@
...
+    ) -> WorkloadIdentityTokenResponse:
+        """
+        Request a workload identity token.
+
+        Makes a POST request to /api/gateway/v1/workload_identity_tokens
+        with the specified claims and scope.
+
+        Args:
+            claims: Dictionary of claims to include in the token
+            scope: Token scope string (e.g., 'read', 'write', 'read write')
```
fincamd: Same comment as above for the scopes placed as examples, this might be misleading to whoever uses the function

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,193 @@
...
+            WorkloadIdentityTokenResponse: Token response with access_token
+
+        Raises:
+            TokenRequestError: If the request fails
+
+        Example:
+            >>> response = client.request_workload_jwt(
+            ...     claims={"sub": "user123", "aud": "my-service"},
+            ...     scope="read write"
+            ... )
```
fincamd, line 159: Same would apply here.

The claims `sub` and `aud` are default claims that are set by the endpoint when generating the JWT. In this client, the `claims` parameter will contain the details of the workload that asks for the JWT (not the filtered claims, but the full json schema). With the minor difference that we will be passing in `audience` value in the request.


```suggestion
            >>> response = client.request_workload_jwt(
            ...     claims={"id": 2, "name": "my-example-job"},
            ...     scope="aap_controller_automation_job"
            ... )
```

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,193 @@
...
+        """
+        request_body = WorkloadIdentityTokenRequest(claims=claims, scope=scope)
```
fincamd: We'll need to pass in the `audience` parameter as well

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,193 @@
...
+            raise TokenRequestError("Response missing 'access_token' field")
+
+        access_token = response_data["access_token"]
+
+        return WorkloadIdentityTokenResponse(
+            access_token=access_token,
+            token_type=response_data.get("token_type", "Bearer"),
+            expires_in=response_data.get("expires_in"),
+            scope=response_data.get("scope", scope),
+        )
```
fincamd: I'm sure we can leverage Serializers created as part of #925 to simplify part of this code.

fincamd (changes requested):
Thanks for taking care of these changes @matoval. I have some questions about some parts of the new client with some minor changes.

ansible_base/resource_registry/workload_identity_types.py
fincamd: These have been defined in #925

ansible_base/resource_registry/workload_identity_types.py
fincamd: These have been defined in #925

ansible_base/resource_registry/workload_identity_client.py
PabloHiro:
```suggestion
    """Token custom scopes string."""
```

ansible_base/resource_registry/workload_identity_client.py
matoval: I addressed all suggestions except for this one, I will make this change once #925 is merged.

ansible_base/resource_registry/service_client.py
```diff
@@ -0,0 +1,150 @@
...
+import time
+from typing import Optional
+
+import requests
+import urllib3
+
+from ansible_base.resource_registry.resource_server import get_service_token
+
+# Disable only insecure request warnings (for verify_https=False scenarios)
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
AlanCoding: Feels like this shouldn't be at import, but maybe this PR is still in-progress?

ansible_base/resource_registry/service_client.py
```diff
@@ -0,0 +1,150 @@
...
+    def refresh_jwt(self) -> None:
+        """
+        Refresh the service token.
+
+        Generates a new service token with the configured expiration.
+        Includes a 2-second buffer to account for slower requests.
+        """
+        # Add a buffer to the token timeout to account for slower requests.
+        self._jwt_timeout = time.time() + (self.jwt_expiration - 2)
+        self._jwt = get_service_token(self.jwt_user_id, expiration=self.jwt_expiration)
```
AlanCoding, line 77: So we make a request to Gateway.

That produces a JWT.

We use this JWT to make a request to Gateway. This will produce a new JWT.

This JWT will be passed to the plugin.

The plugin will use the JWT to make a request to the external service. This will return a secret.

This secret will then be used to make request to some other service, inside of the actual playbook.

ansible_base/resource_registry/service_client.py
matoval: I removed it

ansible_base/resource_registry/service_client.py
matoval: Yes, that's the correct flow! The architecture is:

1. Service → Gateway: Service requests a service JWT (this happens in BaseServiceClient.refresh_jwt())
2. Service → Gateway: Service uses service JWT in X-ANSIBLE-SERVICE-AUTH header to request workload identity JWT from /api/gateway/v1/workload_identity_tokens (this is WorkloadIdentityClient.request_workload_jwt())
3. Service → Plugin: Workload identity JWT is passed to the Ansible plugin/credential type
4. Plugin → External Service: Plugin uses workload identity JWT to authenticate with external service (e.g., HashiCorp Vault) and retrieve secrets
5. Plugin → Playbook: Secrets are injected into the playbook execution

This PR implements step 2 - the client library that services (Controller/EDA) will use to request workload identity JWTs.

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -0,0 +1,187 @@
...
+            ...     claims={"id": 2, "name": "my-example-job"},
+            ...     scope="aap_controller_automation_job",
+            ...     audience="https://vault.example.com"
+            ... )
+        """
+        request_body = WorkloadIdentityTokenRequest(claims=claims, scope=scope, audience=audience)
+        data = request_body._asdict()
+
+        logger.info(f"Requesting workload identity token with scope: {scope}")
+        logger.debug(f"Claims: {claims}")
```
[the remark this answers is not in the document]
matoval: The Inframesh team has decided logging claims in debug isn't an issue, claims structure is set in: https://github.com/ansible/django-ansible-base/tree/devel/ansible_base/lib/workload_identity to see what values will be logged.

ansible_base/resource_registry/service_client.py
AlanCoding: I was wrong. I don't think (1) is correct as stated.

Yes, a JWT is used, but `get_service_token` doesn't make any requests to make that happen. It just re-formats values from settings into JSON (plus some encoding maybe).

The reason I was asking the question is because this "workload identity" thing adds an endpoint for issuing new tokens. And the flow I thought was happening would have meant that an endpoint already existed doing this, but it doesn't.

It seems the reason it's misleading is the `expiration`, suggesting that this matters in some sense from the issuer. But that appears to not matter. It seems to be a self-declared expiration, making it effectively meaningless. Whatever time is saved by the JWT refresh cycle here would only save the python code time for serialization.

matoval: Sorry the wording of the comment and doc string wasn't the clearest. I've updated the comment and docstring to clarify: we use a 2-second buffer (not caching for performance) to prevent the service token from expiring mid-flight while it's being used to authenticate the workload identity token request.

ansible_base/resource_registry/workload_identity_client.py
matoval: @fincamd sorry can you clarify what you mean here? I'm not sure I follow how I would leverage the serializes from #925 here.

john-westcott-iv (commented):
## 📋 Code Review Summary

Solid implementation of the workload identity API client, aligned with ResourceAPIClient and acceptance criteria. Clear types, exceptions, and tests.

**Files Reviewed**: 6 files
**Comments Posted**: 3 review comments

### 🔍 Issues Found
- 0 security/safety issues
- 1 correctness/logic suggestion (buffer edge case)
- 2 code quality / clarity suggestions

### Overall Assessment
LGTM with minor suggestions. No blocking issues.

### General Feedback
- Good test coverage and use of parametrized-style iteration for scopes/claims.
- Pattern consistency with rest_client.py (header name, config factory, token refresh) is helpful.
- Consider adding a note in PR description that the project prefers tox -e py312 per AGENTS.md.

john-westcott-iv (commented):
## 📋 Code Review Summary

Review of the current PR (workload identity client under resource_registry with BaseServiceClient).

**Files Reviewed**: 6 files (resource_registry refactor + workload identity client + tests)

### 🔍 Focus areas
- **Correctness**: Same 2-second buffer in `BaseServiceClient.refresh_jwt` can be negative if `jwt_expiration` &lt; 2; consider `max(0, self.jwt_expiration - 2)` or validating `jwt_expiration >= 2` in `__init__`.
- **Docs**: Consider documenting that `claims` must be JSON-serializable (e.g. no `datetime` or custom objects) in `request_workload_jwt` docstring.
- **Testing**: Project prefers `tox -e py312 -- test_app/...` per AGENTS.md; PR description could mention that.

### Overall
LGTM with minor suggestions. Refactor to `BaseServiceClient` and shared token logic looks good.

ansible_base/resource_registry/workload_identity_client.py
dleehr: I think the suggestion is to use the Django REST Framework [serializers](https://www.django-rest-framework.org/api-guide/serializers/) that were contributed for the server in #925. Specifically, use `WorkloadIdentityTokenRequestSerializer(serializers.Serializer)` and `WorkloadIdentityTokenResponseSerializer(serializers.Serializer)` here (in the client) instead of defining new `WorkloadIdentityTokenRequest(NamedTuple)` and `WorkloadIdentityTokenResponse(NamedTuple)` types.

One benefit is that the serializers already define the request/response payload fields (`scope`/`claims`/`audience` in the request and `jwt` in the response), that the server expects. In addition to that, serializers also describe data types and validation criteria (similar to Django models - you can define if a field is required, allowed to be null, or allowed to be blank). They're really helpful on the server side where these things must be enforced, they simplify a lot of the validation and automate error reporting. We don't have to write a lot of validation logic on the server, we can simply declare what fields need to be present and what the rules are.

Having said all that, serializers are part of the **server**-side framework. They make a ton of sense for a server that needs to accept data, validate it, shuttle it to/from a database, but they're not really intended to be used client-side. They certainly _can_ be used (and we do[^1]), but it means your client depends on all of Django REST Framework and Django. That's not a problem in DAB or AWX (which already depend on these), but as a reviewer I find it unexpected for a client to depend on an entire server framework. Seems like overkill.

In contrast, `namedtuple` is lightweight, python-native, and we have a few instances of DAB using it in clients already, like https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/resource_registry/rest_client.py#L20-L24. We do need to make sure the structures align with the server-side serializers, but IMO, serializers are a server-side implementation detail.

My suggestion is to stick with `namedtuple`, but I'd certainly defer to the component teams that can speak from experience here about using serializers in client-side code.

[^1]: The `rest_client` [uses them](https://github.com/matoval/django-ansible-base/blob/devel/ansible_base/resource_registry/rest_client.py#L196-L213), and there are probably other locations. I think I used them client-side in the PoC.

matoval:
> Review of the current PR (workload identity client under resource_registry with 
> BaseServiceClient).
> 
> Files Reviewed: 6 files (resource_registry refactor + workload identity client + tests)
> 
> 🔍 Focus areas
> Correctness: Same 2-second buffer in BaseServiceClient.refresh_jwt can be negative if jwt_expiration < 2; consider max(0, self.jwt_expiration - 2) or validating jwt_expiration >= 2 in __init__.
> Docs: Consider documenting that claims must be JSON-serializable (e.g. no datetime or custom objects) in request_workload_jwt docstring.
> Testing: Project prefers tox -e py312 -- test_app/... per AGENTS.md; PR description could mention that.
> Overall
> LGTM with minor suggestions. Refactor to BaseServiceClient and shared token logic looks good.

Hey @john-westcott-iv thank you for reviewing! The review mentioned minor suggestions, could you clarify the minor suggestions?

fincamd (approved):
Approving to withdraw requested changes based on discussion during PR review

AlanCoding (approved):

ansible_base/resource_registry/workload_identity_client.py
PabloHiro: The agreement during the review meeting was that either way was fine. Reusing the serializers has the Pro of having to update things only in one place. It has the Con of increasing coupling.

@matoval the PR is fine to merge as is, but feel free to implement Dani's suggestion if you personally think is better

matoval: I'm happy merging what we have currently.

john-westcott-iv:
@matoval I think it was just these two:

Correctness: Same 2-second buffer in BaseServiceClient.refresh_jwt can be negative if jwt_expiration < 2; consider max(0, self.jwt_expiration - 2) or validating jwt_expiration >= 2 in __init__.

Docs: Consider documenting that claims must be JSON-serializable (e.g. no datetime or custom objects) in request_workload_jwt docstring.

zkayyali812 (approved):

ansible_base/resource_registry/service_client.py
zkayyali812: Resolving this after speaking with @tyraziel , and since the above comment captures the initial point, if a future update is needed it will happen in a follow up

PR #925, opened 2026-01-30
[AAP-43414] - Add Workload Identity Token serializers

fincamd:
## Description

- What is being changed?

Adding a couple of Serializer classes for Workload Identity Token requests/responses.

- Why is this change needed?

For OIDC integration with external systems feature.

By creating shared DRF Serializers to be leveraged in ansible applications.

## Type of Change

## Self-Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for Workload Identity Token request and response handling with proper validation for scope, audience, and claims.

* **Tests**
  * Added comprehensive test coverage for Workload Identity Token serializers, including validation scenarios and edge cases.


ansible_base/lib/workload_identity/workload_identity_tokens.py
```diff
@@ -0,0 +1,48 @@
+from django.utils.translation import gettext_lazy as _
+from rest_framework import serializers
+
+
+class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
+    """
+    Serializer for workload identity token request payload.
+    Validates that scope, audience, and claims are properly provided.
```
PabloHiro, line 8: Should we add an additional field for the requested TTL? See:
https://handbook.eng.ansible.com/proposals/ANSTRAT-1019-Configurable-JWT-Expiration-for-OIDC-Integrations

It mentions a field named `workload_ttl_seconds`

fincamd: No, because this is the initial implementation. This will come when we implement tickets output of P4 PR. It's a good suggestion, but I think we have decided to keep the changes separate

PabloHiro (approved):
LGTM, please @fincamd look for approvals from codeowners 🙏

tznamena:
## Code Review

### Overview
This PR adds two well-structured DRF serializer classes for Workload Identity Token functionality. The implementation is clean and focused.

### Files Modified

**`ansible_base/oauth2_provider/serializers/workload_identity_tokens.py`** (+37, new file)

**WorkloadIdentityTokenRequestSerializer:**
- ✅ Clear field validation (required, non-null, non-blank)
- ✅ Comprehensive help text documentation for all fields
- ✅ Three required fields properly validated:
  - `scope`: OIDC scope for the token
  - `audience`: Intended JWT audience (aud claim) 
  - `claims`: Workload details dictionary (enforced non-empty)

**WorkloadIdentityTokenResponseSerializer:**
- ✅ Simple, focused response structure with JWT field
- ✅ Well-documented with help text

### Observations
- Clean implementation following DRF serializer best practices
- Good field-level documentation will help API consumers
- Appropriate validation constraints (non-null, non-blank, non-empty)
- All CI checks passing (Django 4/5, py311/py312, linting, consumer tests)

### Verdict
LGTM - Ready for codeowner approval. The serializers are straightforward and well-documented.

---
*This code review was created by Cursor AI assistant.*

tznamena (approved):

ansible_base/lib/workload_identity/workload_identity_tokens.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+
+    scope = serializers.CharField(
+        required=True,
+        allow_null=False,
+        allow_blank=False,
+        help_text=_("OIDC Scope for the token. Supported scopes can be discovered at the well-known OIDC GW endpoints."),
+    )
```
john-westcott-iv, line 16: Are there going to be models behind these? If so, we should be able to have the fields on the models and not on the serializer right?

BrennanPaciorek: Workload identity tokens are short-lived JWTs, so they are not stored in database, but validated by cryptographic proof. This serializer is here for validating request and response payloads on at least the provider-side I think.

BrennanPaciorek (approved):

BrennanPaciorek (changes requested):
This may need to be a library (in `ansible_base.lib`, if the intent is for this to be consumed by AWX. I believe AWX does not install `ansible_base.oauth2_provider`, which might make this code inaccessible.

PabloHiro:
> This may need to be a library (in `ansible_base.lib`, if the intent is for this to be consumed by AWX. I believe AWX does not install `ansible_base.oauth2_provider`, which might make this code inaccessible.

@BrennanPaciorek  Controller will not consume this directly, but rather use what is present here: https://github.com/ansible/django-ansible-base/pull/924, which is under `ansible_base.resource_registry`

ansible_base/lib/workload_identity/workload_identity_tokens.py
PabloHiro: Brennan is correct 👍 , nothing gets stored in the DB

fincamd: Is there any change needed for this suggestion? Can I resolve this?

BrennanPaciorek:
> Controller will not consume this directly, but rather use what is present here: https://github.com/ansible/django-ansible-base/pull/924, which is under ansible_base.resource_registry

~If this is the case, the serializer for this can move downstream, since we do not have any endpoints that use this in the oauth2_provider DAB app.~

Oh I see what you mean. In this case, we still definitely want to move the serializers to a libarary in `ansible_base.lib`

fincamd:
@BrennanPaciorek
>Oh I see what you mean. In this case, we still definitely want to move the serializers to a libarary in ansible_base.lib

Done, moved into `ansible_base.lib.workload_identity`

jay-steurer:
rerunning jobs

jay-steurer:
Sonar quality gate metrics need to be addressed

BrennanPaciorek (approved):
LGTM. This should be good to merge once code coverage (in this case, some test that just intializes both serializers I think will bring this to 100%)

So just give it a test that provides a good example payload to the serializer and make sure it passes the call to `is_valid`

PabloHiro (approved):

BrennanPaciorek (approved):

zkayyali812 (approved):

ansible_base/lib/workload_identity/workload_identity_tokens.py
```diff
@@ -0,0 +1,37 @@
+from django.utils.translation import gettext_lazy as _
+from rest_framework import serializers
+
+
+class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
+    """
+    Serializer for workload identity token request payload.
+    Validates that scope, audience, and claims are properly provided.
```
AlanCoding, line 8: What would be more helpful to see is what this would be used for. What entity would use this? Is this for the IDP to use? Is it for the client (the relaying party)?

The "token request" is... a request sent _to_ the IDP to get a token? And by token, that's JWT? These are all guesses.

john-westcott-iv (approved):
Approved pending Alans last comment.

AlanCoding:
Does awx-plugins need this data? It is not reasonable to add django-ansible-base to awx-plugins. That is a concern of mine.

Presumably, the awx-plugin new content (the new plugins) will accept the JWT that this produces. Which... I guess that's just...

```python
    jwt = serializers.CharField(
        help_text=_("The signed JWT token for workload identity."),
    )
```

You probably really know better than me, but weird that it's a CharField. I don't think it can be a encrypted or unstructured blob when handed over to the plugin. And I'm confused about the wording. I don't know what a "token for workload identity" is.

From the planning, this looks like a JWT to be used by the lookup plugin to access certain secrets. Maybe this will be used for other things in the future? That would be nice. For now, it seems controller/AWX will get this value, and not care about its contents, because the very next hand-off that matters is that controller just passed it over to the plugin. Took me a while to get clear on this. But I kind of want to know more in this contract serializer, what does workload identity mean in that context? I expect these are general OIDC claims, although only planned for Hashi vault related stuff right now.

AlanCoding:
Confirmed - `jwt` is expected to be encrypted data.

(as in, definitely not json)

AlanCoding:
This is a dependency on another PR, and @john-westcott-iv and @zkayyali812 have approved, and I think it's only used there so you should be good to go with that.

ansible_base/lib/workload_identity/workload_identity_tokens.py
dleehr: It's a serializer for the API in this PR https://github.com/ansible-automation-platform/aap-gateway/pull/1221

AlanCoding (approved):

PR #927, opened 2026-02-04
[POC] [AAP-64471] Add dynamic DB/cache session manager

john-westcott-iv:
## Description
- **What is being changed?** Added a feature flag `FEATURE_GATEWAY_SIDECAR_CACHE_ENABLED` that allows the session store to dynamically switch between cached+DB and DB-only storage modes at runtime.
- **Why is this change needed?** In clustered deployments with sidecar caches, session invalidation (logout) doesn't propagate instantly across nodes. When a user logs out on one node, their session gets cleared from that node's cache and the DB, but other nodes may still have the session cached, causing "zombie sessions" where users appear to still be logged in.
- **How does this change address the issue?** When the feature flag is enabled, sessions bypass the cache entirely and go straight to the DB, ensuring logout is immediately consistent across all cluster nodes.

## Type of Change

## Self-Review Checklist

### Prerequisites
- DAB test environment

### Steps to Test
1. Run the tests: `tox -e py312 -- test_app/tests/lib/sessions/stores/test_cached_dynamic_timeout.py -v`
2. All 10 session tests should pass

### Expected Results
- When `FEATURE_GATEWAY_SIDECAR_CACHE_ENABLED=False` (default): Sessions are stored in both cache and DB
- When `FEATURE_GATEWAY_SIDECAR_CACHE_ENABLED=True`: Sessions are stored in DB only

## Additional Context

This is a **POC** to validate the approach for supporting gateway deployments with distributed sidecar caches where session consistency is critical.

### Key implementation detail
The core logic is just 4 lines:
```python
def _dispatch(self, method_name, *args, **kwargs):
    use_cache = not flag_state('FEATURE_GATEWAY_SIDECAR_CACHE_ENABLED')
    parent = CachedDBSessionStore if use_cache else DBSessionStore
    return getattr(parent, method_name)(self, *args, **kwargs)
```

---

### DIFF
--- a/ansible_base/feature_flags/migrations/0005_manual_20260204.py
+++ b/ansible_base/feature_flags/migrations/0005_manual_20260204.py
@@ -1,0 +1,16 @@
+###
+# Addition of FEATURE_GATEWAY_SIDECAR_CACHE_ENABLED
+###
+
+# FileHash: 91c16e889753dc2aa421c77867db70e86830cd50769c8a2b1d43386561add509
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dab_feature_flags', '0004_remove_dispatcherd_feature_flag'),
+    ]
+
+    operations = [
+    ]
--- a/ansible_base/lib/sessions/stores/cached_dynamic_timeout.py
+++ b/ansible_base/lib/sessions/stores/cached_dynamic_timeout.py
@@ -1,7 +1,9 @@
 import logging
 
 from django.contrib.sessions.backends.cached_db import SessionStore as CachedDBSessionStore
+from django.contrib.sessions.backends.db import SessionStore as DBSessionStore
 from django.utils.translation import gettext_lazy as _
+from flags.state import flag_state
 
 from ansible_base.lib.utils.settings import get_setting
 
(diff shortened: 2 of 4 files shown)


ansible_base/feature_flags/migrations/0005_manual_20260204.py
AlanCoding, line 1: I prefer these to have a descriptive name, but I seem to be losing that battle.

ansible_base/lib/sessions/stores/cached_dynamic_timeout.py
```diff
@@ -21,3 +37,23 @@ def get_session_cookie_age(self):
             )
             timeout = DEFAULT_SESSION_TIMEOUT
         return timeout
+
+    def _dispatch(self, method_name, *args, **kwargs):
+        """Route method calls to cached or DB-only parent based on feature flag."""
+        # Sidecar cache enabled = DB only (no centralized cache)
+        # Sidecar cache disabled = use centralized cache + DB
+        use_cache = not flag_state('FEATURE_GATEWAY_SIDECAR_CACHE_ENABLED')
+        parent = CachedDBSessionStore if use_cache else DBSessionStore
```
AlanCoding, line 46: I thought that the cases were:
 - flag true --> uses the local redis as a cache
 - flag false --> uses the old clustered redis as the cache

What I see here is
 - flag false --> uses the database as the cache

This behavior is actually what I wanted! It makes complete sense to add the sidecar redis, delete the redis cluster, and keep the DB cache (worse performance) as a fallback in case something _goes wrong_ with the sidecar redis, which I find to be entirely reasonable.

But that only makes sense if we are getting rid of the clustered redis. _And this is the best possible plan_. We should do that, although the awkward part is that it isn't really a "feature flag" as opposed to a general setting or behavior-toggle. We would be abandoning the old behavior. But again, this is the _best path forward_. But not the plan you wrote up.

john-westcott-iv: For clarity, DBSessionStore is not a DB cache per say, its just sessions stored in the DB. CachedDBSessionStore is DB stored sessions but sessions are put into the cache to avoid the DB call. In the case of sidecar redis having sessions only in the DB helps prevent the user logout issue and sessions being left around in the cache on other machines (no need to call dispatcherd on a user logout to sync the sessions). In my research, this is what AWX is doing already.

The feature flag would eventually "go away" once side car was GA and the SessionStore class in this file would just inherit from DBSessionStore only.

AlanCoding: > For clarity, DBSessionStore is not a DB cache per say, its just sessions stored in the DB.

Going cross-eyed here.

So if you're using `DBSessionStore`... is it accurate to say that you're not caching sessions?

Yes, sessions are in a table somewhere. That's what they are, and the table comes from some Django contrib app like contrib.auth. So I think it makes sense that doing a kv cache in the database makes absolutely no sense as opposed to just reading the values.

But it sounds like you're saying that the ultimate end-state with Redis sidecar would be that sessions aren't cached. That's not good, right? We need to cache sessions, right?

PR #931, opened 2026-02-06
[AAP-64480] Add audit logging for activity stream changes

john-westcott-iv:
## Description

- **What is being changed?**
  Add audit logging capability to the activity stream that logs model changes to a configurable audit logger.

- **Why is this change needed?**
  To provide comprehensive audit logging for AAP entities, enabling compliance and security auditing requirements.

- **How does this change address the issue?**
  Implements audit logging infrastructure in the AuditableModel signal handlers that can be enabled per-model.

## Type of Change
- [x] Test update

## Self-Review Checklist

### Prerequisites
- Development environment with DAB test_app

### Steps to Test
1. Run the test suite: `tox -e py312 -- -k "test_audit" -v`
2. All new audit logging tests should pass

### Expected Results
- All tests pass
- No regressions in existing activity stream functionality

### New class variables on AuditableModel
- `audit_log_enabled` (default False) - enables/disables audit log messages
- `activity_stream_enabled` (default True) - enables/disables activity stream DB entries

### New logging helper
- Added `log_auth_info()` function for INFO level audit logging

### Audit log message formats
- **create/delete**: Logs full object state as dict
- **update**: Logs one line per changed field with old/new values
- **associate/disassociate**: Logs M2M relationship changes

### Example log output
```
2026-02-06 15:15:02,744 INFO 550e8400-e29b-41d4-a716-446655440000 aap.auth_audit [172.18.0.1] "curl/8.7.1" update User admin changed last_name from 'OldValue' to 'NewValue'
```

---

## Summary by CodeRabbit

* **New Features**
  * Per-model toggles added to control activity stream and audit logging behavior
  * Many-to-many relationship changes are now recorded and logged

* **Bug Fixes**
  * Activity stream entries are only created when activity stream is enabled
  * Better handling of update-field limits and defensive normalization for m2m changes

* **Tests**
  * Extensive tests added for activity stream and audit logging scenarios

### DIFF
--- a/ansible_base/activitystream/models/entry.py
+++ b/ansible_base/activitystream/models/entry.py
@@ -112,6 +112,12 @@ class AuditableModel(models.Model):
     # Adding field names to this list will limit the activity stream changes dictionaries to only include these fields
     activity_stream_limit_field_names = []
 
+    # Controls whether changes to this model are stored in the activity stream
+    activity_stream_enabled = True
+
+    # Controls whether changes to this model are logged to the audit log
+    audit_log_enabled = False
+
     @property
     def activity_stream_entries(self):
         """
--- a/ansible_base/activitystream/signals.py
+++ b/ansible_base/activitystream/signals.py
@@ -2,6 +2,8 @@ import logging
 import threading
 from contextlib import contextmanager
 
+from ansible_base.lib.logging import log_auth_info
+
 logger = logging.getLogger('ansible_base.activitystream.signals')
 
 
@@ -66,11 +68,39 @@ def _store_activitystream_entry(old, new, operation, update_fields=None):
     else:
         content_object = new
 
-    return Entry.objects.create(
-        content_object=content_object,
-        operation=operation,
-        changes=delta.dict(),
-    )
+    # Determine the instance to check attributes on (prefer new, fallback to old)
+    instance_for_check = new if new is not None else old
+
+    if getattr(instance_for_check, 'audit_log_enabled', False):
+        model_name = content_object.__class__.__name__
+        obj_str = str(content_object)
+        changes = delta.dict()
+
+        if operation in ('create', 'delete'):
+            # For create/delete, dump the whole object state as a dict
+            all_fields = {}
+            if operation == 'create':
+                all_fields.update(changes.get('added_fields', {}))
+            else:
+                all_fields.update(changes.get('removed_fields', {}))
+            all_fields.update({k: v[1] if operation == 'create' else v[0] for k, v in changes.get('changed_fields', {}).items()})
+            log_auth_info(f"{operation} {model_name} {obj_str} {all_fields}")
+        else:
+            # For update, emit one line per change
+            for field_name, value in changes.get('added_fields', {}).items():
+                log_auth_info(f"{operation} {model_name} {obj_str} added {field_name}='{value}'")
+            for field_name, value in changes.get('removed_fields', {}).items():
+                log_auth_info(f"{operation} {model_name} {obj_str} removed {field_name} (was '{value}')")
+            for field_name, (old_val, new_val) in changes.get('changed_fields', {}).items():
+                log_auth_info(f"{operation} {model_name} {obj_str} changed {field_name} from '{old_val}' to '{new_val}'")
+
+    if getattr(instance_for_check, 'activity_stream_enabled', True):
+        return Entry.objects.create(
+            content_object=content_object,
+            operation=operation,
+            changes=delta.dict(),
+        )
+    return None
 
 
 def _store_activitystream_m2m(given_instance, model, operation, pk_set, reverse, field_name):
@@ -88,10 +118,20 @@ def _store_activitystream_m2m(given_instance, model, operation, pk_set, reverse,
     entries = []
 
     for instance in instances:
+        content_object = instance if reverse else given_instance
+        related_object = given_instance if reverse else instance
+
+        # Audit logging for m2m changes
+        if getattr(content_object, 'audit_log_enabled', False):
+            content_model_name = content_object.__class__.__name__
+            related_model_name = related_object.__class__.__name__
+            preposition = 'with' if operation == 'associate' else 'from'
+            log_auth_info(f"{operation} {content_model_name} {content_object} {preposition} {related_model_name} {related_object}")
+
         entry = Entry(
-            content_object=instance if reverse else given_instance,
+            content_object=content_object,
             operation=operation,
-            related_content_object=given_instance if reverse else instance,
+            related_content_object=related_object,
             related_field_name=field_name,
             created_by=user,
         )
--- a/ansible_base/activitystream/signals.py
+++ b/ansible_base/activitystream/signals.py
@@ -1,8 +1,15 @@
+from __future__ import annotations
+
 import logging
 import threading
 from contextlib import contextmanager
+from typing import TYPE_CHECKING, Any, Dict, Generator, Optional, Set, Tuple, Type, Union
 
 from ansible_base.lib.logging import log_auth_event
+
+if TYPE_CHECKING:
+    from django.contrib.auth.models import AbstractUser
+    from django.db.models import Model
 
 logger = logging.getLogger('ansible_base.activitystream.signals')
 
@@ -19,7 +26,7 @@ activitystream_enabled = ActivityStreamEnabled()
 
 
 @contextmanager
-def no_activity_stream():
+def no_activity_stream() -> Generator[None, None, None]:
     previous_value = activitystream_enabled.enabled
     activitystream_enabled.enabled = False
     try:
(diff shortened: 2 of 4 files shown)


ansible_base/activitystream/signals.py
```diff
@@ -66,11 +68,39 @@ def _store_activitystream_entry(old, new, operation, update_fields=None):
...
+            all_fields.update({k: v[1] if operation == 'create' else v[0] for k, v in changes.get('changed_fields', {}).items()})
+            log_auth_event(f"{operation} {model_name} {obj_str} {all_fields}")
+        else:
+            # For update, emit one line per change
+            for field_name, value in changes.get('added_fields', {}).items():
+                log_auth_event(f"{operation} {model_name} {obj_str} added {field_name}='{value}'")
+            for field_name, value in changes.get('removed_fields', {}).items():
+                log_auth_event(f"{operation} {model_name} {obj_str} removed {field_name} (was '{value}')")
+            for field_name, (old_val, new_val) in changes.get('changed_fields', {}).items():
+                log_auth_event(f"{operation} {model_name} {obj_str} changed {field_name} from '{old_val}' to '{new_val}'")
```
BrennanPaciorek: We can probably move all of this into it's own function to fix the cognitive complexity issue.

BrennanPaciorek: Is there any risk of this accidentally logging secrets?

BrennanPaciorek (commented):
Need to run formatter, fix sonar, potentially mitigate logging secrets.

ansible_base/activitystream/signals.py
```diff
@@ -66,11 +68,39 @@ def _store_activitystream_entry(old, new, operation, update_fields=None):
     else:
         content_object = new
 
-    return Entry.objects.create(
-        content_object=content_object,
-        operation=operation,
-        changes=delta.dict(),
-    )
+    # Determine the instance to check attributes on (prefer new, fallback to old)
+    instance_for_check = new if new is not None else old
```
bhavenst: Isn't this the same as instance_for_check = content_object ?

ansible_base/activitystream/signals.py
```diff
@@ -66,11 +68,39 @@ def _store_activitystream_entry(old, new, operation, update_fields=None):
...
+
+        if operation in ('create', 'delete'):
+            # For create/delete, dump the whole object state as a dict
+            all_fields = {}
+            if operation == 'create':
+                all_fields.update(changes.get('added_fields', {}))
+            else:
+                all_fields.update(changes.get('removed_fields', {}))
+            all_fields.update({k: v[1] if operation == 'create' else v[0] for k, v in changes.get('changed_fields', {}).items()})
+            log_auth_event(f"{operation} {model_name} {obj_str} {all_fields}")
```
bhavenst: Are we putting it on the caller to add the username making the change?

test_app/tests/activitystream/test_signals.py
```diff
@@ -477,3 +480,384 @@ def test_activitystream_update_fields_none_with_limit_fields():
...
+        assert getattr(AuditableModel, attribute) == expected
+
+
+# =============================================================================
+# Tests for activity_stream_enabled flag
+# =============================================================================
+
+
+@pytest.mark.django_db
+def test_activity_stream_enabled_false_prevents_entry_creation():
```
bhavenst: could be ..._false_on_update to be consistent with other names perhaps?

test_app/tests/activitystream/test_signals.py
```diff
@@ -477,3 +480,384 @@ def test_activitystream_update_fields_none_with_limit_fields():
...
+        animal.age = 5  # This should not be logged
+        animal.save()
+
+        messages = [call[0][0] for call in mock_log.call_args_list]
+
+        # Name should be logged
+        name_logged = any('name' in msg for msg in messages)
+        assert name_logged, f"Expected name in messages: {messages}"
+
+        # Age should NOT be logged (it's excluded)
```
bhavenst, line 778: Could potentially expand this to check some of the disallowed content

Passwords (plain text or hashed)
Security tokens or API keys
Full JWT tokens or SAML assertions
OAuth authorization codes and client id/secrets

Though some of those aren't really applicable w/ just DAB.

ansible_base/activitystream/signals.py
bhavenst: The requirements ask for object IDs as well, which I suppose should be available here?

ansible_base/activitystream/signals.py
bhavenst: What happens here w/ password for instance?  'changed password from $encrypted to $encrypted'?  I guess we want to log the fact that a field was changed even if the value isn't loggable.

john-westcott-iv: There shouldn't be otherwise they would also be exposed in the activity log too. Will confirm and post example output here.

john-westcott-iv: Right now it just shows `$encrypted$` for both values:
```
----------------------------------------
Step 6: Setting password on user
----------------------------------------
2026-02-09 15:29:27,706 INFO      aap.auth_audit [] "" update User audit-log-test-953e4f33 changed password from '$encrypted$' to '$encrypted$'
Set password (initial)
Expected log: update User <username> changed password from '' to '<hash>'

----------------------------------------
Step 7: Updating password on user
----------------------------------------
2026-02-09 15:29:27,817 INFO      aap.auth_audit [] "" update User audit-log-test-953e4f33 changed password from '$encrypted$' to '$encrypted$'
Changed password to a new value
Expected log: update User <username> changed password from '<hash>' to '<hash>'

----------------------------------------
Step 8: Removing password (set_unusable_password)
----------------------------------------
2026-02-09 15:29:27,823 INFO      aap.auth_audit [] "" update User audit-log-test-953e4f33 changed password from '$encrypted$' to '$encrypted$'
Removed password (unusable)
Expected log: update User <username> changed password from '<hash>' to '<unusable>'
```

Is that even usable? Should we just say something like: `update User audit-log-test-953e4f33 changed password`?

john-westcott-iv: See convo above from @BrennanPaciorek, its just showing `$encrypted$` for both values.

john-westcott-iv: Good call, creating `_log_audit_entry` function for refactoring. Will be in next push.

ansible_base/activitystream/signals.py
john-westcott-iv: Yes, good catch, updating to remove `instance_for_check`. Will be in next push.

ansible_base/activitystream/signals.py
john-westcott-iv: Also added, I missed that these required additional items that were not in the filter already. Will be added in next push.

john-westcott-iv: Are we expected to go back through the other audit logs to make sure they have object ID/user making changes (where applicable)? Not sure if any of them do.

test_app/tests/activitystream/test_signals.py
john-westcott-iv: Hmmm... I am questioning if we want to do this here or not? I.e. we should do some /basic/ testing here to ensure a secret field is `$encrypted$` but do we want to test specific flelds on specific models or do we want that to be tested from the applications that house those models? I kind of lean towards the second so that we don't have to update activitystream for every model that has a secret. Thoughts?

test_app/tests/activitystream/test_signals.py
john-westcott-iv: Updated to `test_activity_stream_enabled_false_on_update`, will be in next push.

ansible_base/activitystream/signals.py
BrennanPaciorek: I think its fine so long as we don't end up logging secrets.

How about nested secrets like the GitHub social auth client_secret? iirc, user password being `$encrypted$` is kind of a special case in the serializer at least.

test_app/tests/activitystream/test_signals.py
bhavenst: Sure, that is reasonable to do it at the higher level..

ansible_base/activitystream/signals.py
bhavenst: I would say no since it's not part of the requirements of this ticket, but maybe pinging Ron to have him file another story if needed would be a good idea..

ansible_base/activitystream/signals.py
john-westcott-iv: Authenticators are not AuditableModels so it wouldn't dump that out :)
Per comments below we stated that individual models should be testing to ensure they are not leaking secrets through activity stream. Otherwise we would have to alter activity stream for every model which would be painful.

BrennanPaciorek: > Authenticators are not AuditableModels so it wouldn't dump that out :)

(I probably should have checked that before using it as an example :( )

> Per comments below we stated that individual models should be testing to ensure they are not leaking secrets through activity stream.

This approach is fine by me I think, considering this logging is disabled by default.

BrennanPaciorek: > Is that even usable? Should we just say something like: update User audit-log-test-953e4f33 changed password?

Another thing we can do is key on `encrypted_fields` and `PASSWORD_FIELDS` to produce a unique log for those fields when changed.

ansible_base/activitystream/models/entry.py
```diff
@@ -112,6 +112,12 @@ class Meta:
     # Adding field names to this list will limit the activity stream changes dictionaries to only include these fields
     activity_stream_limit_field_names = []
 
+    # Controls whether changes to this model are stored in the activity stream
+    activity_stream_enabled = True
```
AlanCoding, line 116: so... you're anticipating that some models will be set to log but _not_ save any activity stream entries? Seems a strange preference if done at the model-level. Sure, sometimes people will turn off the DB recording. But weird to subclass the AuditableModel but _not_ write DB entries by default.

john-westcott-iv: Correct, in our future stories we are going to be adding new models which inherit from AuditableModel and the decision was made that those should show up in the audit log but not in the actual activity stream. But solving the logging here seemed like the most reasonable approach instead of having some models go through AuditableModel and some not but for all to have logs.

ansible_base/activitystream/signals.py
bhavenst: I think either way is fine.  $encrypted -> $encrypted is fine and gets the point across.  Or you could intercept and remove the values for encrypted fields "update User <username> changed password."

ansible_base/activitystream/models/entry.py
bhavenst: Not anticipating, that is an explicit requirement.  We are not to add anything new to the activitystream.  When we have cleanup in place, these can be flipped or activity_stream_enabled removed entirely.

john-westcott-iv: @AlanCoding please reopen this if you feel like we need further conversations around this.

ansible_base/activitystream/signals.py
john-westcott-iv: Leaving as is for now.

bhavenst (approved):

PR #932, opened 2026-02-07
[WIP] AAP-57990 Add visible_teams to support org admins all team view

bhavenst:
## Description
Added a visible_teams alongside visible_users to support making all teams viewable in gateway when ORG_ADMINS_CAN_SEE_ALL_USERS=True

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1.  Run UT
or
1. Grab GW partner PR and start gateway
2. Create two orgs, two teams and one normal user.  Make user admin in one org.  Add one team to org.
3. Set ORG_ADMINS_CAN_SEE_ALL_USERS=True
4. Ensure user who is org admin can see both teams
5. Set ORG_ADMINS_CAN_SEE_ALL_USERS=False
6. Ensure user who is org admin can see only the team in the org

### Expected Results
See above.

## Summary by CodeRabbit

* **New Features**
  * Implemented team visibility filtering to ensure users only see teams they have permission to access based on their role and organization assignments.

* **Tests**
  * Added comprehensive test coverage for team visibility across different user roles and permission configurations.

### DIFF
--- a/ansible_base/rbac/api/serializers.py
+++ b/ansible_base/rbac/api/serializers.py
@@ -15,7 +15,7 @@ from ansible_base.lib.utils.auth import get_team_model
 from ansible_base.lib.utils.response import get_relative_url
 from ansible_base.rbac.models import RoleDefinition, RoleTeamAssignment, RoleUserAssignment
 from ansible_base.rbac.permission_registry import permission_registry  # careful for circular imports
-from ansible_base.rbac.policies import check_content_obj_permission, visible_users
+from ansible_base.rbac.policies import can_view_all_users, check_content_obj_permission, visible_teams, visible_users
 from ansible_base.rbac.validators import check_locally_managed, validate_permissions_for_model
 
 from ..models import DABContentType, DABPermission
@@ -109,6 +109,9 @@ class BaseAssignmentSerializer(CommonModelSerializer):
             obj = resource.content_object
             if obj._meta.model_name == 'user':
                 if not visible_users(requesting_user).filter(pk=obj.pk).exists():
+                    raise ObjectDoesNotExist
+            elif obj._meta.model_name == 'team':
+                if not visible_teams(requesting_user).filter(pk=obj.pk).exists():
                     raise ObjectDoesNotExist
             elif not requesting_user.has_obj_perm(obj, 'view'):
                 raise ObjectDoesNotExist
@@ -253,7 +256,10 @@ class RoleTeamAssignmentSerializer(BaseAssignmentSerializer):
         fields = ASSIGNMENT_FIELDS + ['team', 'team_ansible_id']
 
     def get_actor_queryset(self, requesting_user):
-        return permission_registry.team_model.access_qs(requesting_user)
+        if can_view_all_users(requesting_user):
+            return visible_teams(requesting_user)
+        else:
+            return permission_registry.team_model.access_qs(requesting_user)
 
 
 class RoleMetadataSerializer(serializers.Serializer):
--- a/ansible_base/rbac/policies.py
+++ b/ansible_base/rbac/policies.py
@@ -51,6 +51,30 @@ def can_view_all_users(request_user):
     )
 
 
+def visible_teams(request_user, queryset=None) -> QuerySet:
+    """Gives a queryset of teams that another user should be able to view"""
+    team_cls = permission_registry.team_model
+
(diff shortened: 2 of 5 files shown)


john-westcott-iv (commented):
## 📋 Code Review Summary

This PR adds `visible_teams()` alongside `visible_users()` so teams are filtered by organizations the request user can view, and respects `can_view_all_users()` (and thus `ORG_ADMINS_CAN_SEE_ALL_USERS`) for superusers and org admins. Implementation and tests align well with existing `visible_users` patterns.

**Files Reviewed**: 2 files  
**Comments Posted**: 1 style suggestion (below)

### 🔍 Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 1 code quality suggestion

### Overall Assessment
LGTM — one minor style tweak suggested for consistency.

### General Feedback
- `visible_teams` correctly mirrors `visible_users` (unauthenticated → none, `can_view_all_users` → all or passed queryset, otherwise filter by `access_ids_qs(request_user, 'view')`). Logic and edge cases look correct.
- Test coverage is strong: anonymous, superuser, no permissions, view one/multiple orgs, custom queryset parametrized for superuser/org_admin/regular_viewer × `ORG_ADMINS_CAN_SEE_ALL_USERS`, and org admin with True/False. Parametrize and fixtures (`org_admin_rd`) are used appropriately.
- **Style**: In `ansible_base/rbac/policies.py` there are two consecutive blank lines before `def visible_teams` (lines 43–44 in the new code). The rest of the file uses a single blank line between top-level functions. For consistency, remove one of the two blank lines so there is only one blank line between `visible_users` / `can_view_all_users` and `def visible_teams`.

john-westcott-iv (approved):

AlanCoding:
This is boutique logic where it shouldn't be necessary.

The way the system was intended, the right way to do this would be to create a system-wide role with `view_team` permission and give that to all users you want to be able to do this. This is similar to platform auditor, but... team auditor, or whatever you want to call it. This role would be processable and synced to all other services. What you're doing here simply won't work in that sense. The syncing to other services also materially affects functionality and many things won't work without that.

bhavenst:
Thanks @AlanCoding I've been thinking about this and not sure how to do that.  I'm trying to give OrganizationAdmin an additional permission (view all teams) when ORG_ADMINS_CAN_SEE_ALL_USERS is True.  I tried to basically do it the same way as it's being done for [users](https://github.com/ansible-automation-platform/aap-gateway/blob/5cda98fb312663da58cb67eb78b56e959b4c6cf0/aap_gateway_api/views/api/v1/user.py#L31), since that was implemented originally.  Teams were missed, resulting in a difference from 2.4. I took a look at 2.4 code and it looks fairly similar, modifying the queryset filter based on org admin && setting = True.

If such a new role were created, how would it be combined with org admin (which is the role being modified)?  Some kind of auto-role assignment when an org admin is created?

Anyway, do you have some time to walk me through how this would look to do it that way?

AlanCoding:
> I'm trying to give OrganizationAdmin an additional permission (view all teams) when ORG_ADMINS_CAN_SEE_ALL_USERS is True.

There is a domain violation in that sentence itself. You're saying that an admin of organization A should get permission to teams in organization B. No, the system has no mechanism to allow that because it's not _supposed_ to allow that. Permissions a user gets to organization A shouldn't be able to jail-break to outside the scope of that organization. Viewing teams in organization B does.

> I tried to basically do it the same way as it's being done for [users](https://github.com/ansible-automation-platform/aap-gateway/blob/5cda98fb312663da58cb67eb78b56e959b4c6cf0/aap_gateway_api/views/api/v1/user.py#L31), since that was implemented originally.

It's really stretching to say that that prior behavior was not a bug.

> If such a new role were created, how would it be combined with org admin

I get it, the 3-teirs of permissions are too complicated. Those are global, organization-wide, and object-level. But we were merging pre-existing systems that functionally did this, so we wound up with more features than would have otherwise been responsible to add. And `visible_users` broke the rules, as `User` permissions are not tracked in the system. This is not true for the `Team` model.

So is there a way to do a single-assignment of both global and organization permissions? No, there is not, and the danger is that this approach actually makes it all _more_ complicated. In a practical sense, though, I expect what this is really about is the available SSO claims which are not generalized in any way. So you have limited tools with organization admin being one.

PR #940, opened 2026-02-12
AAP-62534 Add get_target_claim_names_to_sub_stubs() classmethod to BaseWorkloadIdentityScope and  AutomationControllerJobScope            

melissalkelly:
These changes were requested by @john-westcott-iv and @dleehr in [#1250](https://github.com/ansible-automation-platform/aap-gateway/pull/1250/changes#r2799953039) under AAP-62534

## Description
What is being changed?
  - Added `generate_sub_claim()` classmethod to BaseWorkloadIdentityScope that encapsulates sub claim generation logic
  - Added `get_target_claim_names_to_sub_stubs()` classmethod to BaseWorkloadIdentityScope and AutomationControllerJobScope
  - The mapping method returns claim names (e.g., CLAIM_JOB_NAME) to their corresponding sub claim stubs (e.g., "job")
  - The generation method uses this mapping to produce the formatted sub claim string (e.g., "job:my-job:organization:my-org:...")

Why is this change needed?
  - Per feedback from @john-westcott-iv in [#1250](https://github.com/ansible-automation-platform/aap-gateway/pull/1250/changes#r2799953039), we should:
    a. Use the claim constants defined in DAB (e.g., AutomationControllerJobScope.CLAIM_JOB_NAME) instead of hardcoded strings
    b. Encapsulate the sub claim generation logic within scope classes rather than spreading it across two repos
    c. Make it automatically extensible when new scopes are added to DAB

How does this change address the issue?
  - Each scope class now defines its own claim-to-stub mapping using class constants
  - The base class provides `generate_sub_claim()` that uses the scope's mapping to build the sub claim string
  - Gateway's JWT creation code can simply call `scope_class.generate_sub_claim(workload_claims)` instead of implementing the logic itself
  - All sub claim generation logic is now encapsulated within django-ansible-base
  - Changes to scope claims in DAB automatically propagate without requiring Gateway code updates

## Type of Change

## Self-Review Checklist

## Testing Instructions
Run the new tests:
```bash
tox -e py312 -- test_app/tests/lib/workload_identity/test_base.py::test_get_target_claim_names_to_sub_stubs test_app/tests/lib/workload_identity/test_controller.py -k "get_target_claim_names_to_sub_stubs or
generate_sub_claim"
```

## Summary by CodeRabbit

* **Refactor**
  * Improved workload-identity scope API to support mapping claim names to sub-stub keys and to produce consistent colon-delimited sub-claim strings for identity claims.

* **Tests**
  * Added and updated tests to validate the claim-to-sub-stub mapping, key validity, and sub-claim generation across full, empty, and missing-value scenarios.

### DIFF
--- a/ansible_base/lib/workload_identity/base.py
+++ b/ansible_base/lib/workload_identity/base.py
@@ -24,15 +24,12 @@ class BaseWorkloadIdentityScope:
         """
         raise NotImplementedError("Subclasses must implement list_claims()")
 
-    @abstractmethod
-    def populate_claims(self, workload_data: dict) -> dict:
+    @classmethod
+    def get_target_claim_names_to_sub_stubs(cls) -> dict[str, str]:
         """
-        Populate claims based on the provided workload data.
+        Return a mapping of claim names to their corresponding sub claim stubs.
 
-        :param workload_data: The workload_data dictionary used to populate claim values.
-                        This workload_data identifies the AAP workload that is requesting a scope.
-        :type workload_data: dict
-        :return: A dictionary mapping claim names to their corresponding values generated from the workload_data.
-        :rtype: dict
+        :return: A dictionary mapping claim names to sub claim stubs.
+        :rtype: dict[str, str]
         """
-        raise NotImplementedError("Subclasses will implement populate_claims() in future iterations")
+        raise NotImplementedError("Subclasses must implement get_target_claim_names_to_sub_stubs()")
--- a/ansible_base/lib/workload_identity/base.py
+++ b/ansible_base/lib/workload_identity/base.py
@@ -25,6 +25,7 @@ class BaseWorkloadIdentityScope:
         raise NotImplementedError("Subclasses must implement list_claims()")
 
     @classmethod
+    @abstractmethod
     def get_target_claim_names_to_sub_stubs(cls) -> dict[str, str]:
         """
         Return a mapping of claim names to their corresponding sub claim stubs.
--- a/ansible_base/lib/workload_identity/base.py
+++ b/ansible_base/lib/workload_identity/base.py
@@ -34,3 +34,36 @@ class BaseWorkloadIdentityScope:
         :rtype: dict[str, str]
         """
         raise NotImplementedError("Subclasses must implement get_target_claim_names_to_sub_stubs()")
+
+    @classmethod
+    def generate_sub_claim(cls, workload_details: dict) -> str:
+        """
+        Generate a sub claim string from workload details using the scope's claim mapping.
+
+        Given a dictionary with the details of a workload, generates a sub claim string with the following format:
+        "job:<job_name>:organization:<organization_name>:project:<project_name>:job_template:<job_template_name>"
+
+        Note: The specified claim names are included in the output sub claim value even if they
+        are empty. Claim validation is expected to take care of doing these checks before this
+        function is called.
+
+        :param workload_details: A dictionary containing the workload details (claim names to values)
+        :type workload_details: dict
+        :return: A string containing the sub claim
+        :rtype: str
+        """
+        target_claim_names_to_sub_stubs = cls.get_target_claim_names_to_sub_stubs()
+        return ":".join([f"{target_claim_names_to_sub_stubs[key]}:{workload_details.get(key, '')}" for key in target_claim_names_to_sub_stubs.keys()])
+
+    @abstractmethod
+    def populate_claims(self, workload_data: dict) -> dict:
+        """
+        Populate claims based on the provided workload data.
+
+        :param workload_data: The workload_data dictionary used to populate claim values.
+                        This workload_data identifies the AAP workload that is requesting a scope.
+        :type workload_data: dict
+        :return: A dictionary mapping claim names to their corresponding values generated from the workload_data.
+        :rtype: dict
+        """
+        raise NotImplementedError("Subclasses will implement populate_claims() in future iterations")
--- a/ansible_base/lib/workload_identity/base.py
+++ b/ansible_base/lib/workload_identity/base.py
@@ -36,24 +36,24 @@ class BaseWorkloadIdentityScope:
         raise NotImplementedError("Subclasses must implement get_target_claim_names_to_sub_stubs()")
 
     @classmethod
-    def generate_sub_claim(cls, workload_details: dict) -> str:
+    def generate_sub_claim(cls, workload_claims: dict) -> str:
         """
-        Generate a sub claim string from workload details using the scope's claim mapping.
+        Generate a sub claim string from workload claims using the scope's claim mapping.
 
-        Given a dictionary with the details of a workload, generates a sub claim string with the following format:
+        Given a dictionary with the claims of a workload, generates a sub claim string with the following format:
         "job:<job_name>:organization:<organization_name>:project:<project_name>:job_template:<job_template_name>"
 
         Note: The specified claim names are included in the output sub claim value even if they
         are empty. Claim validation is expected to take care of doing these checks before this
         function is called.
 
-        :param workload_details: A dictionary containing the workload details (claim names to values)
-        :type workload_details: dict
+        :param workload_claims: A dictionary containing the workload claims (claim names to values)
+        :type workload_claims: dict
         :return: A string containing the sub claim
         :rtype: str
         """
         target_claim_names_to_sub_stubs = cls.get_target_claim_names_to_sub_stubs()
-        return ":".join([f"{target_claim_names_to_sub_stubs[key]}:{workload_details.get(key, '')}" for key in target_claim_names_to_sub_stubs.keys()])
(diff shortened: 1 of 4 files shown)


ansible_base/lib/workload_identity/controller.py
```diff
@@ -43,3 +43,12 @@ class AutomationControllerJobScope(BaseWorkloadIdentityScope):
     @classmethod
     def list_claims(cls) -> list[str]:
         return [getattr(cls, attr) for attr in dir(cls) if attr.startswith('CLAIM_')]
+
+    @classmethod
+    def get_target_claim_names_to_sub_stubs(cls) -> dict[str, str]:
```
john-westcott-iv, line 48: The `AutomationControllerJobScope` class docstring (lines 12-16 on devel) still reads:

```
Note: populate_claims() is not yet implemented and will be added
in a future iteration.
```

Since this PR removes `populate_claims()` from the base class entirely and replaces it with `get_target_claim_names_to_sub_stubs()`, this docstring is now stale. It should be updated or removed to avoid confusion.

Suggested update for the class docstring:

```python
class AutomationControllerJobScope(BaseWorkloadIdentityScope):
    """
    Default scope for AAP Controller automation job workload identity.
    """
```

john-westcott-iv (commented):
## 📋 Code Review Summary

Clean, well-scoped PR that replaces the placeholder `populate_claims()` method with a purposeful `get_target_claim_names_to_sub_stubs()` classmethod, per feedback from @john-westcott-iv and @dleehr. The new method provides a proper API for Gateway to obtain claim-to-stub mappings without hardcoding them downstream.

**Files Reviewed**: 4 files

### 🔍 Issues Found
- 0 security/safety issues
- 1 code quality issue (stale docstring referencing removed method)
- 0 correctness/logic issues

### Overall Assessment
Looks good overall. The implementation is clean, correctly uses class constants via `cls`, proper decorator ordering (`@classmethod` before `@abstractmethod`), and the tests are well-structured — especially `test_get_target_claim_names_to_sub_stubs_keys_are_valid_claims` which validates that mapping keys are actual scope claims. One stale docstring should be cleaned up.

### General Feedback
- The decorator ordering `@classmethod` / `@abstractmethod` is correct per Python docs
- Good test coverage: the exact mapping test and the keys-are-valid-claims validation test complement each other nicely
- Note: `BaseWorkloadIdentityScope` uses `@abstractmethod` but doesn't inherit from `abc.ABC`, so the abstract enforcement is purely via `NotImplementedError` at runtime rather than at class instantiation time. This is pre-existing and not introduced by this PR, just noting for awareness.
- CI is still running at the time of review; linting checks (black, flake8, isort) all pass

---
**Note:** This review was performed with assistance from an AI assistant.

ansible_base/lib/workload_identity/controller.py
melissalkelly: Updated, thanks for catching that.

dleehr (changes requested):
Thanks for driving this forward. I have two pieces of feedback:

1. This PR does address the concern about where the mappings lived, so 👍 to that. However, looking at the change, we'll have the mappings here but the logic still lives in the other repo. Is there a reason for that? What do you think about adding a method here that produces the appropriate `sub` claim for the scope, rather than returning a dict and requiring a different codebase to know what to do with it? I think we have an opportunity here to encapsulate how we build a `sub` claim that is aligned to a scope, rather than spreading that concern across two repos. I guess what I'm suggesting is a method that knows how to populate claims 😉.
2. In addition to the primary change, This PR also removes `populate_claims`, which is an unrelated cleanup. I think we have a team norm to keep PRs focused on a single change. In my experience, when there's more to debate, things often get bogged down. On this point specifically (*it is never called and no longer needed*), I would like to better understand that decision before going ahead. The clean-up may be valid, but I'd suggest breaking out that cleanup into a separate PR. I'm not sure I agree that we'll never need it but I don't think we should be holding up _this_ PR to have that discussion.

melissalkelly:
Updated both. @dleehr 

I initially just moved the mappings because I thought that's what the original comments were asking for, but I can see the value in moving the entire sub claim generation logic here too. I've now added a `generate_sub_claim()` method to `BaseWorkloadIdentityScope` that encapsulates the complete logic. I implemented it as a class method on the base class with the thought that the sub claim format should be consistent across all scopes, each scope just needs to define its own `get_target_claim_names_to_sub_stubs()` mapping. 

I've restored `populate_claims` in this PR. I had initially added it for AAP-62657 but ended up not using it, so I thought it was dead code. But we can discuss that separately.

ansible_base/lib/workload_identity/base.py
```diff
@@ -24,6 +24,37 @@ def list_claims(cls) -> list[str]:
...
+        """
+        raise NotImplementedError("Subclasses must implement get_target_claim_names_to_sub_stubs()")
+
+    @classmethod
+    def generate_sub_claim(cls, workload_claims: dict) -> str:
+        """
+        Generate a sub claim string from workload claims using the scope's claim mapping.
+
+        Given a dictionary with the claims of a workload, generates a sub claim string with the following format:
+        "job:<job_name>:organization:<organization_name>:project:<project_name>:job_template:<job_template_name>"
```
dleehr, line 44: I'd like to suggest a change on this ordering. I don't think we have a clear decision in a proposal on the ordering or structure (just an example), so I think we're OK to iterate/modify. However, before making that change, let's see if a decision exists in the ticket or proposal. Either way, I offer this suggestion as follow-up work (e.g. let's not change it here but create a new Jira story - it's not something to hold this up). I'm glad we have a centralized place for the logic and the mapping 😉.

---

Here's my thinking: The `sub` claim identifies the **subject** of the token. For Vault admins to write effective role bindings/patterns for the `sub`, it will be simpler (and thus more secure) if the details are ordered from the widest to the narrowest. Organization should come first, followed by job template, project, then job.

For inspiration I'm looking to https://developer.hashicorp.com/terraform/cloud-docs/dynamic-provider-credentials/workload-identity-tokens, which structures the sub claim from broadest to narrowest 

Specifically, I think users would prefer to write patterns based on the subject like this:

```
organization:Org-1:*                                # matches any job in Org-1
organization:Org-2:*                                # matches any job in Org-2
organization:Org-2:job_template:Job-Template-B:*    # matches jobs from Job-Template-B in Org-2
```

Rather than

```
job:*:organization:Org-1:*
job:*:organization:Org-2:*
job:*:organization:Org-2:*:project:*:job_template:Job-Template-B
```

I think the former is cleaner and worry that those `*` in the latter _might_ match everything (which would be bad, need to test). For reference, see https://developer.hashicorp.com/terraform/cloud-docs/dynamic-provider-credentials/vault-configuration#create-a-jwt-auth-role for an example showing the JWT Auth role with the `bound_claims` pattern matching an organization, project, and workspace:

```JSON
{
  "policies": ["tfc-policy"],
  "bound_audiences": ["vault.workload.identity"],
  "bound_claims_type": "glob",
  "bound_claims": {
    "sub":
"organization:my-org-name:project:my-project-name:workspace:my-workspace-name:run_phase:*"
  },
  "user_claim": "terraform_full_workspace",
  "role_type": "jwt",
  "token_ttl": "20m"
}
```

ansible_base/lib/workload_identity/base.py
```diff
@@ -24,6 +24,37 @@ def list_claims(cls) -> list[str]:
...
+        are empty. Claim validation is expected to take care of doing these checks before this
+        function is called.
+
+        :param workload_claims: A dictionary containing the workload claims (claim names to values)
+        :type workload_claims: dict
+        :return: A string containing the sub claim
+        :rtype: str
+        """
+        target_claim_names_to_sub_stubs = cls.get_target_claim_names_to_sub_stubs()
+        return ":".join([f"{target_claim_names_to_sub_stubs[key]}:{workload_claims.get(key, '')}" for key in target_claim_names_to_sub_stubs.keys()])
```
dleehr, line 56: Showing my Python age.. but I was looking at this and wondered how we guarantee any predictable order here. We're not using an `OrderedDict` and we're not sorting the `.keys()` list. Yet, it's working as expected 🤔. My 👴 understanding was that python dicts didn't guarantee order. 

All the tests pass with the expected ordering so I thought maybe we were just lucky 🍀. But no, [since 3.7, "Dicts keep insertion order"](https://mail.python.org/pipermail/python-dev/2017-December/151283.html) and GvR made it so. Today I learned (fitting).

So, nothing to see here. play on!

dleehr (approved):
LGTM. Dropping a comment for future enhancement that we should turn into a follow-up item. Plus, something I learned.

PabloHiro (approved):
LGTM, interface is a clear improvement and I doubt we will change it. We can always tweak the implementation later if we are not happy with the `sub` field structure.

fincamd (approved):

AlanCoding (approved):
Found that this is already in the identity provider, so really simple code move, it seems to be

dragid10 (approved):

PR #943, opened 2026-02-13
AAP-65826 - Fallback for timeout if unset

zkayyali812:
## Description

- What is being changed?
Adds a default fallback for the timeout parameter in the service_client.
- Why is this change needed?
When this is unset via component leveraging this, this causes the API request to fail, and the following stacktrace to appear -

```
│ api   File "/opt/aap-gateway/src/aap_gateway_api/views/api/v1/common.py", line 108, in perform_update                                                                                                             │
│ api     self._resources_client.update_resource(                                                                                                                                                                   │
│ api   File "/app/django-ansible-base/ansible_base/resource_registry/rest_client.py", line 103, in update_resource                                                                                                 │
│ api     return self._make_request(action, f"resources/{ansible_id}/", self._get_request_dict(data))                                                                                                               │
│ api            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                               │
│ api   File "/opt/aap-gateway/src/aap_gateway_api/utils/resources_client.py", line 121, in _make_request                                                                                                           │
│ api     responses[service.pk] = super()._make_request(method, path, data, params)                                                                                                                                 │
│ api                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                 │
│ api   File "/app/django-ansible-base/ansible_base/resource_registry/service_client.py", line 134, in _make_request                                                                                                │
│ api     "timeout": self.timeout,                                                                                                                                                                                  │
│ api                ^^^^^^^^^^^^                                                                                                                                                                                   │
│ api AttributeError: 'AllServicesClient' object has no attribute 'timeout' 
```
This adds a default fallback of 30 seconds, if the timeout value is not provided.

## Type of Change

## Self-Review Checklist

### Steps to Test
1. Deploy AAP-Dev with this change
2. Try to toggle a feature flag. It should work (fails without this change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests now include a timeout value only when the service instance has an explicit timeout configured, preventing unintended timeout overrides and preserving expected behavior when no timeout is set.

### DIFF
--- a/ansible_base/resource_registry/service_client.py
+++ b/ansible_base/resource_registry/service_client.py
@@ -131,8 +131,10 @@ class BaseServiceClient:
             "method": method,
             "url": url,
             "verify": self.verify_https,
-            "timeout": self.timeout,
         }
+
+        if hasattr(self, 'timeout'):
+            kwargs["timeout"] = self.timeout
 
         if data:
             kwargs["json"] = data


dleehr (approved):
LGTM. This should address AAP-65826 and AAP-65827 which I just opened due to the same error.

Looks like the issue was introduced in #924 when the `self.timeout` attribute was added[^1]. It's set in the initializer and consumed in some methods. Not all subclasses use the initializer so `self.timeout` was not always set and that caused this `AttributeError`.

I re-reviewed the diff in #924 and don't see any other new attributes added, so this should fix the issue we're seeing today in AAP-65826.

[^1]: Applying the default timeout for this client is a change introduced in #924 for AAP-62691. Request timeout changes don't appear in acceptance criteria or description for AAP-62691 and I don't see an obvious need for it in the workload identity client API. So I don't know why we changed exactly. That said, it's generally a good practice to set a timeout, but we weren't setting it before and this client is used in many places, so I wanted to highlight the change.

ansible_base/resource_registry/service_client.py
```diff
@@ -131,7 +131,7 @@ def _make_request(
             "method": method,
             "url": url,
             "verify": self.verify_https,
-            "timeout": self.timeout,
+            "timeout": getattr(self, 'timeout', 30),
```
dleehr: Actually, can we define a constant or class variable for this `30`? We're assigning this magic number in two places: the initializer and here in the fallback.

Alternatively (and more compatible with previous behavior), we could **only** set `timeout` in the kwargs if it's defined on the class (rather than defaulting it again)

dleehr (changes requested):

ansible_base/resource_registry/service_client.py
zkayyali812: Yes, I think that is a great suggestion, I can capture that update

zkayyali812: Updated!

dleehr (approved):
🚀

fosterseth (approved):

PR #944, opened 2026-02-18
AAP-62532 Add service authorization support to workload identity scopes

melissalkelly:
AAP-62532

## Description               
Add service authorization support to workload identity scopes for AAP-62532.

- Added SERVICE_TYPE_* constants to BaseWorkloadIdentityScope (controller, eda, hub, gateway)
- Added `is_service_allowed(service_name: str)` method to BaseWorkloadIdentityScope (returns False by default - deny all services)
- AutomationControllerJobScope overrides `is_service_allowed()` to allow only controller service

`is_service_allowed` is called here: https://github.com/ansible-automation-platform/aap-gateway/pull/1262   

## Type of Change

## Self-Review Checklist

### DIFF
--- a/ansible_base/lib/workload_identity/base.py
+++ b/ansible_base/lib/workload_identity/base.py
@@ -10,9 +10,13 @@ class BaseWorkloadIdentityScope:
     Base class for OIDC workload identity scopes.
     """
 
+    SERVICE_TYPE_CONTROLLER = "controller"
+    SERVICE_TYPE_EDA = "eda"
+    SERVICE_TYPE_HUB = "hub"
+    SERVICE_TYPE_GATEWAY = "gateway"
+
     name = ""
     description = ""
-    allowed_services = set()
 
     @classmethod
     @abstractmethod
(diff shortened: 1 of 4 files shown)


ansible_base/lib/workload_identity/controller.py
```diff
@@ -17,6 +17,7 @@ class AutomationControllerJobScope(BaseWorkloadIdentityScope):
 
     name = "aap_controller_automation_job"
     description = "Default AAP Controller automation job workload identity"
+    allowed_services = {"controller"}
```
melissalkelly: Implemented as a class attribute rather than an abstract method. An abstract method would enforce that every scope defines allowed services, but it felt unnecessary to have a method that just returns a hardcoded set. Thoughts?

ansible_base/lib/workload_identity/base.py
```diff
@@ -10,6 +10,11 @@ class BaseWorkloadIdentityScope:
     Base class for OIDC workload identity scopes.
     """
 
+    SERVICE_TYPE_CONTROLLER = "controller"
+    SERVICE_TYPE_EDA = "eda"
+    SERVICE_TYPE_HUB = "hub"
+    SERVICE_TYPE_GATEWAY = "gateway"
+
```
dleehr, line 17: There's an `Enum` elsewhere (not in this repo) that defines these service type strings. I don't think it makes sense to redefine (or even define) them in django-ansible-base.

dleehr (changes requested):
Looking at this PR and the one that depends on it, I think it might be better to move this check entirely to other PR instead of duplicating (and maintaining) service type strings in django-ansible-base.

PR #945, opened 2026-02-19
[AAP-64920] Adds audit logging for AAPFlag

huffmanca:
## Description
Updates AAPFlag to extend AuditableModel, specifically with:
- audit_log_enabled = True
- activity_stream_enabled = False

At the moment we want to receive audit logs for this, but not create new activity stream entries

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Start up AAP with these changes.
2. Modify FeatureFlag values (create/update/delete)
3. Ensure that **no** ActivityStream entries are created from these modifications
4. Ensure that the audit log contains messages at the INFO level detailing these modifications

## Additional Context
Leverages the work from PR 931 to handle the heavy lifting.

## Summary by CodeRabbit

* **New Features**
  * Feature flags now enable audit logging by default while suppressing automatic activity-stream entries.

* **Documentation**
  * Clarified that the audit mixin is a pure mixin with no database fields and should be mixed conditionally to avoid environment-dependent migrations.

* **Tests**
  * Added tests for audit logging behavior, absence of activity-stream entries on create, and verification that the audit mixin has no DB fields.

### DIFF
--- a/ansible_base/feature_flags/models/aap_flag.py
+++ b/ansible_base/feature_flags/models/aap_flag.py
@@ -1,9 +1,17 @@
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
 from ansible_base.lib.abstract_models.common import NamedCommonModel
 from ansible_base.resource_registry.fields import AnsibleResourceField
+
+# Conditionally import AuditableModel if activitystream is installed
+_AuditableBase = object
+if 'ansible_base.activitystream' in settings.INSTALLED_APPS:
+    from ansible_base.activitystream.models import AuditableModel
+
+    _AuditableBase = AuditableModel
 
 
 def validate_feature_flag_name(value: str):
@@ -24,7 +32,12 @@ def validate_labels(value):
             raise ValidationError(_("All labels must be strings."))
 
 
-class AAPFlag(NamedCommonModel):
+class AAPFlag(NamedCommonModel, _AuditableBase):
+    # Enable audit log for this model (only if activitystream is installed)
+    audit_log_enabled = True
+    # Disable activity stream for this model
+    activity_stream_enabled = False
+
     class Meta:
         app_label = "dab_feature_flags"
         unique_together = ("name", "condition")
--- a/test_app/tests/feature_flags/models/test_aap_flag_audit_logging.py
+++ b/test_app/tests/feature_flags/models/test_aap_flag_audit_logging.py
@@ -1,0 +1,126 @@
+"""Unit tests for AAPFlag audit logging and activity stream behavior.
+
+Ensures that:
+1. AAPFlag has audit_log_enabled = True (generates audit logs)
+2. AAPFlag has activity_stream_enabled = False (no activity stream entries)
+3. Create operations generate audit logs at INFO level (to aap.auth_audit logger)
+4. Create operations do NOT create activity stream entries
+
+Note: Update and delete behavior is covered by base AuditableModel tests in
+test_app/tests/activitystream/test_signals.py. This provides AAPFlag-specific
+regression coverage for the create operation.
+"""
+
+from unittest import mock
+
+import pytest
+from django.conf import settings
+
+from ansible_base.feature_flags.models import AAPFlag
+
+# Patch where DAB activitystream calls it (signals import log_auth_event at load time)
+# log_auth_event writes to the logger specified by ANSIBLE_BASE_AUTH_AUDIT_LOGGER_NAME
+# (default: 'ansible_base.auth_audit', overridden in AAP Gateway to 'aap.auth_audit')
+AUDIT_LOG_PATCH = "ansible_base.activitystream.signals.log_auth_event"
+
+# Check if activitystream is installed
+ACTIVITYSTREAM_INSTALLED = 'ansible_base.activitystream' in settings.INSTALLED_APPS
+
+
+# -----------------------------------------------------------------------------
+# Basic functionality test: verify AAPFlag works without activitystream
+# -----------------------------------------------------------------------------
+
+
+@pytest.mark.django_db
+def test_aap_flag_basic_crud_without_activitystream():
+    """AAPFlag can be created, read, updated, and deleted regardless of activitystream installation."""
+    # Create
+    flag = AAPFlag.objects.create(
+        name="FEATURE_TEST_BASIC_CRUD_ENABLED",
+        ui_name="Test Basic CRUD",
+        condition="boolean",
+        value="False",
+        support_level="DEVELOPER_PREVIEW",
+        description="Test flag for basic CRUD operations",
+    )
+
+    # Read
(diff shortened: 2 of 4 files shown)


AlanCoding:
FWIW, the direction I would prefer to go, articulated in https://github.com/ansible/django-ansible-base/pull/951, would undo this effectively.

Many applications using the AAPFlag model/application are _not_ using the activitystream app. This makes your approach here at least somewhat awkward. I know, that's because you're using the structure that existing apps have set up, which is exactly the problem, and I'd rather change that.

In the philosophical direction I am going, it would be up to the end service/app to define what models are tracked in activitystream. So accordingly, we would make no effort in DAB to register the AAPFlag model with activitystream. At whatever point an app added activitystream app, it would be on that app to list the flag model in the models to track in activitystream.

But it's also extremely non-obvious whether doing that is desired in all apps, or just in the source-of-truth app for flags. Because having all apps record entries for flag changes means that, in practice, any flag change is recorded in all services. I'm not saying that this is a bad decision, but I want to bring some more transparency to that decision, which is a non-obvious tradeoff.

bhavenst:
@AlanCoding We discussed this in our standup today and would like to continue with the plan as outlined in the proposal/SDP and already partly implemented (audit logger functionality in activitystream + some loggable classes).  Reasons:
-  Sprint is 1/2 over and this must go into the next coming release
-  It has to be backported to 2.6 and 2.5 (so we would prefer to avoid large changes)
-  It follows the approved SDP/proposal
-  The risk of problems is low (note in the class to avoid anyone adding DB fields)
-  Follows existing pattern in DAB as far as optionally extending AuditableModel in various places so not a new problem

OTOH, totally agree with your suggested refactoring, so we will file a ticket to do that and work to get it prioritized.  That would accomplish the decoupling and eliminate the strange dependence on optional apps etc. (would be done for 2.7+).  Your PR #951 looks like a great start to that.

Are you OK with that?

AlanCoding:
So sure, you're basically trying to get this through right now as a gateway-only solution for right now. Nothing else is using activitystream.

AlanCoding (approved):

zkayyali812 (approved):

bhavenst:
Yes, the audit logging has always been gateway only as far as the "feature" requirements.  Other components are welcome to use it if they like, but it's driven by specific customer requirements for gateway..

ansible_base/feature_flags/models/aap_flag.py
```diff
@@ -1,10 +1,18 @@
...
 from django.utils.translation import gettext_lazy as _
 
 from ansible_base.lib.abstract_models.common import NamedCommonModel
 from ansible_base.resource_registry.fields import AnsibleResourceField
 
+# Conditionally import AuditableModel if activitystream is installed
+activitystream = object
+if 'ansible_base.activitystream' in settings.INSTALLED_APPS:
+    from ansible_base.activitystream.models import AuditableModel
+
```
john-westcott-iv, line 13: Nit: the variable name `activitystream` reads like a module or app reference, but it's actually being used as a base class. When you hit `class AAPFlag(NamedCommonModel, activitystream):` below, you have to scroll up to understand what it is. Consider a name that makes the class definition self-documenting:

```suggestion
_AuditableBase = object
if 'ansible_base.activitystream' in settings.INSTALLED_APPS:
    from ansible_base.activitystream.models import AuditableModel

    _AuditableBase = AuditableModel
```

test_app/tests/feature_flags/models/test_aap_flag_audit_logging.py
```diff
@@ -0,0 +1,135 @@
...
+        # Read
+        assert flag.name == "FEATURE_TEST_BASIC_CRUD_ENABLED"
+        assert flag.ui_name == "Test Basic CRUD"
+
+        # Update
+        flag.value = "True"
+        flag.save()
+        flag.refresh_from_db()
+        assert flag.value == "True"
+
```
john-westcott-iv: Nit: since these tests are decorated with `@pytest.mark.django_db`, pytest-django wraps each one in a transaction that gets rolled back automatically. The `try/finally` with `flag.delete()` is unnecessary — the DB state is cleaned up for you. Same applies to the other tests in this file that use this pattern.

```suggestion
    flag = AAPFlag.objects.create(
        name="FEATURE_TEST_BASIC_CRUD_ENABLED",
        ui_name="Test Basic CRUD",
        condition="boolean",
        value="False",
        support_level="DEVELOPER_PREVIEW",
        description="Test flag for basic CRUD operations",
    )

    # Read
    assert flag.name == "FEATURE_TEST_BASIC_CRUD_ENABLED"
    assert flag.ui_name == "Test Basic CRUD"

    # Update
    flag.value = "True"
    flag.save()
    flag.refresh_from_db()
    assert flag.value == "True"
```

john-westcott-iv (approved):
## Code Review Summary

Clean, well-scoped PR. The conditional inheritance pattern is pragmatic given the timeline and backport constraints, and the `AuditableModel` docstring warning is a good defensive addition. Logic is correct, test coverage is solid.

**Files Reviewed**: 4 files

### Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 2 code quality nits (non-blocking)

### Overall Assessment
LGTM — approving. The two comments are minor readability nits, take them or leave them.

### General Feedback
- The `audit_log_enabled = True` / `activity_stream_enabled = False` split is exactly right for the requirement
- The `object` fallback for the mixin is clean (no-op in MRO since `NamedCommonModel` already inherits from `object`)
- The `test_auditable_model_has_no_db_fields` guardrail test is a nice touch

test_app/tests/feature_flags/models/test_aap_flag_audit_logging.py
huffmanca: Removed the try/finally block in these tests.

ansible_base/feature_flags/models/aap_flag.py
huffmanca: Incorporated this suggestion to make it more obvious. There are other places in the codebase where this pattern is used; however, I'm not modifying those in this commit.

test_app/tests/feature_flags/models/test_aap_flag_audit_logging.py
```diff
@@ -0,0 +1,126 @@
...
+    """Creating an AAPFlag emits an audit log record at INFO."""
+    with mock.patch(AUDIT_LOG_PATCH) as log_auth_event:
+        flag = AAPFlag.objects.create(
+            name="FEATURE_TEST_AUDIT_CREATE_ENABLED",
+            ui_name="Test Audit Create",
+            condition="boolean",
+            value="False",
+            support_level="DEVELOPER_PREVIEW",
+            description="Test flag for audit logging on create",
+        )
```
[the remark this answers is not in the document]
huffmanca: Outdated, and caught by linter.

bhavenst (approved):

PR #947, opened 2026-02-23
AAP-64565 Remove fallback authentication

bhavenst:
## Description

Remove DAB support for fallback authentication for 2.7+ (do not backport).

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Nothing to really test, functionality removed.  Run UT to see they aren't broken.
2. 
3. 

### Expected Results
Fallback auth removed, no other changes.

## Summary by CodeRabbit

* **Removed Features**
  * Fallback authenticators are no longer supported for local authentication; the related configuration option has been removed.

* **Documentation**
  * All docs and examples for local fallback authenticators were removed.

* **Tests**
  * Tests covering fallback authenticator behavior and validation were removed; authentication tests were simplified.


john-westcott-iv (approved):

test_app/tests/authentication/authenticator_plugins/test_local.py
```diff
@@ -50,12 +50,7 @@ def test_local_auth_failure(unauthenticated_api_client, local_authenticator, use
     "configuration, expected_status_code",
     [
         ('{}', 201),
-        ('{"fallback_authentication": ["aap_gateway_api.authentication.fallbacks.controller"]}', 201),
         ('{"anything": "here"}', 400),
```
john-westcott-iv, line 55: Consider adding an explicit regression test case for the removed `fallback_authentication` field. Something like:

```suggestion
        ('{}', 201),
        ('{"fallback_authentication": ["some.module.path"]}', 400),
        ('{"anything": "here"}', 400),
```

This payload previously returned 201 when the field existed. Adding it as a negative test ensures that if someone accidentally re-introduces the field (bad revert, merge conflict, etc.), it gets caught immediately.

john-westcott-iv (commented):
Looks good overall -- clean removal, thorough test cleanup, import hygiene is solid.

Two minor items:

1. **Regression test for `fallback_authentication` rejection** (inline comment below) -- since this field *used to* return 201, it's worth having an explicit test case proving it now returns 400. Guards against accidental re-introduction via a bad revert or merge conflict.

2. **Nit on `test_missing_credentials_returns_none` refactor** -- the test was pulled out of the `TestAuthenticateIntegration` class into a standalone function (good, since that class was fallback-specific). The refactored version creates `RequestFactory().post('/login')` inline rather than via a fixture, which is fine but is a minor style difference from the other tests in the file. Not blocking, just noting.

PR #948, opened 2026-02-23
[AAP-65392] Fix OIDC issuer in discovery view

dleehr:
## Description

This PR changes the url pattern for the ConnectDiscoveryInfoView (`.well-known/openid-configuration`) to end with a trailing slash `/`, which matches the implementation in [django-oauth-toolkit 2.3.0](https://github.com/django-oauth/django-oauth-toolkit/blob/2.3.0/oauth2_provider/urls.py#L35), (pinned [here](https://github.com/ansible/django-ansible-base/blob/devel/requirements/requirements_oauth2_provider.in)).

The change is needed to fix a bug where the view returned an invalid issuer and causes OIDC Discovery to fail. Specifically, the `ConnectDiscoveryInfoView` uses a helper function `oidc_issuer()` to derive the issuer URL. In 2.3.0, this function assumes the URL ends with a trailing slash https://github.com/django-oauth/django-oauth-toolkit/blob/2.3.0/oauth2_provider/settings.py#L297 and uses that length. So if ours doesn't include the `/`, that count is off by one and the issuer ends with `/` instead of `/o`.

## Type of Change

## Self-Review Checklist

## Testing Instructions

This can be tested in aap-dev using https://github.com/ansible/aap-dev/pull/448. See section "k8s-vault with aap-dev"

## Additional Context

This URL pattern was discussed in #915: https://github.com/ansible/django-ansible-base/pull/915/changes#r2709849119. I'd prefer without the trailing slash, I'd recommend upgrading django-oauth-toolkit to address that. Since that's a more complex effort I added comments and tests to make sure we consider this when we do upgrade that dependency.

## Summary by CodeRabbit

* **Bug Fixes**
  * OIDC discovery and UserInfo endpoints now require trailing slashes for compatibility with the OAuth2 provider.

* **Tests**
  * Added tests validating OpenID discovery behavior, issuer formatting, 404 behavior when feature flag is off, and URL pattern alignment with the upstream provider.

* **Chores**
  * Added scoped context-manager utilities to temporarily enable or disable feature flags in tests.

* **Refactor**
  * Updated tests to use scoped feature-flag context managers instead of manual enable/disable calls.

### DIFF
--- a/ansible_base/lib/testing/util.py
+++ b/ansible_base/lib/testing/util.py
@@ -1,7 +1,7 @@
 from contextlib import contextmanager
 from pathlib import Path
 
-from flags.state import disable_flag, enable_flag
+from flags.state import disable_flag, enable_flag, flag_state
 from requests import Response
 
 from ansible_base.resource_registry.rest_client import ResourceAPIClient
@@ -79,8 +79,10 @@ class StaticResourceAPIClient(ResourceAPIClient):
 @contextmanager
 def feature_flag_enabled(flag_name):
     """Context manager to temporarily enable a feature flag"""
+    was_enabled = flag_state(flag_name)
     enable_flag(flag_name)
     try:
         yield
     finally:
-        disable_flag(flag_name)
+        if not was_enabled:
+            disable_flag(flag_name)
--- a/ansible_base/lib/testing/util.py
+++ b/ansible_base/lib/testing/util.py
@@ -77,12 +77,27 @@ class StaticResourceAPIClient(ResourceAPIClient):
 
 
 @contextmanager
-def feature_flag_enabled(flag_name):
-    """Context manager to temporarily enable a feature flag"""
+def _feature_flag_state(flag_name, desired_state):
+    """Context manager to temporarily set a feature flag to a desired state"""
     was_enabled = flag_state(flag_name)
-    enable_flag(flag_name)
+    if desired_state:
+        enable_flag(flag_name)
+    else:
+        disable_flag(flag_name)
     try:
         yield
     finally:
-        if not was_enabled:
+        if was_enabled:
+            enable_flag(flag_name)
+        else:
             disable_flag(flag_name)
+
+
+def feature_flag_enabled(flag_name):
+    """Context manager to temporarily enable a feature flag"""
+    return _feature_flag_state(flag_name, desired_state=True)
+
+
+def feature_flag_disabled(flag_name):
+    """Context manager to temporarily disable a feature flag"""
+    return _feature_flag_state(flag_name, desired_state=False)
--- a/ansible_base/lib/testing/util.py
+++ b/ansible_base/lib/testing/util.py
@@ -1,5 +1,7 @@
+from contextlib import contextmanager
 from pathlib import Path
 
+from flags.state import disable_flag, enable_flag
 from requests import Response
 
 from ansible_base.resource_registry.rest_client import ResourceAPIClient
@@ -72,3 +74,13 @@ class StaticResourceAPIClient(ResourceAPIClient):
             response.status_code = 404
 
         return response
+
+
+@contextmanager
+def feature_flag_enabled(flag_name):
+    """Context manager to temporarily enable a feature flag"""
+    enable_flag(flag_name)
+    try:
+        yield
+    finally:
+        disable_flag(flag_name)
--- a/ansible_base/oauth2_provider/urls.py
+++ b/ansible_base/oauth2_provider/urls.py
@@ -39,6 +39,8 @@ oauth_urls = [
     # OIDC endpoints - flag is checked at request time, returns 404 when disabled
     flagged_re_path(
         FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED,
+        # URL patterns below must match installed django-oauth-toolkit version, otherwise discovery fails
+        # See https://github.com/django-oauth/django-oauth-toolkit/blob/2.3.0/oauth2_provider/urls.py#L35
         r"^\.well-known/openid-configuration/$",
         oauth_views.ConnectDiscoveryInfoView.as_view(),
         name="oidc-connect-discovery-info",
--- a/ansible_base/oauth2_provider/urls.py
+++ b/ansible_base/oauth2_provider/urls.py
@@ -39,7 +39,7 @@ oauth_urls = [
     # OIDC endpoints - flag is checked at request time, returns 404 when disabled
     flagged_re_path(
         FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED,
-        r"^\.well-known/openid-configuration$",
+        r"^\.well-known/openid-configuration/$",
         oauth_views.ConnectDiscoveryInfoView.as_view(),
         name="oidc-connect-discovery-info",
     ),
--- a/ansible_base/oauth2_provider/urls.py
+++ b/ansible_base/oauth2_provider/urls.py
@@ -46,7 +46,7 @@ oauth_urls = [
         name="oidc-connect-discovery-info",
     ),
     flagged_re_path(FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED, r"^\.well-known/jwks\.json$", oauth_views.JwksInfoView.as_view(), name="jwks-info"),
-    flagged_re_path(FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED, r"^userinfo$", oauth_views.UserInfoView.as_view(), name="user-info"),
+    flagged_re_path(FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED, r"^userinfo/$", oauth_views.UserInfoView.as_view(), name="user-info"),
 ]
 
 
--- a/test_app/tests/oauth2_provider/test_oidc_urlpatterns.py
+++ b/test_app/tests/oauth2_provider/test_oidc_urlpatterns.py
@@ -1,0 +1,19 @@
+import pytest
+
+from oauth2_provider.urls import oidc_urlpatterns as dot_oidc_urlpatterns
+from ansible_base.oauth2_provider.urls import oauth_urls
+
+dot_oidc_names = {p.name for p in dot_oidc_urlpatterns}
+dab_oidc_names = {
+    getattr(p, 'name', None) for p in oauth_urls
+} & dot_oidc_names
+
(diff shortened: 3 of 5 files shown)


dleehr:
## Testing steps:

Using the `test_app`:

1. Create a private key in `/tmp`: `openssl genrsa -out /tmp/oidc.key 4096`
2. Add the following to `test_app/defaults.py` to enable the feature
```python
with open('/tmp/oidc.key', 'r') as f:
    private_key = f.read()

OAUTH2_PROVIDER = {"OIDC_ENABLED": True, "OIDC_RSA_PRIVATE_KEY": private_key}
FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED = True
```
3. Start the test app: `test_app/scripts/bootstrap.sh` (requires dependencies, see docs at https://github.com/ansible/django-ansible-base/tree/devel/test_app about starting the test app)
4. Fetch the openid-configuration from the test app. Confirm it responds and `issuer` ends with `/o`
```
$ curl -Ls http://localhost:8000/o/.well-known/openid-configuration | jq -r .issuer
http://localhost:8000/o
```
5. Start a vault server (e.g. [start-dev.sh](https://github.com/dleehr/vault-oidc-demo/blob/main/start-dev.sh))
6. Enable JWT auth: `vault auth enable jwt`
7. Tell vault to discover the OIDC provider config from the base (issuer) url `http://localhost:8000/o`
```
$ vault write auth/jwt/config oidc_discovery_url="http://localhost:8000/o"                     
Success! Data written to: auth/jwt/config
```

ansible_base/lib/testing/util.py
```diff
@@ -72,3 +74,15 @@ def _make_request(self, method, path, data=None, params=None, stream=False):
...
+@contextmanager
+def feature_flag_enabled(flag_name):
+    """Context manager to temporarily enable a feature flag"""
+    was_enabled = flag_state(flag_name)
+    enable_flag(flag_name)
+    try:
+        yield
+    finally:
+        if not was_enabled:
+            disable_flag(flag_name)
```
PabloHiro, line 93: In `test_oidc_endpoints_shown_when_flag_enabled` under `test_app/tests/oauth2_provider/views/test_authorization_root.py`, we've simply done:

```
enable_flag("FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED")
try:
    ...
finally:
     disable_flag("FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED")
```

Think you can update that one so there is homogeneity?

john-westcott-iv (commented):
Looks good overall -- the root cause analysis is solid and the fix is correct. A few thoughts:

1. Consider adding a negative test that verifies the endpoint returns 404 when `FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED` is disabled. This would confirm the flag gating works correctly and round out the test coverage for this endpoint.

2. Rather than manually keeping DAB's OIDC URL patterns in sync with DOT, consider adding a test that automatically detects which OIDC endpoints DAB has registered (by intersecting DAB's url names with DOT's `oidc_urlpatterns`) and asserts the regex patterns match. This would have caught the `userinfo` mismatch and would automatically cover any new DOT OIDC endpoints added to DAB in the future -- no hardcoded list to maintain:

```python
from oauth2_provider.urls import oidc_urlpatterns as dot_oidc_urlpatterns
from ansible_base.oauth2_provider.urls import oauth_urls

dot_oidc_names = {p.name for p in dot_oidc_urlpatterns}
dab_oidc_names = {
    getattr(p, 'name', None) for p in oauth_urls
} & dot_oidc_names

@pytest.mark.parametrize("name", sorted(dab_oidc_names))
def test_oidc_url_patterns_match_dot(name):
    """Ensure our OIDC URL patterns stay aligned with the pinned DOT version."""
    dot_pattern = next(p for p in dot_oidc_urlpatterns if p.name == name)
    dab_pattern = next(p for p in oauth_urls if getattr(p, 'name', None) == name)
    assert dab_pattern.pattern._regex == dot_pattern.pattern._regex, (
        f"URL pattern for '{name}' differs from django-oauth-toolkit. "
        f"DAB: {dab_pattern.pattern._regex}, DOT: {dot_pattern.pattern._regex}"
    )
```

3. +1 to PabloHiro's comment about updating `test_oidc_endpoints_shown_when_flag_enabled` to use the new `feature_flag_enabled` context manager.

dleehr:
> Rather than manually keeping DAB's OIDC URL patterns in sync with DOT, consider adding a test that automatically detects which OIDC endpoints DAB has registered (by intersecting DAB's url names with DOT's oidc_urlpatterns) and asserts the regex patterns match. 

I added a test file locally with your suggestion, which ran perfectly. but there's two problems:

1. Our regex pattern for `jwks.json` escapes the `.` but the [d-o-t pattern does not](https://github.com/django-oauth/django-oauth-toolkit/blob/2.3.0/oauth2_provider/urls.py#L39), so the test will fail. I could adopt the "buggy" behavior to pass this test, but that seems like an antipattern. I could also remove this from the set for versions of d-o-t without the fix, maybe I'll do that.
2. The pattern for `userinfo/` doesn't match. I see you mentioned "This would have caught the `userinfo` mismatch" but I don't have context on that. I can change that (we don't have a specific requirement for the `userinfo` URL for this feature but the discovery view needs to be able to `reverse` it), but is something else expecting that?

```
================================================================== short test summary info ===================================================================
FAILED test_app/tests/oauth2_provider/test_oidc_urlpatterns.py::test_oidc_url_patterns_match_dot[jwks-info] - AssertionError: URL pattern for 'jwks-info' differs from django-oauth-toolkit. DAB: ^\.well-known/jwks\.json$, DOT: ^\.well-known/jwks.json$
assert '^\\.well-known/jwks\\.json$' == '^\\.well-known/jwks.json$'
  
  - ^\.well-known/jwks.json$
  + ^\.well-known/jwks\.json$
  ?                   +
FAILED test_app/tests/oauth2_provider/test_oidc_urlpatterns.py::test_oidc_url_patterns_match_dot[user-info] - AssertionError: URL pattern for 'user-info' differs from django-oauth-toolkit. DAB: ^userinfo$, DOT: ^userinfo/$
assert '^userinfo$' == '^userinfo/$'
  
  - ^userinfo/$
  ?          -
  + ^userinfo$
```

ansible_base/lib/testing/util.py
dleehr: Done

dleehr:
Thanks for the feedback @john-westcott-iv and @PabloHiro. This is ready for re-review.

Also I looked into the failing DAB Consumer check. That's a runtime issue in `requests` that checks the version of `chardet`. Not sure where I'd fix that (here or awx) but once `requests` publishes a release with this commit  https://github.com/psf/requests/commit/4bd79e397304d46dfccd76f36c07f66c0295ff82 it should be fixed.

john-westcott-iv:
**Re: `userinfo/` trailing slash**

I don't see anything in the gateway or DAB that references this endpoint directly -- the only consumer appears to be DOT's discovery view using `reverse()`. My comment about `userinfo` was just pointing out the value of the test catching pre-existing drift. The trailing slash change should be safe, and `APPEND_SLASH=True` (Django's default, which neither gateway nor DAB overrides) provides a redirect safety net for any callers using the old path.

**Re: `jwks.json` regex**

Rather than carving out `jwks-info` for DOT < 3.0.0, consider changing the test assertion from regex string equality to a functional match. Instead of `dab_pattern._regex == dot_pattern._regex`, generate the canonical path (e.g. via `reverse()`) and assert that DAB's compiled regex matches it. That way DAB's correctly-escaped `jwks\.json` still passes (it matches the literal path `.well-known/jwks.json`), but you'd still catch real mismatches like the missing trailing slashes. No version-specific exclusions needed, and the test validates what actually matters -- that DAB can route the same URLs DOT expects.

john-westcott-iv (approved):
Approving with minor comments/potential options for the test.

dleehr:
> Rather than carving out jwks-info for DOT < 3.0.0, consider changing the test assertion from regex string equality to a functional match. Instead of dab_pattern._regex == dot_pattern._regex, generate the canonical path (e.g. via reverse()) and assert that DAB's compiled regex matches it.

Thanks for the feedback and appreciate the review/approval. In versions after 2.3.0, some of these url patterns are  `path` instead of `re_path`, so they wouldn't even yield a `._regex` to compile, would they? 

The bug fix itself is a single character, and the rest is to spotlight things to change when we eventually upgrade `django-oauth-toolkit` by failing some focused tests. So I'll plan to move forward without addtional changes.

PabloHiro (approved):

PabloHiro:
For more context about why the regex was different in the first place, it is because the OIDC standard does not include trailing slashes as part of the standard. But I guess the industry has walked a different path and we were trying to be too smart :sweat:

dleehr:
> For more context about why the regex was different in the first place, it is because the OIDC standard does not include trailing slashes as part of the standard. But I guess the industry has walked a different path and we were trying to be too smart 😓

[Yep](https://github.com/ansible/django-ansible-base/pull/915/changes#r2709849119), we were just ahead of our time 😁. In the most recent release of `django-oauth-toolkit` 3.2.0 [the trailing slash is deprecated to align with the OIDC standard](https://github.com/django-oauth/django-oauth-toolkit/blob/3.2.0/oauth2_provider/urls.py#L46-L51) and "shall be removed in a future release":

```python
    # .well-known/openid-configuration/ is deprecated
    # https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig
    # does not specify a trailing slash
    # Support for trailing slash shall be removed in a future release.
    re_path(
        r"^\.well-known/openid-configuration/?$",
```

The corresponding logic in [oidc_issuer()](https://github.com/django-oauth/django-oauth-toolkit/blob/3.2.0/oauth2_provider/settings.py#L317) has no `/`, so that would keep these in agreement:

```python
        return abs_url[: -len("/.well-known/openid-configuration")]
```

We need the trailing slash in this pattern for now so that the version of `oidc_issuer()` we get in 2.4.0 works, but when we upgrade django-oauth-toolkit, we should be able to more closely align with the standard.

dleehr:
The DAB Consumer check now succeeds with the latest [ghcr.io/ansible/awx_devel:devel](https://github.com/ansible/awx/actions/runs/22397762491) build. Yesterday's failures were due to a `tox` dependency causing a runtime assertion to fail in `requests`. `tox` removed the dependency, released a new version, and now the check succeeds again.

PR #950, opened 2026-02-24
AAP-64919 Mask encrypted fields in audit log/activitystream

bhavenst:
## Description

GW preferences contain 'raw_value' model field that is not set as encrypted.  They also contain a flag for encryption though, that results in encrypted values being stored in raw_value.  Those were being dumped in the activitystream and audit log looking like

`changed from '' to '$encrypted$UTF8$AESCBC$Z0FBQUFBQnBuTVBFckM5...'`

which isn't really acceptable.

This PR does additional filtering to handle this case and results in

`changed from '$encrypted$' to '$encrypted$'`

 and prevents info leakage.

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Run GW
2. Change encrypted preference
3. Examine log

### Expected Results
No more cipher/ciphertext info leaked into log or activitystream

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistently mask encrypted values in model diffs (added, removed, modified), honoring instance-level encrypted-field hints and values explicitly marked as encrypted; respects the sanitize toggle.
* **Tests**
  * Added tests covering sanitization for prefixed encrypted values, instance-level encrypted-field names, partial-side sanitization, and disabled-sanitization behavior.

### DIFF
--- a/ansible_base/lib/utils/models.py
+++ b/ansible_base/lib/utils/models.py
@@ -322,7 +322,7 @@ def diff(
     # Get any new fields from the new_fields - old_fields
     for field in new_fields_set - old_fields_set:
         val = fields['new'][field]
-        model_diff.added_fields[field] = ENCRYPTED_STRING if _is_sensitive(new, new_model, field, val, sanitize_encrypted) else val
+        model_diff.added_fields[field] = _sanitize_value(new, new_model, field, val, sanitize_encrypted)
 
     # Find any modified fields from the union of the sets
     for field in new_fields_set & old_fields_set:
@@ -330,8 +330,8 @@ def diff(
             old_val = fields['old'][field]
             new_val = fields['new'][field]
             model_diff.changed_fields[field] = (
-                ENCRYPTED_STRING if _is_sensitive(old, old_model, field, old_val, sanitize_encrypted) else old_val,
-                ENCRYPTED_STRING if _is_sensitive(new, new_model, field, new_val, sanitize_encrypted) else new_val,
+                _sanitize_value(old, old_model, field, old_val, sanitize_encrypted),
+                _sanitize_value(new, new_model, field, new_val, sanitize_encrypted),
             )
 
     return model_diff
--- a/ansible_base/lib/utils/models.py
+++ b/ansible_base/lib/utils/models.py
@@ -166,6 +166,32 @@ def is_encrypted_field(model, field_name):
     return field_name in getattr(model, 'encrypted_fields', [])
 
 
+def _is_sensitive(instance, model, field_name, value, sanitize_encrypted):
+    """Check whether a field value should be replaced with ENCRYPTED_STRING.
+
+    Extends the class-level ``is_encrypted_field`` check with two additional
+    heuristics so that models with *instance-level* encryption (e.g. a
+    ``Preference`` whose ``raw_value`` is only sometimes encrypted) can
+    participate in sanitization:
+
+    1. **Instance attribute** – if the instance carries an
+       ``_encrypted_field_names`` set/list, fields listed there are treated as
+       encrypted.  Models should populate this in ``from_db()`` / ``save()``
+       when they detect that a particular field holds an encrypted value.
+    2. **Value prefix** – if the string value starts with the well-known
+       ``$encrypted$`` marker, it is replaced regardless of field metadata.
+    """
+    if not sanitize_encrypted:
+        return False
+    if is_encrypted_field(model, field_name):
+        return True
+    if instance is not None and field_name in getattr(instance, '_encrypted_field_names', set()):
+        return True
+    if isinstance(value, str) and value.startswith(ENCRYPTED_STRING):
+        return True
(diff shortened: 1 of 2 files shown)


bhavenst:
AWX consumer test (non voting) does not look related to changes.

ansible_base/lib/utils/models.py
```diff
@@ -166,6 +166,32 @@ def is_encrypted_field(model, field_name):
...
+    """
+    if not sanitize_encrypted:
+        return False
+    if is_encrypted_field(model, field_name):
+        return True
+    if instance is not None and field_name in getattr(instance, '_encrypted_field_names', set()):
+        return True
+    if isinstance(value, str) and value.startswith(ENCRYPTED_STRING):
+        return True
+    return False
```
john-westcott-iv: Consider refactoring this to return the sanitized value directly (e.g. rename to `_sanitize_value()`). The function already receives `value` as a parameter, so instead of returning a boolean that forces every call site to do `ENCRYPTED_STRING if _is_sensitive(...) else val`, it can return the appropriate value directly. This eliminates the repeated ternary at all three call sites below and reduces the chance of getting the pattern wrong.

Also worth noting: this function now properly respects the `sanitize_encrypted` flag, which was previously declared as a parameter on `diff()` but never actually referenced in the function body. That's a good fix, but it is a behavioral change — any caller currently passing `sanitize_encrypted=False` was being silently ignored and will now see different results. Might be worth calling out in the PR description.

```suggestion
def _sanitize_value(instance, model, field_name, value, sanitize_encrypted):
    """Return *value* unchanged or ``ENCRYPTED_STRING`` if the field is sensitive.

    Extends the class-level ``is_encrypted_field`` check with two additional
    heuristics so that models with *instance-level* encryption (e.g. a
    ``Preference`` whose ``raw_value`` is only sometimes encrypted) can
    participate in sanitization:

    1. **Instance attribute** -- if the instance carries an
       ``_encrypted_field_names`` set/list, fields listed there are treated as
       encrypted.  Models should populate this in ``from_db()`` / ``save()``
       when they detect that a particular field holds an encrypted value.
    2. **Value prefix** -- if the string value starts with the well-known
       ``$encrypted$`` marker, it is replaced regardless of field metadata.
    """
    if not sanitize_encrypted:
        return value
    if is_encrypted_field(model, field_name):
        return ENCRYPTED_STRING
    if instance is not None and field_name in getattr(instance, '_encrypted_field_names', set()):
        return ENCRYPTED_STRING
    if isinstance(value, str) and value.startswith(ENCRYPTED_STRING):
        return ENCRYPTED_STRING
    return value
```

ansible_base/lib/utils/models.py
```diff
@@ -291,18 +317,22 @@ def diff(
 
     # Get any removed fields from the old_fields - new_fields
     for field in old_fields_set - new_fields_set:
-        model_diff.removed_fields[field] = ENCRYPTED_STRING if is_encrypted_field(old_model, field) else fields['old'][field]
+        val = fields['old'][field]
+        model_diff.removed_fields[field] = ENCRYPTED_STRING if _is_sensitive(old, old_model, field, val, sanitize_encrypted) else val
```
john-westcott-iv: With the suggested `_sanitize_value()` refactor, these two lines collapse to one — and the same applies to the `added_fields` and `changed_fields` blocks below:

```suggestion
        model_diff.removed_fields[field] = _sanitize_value(old, old_model, field, fields['old'][field], sanitize_encrypted)
```

john-westcott-iv (commented):
## Code Review Summary

Good security fix — preventing encrypted ciphertext from leaking into audit logs and activity streams is important. The approach of adding instance-level and prefix-based sensitivity detection is sound and handles the `Preference.raw_value` edge case well.

**Files Reviewed**: 2 files

### Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 2 code quality / design suggestions

### Overall Assessment
Solid PR. The logic is correct and the test coverage is good. Two suggestions:

1. **Refactor `_is_sensitive` to `_sanitize_value`** — have the helper return the appropriate value (`ENCRYPTED_STRING` or the original) instead of a boolean. This eliminates the repeated ternary pattern at every call site and makes the code cleaner.

2. **Document the `sanitize_encrypted` behavioral change** — the `sanitize_encrypted` parameter on `diff()` was previously declared but never referenced in the function body. This PR incidentally fixes that by passing it through to `_is_sensitive`. That's a good fix, but it's a silent behavioral change for any existing callers passing `sanitize_encrypted=False`. Worth a note in the PR description.

### General Feedback
- Test coverage looks comprehensive — prefix detection, instance-level field names, partial sanitization, and the disabled flag are all covered
- The `_encrypted_field_names` convention for instance-level tracking is a reasonable pattern
- The docstring on the helper is thorough

john-westcott-iv (approved):
Approving, refactor is optional.

PR #952, opened 2026-02-24
[AAP-63312] Add workload_ttl_seconds field to WorkloadIdentityTokenRequestSerializer

arrestle:
[AAP-63312] Add workload_ttl_seconds field to WorkloadIdentityTokenRequestSerializer

## Description

**What is being changed?**

Adding optional `workload_ttl_seconds` parameter to `WorkloadIdentityTokenRequestSerializer` to support workload-specific JWT TTL overrides.

Gateway's AAP-63312 implementation requires the ability to accept workload-specific TTL values from API requests. This serializer change enables the API contract for that feature.

Adds a new optional `IntegerField` to the request serializer that validates TTL override values while maintaining backward compatibility (field is optional).

## Type of Change

- [X] Test update

## Self-Review Checklist

## Implementation

**Modified:** `ansible_base/lib/workload_identity/workload_identity_tokens.py` (11 lines)  
**Modified:** `test_app/tests/lib/workload_identity/test_serializers.py` (86 lines)

Added optional `workload_ttl_seconds` field with validation (min_value=0, allow_null, not required).

## Testing Instructions

```bash
# Run serializer tests
tox -e py -- test_app/tests/lib/workload_identity/test_serializers.py -v
```

**Expected:** 29 tests pass (22 existing + 7 new)

## Additional Context

**Backward compatible:** Field is optional, existing requests work unchanged.

**Requires coordinated Gateway PR** to implement TTL calculation logic.

**Related:** AAP-63296 (P4.1), ANSTRAT-1019

---

**Development Note:** This implementation was developed with assistance from Claude AI assistant.

https://handbook.eng.ansible.com/proposals/ANSTRAT-1019-Configurable-JWT-Expiration-for-OIDC-Integrations

**Assisted-by:** Claude

## Summary by CodeRabbit

* **New Features**
  * Optional workload_ttl_seconds parameter to request per-workload token TTL overrides (nullable or omitted to use platform default); accepts a positive integer up to a 24-hour ceiling and applies an automatic 60s clock-skew buffer.

* **Tests**
  * Comprehensive tests covering valid value, zero (rejected), null, omission, negative, exceeding max, and non‑integer inputs to verify behavior and error messages.

### DIFF
--- a/ansible_base/lib/workload_identity/workload_identity_tokens.py
+++ b/ansible_base/lib/workload_identity/workload_identity_tokens.py
@@ -1,6 +1,15 @@
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
 
+# Hard ceiling for workload-specific TTL overrides. 24 hours is intentionally
+# generous — workloads needing longer lifetimes should reconsider their design.
+# Terraform Cloud ties exp to the run timeout with no imposed ceiling for workspace
+# runs, but does enforce a max (30 min) for module test tokens. We follow the
+# bounded approach here as a safety net against misconfigured or unlimited timeouts.
+# See: https://developer.hashicorp.com/terraform/cloud-docs/dynamic-provider-credentials/workload-identity-tokens
+# A preference-driven maximum can replace this constant in a follow-up.
+WORKLOAD_TTL_MAX_SECONDS = 24 * 60 * 60  # 86400 or 24 hours
+
 
 class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
     """
@@ -28,11 +37,12 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
     workload_ttl_seconds = serializers.IntegerField(
         required=False,
         allow_null=True,
-        min_value=0,
+        min_value=1,
+        max_value=WORKLOAD_TTL_MAX_SECONDS,
         help_text=_(
-            "Optional workload-specific TTL override in seconds. "
-            "If provided and > 0, overrides the platform default. "
-            "If omitted or 0, uses platform fallback (jwt_default_ttl_seconds). "
+            f"Optional workload-specific TTL override in seconds (1–{WORKLOAD_TTL_MAX_SECONDS}). "
+            "If provided, overrides the platform default for this token. "
+            "Omit or set to null to use the platform fallback (jwt_default_ttl_seconds). "
             "A 60s clock skew offset is automatically added to all JWTs."
         ),
     )
--- a/ansible_base/lib/workload_identity/workload_identity_tokens.py
+++ b/ansible_base/lib/workload_identity/workload_identity_tokens.py
@@ -3,10 +3,6 @@ from rest_framework import serializers
 
 # Hard ceiling for workload-specific TTL overrides. 24 hours is intentionally
 # generous — workloads needing longer lifetimes should reconsider their design.
-# Terraform Cloud ties exp to the run timeout with no imposed ceiling for workspace
-# runs, but does enforce a max (30 min) for module test tokens. We follow the
-# bounded approach here as a safety net against misconfigured or unlimited timeouts.
-# See: https://developer.hashicorp.com/terraform/cloud-docs/dynamic-provider-credentials/workload-identity-tokens
 # A preference-driven maximum can replace this constant in a follow-up.
 WORKLOAD_TTL_MAX_SECONDS = 24 * 60 * 60  # 86400 or 24 hours
 
--- a/ansible_base/lib/workload_identity/workload_identity_tokens.py
+++ b/ansible_base/lib/workload_identity/workload_identity_tokens.py
@@ -34,7 +34,7 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
         required=False,
         allow_null=True,
         min_value=1,
-        max_value=WORKLOAD_TTL_MAX_SECONDS,
+        max_value=get_setting('ANSIBLE_BASE_WIT_MAX_TOKEN_TTL', WORKLOAD_TTL_MAX_SECONDS),
         help_text=_(
             "Optional workload-specific TTL override in seconds (1-86400). "
             "If provided, overrides the platform default for this token. "
--- a/ansible_base/lib/workload_identity/workload_identity_tokens.py
+++ b/ansible_base/lib/workload_identity/workload_identity_tokens.py
@@ -40,7 +40,7 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
         min_value=1,
         max_value=WORKLOAD_TTL_MAX_SECONDS,
         help_text=_(
-            f"Optional workload-specific TTL override in seconds (1–{WORKLOAD_TTL_MAX_SECONDS}). "
+            "Optional workload-specific TTL override in seconds (1-86400). "
             "If provided, overrides the platform default for this token. "
             "Omit or set to null to use the platform fallback (jwt_default_ttl_seconds). "
             "A 60s clock skew offset is automatically added to all JWTs."
--- a/ansible_base/lib/workload_identity/workload_identity_tokens.py
+++ b/ansible_base/lib/workload_identity/workload_identity_tokens.py
@@ -1,12 +1,13 @@
+from django.core.validators import MinValueValidator
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
 
 from ansible_base.lib.utils.settings import get_setting
 
-# Hard ceiling for workload-specific TTL overrides. 24 hours is intentionally
-# generous — workloads needing longer lifetimes should reconsider their design.
-# A preference-driven maximum can replace this constant in a follow-up.
-WORKLOAD_TTL_MAX_SECONDS = 24 * 60 * 60  # 86400 or 24 hours
+# Default ceiling for workload-specific TTL overrides. 24 hours is intentionally
+# generous -- workloads needing longer lifetimes should reconsider their design.
+# Override with the ANSIBLE_BASE_WIT_MAX_TOKEN_TTL Django setting if needed.
+WORKLOAD_TTL_MAX_SECONDS = 24 * 60 * 60  # 86400
 
 
 class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
@@ -35,8 +36,7 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
     workload_ttl_seconds = serializers.IntegerField(
         required=False,
         allow_null=True,
-        min_value=1,
-        max_value=get_setting('ANSIBLE_BASE_WIT_MAX_TOKEN_TTL', WORKLOAD_TTL_MAX_SECONDS),
+        validators=[MinValueValidator(1)],
         help_text=_(
             "Optional workload-specific TTL override in seconds. "
             "If provided, overrides the platform default for this token. "
@@ -44,6 +44,16 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
             "A 60s clock skew offset is automatically added to all JWTs."
         ),
     )
+
+    def validate_workload_ttl_seconds(self, value):
+        if value is not None:
+            max_ttl = get_setting('ANSIBLE_BASE_WIT_MAX_TOKEN_TTL', WORKLOAD_TTL_MAX_SECONDS)
+            if value > max_ttl:
+                raise serializers.ValidationError(
+                    detail=_('Ensure this value is less than or equal to %(max_value)s.') % {'max_value': max_ttl},
+                    code='max_value',
+                )
+        return value
 
 
 class WorkloadIdentityTokenResponseSerializer(serializers.Serializer):
--- a/ansible_base/lib/workload_identity/workload_identity_tokens.py
+++ b/ansible_base/lib/workload_identity/workload_identity_tokens.py
@@ -1,6 +1,8 @@
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
 
+from ansible_base.lib.utils.settings import get_setting
+
 # Hard ceiling for workload-specific TTL overrides. 24 hours is intentionally
 # generous — workloads needing longer lifetimes should reconsider their design.
 # A preference-driven maximum can replace this constant in a follow-up.
@@ -36,7 +38,7 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
         min_value=1,
         max_value=get_setting('ANSIBLE_BASE_WIT_MAX_TOKEN_TTL', WORKLOAD_TTL_MAX_SECONDS),
         help_text=_(
-            "Optional workload-specific TTL override in seconds (1-86400). "
+            "Optional workload-specific TTL override in seconds. "
             "If provided, overrides the platform default for this token. "
             "Omit or set to null to use the platform fallback (jwt_default_ttl_seconds). "
             "A 60s clock skew offset is automatically added to all JWTs."
--- a/ansible_base/lib/workload_identity/workload_identity_tokens.py
+++ b/ansible_base/lib/workload_identity/workload_identity_tokens.py
@@ -46,6 +46,10 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
     )
 
     def validate_workload_ttl_seconds(self, value):
+        # max_value can't be declared on the field because this module is in
+        # ansible_base/lib/ which must be importable without Django configured
+        # (enforced by the pure-python-imports CI check). Deferring to a
+        # validate method evaluates get_setting() at runtime instead of import time.
         if value is not None:
             max_ttl = get_setting('ANSIBLE_BASE_WIT_MAX_TOKEN_TTL', WORKLOAD_TTL_MAX_SECONDS)
             if value > max_ttl:
--- a/ansible_base/lib/workload_identity/workload_identity_tokens.py
+++ b/ansible_base/lib/workload_identity/workload_identity_tokens.py
@@ -1,4 +1,3 @@
-from django.core.validators import MinValueValidator
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
 
@@ -36,7 +35,7 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
     workload_ttl_seconds = serializers.IntegerField(
         required=False,
         allow_null=True,
-        validators=[MinValueValidator(1)],
+        min_value=1,
         help_text=_(
             "Optional workload-specific TTL override in seconds. "
             "If provided, overrides the platform default for this token. "
--- a/ansible_base/lib/workload_identity/workload_identity_tokens.py
+++ b/ansible_base/lib/workload_identity/workload_identity_tokens.py
@@ -25,6 +25,17 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
         allow_empty=False,
         help_text=_("Workload details to include in the JWT as claims."),
     )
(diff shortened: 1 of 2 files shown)


ansible_base/lib/workload_identity/workload_identity_tokens.py
```diff
@@ -25,6 +25,17 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
...
+        required=False,
+        allow_null=True,
+        min_value=0,
+        help_text=_(
+            "Optional workload-specific TTL override in seconds. "
+            "If provided and > 0, overrides the platform default. "
+            "If omitted or 0, uses platform fallback (jwt_default_ttl_seconds). "
+            "A 60s clock skew offset is automatically added to all JWTs."
+        ),
+    )
```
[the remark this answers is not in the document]
matoval: I have the same concern that there is no max_value.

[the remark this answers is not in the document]
arrestle: Note this was called out in the [handbook](https://handbook.eng.ansible.com/proposals/ANSTRAT-1019-Configurable-JWT-Expiration-for-OIDC-Integrations) under key constraints and approved.

That being said if there is consensus we can revisit.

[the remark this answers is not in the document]
john-westcott-iv: +1 to @matoval and CodeRabbit here. A configurable max actually conforms to "no hard maximum" — if the default ceiling isn't long enough, the admin simply extends it. But it prevents tokens from being valid for effectively infinite durations, which has been a real complaint with the OAuth tokens we generate today.

The serializer is the ideal place for this — it's the first validation layer, so invalid values never reach the view. Even a hardcoded `max_value=86400` (24h) as a starting point would be a big improvement, with a follow-up to make it preference-driven if needed.

john-westcott-iv (commented):
## Code Review Summary

Clean serializer addition — the field definition is well-structured with proper optionality and the test coverage hits all the boundary cases (valid, zero, null, omitted, negative, non-integer).

**Files Reviewed**: 2 files
**Comments Posted**: 1 reply to existing thread

### Issues Found
- 1 security suggestion (no `max_value` — reinforced existing thread with additional context)
- 0 correctness/logic issues
- 0 code quality suggestions

### Overall Assessment
LGTM with the `max_value` addition. Replied to the existing thread with +1 and rationale — a configurable max conforms to "no hard maximum" while preventing effectively infinite token lifetimes (a known complaint from our OAuth token experience).

### General Feedback
- Test assertions on error codes (`min_value`, `invalid`) are a nice touch
- `allow_null=True` + `required=False` combination correctly handles both omission and explicit null

ansible_base/lib/workload_identity/workload_identity_tokens.py
```diff
@@ -25,6 +25,17 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
...
     )
+    workload_ttl_seconds = serializers.IntegerField(
+        required=False,
+        allow_null=True,
+        min_value=0,
+        help_text=_(
+            "Optional workload-specific TTL override in seconds. "
+            "If provided and > 0, overrides the platform default. "
+            "If omitted or 0, uses platform fallback (jwt_default_ttl_seconds). "
+            "A 60s clock skew offset is automatically added to all JWTs."
```
dleehr: Can you help me understand the use case for having `0` be a special case in the serializer and the platform API? I understand why omitting the field in this API request should fall back to a platform-side default, but don't follow why a `0` value is a special case. It means that even after we validate the data in the serializer, we still have to look for a special value. Why do we need to handle `0` differently?

ansible_base/lib/workload_identity/workload_identity_tokens.py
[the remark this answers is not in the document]
dleehr: Thank you @arrestle for highlighting the decision in the proposal. We shouldn't deviate from those decisions without a conversation, but I do think @matoval and @john-westcott-iv raise reasonable concerns about allowing this value to be unlimited, and it seems like a deviation we should consider.

But, before we make that decision, can you provide some context about why it needs to be unlimited in this initial feature?

 @john-westcott-iv offers a suggestion to have a simple 24h limit at first and followed up by a preference. Sounds like that would address his concern. Does that break the feature?

dleehr (changes requested):
Nice use of the serializer `min_value` to let validation handle this. Requesting changes around the special-casing of `0` (also noted in the other PR) and pending a decision on the `max_value`

ansible_base/lib/workload_identity/workload_identity_tokens.py
arrestle: It was in the handbook https://handbook.eng.ansible.com/proposals/ANSTRAT-1019-Configurable-JWT-Expiration-for-OIDC-Integrations line 15. Again thinking we should revisit that see https://github.com/ansible/handbook/pull/1206/ suggested changes.

ansible_base/lib/workload_identity/workload_identity_tokens.py
```diff
@@ -1,6 +1,15 @@
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
 
+# Hard ceiling for workload-specific TTL overrides. 24 hours is intentionally
+# generous — workloads needing longer lifetimes should reconsider their design.
+# Terraform Cloud ties exp to the run timeout with no imposed ceiling for workspace
+# runs, but does enforce a max (30 min) for module test tokens. We follow the
+# bounded approach here as a safety net against misconfigured or unlimited timeouts.
+# See: https://developer.hashicorp.com/terraform/cloud-docs/dynamic-provider-credentials/workload-identity-tokens
+# A preference-driven maximum can replace this constant in a follow-up.
```
dleehr: Let's remove these mentions of Terraform and hashicorp.com from django-ansible-base.

dleehr (changes requested):
Looks good, but the terraform-related comments lack context and may be confusing. Please remove those.

ansible_base/lib/workload_identity/workload_identity_tokens.py
dleehr: Thanks, I hadn't reviewed that for the `0`. Sounds like you agree we don't need to handle `0` explicitly in the serializer/API interface, so I do think it simplifies things to remove that special case.

dleehr (approved):

test_app/tests/lib/workload_identity/test_serializers.py
```diff
@@ -218,6 +219,109 @@ def test_extra_fields_ignored(self):
...
+            'audience': 'https://api.example.com',
+            'claims': {'job_name': 'test-job'},
+            'workload_ttl_seconds': WORKLOAD_TTL_MAX_SECONDS + 1,
+        }
+        serializer = WorkloadIdentityTokenRequestSerializer(data=invalid_data)
+        assert not serializer.is_valid()
+        assert 'workload_ttl_seconds' in serializer.errors
+        assert serializer.errors['workload_ttl_seconds'][0].code == 'max_value'
+
+    def test_workload_ttl_seconds_not_integer_fails(self):
```
dragid10: I wonder how a float would act 🤔

dragid10 (approved):
(as DAB rep)
Overall looks good
Low Care Amount: I see there is a test for not an integer (tests with a string). I'd imagine a float would also be rejected? Do we need a test to check it doesn't slip through via truncation or something?

test_app/tests/lib/workload_identity/test_serializers.py
dleehr: Good question. On that note, we have quite a few separate tests for invalid values of `workload_ttl_seconds` (negative, above max, string, perhaps float) that we can parameterize instead of writing separate test functions.

I've already approved, so this is also low-care for me, but we should try to use table-driven/parameterized tests where possible.

ansible_base/lib/workload_identity/workload_identity_tokens.py
```diff
@@ -25,6 +30,18 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
         allow_empty=False,
         help_text=_("Workload details to include in the JWT as claims."),
     )
+    workload_ttl_seconds = serializers.IntegerField(
+        required=False,
+        allow_null=True,
+        min_value=1,
+        max_value=WORKLOAD_TTL_MAX_SECONDS,
```
john-westcott-iv: Minor suggestion for future-proofing: consider using `get_setting` here so individual services can configure the max TTL via their Django settings without code changes. Keep `WORKLOAD_TTL_MAX_SECONDS` as the hardcoded default, but allow per-service override:

```suggestion
        max_value=get_setting('ANSIBLE_BASE_WIT_MAX_TOKEN_TTL', WORKLOAD_TTL_MAX_SECONDS),
```

This would require adding the import:
```python
from ansible_base.lib.utils.settings import get_setting
```

Note: since this evaluates at class definition time, it's not runtime-dynamic — but it does allow per-service configuration via Django settings, which is the established DAB pattern (see `ID_TOKEN_MAX_AGE` in the OIDC plugin). Low priority, could be a follow-up.

ansible_base/lib/workload_identity/workload_identity_tokens.py
```diff
@@ -25,6 +30,18 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
...
+        help_text=_(
+            "Optional workload-specific TTL override in seconds (1-86400). "
```
john-westcott-iv: The `(1-86400)` range in help_text isn't a pattern used elsewhere in DAB — comparable fields like `ID_TOKEN_MAX_AGE` (OIDC) and `PORT` (TACACS/RADIUS) don't include their ranges in help_text. DRF exposes `min_value`/`max_value` in the API schema automatically, so this is redundant and creates a staleness risk if the bounds ever change.

```suggestion
            "Optional workload-specific TTL override in seconds. "
```

test_app/tests/lib/workload_identity/test_serializers.py
```diff
@@ -218,6 +219,109 @@ def test_extra_fields_ignored(self):
...
+            'scope': 'openid',
+            'audience': 'https://api.example.com',
+            'claims': {'job_name': 'test-job'},
+            'workload_ttl_seconds': 7200,
+        }
+        serializer = WorkloadIdentityTokenRequestSerializer(data=valid_data)
+        assert serializer.is_valid(), f"Serializer errors: {serializer.errors}"
+        assert serializer.validated_data['workload_ttl_seconds'] == 7200
+
+    def test_workload_ttl_seconds_zero_fails(self):
```
john-westcott-iv: +1 to @dragid10's question and @dleehr's follow-up about parameterizing. The four invalid-value tests (`zero_fails`, `negative_fails`, `exceeds_max_fails`, `not_integer_fails`) share identical structure — great candidate for `@pytest.mark.parametrize`. A float case (e.g., `3.14`) would fit naturally into that parameterized set too.

Something like:
```python
@pytest.mark.parametrize(
    'ttl_value, expected_error_code',
    [
        (0, 'min_value'),
        (-100, 'min_value'),
        (WORKLOAD_TTL_MAX_SECONDS + 1, 'max_value'),
        ('not-a-number', 'invalid'),
        (3.14, 'invalid'),
    ],
)
def test_workload_ttl_seconds_invalid_values(self, ttl_value, expected_error_code):
    ...
```

Low priority since you already have approvals, but it'd tighten things up.

john-westcott-iv (approved):
## Code Review Summary

Clean serializer addition with all prior review feedback addressed. The field definition, optionality handling (`allow_null=True` + `required=False`), and test coverage are solid.

**Files Reviewed**: 2 files
**Comments Posted**: 3 inline suggestions

### Issues Found
- 0 security/correctness issues
- 3 minor quality suggestions (all non-blocking)

### Overall Assessment
LGTM — approving with minor suggestions:
1. Consider using `get_setting()` for `max_value` to enable per-service configurability (established DAB pattern)
2. Drop the hardcoded range `(1-86400)` from `help_text` to match DAB conventions and avoid staleness
3. Parameterize the invalid-value tests and add a float case per @dleehr and @dragid10's discussion

---
**Note:** This review was submitted with assistance from Claude AI assistant on behalf of John Westcott.

davemulford (approved):
Changes look good. I see there is a failing test but that's happening outside of the changes you've made, and you're following up with the django-ansible-base team about that.

dleehr (changes requested):
I'm requesting changes to block this merge for now. 

The most recent commits modify the validation logic that was approved and are now using custom logic and a previously-unused class. I need to understand that better before merging.

arrestle:
> Nice use of the serializer `min_value` to let validation handle this. Requesting changes around the special-casing of `0` (also noted in the other PR) and pending a decision on the `max_value`

The validation logic changed because max_value=get_setting(...) on the field was evaluated at class definition time and caused pure-python-imports to fail (Django settings aren’t available at import time).

PabloHiro (approved):

ansible_base/lib/workload_identity/workload_identity_tokens.py
```diff
@@ -35,8 +36,7 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
     workload_ttl_seconds = serializers.IntegerField(
         required=False,
         allow_null=True,
-        min_value=1,
-        max_value=get_setting('ANSIBLE_BASE_WIT_MAX_TOKEN_TTL', WORKLOAD_TTL_MAX_SECONDS),
+        validators=[MinValueValidator(1)],
```
dleehr: Why do we need `validators=[MinValueValidator(1)]` instead of  `min_value=1`? It looked like `min_value=1` was working just fine.

ansible_base/lib/workload_identity/workload_identity_tokens.py
```diff
@@ -45,6 +45,16 @@ class WorkloadIdentityTokenRequestSerializer(serializers.Serializer):
...
 
+    def validate_workload_ttl_seconds(self, value):
+        if value is not None:
+            max_ttl = get_setting('ANSIBLE_BASE_WIT_MAX_TOKEN_TTL', WORKLOAD_TTL_MAX_SECONDS)
+            if value > max_ttl:
+                raise serializers.ValidationError(
+                    detail=_('Ensure this value is less than or equal to %(max_value)s.') % {'max_value': max_ttl},
+                    code='max_value',
+                )
+        return value
```
dleehr, line 59: Can you add a comment that explains why we need a custom function to do this validation?

Reviewing this code, my first question is "Why do we need this? Can't we just use the declarative `max_value=get_setting(...)`?" I don't think the answer is obvious from reading the code, so that's a good case to drop a comment explaining why we had to do it the hard way.

dleehr (commented):
> The validation logic changed because max_value=get_setting(...) on the field was evaluated at class definition time and caused pure-python-imports to fail (Django settings aren’t available at import time).

Thanks for the explanation.

That's unfortunate. I see the simple change in 7b124f8028ae532c566e83caede505aea8210f66 and that looks good. But calling `get_setting` from the class definition causes a linter check to fail: https://github.com/ansible/django-ansible-base/actions/runs/22497968362/job/65177353199. @john-westcott-iv any thoughts on that? I did compare this to the pattern you referenced in https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/authentication/authenticator_plugins/oidc.py#L234, but the difference there is that `get_setting` is called from a function, so it doesn't break the linter. 

I'm OK with the proposed `max_value` validation logic, but please add a comment explaining why we need a custom validation function to validate against a static value.

For the `min_value` change - that stands out a bit more. Does `min_value=1` no longer work? Why do we now need to import and use `MinValueValidator`?

ansible_base/lib/workload_identity/workload_identity_tokens.py
john-westcott-iv: +1 to this. `validators=[MinValueValidator(1)]` is misleading because it doesn't reference the custom `validate_workload_ttl_seconds` method -- reading just the field definition, you'd think there's no upper limit validation at all.

Either:
1. Move *both* min and max checks into `validate_workload_ttl_seconds` so all field validation logic is in one place, or
2. Switch back to `min_value=1` on the field (which is the standard DRF approach) and keep the max in the validate method

Both are better than the current `validators=[MinValueValidator(1)]` which is an unusual way to express a static minimum.

ansible_base/lib/workload_identity/workload_identity_tokens.py
john-westcott-iv: +1 -- a comment here is important.

The reason `max_value=get_setting(...)` can't be used declaratively on the field is that this serializer lives in `ansible_base/lib/`, which is tested by the `pure-python-imports` CI check. That check imports all `lib/` modules without Django configured, and `get_setting()` touches `django.conf.settings` -- so calling it at class definition time (which happens at import time) causes an `UNEXPECTED FAIL`.

Something like:
```suggestion
    def validate_workload_ttl_seconds(self, value):
        # max_value can't be declared on the field because this module is in
        # ansible_base/lib/ which must be importable without Django configured
        # (enforced by the pure-python-imports CI check). Deferring to a
        # validate method evaluates get_setting() at runtime instead of import time.
        if value is not None:
            max_ttl = get_setting('ANSIBLE_BASE_WIT_MAX_TOKEN_TTL', WORKLOAD_TTL_MAX_SECONDS)
            if value > max_ttl:
                raise serializers.ValidationError(
                    detail=_('Ensure this value is less than or equal to %(max_value)s.') % {'max_value': max_ttl},
                    code='max_value',
                )
        return value
```

Side note: per `docs/apps_or_lib.md`, criterion #4 for when a feature should be a Django app is "Does the feature require settings?" -- and with `ANSIBLE_BASE_WIT_MAX_TOKEN_TTL`, the answer is now yes. The base class and scope definitions in this package are legitimately pure-Python lib material, but the serializer (with DRF dependency + configurable settings) is starting to scope-creep beyond what `lib/` is meant for. Might be worth a follow-up discussion about whether `workload_identity_tokens.py` should live in a proper DAB app at some point.

ansible_base/lib/workload_identity/workload_identity_tokens.py
arrestle: gotcha!

ansible_base/lib/workload_identity/workload_identity_tokens.py
arrestle: Createed [slack thread](https://redhat-internal.slack.com/archives/C0A048CSKHP/p1772466407986739) for followup.

dleehr: Thought I posted a comment here but apparently I closed the tab too early. I created AAP-67004 to follow this up. The point on `apps_or_lib` is a great call. I think we can probably move these serializers out of the repo and have them live with the views but otherwise everything else can stay in `lib`

ansible_base/lib/workload_identity/workload_identity_tokens.py
dleehr: @arrestle sorry, what's the resolution here? 

I asked why `MinValueValidator` is necessary, @john-westcott-iv had the same question and proposed two paths forward. I don't see any changes here in the PR to implement either suggestion, but now the comment thread is resolved with `validators=[MinValueValidator(1)]`

dleehr (changes requested):
I see that I was asked to re-review. The thread about `MinValueValidator` was resolved without addressing SME feedback from @john-westcott-iv. I followed up here and in slack see https://github.com/ansible/django-ansible-base/pull/952/changes#r2872845254.

Requesting changes since this is still outstanding.

ansible_base/lib/workload_identity/workload_identity_tokens.py
arrestle: ah. needed to push.

dleehr (approved):
Looks great!

PR #954, opened 2026-02-25
[AAP-63314] P4.4: Add workload_ttl_seconds to WorkloadIdentityClient

arrestle:
[AAP-63314] P4.4: Add workload_ttl_seconds to WorkloadIdentityClient

## Description

**What is being changed?**

Adding optional `workload_ttl_seconds` parameter to the `WorkloadIdentityClient.request_workload_jwt()` method and `WorkloadIdentityTokenRequest` NamedTuple so callers can pass a workload-specific TTL hint to the Gateway.

**NOTE: 0 → None change:** The client previously used `workload_ttl_seconds: int = 0` for platform fallback. Per the [ANSTRAT-1019 proposal PR#1206](https://github.com/ansible/handbook/pull/1206), `0` is invalid (Gateway serializer rejects it); omit or `null` signals platform fallback. Updated to `int | None = None` and omit the field when `None`.  This is because `0` is the same as **infinite**, which has caused issues in the past.

The Gateway (AAP-63312) can now calculate JWT TTL based on workload duration, but it requires the TTL to be sent in the request. This change extends the client library so Controller (AWX) can supply `job.timeout` when requesting JWTs.

- `WorkloadIdentityTokenRequest` NamedTuple: adds `workload_ttl_seconds: int = 0` (default 0 = use platform fallback)
- `request_workload_jwt()`: accepts `workload_ttl_seconds: int = 0` and includes it in the POST body
- Backward compatible — existing callers without the argument send `0`, triggering Gateway's platform-default fallback

**Relationship to other PRs:**

This is the **client half** of a two-part change:
1. **This PR (DAB):** exposes the parameter in the client library
2. **AWX PR (AAP-63314):** calls this method with `job.timeout` — see `awx/AAP-63314.pr.md`

Both depend on AAP-63312 (Gateway + serializer) being merged first so the endpoint accepts the parameter.

## Type of Change

- [X] Test update

## Self-Review Checklist

## Implementation

**Modified:** `ansible_base/resource_registry/workload_identity_client.py` (~20 lines)  
**Modified:** `test_app/tests/resource_registry/test_workload_identity_client.py` (+5 tests)

## Testing Instructions

```bash
tox -e py -- test_app/tests/resource_registry/test_workload_identity_client.py -v
```

**Expected:** All tests pass. New tests:
- `test_request_type_with_custom_ttl`
- `test_request_workload_jwt_workload_ttl_seconds[custom-ttl-sent-to-gateway]`
- `test_request_workload_jwt_workload_ttl_seconds[zero-ttl-uses-platform-default]`

**Related:** AAP-63312 (Gateway TTL calculation), AAP-63296 (P4.1), ANSTRAT-1019

---

**Assisted-by:** Claude

## Summary by CodeRabbit

* **New Features**
  * Workload identity tokens can include a configurable TTL (seconds); when provided it’s forwarded in requests, otherwise platform default is used.

* **Bug Fixes / Validation**
  * Negative TTL values are rejected to prevent invalid token requests.

* **Documentation**
  * Examples and usage guidance updated to show specifying TTL and its behavior when omitted.

* **Tests**
  * Unit tests added/refactored to cover default (omitted) and custom TTL forwarding and payload assertions.

### DIFF
--- a/ansible_base/resource_registry/workload_identity_client.py
+++ b/ansible_base/resource_registry/workload_identity_client.py
@@ -170,6 +170,8 @@ class WorkloadIdentityClient(BaseServiceClient):
             ...     workload_ttl_seconds=3600,
             ... )
         """
+        if workload_ttl_seconds < 0:
+            raise ValueError(f"workload_ttl_seconds must be >= 0, got {workload_ttl_seconds}")
         request_body = WorkloadIdentityTokenRequest(
             claims=claims,
             scope=scope,
--- a/ansible_base/resource_registry/workload_identity_client.py
+++ b/ansible_base/resource_registry/workload_identity_client.py
@@ -29,11 +29,12 @@ class WorkloadIdentityTokenRequest(NamedTuple):
     audience: str
     """Audience for the token - the external service that will validate it."""
 
-    workload_ttl_seconds: int = 0
+    workload_ttl_seconds: int | None = None
     """Optional workload-specific TTL override in seconds.
 
     If > 0, the Gateway uses this as the base TTL instead of the platform default.
-    Pass 0 (or omit) to use the platform fallback (jwt_default_ttl_seconds).
+    Pass None or omit to use the platform fallback (jwt_default_ttl_seconds).
+    Note: 0 is invalid — Gateway serializer rejects it (min_value=1); use None or omit.
     """
 
 
@@ -140,7 +141,7 @@ class WorkloadIdentityClient(BaseServiceClient):
         claims: dict,
         scope: str,
         audience: str,
-        workload_ttl_seconds: int = 0,
+        workload_ttl_seconds: int | None = None,
     ) -> WorkloadIdentityTokenResponse:
         """
         Request a workload identity token.
@@ -154,13 +155,13 @@ class WorkloadIdentityClient(BaseServiceClient):
             audience: Audience for the token - the external service that will validate it
             workload_ttl_seconds: Optional TTL override in seconds. If > 0, the Gateway
                 uses this as the base TTL instead of the platform default (jwt_default_ttl_seconds).
-                Pass 0 to use the platform fallback.
+                Pass None or omit to use the platform fallback. Note: 0 is invalid (Gateway rejects it).
 
         Returns:
             WorkloadIdentityTokenResponse: Token response with JWT
 
         Raises:
-            ValueError: If workload_ttl_seconds is negative
+            ValueError: If workload_ttl_seconds is 0 or negative
             TokenRequestError: If the request fails
 
         Example:
@@ -171,15 +172,17 @@ class WorkloadIdentityClient(BaseServiceClient):
             ...     workload_ttl_seconds=3600,
             ... )
         """
-        if workload_ttl_seconds < 0:
-            raise ValueError(f"workload_ttl_seconds must be >= 0, got {workload_ttl_seconds}")
-        request_body = WorkloadIdentityTokenRequest(
-            claims=claims,
-            scope=scope,
-            audience=audience,
-            workload_ttl_seconds=workload_ttl_seconds,
-        )
-        data = request_body._asdict()
+        if workload_ttl_seconds is not None and workload_ttl_seconds < 1:
+            raise ValueError(
+                f"workload_ttl_seconds must be None (platform fallback) or >= 1, got {workload_ttl_seconds}"
+            )
+        data = {
+            "claims": claims,
+            "scope": scope,
+            "audience": audience,
+        }
+        if workload_ttl_seconds is not None and workload_ttl_seconds > 0:
+            data["workload_ttl_seconds"] = workload_ttl_seconds
 
         logger.info(f"Requesting workload identity token with scope: {scope}")
         logger.debug(f"Claims: {claims}")
--- a/ansible_base/resource_registry/workload_identity_client.py
+++ b/ansible_base/resource_registry/workload_identity_client.py
@@ -29,6 +29,13 @@ class WorkloadIdentityTokenRequest(NamedTuple):
     audience: str
     """Audience for the token - the external service that will validate it."""
 
+    workload_ttl_seconds: int = 0
+    """Optional workload-specific TTL override in seconds.
+
+    If > 0, the Gateway uses this as the base TTL instead of the platform default.
+    Pass 0 (or omit) to use the platform fallback (jwt_default_ttl_seconds).
+    """
+
 
 class WorkloadIdentityTokenResponse(NamedTuple):
     """Response from workload identity token endpoint."""
@@ -133,6 +140,7 @@ class WorkloadIdentityClient(BaseServiceClient):
         claims: dict,
         scope: str,
         audience: str,
+        workload_ttl_seconds: int = 0,
     ) -> WorkloadIdentityTokenResponse:
         """
         Request a workload identity token.
@@ -144,6 +152,9 @@ class WorkloadIdentityClient(BaseServiceClient):
             claims: Dictionary containing workload details (e.g., job ID, name)
             scope: Token custom scopes string (e.g., 'aap_controller_automation_job')
             audience: Audience for the token - the external service that will validate it
+            workload_ttl_seconds: Optional TTL override in seconds. If > 0, the Gateway
+                uses this as the base TTL instead of the platform default (jwt_default_ttl_seconds).
+                Pass 0 to use the platform fallback.
 
         Returns:
             WorkloadIdentityTokenResponse: Token response with JWT
@@ -155,10 +166,16 @@ class WorkloadIdentityClient(BaseServiceClient):
             >>> response = client.request_workload_jwt(
             ...     claims={"id": 2, "name": "my-example-job"},
             ...     scope="aap_controller_automation_job",
-            ...     audience="https://vault.example.com"
+            ...     audience="https://vault.example.com",
+            ...     workload_ttl_seconds=3600,
             ... )
         """
-        request_body = WorkloadIdentityTokenRequest(claims=claims, scope=scope, audience=audience)
+        request_body = WorkloadIdentityTokenRequest(
+            claims=claims,
+            scope=scope,
+            audience=audience,
+            workload_ttl_seconds=workload_ttl_seconds,
+        )
         data = request_body._asdict()
 
         logger.info(f"Requesting workload identity token with scope: {scope}")
--- a/ansible_base/resource_registry/workload_identity_client.py
+++ b/ansible_base/resource_registry/workload_identity_client.py
@@ -160,6 +160,7 @@ class WorkloadIdentityClient(BaseServiceClient):
             WorkloadIdentityTokenResponse: Token response with JWT
 
         Raises:
+            ValueError: If workload_ttl_seconds is negative
             TokenRequestError: If the request fails
 
         Example:
--- a/ansible_base/resource_registry/workload_identity_client.py
+++ b/ansible_base/resource_registry/workload_identity_client.py
@@ -173,9 +173,7 @@ class WorkloadIdentityClient(BaseServiceClient):
             ... )
         """
         if workload_ttl_seconds is not None and workload_ttl_seconds < 1:
-            raise ValueError(
-                f"workload_ttl_seconds must be None (platform fallback) or >= 1, got {workload_ttl_seconds}"
-            )
+            raise ValueError(f"workload_ttl_seconds must be None (platform fallback) or >= 1, got {workload_ttl_seconds}")
         data = {
             "claims": claims,
             "scope": scope,
--- a/test_app/tests/resource_registry/test_workload_identity_client.py
+++ b/test_app/tests/resource_registry/test_workload_identity_client.py
@@ -232,7 +232,7 @@ class TestWorkloadIdentityClient:
         """Test that workload_ttl_seconds=0 or negative raises ValueError (Gateway rejects 0)."""
         client = WorkloadIdentityClient(base_url="https://gateway.example.com")
 
-        with pytest.raises(ValueError, match="must be None.*or >= 1"):
+        with pytest.raises(ValueError, match=r"must be None.*or >= 1"):
             client.request_workload_jwt(
                 claims={"id": 1, "name": "test-job"},
                 scope="aap_controller_automation_job",
--- a/test_app/tests/resource_registry/test_workload_identity_client.py
+++ b/test_app/tests/resource_registry/test_workload_identity_client.py
@@ -29,6 +29,18 @@ class TestWorkloadIdentityTokenTypes:
         assert request.claims == {"id": 2, "name": "my-example-job"}
         assert request.scope == "aap_controller_automation_job"
         assert request.audience == "https://vault.example.com"
+        assert request.workload_ttl_seconds == 0
+
+    def test_request_type_with_custom_ttl(self):
+        """Test that workload_ttl_seconds can be set on the request type."""
+        request = WorkloadIdentityTokenRequest(
+            claims={"id": 2, "name": "my-example-job"},
+            scope="aap_controller_automation_job",
+            audience="https://vault.example.com",
+            workload_ttl_seconds=3600,
+        )
+
+        assert request.workload_ttl_seconds == 3600
 
     def test_request_type_as_dict(self):
         """Test that WorkloadIdentityTokenRequest can be converted to dict."""
@@ -43,6 +55,7 @@ class TestWorkloadIdentityTokenTypes:
             "claims": {"id": 1, "name": "test-job"},
             "scope": "aap_controller_automation_job",
             "audience": "https://vault.example.com",
+            "workload_ttl_seconds": 0,
         }
 
     def test_response_type_creation(self):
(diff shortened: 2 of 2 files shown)


ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -155,10 +166,16 @@ def request_workload_jwt(
             >>> response = client.request_workload_jwt(
             ...     claims={"id": 2, "name": "my-example-job"},
             ...     scope="aap_controller_automation_job",
-            ...     audience="https://vault.example.com"
+            ...     audience="https://vault.example.com",
+            ...     workload_ttl_seconds=3600,
             ... )
         """
-        request_body = WorkloadIdentityTokenRequest(claims=claims, scope=scope, audience=audience)
+        request_body = WorkloadIdentityTokenRequest(
```
john-westcott-iv: +1 on CodeRabbit's validation suggestion, but I'd go simpler. The `isinstance`/`bool` checks are Python trivia that type checkers handle — the key runtime invariant is non-negativity.

Since the AWX caller will pass `job.timeout` here, a guard prevents the subtle case where a negative timeout (valid in AWX as "no timeout") gets forwarded as a negative TTL, hitting the Gateway serializer and returning a less-obvious 400.

```suggestion
        if workload_ttl_seconds < 0:
            raise ValueError(f"workload_ttl_seconds must be >= 0, got {workload_ttl_seconds}")
        request_body = WorkloadIdentityTokenRequest(
```

If you add this, also update the `Raises` docstring section to include `ValueError`.

test_app/tests/resource_registry/test_workload_identity_client.py
```diff
@@ -175,16 +187,43 @@ def test_request_workload_jwt_success(self, mock_request, mock_get_service_token
...
+            "workload_ttl_seconds": 0,
         }
         assert call_kwargs["headers"]["X-ANSIBLE-SERVICE-AUTH"] == "service-token"
         assert call_kwargs["verify"] is True
 
+    @pytest.mark.parametrize(
+        "workload_ttl_seconds,expected",
+        [
+            (3600, 3600),
+            (0, 0),
```
john-westcott-iv: Nit: if validation is added per the source comment, consider extending this parametrize with a negative-value error case to lock the behavior:

```python
@pytest.mark.parametrize(
    "workload_ttl_seconds,expected",
    [
        (3600, 3600),
        (0, 0),
    ],
    ids=["custom-ttl-sent-to-gateway", "zero-ttl-uses-platform-default"],
)
```

Plus a separate test:
```python
def test_request_workload_jwt_negative_ttl_raises(self):
    client = WorkloadIdentityClient(base_url="https://gateway.example.com")
    with pytest.raises(ValueError, match="must be >= 0"):
        client.request_workload_jwt(
            claims={"id": 1, "name": "test"},
            scope="aap_controller_automation_job",
            audience="https://vault.example.com",
            workload_ttl_seconds=-1,
        )
```

john-westcott-iv (commented):
## 📋 Code Review Summary

Clean, backward-compatible extension of the workload identity client to support workload-specific TTL hints. The `NamedTuple` addition with a default of `0` preserves backward compatibility perfectly, and the test refactoring with `_setup_mock_response` is a nice DRY improvement that makes the test class easier to extend.

**Files Reviewed**: 2 files

### 🔍 Issues Found
- 0 security/safety issues
- 1 correctness suggestion (input validation)
- 1 test coverage suggestion

### Overall Assessment
LGTM with minor suggestions. The core change is solid and well-tested.

### General Feedback
- Backward compatibility is well-handled with the `int = 0` default
- Good docstrings clearly explaining the `0` vs `> 0` semantics
- The `_setup_mock_response` helper reduces test boilerplate nicely
- The parametrized TTL test covers the two documented behaviors cleanly

john-westcott-iv (approved):
Conditionally approving based on the negative number conversation. This is technically correct for what its doing pending that conversation.

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -29,6 +29,13 @@ class WorkloadIdentityTokenRequest(NamedTuple):
     audience: str
     """Audience for the token - the external service that will validate it."""
 
+    workload_ttl_seconds: int = 0
+    """Optional workload-specific TTL override in seconds.
+
+    If > 0, the Gateway uses this as the base TTL instead of the platform default.
+    Pass 0 (or omit) to use the platform fallback (jwt_default_ttl_seconds).
```
dleehr: This doesn't agree with what was merged in #952: 

https://github.com/ansible/django-ansible-base/blob/56e1af06f4162606f5e637ac8c36921562086bd9/ansible_base/lib/workload_identity/workload_identity_tokens.py#L35-L38

> Pass 0 (or omit) to use the platform fallback (jwt_default_ttl_seconds).

`0` is an invalid value

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -133,6 +140,7 @@ def request_workload_jwt(
         claims: dict,
         scope: str,
         audience: str,
+        workload_ttl_seconds: int = 0,
```
dleehr: Since `0` is an invalid value, perhaps this should default to `None`?

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -144,21 +152,33 @@ def request_workload_jwt(
...
             ...     claims={"id": 2, "name": "my-example-job"},
             ...     scope="aap_controller_automation_job",
-            ...     audience="https://vault.example.com"
+            ...     audience="https://vault.example.com",
+            ...     workload_ttl_seconds=3600,
             ... )
         """
-        request_body = WorkloadIdentityTokenRequest(claims=claims, scope=scope, audience=audience)
+        if workload_ttl_seconds < 0:
+            raise ValueError(f"workload_ttl_seconds must be >= 0, got {workload_ttl_seconds}")
```
dleehr: We have validation logic in the serializer that will enforce this server-side. Do we also need it here?

test_app/tests/resource_registry/test_workload_identity_client.py
```diff
@@ -29,6 +29,18 @@ def test_request_type_creation(self):
         assert request.claims == {"id": 2, "name": "my-example-job"}
         assert request.scope == "aap_controller_automation_job"
         assert request.audience == "https://vault.example.com"
+        assert request.workload_ttl_seconds == 0
```
dleehr: `0` is an invalid value, see

https://github.com/ansible/django-ansible-base/blob/56e1af06f4162606f5e637ac8c36921562086bd9/test_app/tests/lib/workload_identity/test_serializers.py#L239

dleehr (changes requested):
Requesting changes around the handling of `workload_ttl_seconds=0`. The implementation here doesn't align with what was merged in #952.

In the serializer, we now enforce a minimum value of `1` for `workload_ttl_seconds`. But this PR states that `0` is a valid value to send (and that it has a special meaning server-side). Neither of those are the case, so please update this logic to align with what you did in #952. I provided some line-level feedback on some of the invalid references but not all.

---

Related question (but not inherently blocking approval): Where in the flow should we catch a `0` and treat it as a special case? I understand that many (most) workloads will report a duration (job timeout) of `0` and we need to special case that before we make the request (since `0` will be rejected).

melissalkelly (commented):
Good work! The only thing I noticed was also pointed out by Dan, the payload construction not matching the gateway contract. Passing 0 will raise a 400 validation error, so `workload_ttl_seconds` could either be omitted from the payload or set to null instead of passing 0. Everything else looked good to me.
edit: also added a small comment to increase test coverage

test_app/tests/resource_registry/test_workload_identity_client.py
```diff
@@ -29,6 +29,18 @@ def test_request_type_creation(self):
...
+
+    def test_request_type_with_custom_ttl(self):
```
melissalkelly, line 34: Sonarqube is failing since the ValueError line is not covered. Adding a test that verifies `ValueError` is raised when `workload_ttl_seconds` is negative should bring the coverage to 100%.

test_app/tests/resource_registry/test_workload_identity_client.py
arrestle: There was a change to the requirements see AAP-63312 (Gateway rejects 0) in favor of null (0 = infinite, null is safer) 

see test_request_workload_jwt_rejects_invalid_ttl as an alternative to above.

arrestle:
> Requesting changes around the handling of `workload_ttl_seconds=0`. The implementation here doesn't align with what was merged in #952.
> 
> In the serializer, we now enforce a minimum value of `1` for `workload_ttl_seconds`. But this PR states that `0` is a valid value to send (and that it has a special meaning server-side). Neither of those are the case, so please update this logic to align with what you did in #952. I provided some line-level feedback on some of the invalid references but not all.
> 
> Related question (but not inherently blocking approval): Where in the flow should we catch a `0` and treat it as a special case? I understand that many (most) workloads will report a duration (job timeout) of `0` and we need to special case that before we make the request (since `0` will be rejected).

The implementation has been updated to match #952:
0 is no longer treated as valid. The client now uses workload_ttl_seconds: int | None = None. Passing 0 (or any negative value) raises ValueError with a message that 0 is invalid and only None or >= 1 are accepted.
Platform fallback. When the argument is None (or omitted), workload_ttl_seconds is left out of the request body. The Gateway then uses its platform default, per the ANSTRAT-1019 proposal.
Where to handle job.timeout == 0. That special case is handled in the caller (e.g. AWX), before the client is called. For job.timeout == 0, AWX passes None, so the client omits the field. Any caller that passes 0 into the client gets ValueError; the client does not treat 0 as a sentinel for platform fallback.

See handbook PR https://github.com/ansible/handbook/pull/1206

test_app/tests/resource_registry/test_workload_identity_client.py
arrestle: Thank you Melissa, it's running again I'll check when it's done. Lots of impacts from P4.1-3 changes.

ansible_base/resource_registry/workload_identity_client.py
arrestle: I recommend we keep it here: Callers see an immediate ValueError instead of waiting for an HTTP 400 from the Gateway. The error is clearer and in their stack trace.

test_app/tests/resource_registry/test_workload_identity_client.py
```diff
@@ -138,53 +162,91 @@ def test_service_token_property_auto_refresh(self, mock_get_service_token):
...
+            assert "workload_ttl_seconds" not in json_body
+        else:
+            assert json_body["workload_ttl_seconds"] == expected_in_json
+
+    @pytest.mark.parametrize("invalid_ttl", [0, -1, -3600])
+    def test_request_workload_jwt_rejects_invalid_ttl(self, invalid_ttl):
+        """Test that workload_ttl_seconds=0 or negative raises ValueError (Gateway rejects 0)."""
+        client = WorkloadIdentityClient(base_url="https://gateway.example.com")
+
+        with pytest.raises(ValueError, match="must be None.*or >= 1"):
```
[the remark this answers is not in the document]
arrestle: updated.

test_app/tests/resource_registry/test_workload_identity_client.py
arrestle: fixed

ansible_base/resource_registry/workload_identity_client.py
arrestle: correct this should all be resolved now. Let me know if you need a specific change.

ansible_base/resource_registry/workload_identity_client.py
arrestle: workload_ttl_seconds: int | None = None,

PabloHiro (approved):
LGTM

davemulford (approved):
Looks good.

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -144,22 +153,34 @@ def request_workload_jwt(
...
-        request_body = WorkloadIdentityTokenRequest(claims=claims, scope=scope, audience=audience)
-        data = request_body._asdict()
+        if workload_ttl_seconds is not None and workload_ttl_seconds < 1:
+            raise ValueError(f"workload_ttl_seconds must be None (platform fallback) or >= 1, got {workload_ttl_seconds}")
+        data = {
+            "claims": claims,
+            "scope": scope,
+            "audience": audience,
+        }
+        if workload_ttl_seconds is not None and workload_ttl_seconds > 0:
```
dleehr, line 182: Nit: we can simplify this:

```suggestion
        if workload_ttl_seconds:
```

Proof:

```python
Python 3.13.1 (main, Jan 27 2025, 14:38:42) [Clang 16.0.0 (clang-1600.0.26.6)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> x = 3 
>>> if x:
...     print('yes')
...     
yes
>>> x = 0
>>> if x:
...     print('yes')
...     
>>> x = None
>>> if x:
...     print('yes')
...     
>>> 
```

test_app/tests/resource_registry/test_workload_identity_client.py
```diff
@@ -138,53 +162,89 @@ def test_service_token_property_auto_refresh(self, mock_get_service_token):
...
+        None (or omit) signals the Gateway to use its platform fallback (jwt_default_ttl_seconds).
+        """
+        self._setup_mock_response(mock_request, mock_get_service_token)
+
+        kwargs = {
+            "claims": {"id": 1, "name": "test-job"},
+            "scope": "aap_controller_automation_job",
+            "audience": "https://vault.example.com",
+        }
+        if workload_ttl_seconds is not None:
```
dleehr: nit:

```suggestion
        if workload_ttl_seconds:
```

ansible_base/resource_registry/workload_identity_client.py
```diff
@@ -144,22 +153,34 @@ def request_workload_jwt(
...
             ...     scope="aap_controller_automation_job",
-            ...     audience="https://vault.example.com"
+            ...     audience="https://vault.example.com",
+            ...     workload_ttl_seconds=3600,
             ... )
         """
-        request_body = WorkloadIdentityTokenRequest(claims=claims, scope=scope, audience=audience)
-        data = request_body._asdict()
+        if workload_ttl_seconds is not None and workload_ttl_seconds < 1:
+            raise ValueError(f"workload_ttl_seconds must be None (platform fallback) or >= 1, got {workload_ttl_seconds}")
```
dleehr, line 176: I left a comment earlier, but since this is enforced by the serializer on the server-side, do we need to check it here in the client?

dleehr (approved):

melissalkelly (approved):

test_app/tests/resource_registry/test_workload_identity_client.py
```diff
@@ -29,6 +29,18 @@ def test_request_type_creation(self):
         assert request.claims == {"id": 2, "name": "my-example-job"}
         assert request.scope == "aap_controller_automation_job"
         assert request.audience == "https://vault.example.com"
+        assert request.workload_ttl_seconds is None
+
+    def test_request_type_with_custom_ttl(self):
+        """Test that workload_ttl_seconds can be set on the request type."""
+        request = WorkloadIdentityTokenRequest(
```
dragid10, line 36: **Low Care Amount:** I could be missing it, but I'm not seeing `WorkloadIdentityTokenRequest` be used in the actual client code anymore. is that just leftover? Should it be removed? Will it perhaps be used in type hints or something?

Either way, it don't think that this hurts anything

dragid10 (approved):
(As DAB owner)
I +1 Dan's improvement comments, but they're a Low Care Amount for me

ansible_base/resource_registry/workload_identity_client.py
arrestle: It's true it's simpler, but I feel the readability is improved with the existing code, so I vote we keep it the way it is.

ansible_base/resource_registry/workload_identity_client.py
arrestle: This improves error handling: "Fail fast – Invalid values (0, negative) are caught before a request, so callers get an immediate ValueError instead of a 400 from the server."

test_app/tests/resource_registry/test_workload_identity_client.py
arrestle: This seems like a good suggestion - improves clarity.

test_app/tests/resource_registry/test_workload_identity_client.py
arrestle: Kept it for schema documentation and so tests can assert the request shape. Not used when building the request today, but it makes the contract explicit and could be used if we add typed request builders later.

dleehr: That's a great catch @dragid10 🙌 ! 

Looks like we're defining the `WorkloadIdentityTokenRequest` and `WorkloadIdentityTokenResponse` named tuples here but only using the `WorkloadIdentityTokenResponse` type in the client:

https://github.com/arrestle/django-ansible-base/blob/e88d91784e1423db35835c4fa4e8a51ad550ba65/ansible_base/resource_registry/workload_identity_client.py#L131-L136

That omission is outside the scope of this PR, but that's certainly something we missed in the earlier PRs around this client.

This unit test is confirming the functionality of the `WorkloadIdentityTokenRequest` constructor with the new field, so that's a good test 👍 

I'll create a follow-up story to clean that up.

PR #956, opened 2026-02-25
[AAP-64918] Adds AuditableModel for Role*Assignment

huffmanca:
## Description

Adds AuditableModel to AssignmentBase so that it can be leveraged by RoleUserAssignment and RoleTeamAssignment. This is configured to have audit logs created for these classes without updating the ActivityStream.

These are similar changes to PR 945 , just for the RoleUserAssignment / RoleTeamAssignment (through AssignmentBase).

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

- [x] I have performed a self-review of my (and Claude's) code

### Steps to Test
1. Start an AAP instance with these changes.
2. Create RoleUserAssignment/RoleTeamAssignment objects.
3. Ensure that the Gateway logs have entries indicating that the models were created. Sample log lines are described below.
4. Ensure that no ActivityStream entries are created.

### Expected Results
We should see lines similar to the following in the logs:
```
myaap-gateway-5fc8fd7844-tsl9g api 2026-02-25T14:48:10.022072188-05:00 2026-02-25 19:48:10,021 INFO     3c800b99-b104-428e-8046-eb4fc65d907b aap.auth_audit [10.244.0.1] "Mozilla/5.0 (X11; Linux x86_64; rv:146.0) Gecko/20100101 Firefox/146.0" User: admin create RoleUserAssignment RoleUserAssignment object (18) (18) {'object_role': '4', 'created': '2026-02-25T19:48:10.020024+00:00', 'content_type': '11', 'object_id': '214', 'role_definition': '2', 'id': '18', 'created_by': '2', 'user': '18'}
```

No ActivityStream entries should be created

## Additional Context
As mentioned in the description, this is similar to the approach in PR 945 . 

## Summary by CodeRabbit

* **New Features**
  * Audit logging enabled for role assignment operations. Activity stream integration is conditional and remains disabled by default when the activity stream feature isn't available.
* **Tests**
  * Added tests confirming audit logs are emitted on creation and deletion of role assignments and that no activity stream entries are produced whether the activity stream is present or not.

### DIFF
--- a/ansible_base/rbac/models/role.py
+++ b/ansible_base/rbac/models/role.py
@@ -434,9 +434,10 @@ class AssignmentBase(ImmutableCommonModel, ObjectRoleFields, _AuditableBase):
     both models are immutable, making caching easy.
     """
 
-    # Enable audit log for this model (only if activitystream is installed)
+    # When activitystream is installed (_AuditableBase = AuditableModel), these flags
+    # configure AuditableModel to enable audit logging and suppress activity stream entries.
+    # When activitystream is absent, _AuditableBase = object and these flags are inert.
     audit_log_enabled = True
-    # Disable activity stream for this model
     activity_stream_enabled = False
 
     object_role = models.ForeignKey(
--- a/ansible_base/rbac/models/role.py
+++ b/ansible_base/rbac/models/role.py
@@ -31,6 +31,13 @@ from .content_type import DABContentType
 from .fields import FederatedForeignKey
 from .permission import DABPermission
 
+# Conditionally import AuditableModel if activitystream is installed
+_AuditableBase = object
+if 'ansible_base.activitystream' in settings.INSTALLED_APPS:
+    from ansible_base.activitystream.models import AuditableModel
+
+    _AuditableBase = AuditableModel
+
 logger = logging.getLogger('ansible_base.rbac.models')
 
 
@@ -419,13 +426,18 @@ class ObjectRoleFields(models.Model):
         return RoleEvaluation._meta.get_field('object_id').to_python(self.object_id)
 
 
-class AssignmentBase(ImmutableCommonModel, ObjectRoleFields):
+class AssignmentBase(ImmutableCommonModel, ObjectRoleFields, _AuditableBase):
     """
     This uses some parts of CommonModel to save metadata like documenting
     the user who assigned the permission and timestamp when it happened.
     This caches ObjectRole fields for purposes of serializers,
     both models are immutable, making caching easy.
     """
+
+    # Enable audit log for this model (only if activitystream is installed)
+    audit_log_enabled = True
+    # Disable activity stream for this model
+    activity_stream_enabled = False
 
     object_role = models.ForeignKey(
         'dab_rbac.ObjectRole', on_delete=models.CASCADE, editable=False, null=True, help_text=_("A roll-up of the fields (role_definition, content_type).")
--- a/test_app/tests/rbac/models/test_assignment_audit_logging.py
+++ b/test_app/tests/rbac/models/test_assignment_audit_logging.py
@@ -28,29 +28,21 @@ ACTIVITYSTREAM_INSTALLED = 'ansible_base.activitystream' in settings.INSTALLED_A
 # -----------------------------------------------------------------------------
 
 
-@pytest.mark.django_db
-@pytest.mark.skipif(not ACTIVITYSTREAM_INSTALLED, reason="activitystream not installed")
 def test_role_user_assignment_has_audit_log_enabled():
     """RoleUserAssignment has audit_log_enabled = True to enable audit logging."""
     assert getattr(RoleUserAssignment, "audit_log_enabled", False) is True
 
 
-@pytest.mark.django_db
-@pytest.mark.skipif(not ACTIVITYSTREAM_INSTALLED, reason="activitystream not installed")
 def test_role_user_assignment_has_activity_stream_disabled():
     """RoleUserAssignment has activity_stream_enabled = False to disable activity stream."""
     assert getattr(RoleUserAssignment, "activity_stream_enabled", True) is False
 
 
-@pytest.mark.django_db
-@pytest.mark.skipif(not ACTIVITYSTREAM_INSTALLED, reason="activitystream not installed")
 def test_role_team_assignment_has_audit_log_enabled():
     """RoleTeamAssignment has audit_log_enabled = True to enable audit logging."""
     assert getattr(RoleTeamAssignment, "audit_log_enabled", False) is True
 
 
-@pytest.mark.django_db
-@pytest.mark.skipif(not ACTIVITYSTREAM_INSTALLED, reason="activitystream not installed")
 def test_role_team_assignment_has_activity_stream_disabled():
     """RoleTeamAssignment has activity_stream_enabled = False to disable activity stream."""
     assert getattr(RoleTeamAssignment, "activity_stream_enabled", True) is False
--- a/test_app/tests/rbac/models/test_assignment_audit_logging.py
+++ b/test_app/tests/rbac/models/test_assignment_audit_logging.py
@@ -5,6 +5,8 @@ Ensures that:
 2. Both assignment classes have activity_stream_enabled = False (no activity stream entries)
 3. Create operations generate audit logs at INFO level
 4. Create operations do NOT create activity stream entries
+5. Delete operations generate audit logs at INFO level
+6. Delete operations do NOT create activity stream entries
 """
 
 from unittest import mock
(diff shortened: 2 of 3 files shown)


AlanCoding:
This is different from https://github.com/ansible/django-ansible-base/pull/945, where _all_ of the components are using the RBAC app.

I know this is a lost cause, but we ideally shouldn't be referencing settings on import. This makes the entire class structure dynamic, which is a dramatic level of debt.

huffmanca:
@AlanCoding ,

Since it's used by all components, then I'm assuming we can safely drop the conditional import check. Once that's done, is there a better way to meet the current requirement/proposal for having these audit logs (but no activity stream) for RoleUserAssignment and RoleTeamAssignment? 

I don't mind reworking the PR, but since we're moving forward (at least for immediacy + backporting to 2.5 and 2.6) with the logging controlled by AuditableModel, that's the direction I'd like to use for now. Then fix it in 2.7 with something like [951](https://github.com/ansible/django-ansible-base/pull/951) .

jay-steurer:
## Comprehensive PR Review Summary

Thank you for this well-structured implementation of audit logging for role assignments! The core functionality is solid and follows good Django patterns. However, I've identified several critical issues that should be addressed before merging, particularly around error handling and silent failure modes.

## 🚨 Critical Issues (Must Fix)

### 1. **Silent Security Degradation**
The conditional inheritance pattern silently disables audit logging when `activitystream` is not installed, without any warnings. This could cause compliance violations in production.

**Risk**: Operators may believe audit logging is active when it's completely disabled.

### 2. **Unprotected Signal Handlers** 
Audit logging failures will crash database transactions instead of being handled gracefully.

### 3. **Missing Deletion Tests**
No tests verify audit logs are created when role assignments are deleted (the primary mutation after creation).

## 📋 Required Actions

**Before Merge:**
- [ ] Add error handling to prevent audit failures from killing transactions
- [ ] Either fail loudly when activitystream missing OR log prominent warnings  
- [ ] Add tests for assignment deletion scenarios
- [ ] Fix misleading `audit_log_enabled = True` when logging is non-functional

**Strongly Recommended:**
- [ ] Add cascade deletion audit tests (user/team/resource deletion)
- [ ] Enhance test assertions to verify logged data accuracy
- [ ] Address code style issues (spacing, constants, type annotations)

## 🌟 Strengths
- Excellent test organization with clear documentation
- Proper conditional test skipping
- Follows established patterns from PR #945  
- Good separation between audit and activity stream concerns

I'll add specific line-by-line comments with detailed recommendations. The security aspects are the highest priority - this feature must fail safely rather than silently.

Great work overall! The implementation approach is sound, it just needs stronger defensive programming.

ansible_base/rbac/models/role.py
```diff
@@ -31,6 +31,13 @@
 from .fields import FederatedForeignKey
 from .permission import DABPermission
 
+# Conditionally import AuditableModel if activitystream is installed
+_AuditableBase = object
+if 'ansible_base.activitystream' in settings.INSTALLED_APPS:
+    from ansible_base.activitystream.models import AuditableModel
+
+    _AuditableBase = AuditableModel
```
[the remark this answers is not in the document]
huffmanca: I've discussed this with others, and as of right now this is the intended behavior if a component doesn't use activitystream in their INSTALLED_APPS.

AlanCoding:
> Since it's used by all components, then I'm assuming we can safely drop the conditional import check.

I am saying that all components use the RBAC app, but only Gateway uses the activitystream app. As far as I'm concerned, the activitystream app can be classified as Gateway-internal code that just happens to live here. You have to have the conditional (absent modifying activitystream logic), because of the projects that have RBAC and do not have activitystream. Otherwise, you'll get errors from importing from an app that is not in INSTALLED_APPS. So no, the conditional can't just be dropped because that will not work.

I don't deny this will work (adding the abstract parent will not generate new migrations in the current form). But this introduces a practice we should try very hard to stay away from, and the only reason it is needed is because of hesitancy to change the logic in activitystream app. So yes, clearly my preference, would be to restructure the activitystream app's logic so we don't have to do this, even for a backport, that should not present problems.

jeffh-oss:
> > Since it's used by all components, then I'm assuming we can safely drop the conditional import check.
> 
> I am saying that all components use the RBAC app, but only Gateway uses the activitystream app. As far as I'm concerned, the activitystream app can be classified as Gateway-internal code that just happens to live here. You have to have the conditional (absent modifying activitystream logic), because of the projects that have RBAC and do not have activitystream. Otherwise, you'll get errors from importing from an app that is not in INSTALLED_APPS. So no, the conditional can't just be dropped because that will not work.
> 
> I don't deny this will work (adding the abstract parent will not generate new migrations in the current form). But this introduces a practice we should try very hard to stay away from, and the only reason it is needed is because of hesitancy to change the logic in activitystream app. So yes, clearly my preference, would be to restructure the activitystream app's logic so we don't have to do this, even for a backport, that should not present problems.

@AlanCoding could we move forward as is and write up a jira issue for your suggested changes at a later time? I believe what @huffmanca has done is in line with what was outlined in the [SDP](https://handbook.eng.ansible.com/System-Design-Plans/0056-Comprehensive-Gateway-Audit-Logging) for audit logging and the activity stream?

john-westcott-iv (approved):
## Code Review Summary

This PR adds `AuditableModel` to `AssignmentBase` so `RoleUserAssignment` and `RoleTeamAssignment` generate audit logs on create/delete while suppressing activity stream entries. The implementation follows the same conditional inheritance pattern established in PR #945 (AAPFlag), which has already been merged.

**Files Reviewed**: 2 files
**Comments Posted**: 2 inline comments (non-blocking)

### Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 2 code quality nits (non-blocking)

### Overall Assessment
Approving. The code is correct, tests are thorough (attribute verification, create/delete audit logs, activity stream absence checks), and all CI passes.

### Alan's Architectural Concerns
I agree with @AlanCoding's feedback that referencing `settings` on import to make the class structure dynamic is tech debt, and that the direction in PR #951 is preferable long-term. However, this follows the same pattern we already merged in #945, so blocking on architectural concerns here would be inconsistent. Both #945 and this PR should be revisited when #951 (or similar) lands in 2.7+.

@AlanCoding -- one question: do you know if any of the other services (Controller, Hub, EDA) have plans to adopt `activitystream` from DAB? The intent was always for this to be reusable across services, and that context would inform how urgently we should prioritize the #951 refactoring.

---
**Note:** This review was submitted with assistance from Claude AI assistant on behalf of John Westcott.

ansible_base/rbac/models/role.py
```diff
@@ -419,14 +426,19 @@ def cache_id(self):
...
-class AssignmentBase(ImmutableCommonModel, ObjectRoleFields):
+class AssignmentBase(ImmutableCommonModel, ObjectRoleFields, _AuditableBase):
     """
     This uses some parts of CommonModel to save metadata like documenting
     the user who assigned the permission and timestamp when it happened.
     This caches ObjectRole fields for purposes of serializers,
     both models are immutable, making caching easy.
     """
 
+    # Enable audit log for this model (only if activitystream is installed)
```
john-westcott-iv: Nit: This comment says "only if activitystream is installed" but the attribute is set unconditionally on `AssignmentBase` regardless of whether `_AuditableBase` resolves to `AuditableModel` or `object`. The flag is always present — it just has no effect when `_AuditableBase = object` because no signal handler consumes it.

Same issue exists in the merged #945 code on `AAPFlag`. Non-blocking, but worth fixing in both places for accuracy.

```suggestion
    # When activitystream is installed (_AuditableBase = AuditableModel), these flags
    # configure AuditableModel to enable audit logging and suppress activity stream entries.
    # When activitystream is absent, _AuditableBase = object and these flags are inert.
    audit_log_enabled = True
    activity_stream_enabled = False
```

test_app/tests/rbac/models/test_assignment_audit_logging.py
```diff
@@ -0,0 +1,200 @@
...
+ACTIVITYSTREAM_INSTALLED = 'ansible_base.activitystream' in settings.INSTALLED_APPS
+
+
+# -----------------------------------------------------------------------------
+# Verify both assignment classes have correct attribute settings
+# -----------------------------------------------------------------------------
+
+
+@pytest.mark.django_db
+@pytest.mark.skipif(not ACTIVITYSTREAM_INSTALLED, reason="activitystream not installed")
```
john-westcott-iv: Nit: The four attribute-only tests (lines 32-57) only read class-level attributes via `getattr()` — no database access, no dependency on activitystream being installed. Both `@pytest.mark.django_db` and `@pytest.mark.skipif(not ACTIVITYSTREAM_INSTALLED)` are unnecessary here and could be removed. The attributes are set unconditionally on `AssignmentBase` regardless of `_AuditableBase`.

Non-blocking, but removing the decorators would make these tests run faster (no DB setup) and provide coverage even when activitystream isn't installed.

bhavenst (approved):

test_app/tests/rbac/models/test_assignment_audit_logging.py
huffmanca: Updated. These tests all pass locally with these changes.

ansible_base/rbac/models/role.py
huffmanca: I've updated the comment here and in AAPFlag. Note that AAPFlag was done in a separate commit in case we don't want to include it as part of this PR.

PR #957, opened 2026-02-26
Configure DAB so custom OIDC scopes and claims are supported

melissalkelly:
AAP-60507

## Description
### What is being changed?                                                                                                                                                                                            
The OIDC discovery endpoint URL pattern now uses a custom ConnectDiscoveryInfoView (if available) instead of always using the default django-oauth-toolkit view. Falls back to the default view when custom view is not present.                                                                                                                                                                                           
                  
Gateway needs to extend the OIDC discovery endpoint to dynamically populate scopes_supported and claims_supported with workload identity scopes and claims. Since DAB owns the OAuth2/OIDC URL routing, the custom view must be wired up here.

Allows Gateway to override the discovery view to include workload identity data in the response, while maintaining backwards compatibility for standalone DAB usage via the try/except fallback.


## Type of Change
- [x] Configuration change

## Self-Review Checklist
<!-- These items help ensure quality - they complement our automated CI checks -->

## Testing Instructions
See testing in corresponding PR for AAP-60507




<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Refactor**
  * OAuth2/OpenID Connect discovery endpoint now prefers a custom gateway discovery view when available, with an automatic fallback to the standard provider. This adds runtime flexibility so deployments with a gateway use its discovery behavior without impacting others.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/ansible_base/oauth2_provider/urls.py
+++ b/ansible_base/oauth2_provider/urls.py
@@ -8,7 +8,12 @@ from ansible_base.lib.routers import AssociationResourceRouter
 from ansible_base.oauth2_provider import views as oauth2_provider_views
 from ansible_base.oauth2_provider.apps import Oauth2ProviderConfig
 
-if importlib.util.find_spec("aap_gateway_api.views.oidc_discovery") is not None:
+try:
+    _gateway_spec = importlib.util.find_spec("aap_gateway_api.views.oidc_discovery")
+except ModuleNotFoundError:
+    _gateway_spec = None
+
+if _gateway_spec is not None:
     from aap_gateway_api.views.oidc_discovery import ConnectDiscoveryInfoView
 else:
     ConnectDiscoveryInfoView = oauth_views.ConnectDiscoveryInfoView
--- a/ansible_base/oauth2_provider/urls.py
+++ b/ansible_base/oauth2_provider/urls.py
@@ -6,6 +6,11 @@ from ansible_base.lib.routers import AssociationResourceRouter
 from ansible_base.oauth2_provider import views as oauth2_provider_views
 from ansible_base.oauth2_provider.apps import Oauth2ProviderConfig
 
+try:
+    from aap_gateway_api.views.oidc_discovery import ConnectDiscoveryInfoView
+except ImportError:
+    ConnectDiscoveryInfoView = oauth_views.ConnectDiscoveryInfoView
+
 app_name = Oauth2ProviderConfig.label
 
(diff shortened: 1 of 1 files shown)


ansible_base/oauth2_provider/urls.py
```diff
@@ -6,6 +8,16 @@
 from ansible_base.oauth2_provider import views as oauth2_provider_views
 from ansible_base.oauth2_provider.apps import Oauth2ProviderConfig
 
+try:
+    _gateway_spec = importlib.util.find_spec("aap_gateway_api.views.oidc_discovery")
+except ModuleNotFoundError:
+    _gateway_spec = None
+
+if _gateway_spec is not None:
+    from aap_gateway_api.views.oidc_discovery import ConnectDiscoveryInfoView
```
dleehr, line 17: Nice ingenuity here in getting his to work, but I don't think we should be importing gateway from DAB. There's a pretty clear dependency relationship already and this is not a change we should make. Let's find a better way to achieve the scopes/claims customization.

dleehr (changes requested):
Requesting changes around this import pattern, I don't think we should pursue that.

I see that the jira story suggests overriding the discovery view. But in my read of the docs django-oauth-toolkit provides mechanisms to supply scopes/claims **without** rewriting the views: https://django-oauth-toolkit.readthedocs.io/en/latest/oidc.html#using-oidc-scopes-to-determine-which-claims-are-returned

I have this working in a proof-of-concept branch: [AAP-60507-scopes-claims-poc](https://github.com/ansible/django-ansible-base/compare/devel...dleehr:django-ansible-base:AAP-60507-scopes-claims-poc). Specifically, see commit https://github.com/ansible/django-ansible-base/commit/9c68a9a05c6ff8351e485ae660648f0f27686efc from Jan 25.

In summary, we should be able to achieve this by referencing a custom validator class and the scope definitions it in `settings` like this:

```
    "OAUTH2_VALIDATOR_CLASS": "ansible_base.workload_identity.CustomValidator",
    "SCOPES": {"some-custom-scope": "Custom Scope Description"},
```

In my example, I have `SCOPES` defined in settings, but your `SCOPES_BACKEND_CLASS` in the other PR will achieve the same. 

Now, if that doesn't do what we need, we can still write a custom `ConnectDiscoveryInfoView`, but let's make sure we're not importing from gateway here.

PR #959, opened 2026-03-02
AAP-45047: Fix allow map to process grant case in create_claims

doziedotdev:
## Description
- What is being changed?
  - The `create_claims()` function in `claims.py` now correctly handles `map_type=allow` when the trigger fires (`has_permission=True`), allowing a later allow map to override an earlier deny.
- Why is this change needed?
  - The allow map only handled the deny case (`has_permission=False`). When a trigger fired and `has_permission=True`, the code fell through to the catch-all else branch, logging "does not know how to be processed" and never setting `access_allowed = True`. This broke the documented deny-all + allow-override pattern.
  - Changes the conditional from `if auth_map.map_type == 'allow' and not has_permission:` to `if auth_map.map_type == 'allow':` and sets `access_allowed = has_permission`. This handles both grant and deny cases and respects map ordering (later maps override earlier ones).

Fixes: AAP-45047, AAP-45394

## Type of Change
- [x] Test update

## Self-Review Checklist

### Prerequisites
Any authenticator (LDAP or Local Database).

### Steps to Test
1. Create two authenticator maps on an authenticator:
   - "Deny All": map_type=allow, triggers={"never": {}}, order=1
   - "Allow Override": map_type=allow, triggers={"always": {}}, order=2
2. Attempt to log in as any user through that authenticator

### Expected Results
- Login succeeds. The allow override at order 2 overrides the deny at order 1.
- No "does not know how to be processed" error in logs.

### Before this fix
- Login fails (HTTP 401)
- Log: `ERROR Map type allow of rule Allow Override does not know how to be processed`
- Log: `WARNING User <username> failed an allow map and was denied access`

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected authentication claims handling to ensure allow-type rules consistently and directly reflect permission decisions across all scenarios.

* **Tests**
  * Added comprehensive regression tests for authentication claims behavior, including group-based and attribute-based authorization, deny-all override patterns, and edge-case interactions.


doziedotdev:
Hi @john-westcott-iv sorry about the ping

Could you help review this when you get a chance? You identified the root cause line in AAP-45394. This PR fixes the allow map handling so it correctly grants access instead of only processing the deny case. 6+ support cases linked across AAP-45047 and AAP-45394, both open ~10 months.

ansible_base/authentication/utils/claims.py
```diff
@@ -120,9 +120,8 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
         rule_responses.append({mpk: has_permission, 'enabled': auth_map.enabled})
 
         understood_map = False
-        if auth_map.map_type == 'allow' and not has_permission:
-            # If any rule does not allow we don't want to return this to true
-            access_allowed = False
+        if auth_map.map_type == 'allow':
+            access_allowed = has_permission
```
john-westcott-iv, line 124: Clean fix. This now mirrors the `is_superuser` pattern on line 127 (`is_superuser = has_permission`), making the handling consistent across map types. The "last evaluated map wins" semantic is correct for ordered map evaluation.

The old code's `and not has_permission` guard meant allow maps could only _deny_ — they could never _grant_ access back. This simple change restores the intended bidirectional behavior.

test_app/tests/authentication/utils/test_claims.py
```diff
@@ -274,6 +286,107 @@ def test_create_claims_bad_map_type_logged(
...
+    # Map 2 (order 2): allow override
+    local_authenticator_map_1.triggers = {"always": {}}
+    local_authenticator_map_1.map_type = "allow"
+    local_authenticator_map_1.order = 2
+    local_authenticator_map_1.save()
+
+    authenticator = local_authenticator_map.authenticator
+    res = claims.create_claims(authenticator, "username", {}, [])
+
+    assert res["access_allowed"] is True
```
john-westcott-iv, line 333: Nit (non-blocking): Consider adding a reverse-order test — "allow-all (order=1) then deny-all (order=2)" — to verify that a later deny still correctly overrides an earlier allow. This would complete the ordering matrix:

| Map 1 | Map 2 | Expected |
|-------|-------|----------|
| deny-all | allow-override | `True` ✅ (this test) |
| allow-all | deny-override | `False` (not tested) |

This is minor since the deny path was already working before this fix, but it would document the expected bidirectional "last map wins" contract and guard against future regressions.

john-westcott-iv (commented):
## 📋 Code Review Summary

This PR fixes a bug where `map_type='allow'` in `create_claims()` only handled the deny case (`has_permission=False`). When a trigger fired and granted permission (`has_permission=True`), the code fell through to the catch-all `else` branch, logging an erroneous "does not know how to be processed" error and never setting `access_allowed = True`. This broke the documented deny-all + allow-override pattern (equivalent to `AUTH_LDAP_REQUIRE_GROUP`).

**Files Reviewed**: 2 files

### 🔍 Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 1 minor suggestion (non-blocking)
- 1 positive observation

### Overall Assessment
**LGTM** — This is a clean, minimal, and correct fix. The code change is a single-line semantic improvement that makes `allow` map handling consistent with `is_superuser`. The tests are comprehensive, covering the main fix scenario, the inverse (deny not overridden without match), group-based matching, and error-log regression. Well done.

### General Feedback
- The fix correctly implements "last evaluated map wins" semantics for `allow` maps, which is consistent with how `is_superuser` maps already work (line 127: `is_superuser = has_permission`)
- Test coverage is thorough: basic parametrized case, error-log regression (AAP-45047), deny-all override (AAP-45394), group-based positive and negative cases
- The removed comment ("If any rule does not allow we don't want to return this to true") was documenting the buggy one-way-deny-only behavior — removing it is the right call
- 6+ support cases across two JIRA tickets — great to see this finally fixed

bhavenst:
Hi @doziedotdev — thanks for the fix here. The production code change is identical to what's in #995, which also includes additional test coverage (reverse-order test, revoke+deny override, 3-map last-writer-wins) and now also incorporates your `test_create_claims_allow_grant_no_error_logged` regression test for AAP-45047.

Since #995 is a strict superset of this PR's changes, this one can be closed once #995 merges. Appreciate the work tracking this down — credit to you for identifying the root cause.

doziedotdev:
No worries. Thanks Bryan!

PR #960, opened 2026-03-05
[AAP-66890] Re-order subject claim in workload identity token to be org->job

melissalkelly:
[AAP-66890](https://issues.redhat.com/browse/AAP-66890)

## Description

**What is being changed?**
  - Reordering the workload identity token `sub` claim to `workload_type:<workload>organization:<org>:job_template:<template>`
  - Removed controller-specific examples from base class docstrings
  
  - The sub claim should identify the immutable principal identity, not mutable attributes
  - Mutable attributes like project, inventory, job_type remain available as separate claims for use in Vault bound claims
  
  - Implements minimal principal identity pattern: organization -> job_template
  - Aligns with team decision after discussion in [Slack thread](https://redhat-internal.slack.com/archives/C0A048CSKHP/p1772659828705009)

## Type of Change
- [x] Test update

## Self-Review Checklist

### Steps to Test
Unit tests:
  1. Run workload identity tests: python -m pytest test_app/tests/lib/workload_identity/test_controller.py -v
  2. Verify all tests pass
  3. Check that generate_sub_claim produces format: organization:<org>:job_template:<template>
End to end tests see Jira ticket

## Summary by CodeRabbit

* **Bug Fixes**
  * Shortened workload identity sub-claims by removing job and project components, producing more concise claim strings.
  * Sub-claim values now consistently prefixed with "workload_type:{name}:" and enforce a stable segment order.

* **Tests**
  * Updated unit tests to align with the new sub-claim formats and mappings.

* **Documentation**
  * Updated claim-generation docs to reflect the new prefixing and simplified mappings.

### DIFF
--- a/ansible_base/lib/workload_identity/base.py
+++ b/ansible_base/lib/workload_identity/base.py
@@ -56,9 +56,7 @@ class BaseWorkloadIdentityScope:
         target_claim_names_to_sub_stubs = cls.get_target_claim_names_to_sub_stubs()
         base_sub = ":".join([f"{target_claim_names_to_sub_stubs[key]}:{workload_claims.get(key, '')}" for key in target_claim_names_to_sub_stubs.keys()])
 
-        if cls.name:
-            return f"workload_type:{cls.name}:{base_sub}"
-        return base_sub
+        return f"workload_type:{cls.name}:{base_sub}"
 
     @abstractmethod
     def populate_claims(self, workload_data: dict) -> dict:
--- a/ansible_base/lib/workload_identity/base.py
+++ b/ansible_base/lib/workload_identity/base.py
@@ -40,7 +40,7 @@ class BaseWorkloadIdentityScope:
         """
         Generate a sub claim string from workload claims using the scope's claim mapping.
 
-        Constructs a colon-delimited string prefixed with "service:{service_name}:" followed by
+        Constructs a colon-delimited string prefixed with "workload_type:{scope_name}:" followed by
         the claim names and values specified by the scope's get_target_claim_names_to_sub_stubs()
         method. The order is determined by the subclass implementation.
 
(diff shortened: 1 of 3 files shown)


PabloHiro (approved):
LGTM!

dleehr (approved):
LGTM

Only question I have - are there any tests in other repos that will break because of this change? Hopefully not, but keep an eye out after merge.

davemulford (approved):

ansible_base/lib/workload_identity/base.py
```diff
@@ -40,20 +40,26 @@ def generate_sub_claim(cls, workload_claims: dict) -> str:
...
         :type workload_claims: dict
-        :return: A string containing the sub claim
+        :return: A string containing the sub claim prefixed with service identifier
         :rtype: str
         """
         target_claim_names_to_sub_stubs = cls.get_target_claim_names_to_sub_stubs()
-        return ":".join([f"{target_claim_names_to_sub_stubs[key]}:{workload_claims.get(key, '')}" for key in target_claim_names_to_sub_stubs.keys()])
+        base_sub = ":".join([f"{target_claim_names_to_sub_stubs[key]}:{workload_claims.get(key, '')}" for key in target_claim_names_to_sub_stubs.keys()])
+
+        if cls.name and '_' in cls.name:
```
melissalkelly: I wasn't sure if we wanted to define a new class variable `prefix_name` in the scope classes or if we wanted to slice the name out of the existing `name` variable. It felt a bit redundant to define `prefix_name` when we already have `name`.

dleehr: I agree we don't need to define a new variable and we shoud re-use `name`. Actually I'd advocate for further simplifying. I may have been overly-specific suggesting `service` here. I didn't mean to imply we should transform the name, I think `aap_controller_automation_job` better describes the workload than just `controller`.

Maybe `workload_type` is more appropriate than `service` then.

melissalkelly: Updated to use workload type instead of service, and full scope name.

dleehr (approved):

AlanCoding (approved):
Your team owns the contract for this, the window for changing any of the schema remains wide open for now

PR #961, opened 2026-03-05
AAP-57990 Add visible_teams to support org admins all team view

bhavenst:
## Description
Added a visible_teams alongside visible_users to support making all teams viewable in gateway when ORG_ADMINS_CAN_SEE_ALL_USERS=True

Also enables org role assignments to teams not in the org.

See former version of this PR for commentary: https://github.com/ansible/django-ansible-base/pull/932

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1.  Run UT
or
1. Grab GW partner PR and start gateway
2. Create two orgs, two teams and one normal user.  Make user admin in one org.  Add one team to org.
3. Set ORG_ADMINS_CAN_SEE_ALL_USERS=True
4. Ensure user who is org admin can see both teams
5. Set ORG_ADMINS_CAN_SEE_ALL_USERS=False
6. Ensure user who is org admin can see only the team in the org

### Expected Results
See above.

## Summary by CodeRabbit

* **New Features**
  * Organization-aware team visibility added (visible teams respect user permissions and org access).
  * Role-to-team assignment now respects org-admin visibility rules, honoring ORG_ADMINS_CAN_SEE_ALL_USERS.

* **Tests**
  * Extensive tests added for team visibility and role-assignment across user roles, organizations, and configuration scenarios.

### DIFF
--- a/ansible_base/rbac/api/serializers.py
+++ b/ansible_base/rbac/api/serializers.py
@@ -15,7 +15,7 @@ from ansible_base.lib.utils.auth import get_team_model
 from ansible_base.lib.utils.response import get_relative_url
 from ansible_base.rbac.models import RoleDefinition, RoleTeamAssignment, RoleUserAssignment
 from ansible_base.rbac.permission_registry import permission_registry  # careful for circular imports
-from ansible_base.rbac.policies import check_content_obj_permission, visible_users
+from ansible_base.rbac.policies import can_view_all_users, check_content_obj_permission, visible_teams, visible_users
 from ansible_base.rbac.validators import check_locally_managed, validate_permissions_for_model
 
 from ..models import DABContentType, DABPermission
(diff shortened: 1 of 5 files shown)


bhavenst:
@AlanCoding It's back! :) We have this item in our sprint as a high priority from several customers.  Please let me know if this is acceptable or alternatively would you mind putting up a PR to accomplish the same thing in a better way?

This code along w/ GW PR https://github.com/ansible-automation-platform/aap-gateway/pull/1242 is on this RPM env for test if you like: https://54.227.202.33 (normal Jenkins PW)

john-westcott-iv (approved):

AlanCoding (changes requested):
You are modifying the view/serializers at `ansible_base/rbac/api/serializers.py`, and doing it to allow the team model, specifically, through.

But I think you should, instead, focus on modifying `ansible_base/rbac/service_api/serializers.py`, and do this generally to _any_ actors, meaning user & team models. Because in both cases, those models are not owned by the service. So this creates a clean separation of responsibilities. The resource provider server is responsible for checking user/team visibility, the service is just filling in unknown information, occasionally, which is what organization an object is in mostly.

This is also why I think `visible_teams` doesn't need to be, and shouldn't be, in this PR. It confuses the issue. My statement is that this should _not check the actor visibility in internal requests_. However, what you are doing is _not checking team permission in public requests_, which is not actually solving the issue either, because those policies are not implemented in other apps. We have a way forward where those apps don't need to check this.

bhavenst:
Doing this in gateway instead.

PR #967, opened 2026-03-22
[AAP-69161] Fix TOCTOU race in RBAC caching causing FK violations under concurrency

john-westcott-iv:
## Description

- **What:** Fix a TOCTOU (Time-of-Check-Time-of-Use) race condition in `compute_team_member_roles()` and `compute_object_role_permissions()` in `ansible_base/rbac/caching.py`.
- **Why:** Under concurrent API operations (e.g. parallel test workers or simultaneous role assignment + org deletion), `compute_team_member_roles()` collects ObjectRole IDs in a read phase, then later writes them to the `dab_rbac_objectrole_provides_teams` M2M table. A concurrent committed transaction (e.g. org deletion cascading through `rbac_post_delete_remove_object_roles`) can delete those ObjectRoles between the two phases, causing PostgreSQL to reject the M2M insert with: `insert or update on table "dab_rbac_objectrole_provides_teams" violates foreign key constraint — Key (objectrole_id)=(NNN) is not present in table "dab_rbac_objectrole"`.
- **How:** Wrap the `team.member_roles.add()` and `RoleEvaluation.bulk_create()` calls in database savepoints (`transaction.atomic()`). If an `IntegrityError` from a stale FK reference is caught, re-filter to only ObjectRole IDs that still exist and retry. The savepoint ensures the outer transaction is not aborted. If the retry also fails (extremely unlikely), a warning is logged and the state self-corrects on the next recompute.

The same class of bug also exists in `compute_object_role_permissions()` — the `bulk_create` for `RoleEvaluation` rows references ObjectRoles via the `role_id` FK, and `ignore_conflicts=True` only suppresses unique-constraint violations, not FK violations. This PR fixes both locations.

## Type of Change

## Self-Review Checklist

### Prerequisites
- PostgreSQL database (not SQLite — FK enforcement differs)
- Multiple concurrent database connections (e.g. `pytest -n auto`)

### Steps to Test
1. Run the DAB RBAC test suite: `ANSIBLE_BASE_TEST_DIRS="" tox -e py312 -- test_app/tests/rbac/ -v`
2. To reproduce the original race: run AAP platform-services-test-suite with `pytest -n auto` (12+ parallel workers). The race manifests when one worker creates a `role_user_assignment` (triggering `compute_team_member_roles`) while another worker's test cleanup deletes an organization (triggering cascade ObjectRole deletion).

### Expected Results
- All RBAC tests pass
- No `IntegrityError` FK violations on `dab_rbac_objectrole_provides_teams` under concurrent operations
- If a race does occur, it is caught gracefully with a log warning and self-corrects on the next recompute

### Root Cause Analysis
The `transaction.atomic()` wrapper in the serializer's `create()` method does not prevent this race because the competing DELETE is in a **different database connection's already-committed transaction**. The fix uses savepoints (nested `transaction.atomic()`) to create a safe rollback boundary around the specific write operations, combined with a re-validation + retry pattern.

### Why not SELECT FOR UPDATE?
`SELECT FOR UPDATE` on the ObjectRole rows would be heavyweight and could introduce deadlocks in this global recompute function, which iterates all teams. The catch-and-retry approach is simpler, has minimal performance impact in the common (non-racy) case, and the system self-heals because deletion triggers also call `compute_team_member_roles()`.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of team role adds and bulk evaluation writes against race conditions when referenced roles are deleted concurrently: operations now detect stale references, retry once after filtering out removed roles, and log persistent failures without changing public APIs.
* **Tests**
  * Added comprehensive tests for error classification, retry/filter behavior, non-retryable error propagation, and logging of persistent failures.

### DIFF
--- a/ansible_base/rbac/caching.py
+++ b/ansible_base/rbac/caching.py
@@ -4,6 +4,8 @@ from typing import Optional
 from uuid import UUID
 
 from django.conf import settings
+from django.db import transaction
+from django.db.utils import IntegrityError
 
 from ansible_base.rbac.models import ObjectRole, RoleDefinition, RoleEvaluation, RoleEvaluationUUID
 from ansible_base.rbac.permission_registry import permission_registry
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):
Philosophically, I'm totally aligned with this. We recently had a fix that was extremely similar in nature for assignment to creator roles (well, role creation supporting that).

This will be a little bit chattier by adding more transaction blocks. But this is only when doing a write to begin with, and only when RoleEvaluation entries are inserted specifically due to these signals.

PR #799, opened 2025-08-12
Observability and profiling middleware

kdelee:
- What is being changed?
  This commit introduces a ObservabilityMiddleware that provides request tracing, performance profiling, and secure SQL query commenting. This consists of three components (ProfileRequestMiddleware, SQLProfilingMiddleware, TraceContextMiddleware) to simplify testing and maintenance, with a single public middleware for apps to consume.

   - Why is this change needed?
  The goal is to provide a consistent way for any django-ansible-base consumer to gain performance and debugging insights. Many of these features were previously implemented in AWX, but not available in other apps.

   - How does this change address the issue?
  It creates a single middleware (ObservabilityMiddleware) that developers can add to the top of their MIDDLEWARE list. This ensures the correct ordering of tracing, profiling, and SQL commenting.

  Type of Change
  
   - [ ] Bug fix (non-breaking change which fixes an issue)
   - [x] New feature (non-breaking change which adds functionality)
   - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
   - [x] Documentation update
   - [x] Test update
   - [ ] Refactoring (no functional changes)
   - [ ] Development environment change
   - [x] Configuration change

  Self-Review Checklist
  
   - [x] I have performed a self-review of my code
   - [x] I have added relevant comments to complex code sections
   - [x] I have updated documentation where needed
   - [x] I have considered the security impact of these changes
   - [x] I have considered performance implications
   - [x] I have thought about error handling and edge cases
   - [x] I have tested the changes in my local environment

  Testing Instructions
  
  
  Prerequisites
  
   - In your Django settings (e.g., test_app/defaults.py), ensure the new middleware is at the top of the MIDDLEWARE list:

   1   MIDDLEWARE = [
   2       'ansible_base.lib.middleware.observability.ObservabilityMiddleware',
   3       ...
   4   ]

   - To test all features, enable the following settings:

ANSIBLE_BASE_PROFILE_TIMING=True \
ANSIBLE_BASE_PROFILE_NODE=True \
ANSIBLE_BASE_CPROFILE_REQUESTS=True \
ANSIBLE_BASE_SQL_PROFILING=True

Note: ANSIBLE_BASE_CPROFILE_REQUESTS has significant performance implications and is intended for live debugging, not for permanent use in production.

To verify the SQL comments, you must have access to the database logs (e.g., by setting log_statement=all or enabling slow query logging in postgresql.conf).
[...]

PR #974, opened 2026-04-07
[AAP-70480] Introduces a profiling middleware to obtain performance data from endpoint calls

huffmanca:
## Description
This extends https://github.com/ansible/django-ansible-base/pull/799 to include changes from Brennan and also switches to use a static setting (instead of a dynamic one) to avoid performance overhead even when disabled.

The goal is to be able to support obtaining .prof files so that we can perform performance analysis on problematic endpoints.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Define the `PROFILING_ENABLED=true` variable in your AAP instance.
2. Start the AAP instance.
3. Hit one of the endpoints not in the DEFAULT_EXCLUDE_PATHS list.

### Expected Results
Confirm that you see the profiling values in the headers and .prof files were created (default path = `/tmp`).

## Additional Context
This was made into a static setting, instead of using one of the dynamic_settings, so that we can use a simple dict lookup instead of the `get_setting()` method on every request (even when disabled) which would introduce overhead.

### Required Actions

## Summary by CodeRabbit

* **New Features**
  * Request tracing with X-Request-ID propagation and a context manager/decorator for per-execution trace IDs
  * Middleware-based observability: optional profiling, wall-clock timing, cProfile output files, and profiling response headers
  * SQL query metrics with safe query annotation and configurable path exclusions

* **Documentation**
  * Detailed observability middleware guide covering configuration, profiling modes, cProfile file handling, and analysis tips

* **Tests**
  * Extensive tests for tracing, middleware, profiling, SQL metrics, and end-to-end behavior


ansible_base/lib/utils/views/ansible_base.py
```diff
@@ -51,12 +48,7 @@ def finalize_response(self, request, response, *args, **kwargs):
             response['X-API-Product-Version'] = version
 
         response['X-API-Product-Name'] = get_setting('ANSIBLE_BASE_PRODUCT_NAME', _('Unnamed'))
-        response['X-API-Node'] = get_setting('CLUSTER_HOST_ID', _('Unknown'))
-
-        time_started = getattr(self, 'time_started', None)
-        if time_started:
-            time_elapsed = time.time() - self.time_started
-            response['X-API-Time'] = '%0.3fs' % time_elapsed
+        # Note: X-API-Time and X-API-Node are added by ObservabilityMiddleware when profiling is enabled
```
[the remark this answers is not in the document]
huffmanca: I've modified this back to the View, and added `X-API-Total-Time` to the profiling middleware (since it's a better representation of overall time spent) to prevent overwriting the previous header and introduce a regression.

ansible_base/lib/middleware/profiling/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+When the `ANSIBLE_BASE_PROFILE_TIMING` setting is enabled, the middleware adds:
+
+*   `X-API-Time`: The total time taken to process the request, in seconds.
+
+```python
+# settings.py
+ANSIBLE_BASE_PROFILE_TIMING = True
+```
+
```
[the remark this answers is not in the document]
huffmanca: Docs have been updated

ansible_base/lib/middleware/profiling/profile_request.py
```diff
@@ -0,0 +1,174 @@
...
+                os.makedirs(output_dir, exist_ok=True)
+                filename = f"cprofile-{profile_id}.prof"
+                cprofile_filename = os.path.join(output_dir, filename)
+                self.prof.dump_stats(cprofile_filename)
+            except OSError:
+                logger.warning(f"Failed to write cProfile output to {output_dir}, falling back to {tempfile.gettempdir()}")
+                output_dir = tempfile.gettempdir()
+                filename = f"cprofile-{profile_id}.prof"
+                cprofile_filename = os.path.join(output_dir, filename)
+                self.prof.dump_stats(cprofile_filename)
```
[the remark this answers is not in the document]
huffmanca: Addressed. I hesitated to include this, as a temporary system directory not being writeable after including an optional parameter seems a bit like overkill, but we also don't want the directory to fill up with a ton of .prof files and then crash request handling.

test_app/tests/lib/middleware/test_profiling_middleware.py
```diff
@@ -0,0 +1,277 @@
...
+        profiler = DABProfiler(output_dir='/nonexistent/unwritable/path')
+        profiler.start()
+        profile_id = uuid.uuid4()
+        elapsed, cprofile_filename = profiler.stop(profile_id=profile_id)
+
+        self.assertIsNotNone(elapsed)
+        self.assertIsNotNone(cprofile_filename)
+        self.assertTrue(cprofile_filename.startswith(tempfile.gettempdir()))
+        self.assertTrue(os.path.exists(cprofile_filename))
+        os.remove(cprofile_filename)
```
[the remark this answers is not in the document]
huffmanca: Updated this to no longer rely on a non-privileged user. I've modified this to raise an Error every time (which doesn't validate the permission errors, but does validate the code fallback).

ansible_base/lib/middleware/profiling/profile_request.py
```diff
@@ -0,0 +1,202 @@
...
+            # Get output directory: explicit override > dynamic setting > system temp
+            output_dir = self.output_dir if self.output_dir else getattr(settings, 'PROFILING_CPROFILE_DIR', None)
+            if not output_dir:
+                output_dir = tempfile.gettempdir()
+
+            try:
+                os.makedirs(output_dir, exist_ok=True)
+                filename = f"cprofile-{profile_id}.prof"
+                cprofile_filename = os.path.join(output_dir, filename)
+                self.prof.dump_stats(cprofile_filename)
```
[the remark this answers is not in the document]
huffmanca: Added a comment. Profile_id should be a UUID, and it's intentionally being used to help correlate .prof files with incoming requests.

ansible_base/lib/middleware/profiling/profile_request.py
```diff
@@ -0,0 +1,202 @@
...
+
+        elapsed, cprofile_filename = self.profiler.stop(profile_id=request_id)
+
+        if elapsed is not None:
+            response['X-API-Total-Time'] = f'{elapsed:.3f}s'
+
+        if cprofile_filename:
+            response['X-API-CProfile-File'] = cprofile_filename
+
+        return response
```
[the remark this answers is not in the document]
huffmanca: Added a try/except block that stops the Profiler on failed requests.

ansible_base/lib/logging/context.py
```diff
@@ -0,0 +1,50 @@
...
+        self.tokens = []
+
+        if trace_id:
+            try:
+                # Validate that the provided header is a valid UUID
+                uuid.UUID(trace_id)
+                self.trace_id = trace_id
+            except (ValueError, TypeError):
+                # If it's not a valid UUID, discard it and we'll generate a new one
+                self.trace_id = str(uuid.uuid4())
```
AlanCoding, line 28: Somewhere, there is a header that gives the request id. I know I got a lot of requests for this with the task system, saying that the UUID should persist to task the request initiates, if the request starts a background task.

I am failing to find a reference now, but it seems like the most logical id to put on this profiling data is the already-standardized request-tracking UUID that already exists. I would appeal to @kdelee to get what that thing actually is, as I do not remember. It is returned in a header or something like that.

ansible_base/lib/middleware/profiling/profile_request.py
```diff
@@ -0,0 +1,207 @@
...
+        X-API-Total-Time: Wall-clock request duration (e.g. "0.045s").
+        X-API-CProfile-File: Filesystem path to the .prof file on the server.
+    """
+
+    def __init__(self, get_response=None):
+        self.get_response = get_response
+        self.profiler = DABProfiler()
+
+    def __call__(self, request):
+        self.profiler.start()
```
AlanCoding: If you look at the AWX version of this same thing.

https://github.com/ansible/awx/blob/devel/awx/main/middleware.py#L49

There is some work for enablement. That is, dynamic enablement of the middleware.

As written, it's a struggle for me to understand how this would be reasonably used. Even if you put it in a settings override file, you're going to have to know the set of existing middleware and I do not believe this is automatic or easy. Then you restart services, then maybe it'll start working.

Perhaps something in the `docs/` folder might help here, because I don't even see a use case that really makes sense to me.

Maybe a good practice would be to keep this middleware, and then have another "conditional profiler" middleware that wrapps it. Because we are still not on the same DB settings system, and this could allow AWX to delete its logic and only replace the wrapper middleware.

ansible_base/lib/middleware/profiling/README.md
AlanCoding, line 1: ok, you do have a doc. But now I find this location very non-standard. Using the `docs/` folder would be better.

Or else, delete the `docs/` folder and restructure all of the markdown docs into the code.

ansible_base/lib/middleware/profiling/README.md
```diff
@@ -0,0 +1,115 @@
...
+    'ansible_base.lib.middleware.observability.ObservabilityMiddleware',
+    ...
+]
+```
+
+## Settings
+
+| Setting | Default | Description |
+|---|---|---|
+| `PROFILING_ENABLED` | `False` | Master switch. When `False`, only trace context (X-Request-ID) is applied with zero overhead. Requires a restart to take effect. |
```
AlanCoding: am I crazy? I can not find where this setting is referenced anywhere in the code.

ansible_base/lib/logging/context.py
huffmanca: This UUID should be the same as the X-Request-ID from the headers, and it appears to be consistently used across these areas:
* _TraceContextMiddleware (what we're using here)
* RequestIdFilter
* JWTCommonAuth
The goal is that we should be able to correlate the generated .prof files with the X-Request-ID in the logs.

ansible_base/lib/middleware/profiling/README.md
huffmanca: Moved this file into docs and deleted this file.

ansible_base/lib/middleware/profiling/README.md
huffmanca: Previously these were behind comments; they've been renamed to better follow convention (starting with `ANSIBLE_BASE_`, and also are explicitly referenced instead of being abstracted away into a constant that's referenced elsewhere.

ansible_base/lib/middleware/profiling/profile_request.py
huffmanca: Previously we were adding the middleware in the Gateway defaults, but I've moved it so that the `ObservabilityMiddleware` is enabled by default, but the profiling is gated behind the (now) `ANSIBLE_BASE_PROFILING_ENABLED` setting.

> As written, it's a struggle for me to understand how this would be reasonably used. Even if you put it in a settings override file, you're going to have to know the set of existing middleware and I do not believe this is automatic or easy. Then you restart services, then maybe it'll start working.

My hope is that we can toggle this on for debugging purposes - if we see slow requests, then we can enable the setting, restart AAP, capture the .prof files, and then disable it and analyze the files. Ideally, I'd like this to eventually be a fairly automated flow. We've started seeing performance complaints for various endpoints, and so attempting to get profiling data from those endpoints to address/optimize the underlying issue is the main motivator here.

> Because we are still not on the same DB settings system, and this could allow AWX to delete its logic and only replace the wrapper middleware.

I attempted to rework this so that similar data is gated behind the `ANSIBLE_BASE_PROFILING_ENABLED` and keep the same headers that are being used in AWX. 

Would love more thoughts here, though, if there are ways we can achieve the ^^ workflow in a better or more consistent way.

ansible_base/lib/dynamic_config/settings_logic.py
```diff
@@ -144,6 +144,13 @@ def get_mergeable_dab_settings(settings: dict) -> dict:  # NOSONAR
         # these will be combined with the actual DRF defaults in our base view
         dab_data['ANSIBLE_BASE_CUSTOM_VIEW_FILTERS'] = dab_data['ANSIBLE_BASE_ALL_REST_FILTERS']
 
+    # ObservabilityMiddleware provides request tracing (X-Request-ID) and timing (X-API-Total-Time)
+    # on every request. Set ANSIBLE_BASE_PROFILING_ENABLED=True for cProfile .prof files,
+    # and/or ANSIBLE_BASE_PROFILING_SQL_ENABLED=True for SQL query metrics.
+    observability_mw = 'ansible_base.lib.middleware.observability.ObservabilityMiddleware'
+    if observability_mw not in middleware:
+        middleware.insert(0, observability_mw)
```
AlanCoding: I get that we have a tool that can automatically insert into the middleware list, but I don't think it's a good idea to use it. There are a lot of reasons that an app maintainer needs to be able to control the ordering of their middleware. Yeah, they can go add it manually and then if everyone is compliant we can then remove this. That is the typical pattern for tech debt that we see, and it will never get done.

I would prefer to never add this auto-add logic, even if it slows down the rollout to additional components. Because that rollout needs more thought in most cases anyway.

huffmanca: Fair enough - I've removed the dynamic insertion of it here. If a component wants to use it, then they can add it to their Middleware list.

ansible_base/lib/middleware/observability.py
```diff
@@ -0,0 +1,59 @@
+"""
+A single middleware to provide a unified observability layer, ensuring that context,
+profiling, and SQL metrics are captured in the correct order.
+"""
+
+import logging
+
+from django.conf import settings
+
+from .profiling.profile_request import _ProfileRequestMiddleware, _SQLProfilingMiddleware
```
bhavenst, line 10: **Critical: Missing `__init__.py` for the new `profiling` package.**

This import requires `ansible_base/lib/middleware/profiling/` to be a proper Python package, but no `__init__.py` is included in the PR. The sibling directory `ansible_base/lib/middleware/logging/` has one, establishing the project convention.

Without it, imports may fail depending on the tool/configuration (pytest, mypy, Django's `import_string()` used by the test `MIDDLEWARE` lists).

**Fix:** Add an empty `ansible_base/lib/middleware/profiling/__init__.py`.

docs/lib/observability.md
```diff
@@ -0,0 +1,128 @@
+# Observability Middleware
+
+The `ObservabilityMiddleware` provides request tracing, timing, cProfile analysis, and SQL query metrics through a single middleware entry point.
+
+## Setup
+
+The middleware is **automatically installed** at the top of the `MIDDLEWARE` stack by DAB's `settings_logic`. No manual configuration is needed.
```
bhavenst: **Critical: This statement is incorrect.** `settings_logic.py` only auto-installs `SocialExceptionHandlerMiddleware` and `AuthenticatorBackendMiddleware` — there is no code that auto-installs `ObservabilityMiddleware`. Users following this doc will find no observability features work.

Either add the auto-install logic to `settings_logic.py` as part of this PR, or rewrite this section to show manual `MIDDLEWARE` configuration (which matches how the tests configure it via `override_settings`).

docs/lib/observability.md
```diff
@@ -0,0 +1,128 @@
...
+## X-Request-ID and Trace Propagation
+
+The `X-Request-ID` header is the canonical request-tracking identifier across the platform. The middleware:
+
+1. Reads the incoming `X-Request-ID` header (if present and a valid UUID).
+2. Generates a new UUID if the header is missing or invalid.
+3. Stores the ID in a `ContextVar` (`trace_id_var`) accessible throughout the request lifecycle.
+4. Returns the ID on the response as `X-Request-ID`.
+
+This is the same UUID available to logging via the `RequestIdFilter`, and it can be propagated to background tasks via the `trace_context` context manager (see below).
```
bhavenst: **Critical: This claim is misleading.** The existing `RequestIdFilter` in `ansible_base/lib/logging/filters/request_id.py` reads `X-Request-ID` from `request.META` via a `thread_local.request` mechanism — it does **not** read from `trace_id_var`. These are two separate mechanisms.

When no header is sent, the middleware generates a UUID in `trace_id_var`, but `RequestIdFilter` will return an empty string. They only agree when the client sends a valid UUID header.

Either update `RequestIdFilter` to read from `trace_id_var` (making this statement true), or rewrite this sentence to accurately describe the relationship.

ansible_base/lib/logging/context.py
```diff
@@ -0,0 +1,52 @@
...
+        # Reset the context variables to their previous state
+        for token in self.tokens:
+            var = token.var
+            var.reset(token)
+
+    def __call__(self, func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            with self:
+                return func(*args, **kwargs)
```
bhavenst, line 55: **Critical: Concurrency hazard with decorator reuse.**

When used as `@trace_context(origin='foo')`, the same instance serves every invocation, sharing `self.tokens`. If the decorated function is called concurrently (even in asyncio on a single thread), `__exit__` iterates over *all* tokens and could drain tokens from a different invocation.

**Fix:** Create a fresh instance per invocation:
```python
def __call__(self, func):
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        ctx = trace_context(origin=self.origin, trace_id=self.trace_id)
        with ctx:
            return func(*args, **kwargs)
    return wrapper
```

ansible_base/lib/middleware/profiling/profile_request.py
```diff
@@ -0,0 +1,218 @@
...
+                try:
+                    output_dir = tempfile.gettempdir()
+                    filename = f"cprofile-{profile_id}.prof"
+                    cprofile_filename = os.path.join(output_dir, filename)
+                    self.prof.dump_stats(cprofile_filename)
+                except OSError:
+                    logger.warning(f"Failed to write cProfile output to fallback {output_dir}, discarding profile data")
+                    cprofile_filename = None
+
+        return elapsed, cprofile_filename
```
bhavenst, line 80: **Critical: `stop()` does not reset internal state.**

`self.start_time` and `self.prof` retain values from the completed run. Since `_ProfileRequestMiddleware` reuses the same `DABProfiler` per-thread via `threading.local`, a double `stop()` call would return stale timing (measured from the *original* `start_time`), and the `.prof` file could be written twice with different data.

**Fix:** Add before the return:
```python
self.start_time = None
self.prof = None
return elapsed, cprofile_filename
```

bhavenst (commented):
## PR Review Summary

Thanks for this PR! The observability middleware architecture is well-designed — the facade pattern composing three concerns (trace context, cProfile, SQL metrics) into a single middleware entry point is clean and prevents misconfiguration. The defense-in-depth SQL sanitization, proper `contextvars` usage for trace propagation, and the `threading.local` approach for cProfile thread safety are all solid choices.

I've left inline comments on specific issues. Here's the high-level breakdown:

### Critical (5)
1. Missing `__init__.py` for the `profiling` package *(inline)*
2. Documentation claims auto-install that doesn't exist *(inline)*
3. Documentation claims `RequestIdFilter` integration that doesn't exist *(inline)*
4. `trace_context` decorator reuse shares mutable state — concurrency hazard *(inline)*
5. `DABProfiler.stop()` doesn't reset state between requests *(inline)*

### Important (4)
6. `trace_context.__enter__` doesn't return `self` *(inline)*
7. Inconsistent `ValueError`/`TypeError` catching between `trace_context` and `_TraceContextMiddleware` *(inline)*
8. Profile data path silently discarded on exception *(inline)*
9. `DABProfiler.__init__` accepts `*args, **kwargs` but only uses `output_dir` *(inline)*

### Test Coverage Gaps

The test suite is solid overall (good feature-flag matrix, context isolation, malicious input handling). Key gaps to consider adding:

- **Exception paths**: No tests verify that context variables are properly reset when `get_response()` raises in `_TraceContextMiddleware`, that the profiler is cleaned up on exception in `_ProfileRequestMiddleware`, or that exceptions propagate cleanly through `_SQLProfilingMiddleware`. These are the most likely source of production issues since error paths run precisely when things are already going wrong.
- **Thread safety of `_ProfileRequestMiddleware(threading.local)`**: The `threading.local` inheritance prevents cross-thread profiler corruption, but there's no test verifying this property. The `trace_id_var` thread safety test exists in `test_context.py` but uses a completely different mechanism (`contextvars`).
- **`DABProfiler` graceful degradation**: When another cProfile is already active, `start()` silently falls back to timing-only — this path isn't tested. Also, `stop()` without a preceding `start()` isn't tested.

### Minor Items (not inlined)

- `_sanitize_for_sql_comment` docstring says "Removes any `*/` sequences" but line 141 also removes `/*` — the docstring should mention both.
- The comment on line 23 of `context.py` says "Validate that the provided *header* is a valid UUID" but `trace_context` is for non-web contexts — `trace_id` is a plain string parameter. Looks like a copy-paste from `_TraceContextMiddleware`.
- Several inline comments in `request_context.py` restate WHAT rather than WHY (e.g., `# Get the request ID from the header` above `request.headers.get('x-request-id')`). These add maintenance burden without aiding comprehension.
- `_is_excluded` reads settings via `getattr` on every request. Minor, but caching in `__init__` would be consistent with the PR description's philosophy of avoiding per-request overhead.

### Strengths

- Clean middleware composition with fixed chain order enforced at construction
- Proper use of `contextvars.Token.reset()` for nested context restoration
- Excellent defense-in-depth comment on `_sanitize_for_sql_comment` explaining why sanitization exists even for trusted input
- The chain order comment in `observability.py` is valuable since reading nested constructors backwards is error-prone
- Thorough test matrix covering all four combinations of `PROFILING_ENABLED` × `SQL_ENABLED`
- `DABProfiler` docstring with the output directory resolution order is concise and exactly the right level of detail
- Graceful handling of "another profiler already active" in `DABProfiler.start()` with a clear comment explaining the gRPC thread pool scenario

*Review generated with Claude Code*

ansible_base/lib/logging/context.py
```diff
@@ -0,0 +1,52 @@
...
+        else:
+            self.trace_id = str(uuid.uuid4())
+
+    def __enter__(self):
+        # Set the trace ID for this context
+        self.tokens.append(trace_id_var.set(self.trace_id))
+
+        # Set the origin (e.g., 'dispatcher')
+        if self.origin:
+            self.tokens.append(origin_var.set(self.origin))
```
bhavenst, line 38: **Important: `__enter__` does not return `self`.**

Per the context manager protocol (PEP 343), `__enter__` should return the value bound by the `as` clause. Currently `with trace_context(...) as ctx:` binds `ctx` to `None`. Since this is a public API documented for downstream use in `docs/lib/observability.md`, this will cause confusing `AttributeError`s for anyone using the `as` clause.

**Fix:** Add `return self` at the end of this method.

ansible_base/lib/middleware/request_context.py
```diff
@@ -0,0 +1,41 @@
...
+        # Get the request ID from the header
+        header_trace_id = request.headers.get('x-request-id')
+        trace_id = None
+
+        if header_trace_id:
+            try:
+                # Validate that the provided header is a valid UUID
+                uuid.UUID(header_trace_id)
+                trace_id = header_trace_id
+            except ValueError:
```
bhavenst, line 23: **Important: Inconsistent exception handling.** This catches only `ValueError`, but the equivalent code in `trace_context.__init__` (context.py:26) catches `(ValueError, TypeError)`. While `request.headers.get()` always returns a string or `None` (and `None` is guarded by the `if header_trace_id` check), the inconsistency could confuse maintainers.

Consider either catching `(ValueError, TypeError)` in both places for consistency, or adding a brief comment here explaining why `TypeError` is unnecessary in this context.

ansible_base/lib/middleware/profiling/profile_request.py
```diff
@@ -0,0 +1,218 @@
...
+        if cprofile_enabled:
+            self.profiler.start()
+
+        start_time = time.time()
+
+        try:
+            response = self.get_response(request)
+        except Exception:
+            if cprofile_enabled:
+                self.profiler.stop(profile_id=request_id)
```
bhavenst: **Important: Profile data path silently discarded.**

When `get_response` raises, the `.prof` file is written to disk but its path is thrown away (not assigned). For erroring requests, the profile data is *especially* valuable for debugging, but the operator has no way to locate it.

Consider logging the path:
```python
_, path = self.profiler.stop(profile_id=request_id)
if path:
    logger.info("Request raised an exception; cProfile data saved to: %s", path)
```

ansible_base/lib/middleware/profiling/profile_request.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+class DABProfiler:
+    """cProfile wrapper that writes .prof files to a configurable directory.
+
+    Output directory resolution: explicit override > ANSIBLE_BASE_PROFILING_CPROFILE_DIR setting > system temp.
+    Falls back to the system temp directory on write errors.
+    """
+
+    def __init__(self, *args, **kwargs):
```
bhavenst, line 26: **Suggestion:** `*args, **kwargs` accepts arbitrary positional and keyword arguments silently — e.g., `DABProfiler("foo", bar=True)` succeeds without error, masking misuse. Since only `output_dir` is used, consider:
```python
def __init__(self, *, output_dir=None):
```

ansible_base/lib/middleware/profiling/profile_request.py
```diff
@@ -0,0 +1,218 @@
...
+
+    The timing header (X-API-Total-Time) is always set. cProfile output is
+    only produced when the ANSIBLE_BASE_PROFILING_ENABLED setting is True.
+
+    Response headers:
+        X-API-Total-Time: Wall-clock request duration (e.g. "0.045s"). Always set.
+        X-API-Profile-File: Filesystem path to the .prof file on the server (cProfile only).
+    """
+
+    def __init__(self, get_response=None):
```
bhavenst, line 94: **Suggestion:** `get_response=None` means construction succeeds but `__call__` will raise `TypeError: 'NoneType' object is not callable`. Failing at construction time is easier to debug — consider removing the default:
```python
def __init__(self, get_response):
```

Also: inheriting from `threading.local` is the right call for WSGI thread safety, but it's non-obvious enough that a brief class-level comment would help prevent a future maintainer from "simplifying" it away. Something like: *"Inherits from `threading.local` so that `self.profiler` is per-thread, preventing race conditions in multi-threaded WSGI servers where Django creates a single middleware instance shared across threads."*

AlanCoding (approved):

ansible_base/lib/middleware/observability.py
huffmanca: Included this file.

docs/lib/observability.md
huffmanca: Updated to indicate how to add this Middleware (removed per Alan's most recent comment).

docs/lib/observability.md
huffmanca: Fixed this to clarify the two different systems.

ansible_base/lib/logging/context.py
huffmanca: Fixed.

ansible_base/lib/middleware/profiling/profile_request.py
huffmanca: Good catch. fixed.

ansible_base/lib/logging/context.py
huffmanca: Added `return self` to this method.

ansible_base/lib/middleware/request_context.py
huffmanca: I'm skipping this for now. There's a safeguard already in place, which is acknowledged by this comment.

ansible_base/lib/middleware/profiling/profile_request.py
huffmanca: Added this so the path is logged.

ansible_base/lib/middleware/profiling/profile_request.py
huffmanca: Skipping this for now, since it's an internal class.

ansible_base/lib/middleware/profiling/profile_request.py
huffmanca: I'm skipping these for now.

bhavenst (commented):
## Re-review: Feedback Addressed

Thanks for the quick turnaround! The critical issues from the first review have all been addressed:

- ✅ `__init__.py` added for the `profiling` package
- ✅ Docs rewritten with correct manual `MIDDLEWARE` setup (no more false auto-install claim)
- ✅ `RequestIdFilter` relationship accurately described
- ✅ `trace_context.__call__` now creates a fresh instance per invocation (concurrency fix)
- ✅ `DABProfiler.stop()` now resets `start_time` and `prof` to `None`
- ✅ `trace_context.__enter__` returns `self`
- ✅ Profile data path logged on exception in `_ProfileRequestMiddleware`

### Remaining Items (non-blocking)

A few lower-priority items from the original review were not addressed. These are suggestions, not blockers:

1. **`_TraceContextMiddleware` catches only `ValueError`** (`request_context.py:23`) while `trace_context.__init__` catches `(ValueError, TypeError)`. Minor inconsistency — `request.headers.get()` always returns `str | None` so `TypeError` can't happen here, but a one-line comment explaining why would prevent future confusion.

2. **`DABProfiler.__init__(self, *args, **kwargs)`** (`profile_request.py:25`) — still accepts arbitrary args silently. Low risk since this is mostly used internally, but `def __init__(self, *, output_dir=None):` would be cleaner.

3. **`_ProfileRequestMiddleware.__init__` defaults `get_response=None`** (`profile_request.py:93`) — will raise `TypeError` at call time rather than construction. Minor.

4. **`_sanitize_for_sql_comment` docstring** (`profile_request.py:272`) — still says "Removes any `*/` sequences" but the code also removes `/*`. Trivial fix.

### Overall

The PR is in good shape. The architecture is solid, the critical fixes landed cleanly, and the remaining items are minor enough to address in a follow-up if desired.

*Review generated with Claude Code*

bhavenst (approved):

bhavenst (approved):

PR #977, opened 2026-04-09
[AAP-71514] Address deletion with partially completed requests in resource_sync 

huffmanca:
## Description

Fixes an issue where deletions were performed on get_remote_assignments when the requests were paginated and errored out during pagination.

We saw an error where, with a large number of role_user_assignments, Controller was deleting these assignments during the background sync job. This appears to be because the background sync errored part-way through the request, so only some of the pages had data, and the rest were cut off (for instance, we fetched the first 40 pages, but the remainder weren't included in the response, so were deleted as they were missing from the set).

This commit attempts to resolve that issue. 

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites
* Have a large number of role_user_assignments, forcing the background sync job to paginate over them.

### Steps to Test
1. Wait for the background sync to begin (should run every 15 minutes).
2. Bring the Gateway node down, so that the request fails part-way through.
3. Confirm that no role_user_assignments were deleted, even on the failed sync.

### Expected Results
We may create on an error, but we shouldn't delete role_user_assignments (or other resource syncs) if there's an error during the requests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Assignment sync now preserves partial remote results and defers deletions when remote fetches are incomplete or fail; creations from available remote data are still applied and a warning plus “Skipping assignment deletions” is logged to prevent accidental removals.

* **Tests**
  * Added tests covering pagination success, pagination failures/exceptions, incomplete fetch behavior, deferred-deletion handling, and creation-from-partial-data.

### DIFF
--- a/ansible_base/resource_registry/tasks/sync.py
+++ b/ansible_base/resource_registry/tasks/sync.py
@@ -182,39 +182,64 @@ class RemoteAssignmentResult:
     is_complete: bool = False
 
 
-def _paginate_assignments(list_fn, actor_id_key: str, assignment_type: str, assignments: set[AssignmentTuple]) -> bool:
-    """Paginate a single assignment type endpoint, adding results to *assignments*.
+class RemoteAssignmentFetcher:
+    """Fetches role assignments from a remote resource server with pagination.
 
-    Returns True if all pages were fetched successfully, False on any error.
+    Collects user and team assignments into a single set.  If any page
+    request fails the fetcher stops early and marks the result as
+    incomplete so the caller can skip deletions safely.
     """
-    page = 1
-    try:
-        while True:
-            resp = list_fn(filters={'page': page})
-            if resp.status_code != 200:
-                logger.warning(f"Failed to fetch {assignment_type} assignments page {page}: HTTP {resp.status_code}")
-                return False
 
-            data = resp.json()
-            for assignment in data.get('results', []):
-                ansible_id_or_pk = assignment.get('object_ansible_id') or assignment.get('object_id')
-                assignments.add(
-                    AssignmentTuple(
-                        actor_ansible_id=assignment[actor_id_key],
-                        ansible_id_or_pk=ansible_id_or_pk,
-                        role_definition_name=assignment['role_definition'],
-                        assignment_type=assignment_type,
+    def __init__(self, api_client: ResourceAPIClient):
+        self.api_client = api_client
+        self.assignments: set[AssignmentTuple] = set()
+
+    def fetch(self) -> RemoteAssignmentResult:
+        """Paginate user then team assignments and return the result.
+
+        If user pagination fails, team pagination is skipped entirely
+        because the result will be incomplete regardless.
+        """
+        users_ok = self._paginate(self.api_client.list_user_assignments, 'user_ansible_id', 'user')
+        if not users_ok:
+            return RemoteAssignmentResult(assignments=self.assignments, is_complete=False)
+
+        teams_ok = self._paginate(self.api_client.list_team_assignments, 'team_ansible_id', 'team')
+        return RemoteAssignmentResult(assignments=self.assignments, is_complete=teams_ok)
+
+    def _paginate(self, list_fn, actor_id_key: str, assignment_type: str) -> bool:
+        """Paginate a single assignment endpoint, adding results to ``self.assignments``.
+
+        Returns True if all pages were fetched successfully, False on any error.
+        """
+        page = 1
+        try:
+            while True:
+                resp = list_fn(filters={'page': page})
+                if resp.status_code != 200:
+                    logger.warning(f"Failed to fetch {assignment_type} assignments page {page}: HTTP {resp.status_code}")
+                    return False
+
+                data = resp.json()
+                for assignment in data.get('results', []):
+                    ansible_id_or_pk = assignment.get('object_ansible_id') or assignment.get('object_id')
+                    self.assignments.add(
+                        AssignmentTuple(
+                            actor_ansible_id=assignment[actor_id_key],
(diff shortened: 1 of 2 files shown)


ansible_base/resource_registry/tasks/sync.py
```diff
@@ -169,75 +169,94 @@ def get_content_object(role_definition, assignment_tuple: AssignmentTuple) -> An
...
+                        f"Error fetching {assignment_type} assignments page {page}: "
+                        f"{e} (attempt {attempt + 1}/{max_retries + 1})"
                     )
+                    resp = None
 
-                # Check if there's a next page
-                if not team_data.get('next'):
-                    break
+                if attempt < max_retries:
+                    time.sleep(2**attempt)
```
AlanCoding: worried about this. I suspect it would be better to just give up. This will run periodically anyway.

huffmanca: Sounds good. I dropped the latest commit which introduced the retry mechanism.

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -169,75 +169,67 @@ def get_content_object(role_definition, assignment_tuple: AssignmentTuple) -> An
...
+    """Fetch remote assignments from the resource server and convert to tuples.
+
+    Returns a ``RemoteAssignmentResult`` so the caller can distinguish a
+    complete fetch from a partial one (e.g. due to HTTP errors or
+    timeouts mid-pagination).
+    """
+    assignments: set[AssignmentTuple] = set()
+
+    users_ok = _paginate_assignments(api_client.list_user_assignments, 'user_ansible_id', 'user', assignments)
+    teams_ok = _paginate_assignments(api_client.list_team_assignments, 'team_ansible_id', 'team', assignments)
```
AlanCoding: It's a little weird of a function signature that:
 - the first argument takes a callable, and there are a million assumptions about that callable
 - the `assignments` is passed as an argument and mutated by the method

I still don't think you should change the overall structure where it returns a boolean. But it would be better as a class structure. The way you did this, `RemoteAssignmentResult` becomes a rump construct.

And flowing from that, if `users_ok` is `False`, it seems there is no reason to paginate the team assignments. You should just quit. Because you're not going to do anything with the result anyway.

AlanCoding (approved):
My comment is _mostly_ a code style thing. I am okay for it to be non-blocking. This accomplishes the basic goal, which is very important to ship.

ansible_base/resource_registry/tasks/sync.py
huffmanca: This was changed to use a class structure and also return early if `users_ok` is `False`. I can revert these changes if it ends up getting too unwieldy, but since I'd like to get this backported to 2.6, doing this the "right" way would be my preferred option if at all possible.

AlanCoding (approved):

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -169,75 +169,87 @@ def get_content_object(role_definition, assignment_tuple: AssignmentTuple) -> An
...
-                logger.debug(f"Fetching next page {page} of team assignments")
-            else:
-                logger.warning(f"Failed to fetch team assignments page {page}: HTTP {team_resp.status_code}")
-                break
-    except Exception as e:
-        logger.warning(f"Failed to fetch remote team assignments: {e}")
+                logger.debug(f"Fetching next page {page} of {assignment_type} assignments")
+        except Exception as e:
+            logger.exception(f"Failed to fetch remote {assignment_type} assignments: {e}")
+            return False
```
john-westcott-iv, line 243: **Nit (non-blocking):** `logger.exception()` already appends the full exception info (type, message, traceback) to the log record. Including `{e}` in the format string causes the exception message to appear twice in the output. Cleaner to drop it:

```suggestion
            logger.exception(f"Failed to fetch remote {assignment_type} assignments")
```

john-westcott-iv (commented):
## 📋 Code Review Summary

Critical bug fix that prevents accidental deletion of role assignments when remote API pagination fails mid-request. The `RemoteAssignmentResult` dataclass with `is_complete` flag (defaulting to `False`) is a clean, safe-by-default design. The `RemoteAssignmentFetcher` class nicely encapsulates pagination with DRY handling of user and team assignment types.

**Files Reviewed**: 2 files

### 🔍 Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 1 code quality suggestion (minor nit, non-blocking)

### Overall Assessment
**LGTM** — This is an important safety fix. The core logic is correct: `is_complete` defaults to `False` (safe), deletions only happen on a complete fetch, creations are safe on partial fetches. The short-circuit optimization (skip team pagination when user pagination fails) is a good addition.

### General Feedback
- Good safe-by-default: `is_complete: bool = False` means any unexpected code path that doesn't explicitly set `True` will prevent deletions
- The `_apply_assignment_changes` helper fixes the pre-existing f-string bug (missing `f` prefix on continuation strings) from the old DELETED/CREATED log messages
- Clean refactoring from duplicated user/team pagination loops into the generic `_paginate` method
- Test coverage is solid: parametrized failure modes (HTTP error + exception), incomplete fetch skip, complete success path, and verification that team pagination is skipped on user failure
- 93.7% SonarCloud coverage on new code is well above the 80% threshold

john-westcott-iv (approved):
Do we want to increase the page size while we are mucking around in here?

ansible_base/resource_registry/tasks/sync.py
huffmanca: Dropped this here to use the suggested fix.

PR #978, opened 2026-04-10
[AAP-68463] fix: recurring error with RoleDefinition sync

ttuffin:
## Description

- What is being changed?
Add local role name filter to get_remote_assignments() to resolve a bug where AAP components were reporting sync errors caused by Gateway returning assignments for all services, not just the ones specific to that component, which produces the below error in the logs when the periodic sync is performed.
```
pulp [7b6124db8e5b43638c1aa5276c3318ba]: ansible_base.resources_api.tasks.sync:ERROR: Failed to create assignment AssignmentTuple(actor_ansible_id='1d5643cf-e316-4405-8957-5c7ffd05b354', ansible_id_or_pk='1', role_definition_name='Credential Admin', assignment_type='user'): RoleDefinition matching query does not exist.
```

- Why is this change needed?
To address the bug reported in https://redhat.atlassian.net/browse/AAP-68463

Introduces local role name filtering in `get_remote_assignments()` to prevent cross-service assignments from entering the sync diff for each service, which eliminates the sync errors.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Run a new devel deployment and execute resource sync
2. Review one of the component log (EDA/Hub)

### Expected Results
- The "RoleDefinition matching query does not exist" should no longer be reported in component logs. 
- If debug logging is enabled, a "Skipping remote user assignment with unknown local role" message will be present indicate an unknown role was skipped for that components.

### Required Actions

- [ ] Blocked by PR/MR: #XXX
  

## Summary by CodeRabbit

* **Bug Fixes**
  * Sync now excludes remote assignments that reference roles not present locally; these are skipped and recorded, improving data integrity and preventing sync errors.

* **Tests**
  * Added database-backed tests to verify unknown-role assignments are filtered, that only valid assignments remain, and that sync reports completion.

### DIFF
--- a/ansible_base/resource_registry/tasks/sync.py
+++ b/ansible_base/resource_registry/tasks/sync.py
@@ -200,6 +200,10 @@ class RemoteAssignmentFetcher:
         If user pagination fails, team pagination is skipped entirely
         because the result will be incomplete regardless.
         """
+        from ansible_base.rbac.models.role import RoleDefinition
+
+        self.local_role_names: set[str] = set(RoleDefinition.objects.values_list('name', flat=True))
+
         users_ok = self._paginate(self.api_client.list_user_assignments, 'user_ansible_id', 'user')
         if not users_ok:
             return RemoteAssignmentResult(assignments=self.assignments, is_complete=False)
@@ -222,12 +226,16 @@ class RemoteAssignmentFetcher:
 
                 data = resp.json()
                 for assignment in data.get('results', []):
+                    role_name = assignment['role_definition']
+                    if role_name not in self.local_role_names:
+                        logger.debug(f"Skipping remote {assignment_type} assignment with unknown local role: {role_name}")
+                        continue
                     ansible_id_or_pk = assignment.get('object_ansible_id') or assignment.get('object_id')
                     self.assignments.add(
                         AssignmentTuple(
                             actor_ansible_id=assignment[actor_id_key],
                             ansible_id_or_pk=ansible_id_or_pk,
-                            role_definition_name=assignment['role_definition'],
+                            role_definition_name=role_name,
                             assignment_type=assignment_type,
                         )
                     )
--- a/test_app/tests/resource_registry/test_resource_sync.py
+++ b/test_app/tests/resource_registry/test_resource_sync.py
@@ -442,9 +442,11 @@ def _mock_response(status_code=200, body=None):
     return resp
 
 
+@pytest.mark.django_db
 @pytest.mark.parametrize("failure_mode", ["http_error", "exception"])
 def test_get_remote_assignments_incomplete_on_failure(failure_mode):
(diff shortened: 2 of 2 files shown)


test_app/tests/resource_registry/test_resource_sync.py
```diff
@@ -370,3 +371,57 @@ def test_role_assignment_resource_sync(mock_delete, mock_create, static_api_clie
         assert executor.results["assignments_created"] == [0]
         assert executor.results["assignments_deleted"] == [1]
         assert executor.results["assignment_errors"] == [0]
+
+
+@pytest.mark.django_db
+def test_get_remote_assignments_filters_unknown_roles(static_api_client):
```
AlanCoding, line 494: Other than the way it is written, this does not appear to be different in function from the existing `test_resource_sync_create_non_local_role_definition` inside of this file.

I knew this was previously an issue, and https://github.com/ansible/django-ansible-base/pull/811 did what this PR is also trying to do. Still trying to understand what the disconnect is.

AlanCoding: Oh, that one was for _role definitions_ and this one is for _assignment_ syncing. Ok, this makes sense, as the next fix along that path.

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -182,13 +189,17 @@ def get_remote_assignments(api_client: ResourceAPIClient) -> set[AssignmentTuple
             if user_resp.status_code == 200:
                 user_data = user_resp.json()
                 for assignment in user_data.get('results', []):
+                    role_name = assignment['role_definition']
+                    if role_name not in local_role_names:
+                        logger.debug(f"Skipping remote user assignment with unknown local role: {role_name}")
+                        continue
```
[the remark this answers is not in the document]
AlanCoding: The response is produced by the same views as defined here. Those aren't just randomly going to not have a field given in the model/serializer. This can safely proceed with the assumption that it is same-code generated.

AlanCoding (approved):
The bones of the patch are actually pretty simple, and it seems to be the best option available.

AlanCoding:
Conflicts are not surprising due to https://github.com/ansible/django-ansible-base/pull/977 @huffmanca hitting around the same functionality.

That moved around some code blocks. High-level, this patch is absolutely necessary, this issue wasn't touched by the other patch.

ttuffin:
I have rebased and updated a couple of existing tests because:
a) the RoleDefinition call in `fetch()` queries the database for filtering, so these tests now need the `pytest.mark.django_db` marker
b) we need to ensure the "team_member" exists for filtering

AlanCoding (approved):

PR #840, opened 2025-09-16
AAP-53611 Introduce a bulk re-computation context manager for RBAC

AlanCoding:
## Description
  This PR implements a bulk_rbac_caching context manager to optimize performance during bulk RBAC permission
  operations. Currently, each call to give_permission or remove_permission triggers expensive cache updates via
  update_after_assignment, which calls compute_team_member_roles() and compute_object_role_permissions(). During
  migration processes or other bulk operations, this results in excessive resource consumption as these expensive
  functions are called repeatedly for each individual permission change.

  The new context manager defers these cache updates during bulk operations by collecting update parameters in
  thread-local storage and performing a single global update when exiting the context. This maintains all existing
  functionality and correctness while providing significant performance improvements for bulk scenarios. The
  implementation includes comprehensive tests covering deferred updates, parameter collection, nested contexts,
  exception handling, and performance verification. Usage is simple: wrap bulk operations with with 
  bulk_rbac_caching(): and cache updates will be automatically batched and executed once at the end.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test update
- [ ] Refactoring (no functional changes)
- [ ] Development environment change
- [ ] Configuration change

PR #979, opened 2026-04-13
[AAP-70854] Optimize JWT claims reconciliation performance

john-westcott-iv:
## Summary

- **Wrap `save_user_claims()` in `no_reverse_sync()`** to suppress redundant HTTP calls back to Gateway during stub Org/Team creation triggered by `get_or_create_resource()` post_save signals
- **Wrap in `transaction.atomic()`** so partial failures roll back cleanly instead of leaving users in an inconsistent permission state (especially important for services without `ATOMIC_REQUESTS`)
- **Add `defer_role_evaluation()` context manager** in `triggers.py` that collects all affected `ObjectRole` instances and processes `compute_object_role_permissions()` in a single batch on context exit, reducing N individual compute+write cycles to one
- **Add timing and grant/removal counters** to `save_user_claims()` and `process_rbac_permissions()` for production observability of claims reconciliation duration

## Problem

When a user with many role assignments (e.g., superusers with hundreds of explicit assignments) authenticates via JWT to a service, `save_user_claims()` executes N individual `give_permission()`/`remove_permission()` calls. Each call triggers:
1. An individual `RoleEvaluation` cache recompute (`compute_object_role_permissions()`)
2. Potential `post_save` signals from stub Org/Team creation that fire HTTP reverse sync calls back to Gateway

This caused timeouts during authentication and occasional permission loss when reconciliation didn't complete.

### `ansible_base/rbac/triggers.py`
- New `_DeferredEvaluationState` class and `defer_role_evaluation()` context manager
- Modified `update_after_assignment()` to defer when evaluation is active
- Supports nesting by saving/restoring previous state

### `ansible_base/rbac/claims.py`
- `save_user_claims()` now wrapped in `no_reverse_sync()`, `transaction.atomic()`, and `defer_role_evaluation()`
- Added grant/removal counters and timing log

### `ansible_base/jwt_consumer/common/auth.py`
- Added timing breakdown in `process_rbac_permissions()`: fetch time, save time, total time

### New Tests
- `test_app/tests/rbac/test_defer_role_evaluation.py` — 14 tests covering context manager state, nesting, exception handling, deferred batching, and integration with give/remove permission
- `test_app/tests/rbac/test_save_user_claims_optimizations.py` — 6 tests verifying atomicity, reverse sync suppression, deferred evaluation, and logging
- `test_app/tests/jwt_consumer/common/test_auth_timing.py` — 4 tests verifying timing logs in process_rbac_permissions

## Test plan

- [x] All 184 targeted tests pass (ran with `py312-sqlite`)
- [ ] Existing `test_claims.py`, `test_triggers.py`, and `test_auth.py` tests continue to pass
- [ ] CI pipeline passes
- [ ] Verify timing logs appear in service logs during JWT authentication

---

## Release Notes

* **Bug Fixes**
  * Improved robustness of role-based access control (RBAC) operations by detecting and handling stale data conditions in concurrent scenarios.
  * Enhanced transaction atomicity for permission management operations.

* **Performance**
  * Added timing instrumentation to JWT claims reconciliation and permission synchronization for diagnostic insights.

* **Tests**
  * Comprehensive test coverage added for RBAC reliability, caching behavior, and optimization features.

### DIFF
--- a/ansible_base/jwt_consumer/common/auth.py
+++ b/ansible_base/jwt_consumer/common/auth.py
@@ -1,4 +1,5 @@
 import logging
+import time
 import uuid
 from datetime import datetime
 from typing import Optional
@@ -286,18 +287,27 @@ class JWTCommonAuth:
 
         # Claims hash mismatch - fetch from gateway
         logger.info(f"Claims hash mismatch for user {user_ansible_id}. JWT: {jwt_claims_hash}, Local: {local_claims_hash}. Fetching from gateway.")
+        reconcile_start = time.monotonic()
         try:
+            fetch_start = time.monotonic()
             gateway_claims = self._fetch_jwt_claims_from_gateway(user_ansible_id)
-            # Extract claims structure from gateway response
+            fetch_elapsed = time.monotonic() - fetch_start
+
             objects = gateway_claims.get('objects', {})
             object_roles = gateway_claims.get('object_roles', {})
             global_roles = gateway_claims.get('global_roles', [])
 
-            # Process the RBAC permissions with the gateway claims
+            save_start = time.monotonic()
             save_user_claims(self.user, objects, object_roles, global_roles)
+            save_elapsed = time.monotonic() - save_start
 
-            # Update cache with the new hash
             self.cache.cache_claims_hash(user_ansible_id, jwt_claims_hash)
+
+            total_elapsed = time.monotonic() - reconcile_start
+            logger.info(
+                f"Claims reconciliation for {user_ansible_id}: "
+                f"fetch={fetch_elapsed:.3f}s, save={save_elapsed:.3f}s, total={total_elapsed:.3f}s"
+            )
         except GatewayLockedException:
             if self.token.get('user_data', {}).get("is_superuser", False) is False:
                 self.log_and_raise(
--- a/ansible_base/rbac/caching.py
+++ b/ansible_base/rbac/caching.py
@@ -4,6 +4,8 @@ from typing import Optional
 from uuid import UUID
 
 from django.conf import settings
+from django.db import transaction
+from django.db.utils import IntegrityError
 
 from ansible_base.rbac.models import ObjectRole, RoleDefinition, RoleEvaluation, RoleEvaluationUUID
 from ansible_base.rbac.permission_registry import permission_registry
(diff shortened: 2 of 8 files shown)


huffmanca (commented):
Nice work on this — the batching approach and TOCTOU guards are solid. One thing I noticed:

### `defer_role_evaluation()` runs `compute_*` in two cases where it doesn't need to

The `finally` block in `defer_role_evaluation()` unconditionally calls `compute_team_member_roles()` / `compute_object_role_permissions()`:

1. **On inner context exit when nested** — if `previous_enabled` is `True`, the inner context flushes its roles immediately instead of merging them into the parent's set. This defeats the batching optimization (though nesting isn't used in any current call site, so it's theoretical today).

2. **On exception** — since `save_user_claims()` wraps everything in `transaction.atomic()`, an exception rolls back the DB changes, but `compute_*` still runs against that rolled-back state. The results get corrected on next recompute, but it's throwaway work.

Neither is a correctness issue — `compute_*` is idempotent and the system self-corrects. But both are straightforward to fix by checking whether to merge vs. flush:

```python
completed = False
try:
    yield
    completed = True
finally:
    deferred_roles = _deferred_evaluation.object_roles
    deferred_teams = _deferred_evaluation.needs_team_recompute
    _deferred_evaluation.enabled = previous_enabled
    _deferred_evaluation.object_roles = previous_roles
    _deferred_evaluation.needs_team_recompute = previous_teams

    if previous_enabled:
        previous_roles.update(deferred_roles)
        _deferred_evaluation.needs_team_recompute |= deferred_teams
    elif completed:
        if deferred_teams:
            compute_team_member_roles()
        if deferred_roles:
            compute_object_role_permissions(object_roles=deferred_roles)
```

AlanCoding:
> Wrap save_user_claims() in no_reverse_sync() to suppress redundant HTTP calls back to Gateway during stub Org/Team creation triggered by get_or_create_resource() post_save signals

This is bananas. I can hardly believe this was actually happening. This is gargantuan red flag territory.

Maybe it would be better to cut out a smaller portion of this patch doing this stuff. Because the rest of it goes really deep, and slapping on `no_reverse_sync` is super easy to get moving, and would be good to measure improvement for independent of the rest. In terms of priority of the patch, there's an extremely clear cliff here, so splitting it up is often a good idea in those cases.

AlanCoding:
This also seems to directly overlap with my attempt at bulk re-computation here.

https://github.com/ansible/django-ansible-base/pull/840

You did `defer_role_evaluation()`, compare to `bulk_rbac_caching()` there. These are mostly trying to do the same thing, but there are 1 or 2 extremely critical logical differences. The other one throws up its hands and does the global `compute_object_role_permissions()` if any team roles change, whereas yours doesn't. That, and other things, need more thought about what gets parity with what already exists as the incremental update.

PR #982, opened 2026-04-17
[AAP-73865] Add observability app

chrismeyersfsu:
## Description

- What is being changed?
Add in a telemetry app

- Why is this change needed?
Consistent open telemetry logging across services that use django-ansible-base.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. run the dev server
2. Hit some api endpoints

### Expected Results
OTLP traces and logs displayed in the two text boxes in the homepage of the runserver app

### Required Actions

- [ ] Blocked by PR/MR: #XXX
  

### Screenshots/Logs

image

## Summary by CodeRabbit

* **New Features**
  * App-level OpenTelemetry observability: automatic tracing/logging, test-name header propagation, in-app OTLP ingestion endpoints, background test OTLP server, and a UI showing recent traces/logs.

* **Tests**
  * In-memory span/log exporters fixture and a test asserting spans are emitted.

* **Documentation**
  * New observability guide with setup, configuration, and advanced usage.

* **Chores**
  * Optional observability dependency group and defaults updated (app registration, batch delay, logging null handler).

---

> [!NOTE]
> **Medium Risk**
> Introduces global OpenTelemetry provider setup and logging handler attachment during app startup, which can affect runtime performance and log/trace volume if enabled/misconfigured.
> 
> **Overview**
> Adds a new `ansible_base.observability` Django app that auto-runs `setup_observability()` on startup to configure OTLP span + log export, instrument Django/requests/gRPC/psycopg, and attach an OTEL `LoggingHandler` (including to `propagate=False` loggers).
> 
> Implements request-context header capture/forwarding via hooks (`ANSIBLE_OBSERVABILITY_CAPTURE_HEADERS`) and documents configuration in `docs/apps/observability.md`.
> 
> Extends the `test_app` to exercise the feature (embedded OTLP gRPC receiver, new `/api/v1/otel/*` endpoints, and homepage UI polling traces/logs + a header-propagation demo) and adds observability-focused tests/fixtures.
> 
> Adds an `observability` optional dependency set (and includes it in `all`/tox requirements) plus pinned OTEL/grpc dependencies in `requirements_all.txt`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1308e01895b0dc25af17921d652d94bc4ca13904. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


test_app/views.py
```diff
@@ -245,6 +245,32 @@ class TestViewSetWithBothAttributes(TestAppViewSet):
...
+
+
+def otel_logs(request):
+    from django.http import JsonResponse
+
+    from test_app import otlp_server
+
+    with otlp_server._lock:
+        data = list(otlp_server.recent_logs)
+    return JsonResponse({"logs": data})
```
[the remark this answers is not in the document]
chrismeyersfsu: The `test_app` is for developers only.

ansible_base/observability/instrument.py
```diff
@@ -0,0 +1,100 @@
...
+        from opentelemetry.instrumentation.psycopg2 import Psycopg2Instrumentor
+
+        Psycopg2Instrumentor().instrument()
+    except ModuleNotFoundError:
+        try:
+            from opentelemetry.instrumentation.psycopg import PsycopgInstrumentor
+
+            PsycopgInstrumentor().instrument()
+        except ModuleNotFoundError:
+            logging.getLogger(__name__).warning("psycopg2 nor psycopg found. Failed to instrument psycopg.")
```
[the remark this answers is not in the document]
chrismeyersfsu: The part about required dependencies is true. But the other part, lies.

ansible_base/observability/apps.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ObservabilityConfig(AppConfig):
+    default_auto_field = 'django.db.models.BigAutoField'
+    name = 'ansible_base.observability'
+    label = 'dao_opentelemetry'
+    verbose_name = 'Auto-Instrumented OpenTelemetry'
+
+    def ready(self):
+        setup_observability()
```
bhavenst, line 13: **Critical: Unhandled failure here will crash Django startup.**

`setup_observability()` can raise on missing deps, unreachable OTLP endpoint, version mismatch, or double-initialization (`DjangoInstrumentor` raises `RuntimeError` if already instrumented). An observability module should degrade gracefully to "no telemetry" rather than preventing the application from starting.

Suggestion:
```python
def ready(self):
    try:
        setup_observability()
    except Exception:
        import logging
        logging.getLogger(__name__).exception(
            "Failed to initialize OpenTelemetry observability. "
            "The application will continue without telemetry. "
            "Check OTEL_EXPORTER_OTLP_ENDPOINT and installed dependencies."
        )
    super().ready()
```

ansible_base/observability/instrument.py
```diff
@@ -0,0 +1,100 @@
...
+
+def _outgoing_request_hook(span, request):
+    if test_name := _active_test_name.get():
+        request.headers['X-Test-Name'] = test_name
+
+
+def setup_observability(service_name=None, instrument_non_propagating: bool = True, span_exporter=None, log_exporter=None):
+
+    service_name = service_name or os.environ.get("OTEL_SERVICE_NAME", "aap-generic")
+
```
bhavenst, line 108: **Critical: No idempotency guard — repeated calls cause handler accumulation and crashes.**

`setup_observability()` is public API (exported via `__init__.py`, documented in "Advanced usage") and also called from `AppConfig.ready()`. If called twice:
- `DjangoInstrumentor().instrument()` raises `RuntimeError`
- Duplicate `LoggingHandler` on root logger → every log record exported N times
- Duplicate `BatchSpanProcessor` → every span exported N times
- `trace.set_tracer_provider()` silently ignores the second call, leaking the new provider

Suggestion:
```python
_initialized = False

def setup_observability(...):
    global _initialized
    if _initialized:
        return
    _initialized = True
    # ... rest of function
```

ansible_base/observability/instrument.py
```diff
@@ -0,0 +1,100 @@
...
+        for _name, _logger in logging.Logger.manager.loggerDict.items():
+            if isinstance(_logger, logging.Logger) and not _logger.propagate and not (_name == 'opentelemetry' or _name.startswith('opentelemetry.')):
+                _logger.addHandler(handler)
+
+
+_active_test_name: contextvars.ContextVar[str] = contextvars.ContextVar('active_test_name', default='')
+
+
+def _request_hook(span, request):
+    test_name = request.META.get('HTTP_X_TEST_NAME', '')
```
bhavenst: **Important: `_active_test_name` ContextVar can leak stale values in threaded WSGI deployments.**

In servers using thread pools (e.g., gunicorn sync workers), a thread serves multiple sequential requests. The ContextVar from request N persists into subsequent code running on the same thread. While request N+1's `_request_hook` resets it, `_outgoing_request_hook` runs for *any* outgoing HTTP call from any code path — background tasks, management commands, or Celery workers sharing the instrumented `requests` library could pick up a stale value and inject `X-Test-Name` into unrelated outgoing requests.

Consider scoping the reset more explicitly (e.g., via Django middleware that resets the token on response, or using `_active_test_name.set('')` in a response hook).

ansible_base/observability/instrument.py
```diff
@@ -0,0 +1,100 @@
...
+
+    resource = _setup_tracing(service_name, span_exporter=span_exporter)
+    _setup_logging(resource, instrument_non_propagating=instrument_non_propagating, log_exporter=log_exporter)
+
+    DjangoInstrumentor().instrument(request_hook=_request_hook)
+    try:
+        from opentelemetry.instrumentation.psycopg2 import Psycopg2Instrumentor
+
+        Psycopg2Instrumentor().instrument()
+    except ModuleNotFoundError:
```
bhavenst, line 69: **Important: Catch blocks are too narrow — only `ModuleNotFoundError` is caught.**

If psycopg2 is installed but `Psycopg2Instrumentor().instrument()` fails for another reason (`RuntimeError` from double-instrumentation, `TypeError` from version mismatch), the error propagates up instead of falling through to try psycopg.

Consider broadening to `except Exception` and adding `exc_info=True` to the warning so the actual failure reason is logged:
```python
except Exception:
    try:
        from opentelemetry.instrumentation.psycopg import PsycopgInstrumentor
        PsycopgInstrumentor().instrument()
    except Exception:
        logging.getLogger(__name__).warning(
            "Failed to instrument database driver (psycopg2/psycopg). "
            "Database spans will not be collected.",
            exc_info=True,
        )
```

docs/apps/observability.md
```diff
@@ -0,0 +1,89 @@
...
+        ...
+    },
+    'loggers': {
+        'opentelemetry.exporter.otlp.proto.grpc.exporter': {
+            'handlers': ['null'],
+            'propagate': False,
+        },
+        ...
+    },
+}
```
bhavenst, line 65: **Important: This silences the only error signal for unreachable OTLP endpoints.**

The `BatchSpanProcessor`/`BatchLogRecordProcessor` silently drop telemetry when the collector is unreachable. The only indication is log output from this specific logger. Routing it to `NullHandler` means operators will have zero visibility that their telemetry pipeline is broken — this could go undetected for weeks.

Consider recommending a level filter (`WARNING` or `ERROR`) instead of `NullHandler` so transient retry noise is suppressed but persistent failures remain visible:
```python
'opentelemetry.exporter.otlp.proto.grpc.exporter': {
    'handlers': ['console'],
    'level': 'WARNING',
    'propagate': False,
},
```

bhavenst (commented):
## Review Notes

Thanks for this — the architecture is clean and follows the existing optional-app pattern well. The dependency injection on `setup_observability()` is a nice touch for testability, and the non-propagating logger handling is thoughtful.

A few items below ranging from critical (startup safety) to suggestions. I focused on the most impactful ones.

### Test Coverage

The single test (`test_spans_emitted`) verifies the TracerProvider was initialized, but leaves the majority of the module's behavioral surface untested. For a module that modifies global Python logging state and instruments multiple libraries at startup, I'd suggest adding tests for at least:

1. **Log record export** — emit a log via `logging.getLogger().warning()`, verify it appears in the `InMemoryLogRecordExporter`
2. **Non-propagating logger handling** — create a `propagate=False` logger, verify its records are exported, and verify `opentelemetry.*` loggers are excluded (the infinite recursion guard)
3. **X-Test-Name header propagation** — send a request with `HTTP_X_TEST_NAME` via Django test client, verify the `test.name` span attribute is set
4. **psycopg2/psycopg fallback chain** — mock `ModuleNotFoundError` to exercise all 3 branches
5. **Idempotency** — call `setup_observability()` twice, verify no duplicate spans/crashes

Also note: the `otel_memory_exporters` fixture adds processors that are never removed or shut down on teardown (only the exporter data is cleared). Over a large test suite, processors will accumulate on the global providers. Consider calling `span_processor.shutdown()` / `log_processor.shutdown()` in teardown.

test_app/tests/observability/test_instrument.py
AlanCoding, line 1: really weird to put your tests here when all other tests are in `test_app/tests/`

ansible_base/observability/apps.py
chrismeyersfsu: I would think we want exceptions raised for missing dependencies.

Unreachable OTLP endpoint isn't going to get raised here. Logs and traces gets flushed in batches, async, in background threads.

version mismatch, same with dependencies.

Double-initialization, fix the code and don't do that. Error being raised is preferred.

On the flip side, this is in the `ready()` path. Let's say an opentelemetry package X version is bumped and some interface we are using changes. The nightly run would fail _pretty_ early.

I vote fail early. One day, if/when our logs are clean and we read them then I'd be pro resiliency with an error message.

ansible_base/observability/instrument.py
chrismeyersfsu: I'd rather keep the exception targeted.

test_app/tests/observability/test_instrument.py
chrismeyersfsu: moved

ansible_base/observability/instrument.py
chrismeyersfsu: good feedback, resolved.

test_app/views.py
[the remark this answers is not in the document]
chrismeyersfsu: resolved

ansible_base/observability/apps.py
bhavenst: Sounds reasonable..

AlanCoding (approved):

AlanCoding (approved):

PR #992, opened 2026-05-04
[AAP-69617] Optimize related_fields() and get_summary_fields() to avoid N+1 FK queries

huffmanca:
related_fields() was loading each ForeignKey object from the database just to build a URL. This caused N lazy-load queries per object in list views. Now uses the raw FK id (field.attname) and the related model's URL pattern to construct URLs without loading the object. Falls back to loading the object for models that use get_absolute_url.

get_summary_fields() was loading FK objects even when the FK was null. Now checks the raw FK id first to skip null FKs without a query.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. 
2. 
3. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX
  

## Summary by CodeRabbit

* **Performance Improvements**
  * Optimized related-field URL generation and summary-field assembly to avoid unnecessary object loading and extra database queries; gracefully skips null relations while preserving visible output.

* **Tests**
  * Added tests verifying related-field and summary-field outputs remain correct and that query counts are reduced in null-FK and preloaded (select_related) scenarios.

### DIFF
--- a/ansible_base/lib/abstract_models/common.py
+++ b/ansible_base/lib/abstract_models/common.py
@@ -209,12 +209,15 @@ class AbstractCommonModel(models.Model):
         for field in self._meta.fields:
             if field.name in self.ignore_relations:
                 continue  # This check is beneficial before getattr to prevent some error cases
-            if isinstance(field, models.ForeignObject) and getattr(self, field.name):
-                # ignore relations on inherited django models
-                if field.name.endswith("_ptr"):
-                    continue
-                if hasattr(getattr(self, field.name), 'summary_fields'):
-                    response[field.name] = getattr(self, field.name).summary_fields()
+            # Skip non-FK fields and inherited model pointers (e.g. user_ptr)
+            if not isinstance(field, models.ForeignObject) or field.name.endswith("_ptr"):
+                continue
+            # Check the raw FK id to skip null FKs without loading the related object
+            if getattr(self, field.attname, None) is None:
+                continue
+            related_obj = getattr(self, field.name)
+            if related_obj and hasattr(related_obj, 'summary_fields'):
+                response[field.name] = related_obj.summary_fields()
         return response
 
     def related_fields(self, request):
@@ -228,9 +231,16 @@ class AbstractCommonModel(models.Model):
             if not isinstance(field, models.ForeignKey) or field.name.endswith("_ptr"):
                 continue
 
-            if obj := getattr(self, field.name):
-                if related_url := get_url_for_object(obj):
-                    response[field.name] = related_url
+            # Use the raw FK id (e.g. modified_by_id) to build the URL without
+            # loading the related object from the database.
+            fk_id = getattr(self, field.attname)
+            if fk_id is not None:
+                related_model = field.related_model
+                basename = get_cls_view_basename(related_model)
+                try:
+                    response[field.name] = get_relative_url(f'{basename}-detail', kwargs={'pk': fk_id})
+                except NoReverseMatch:
+                    pass
 
         basename = get_cls_view_basename(self.__class__)
 
--- a/test_app/tests/lib/abstract_models/test_common_query_optimization.py
+++ b/test_app/tests/lib/abstract_models/test_common_query_optimization.py
@@ -1,0 +1,111 @@
+import pytest
+from django.db import connection
+from django.test.utils import CaptureQueriesContext
+
+from ansible_base.lib.abstract_models.common import get_url_for_object
+from test_app.models import Organization, Team, User
+
+
+@pytest.mark.django_db
+class TestRelatedFieldsQueryOptimization:
+    """Verify that related_fields() builds correct URLs using raw FK IDs
+    without loading related objects from the database."""
+
+    def test_related_fields_urls_match_original_method(self, system_user):
+        """The optimized related_fields() must produce the same URLs as
+        loading the full FK object and calling get_url_for_object()."""
+        org = Organization.objects.create(name="test-org")
+        team = Team.objects.create(name="test-team", organization=org)
+
+        related = team.related_fields(None)
+
+        # The organization FK should produce a valid URL
+        assert "organization" in related
+        expected_url = get_url_for_object(org)
+        assert related["organization"] == expected_url
+
+    def test_related_fields_skips_null_fks(self, system_user):
+        """Null FK fields should not appear in related_fields output."""
+        org = Organization.objects.create(name="test-org")
+        # Force a known-null FK to make the assertion non-vacuous
+        Organization.objects.filter(pk=org.pk).update(modified_by=None)
+        org.refresh_from_db()
+        assert org.modified_by_id is None
+
+        related = org.related_fields(None)
+
+        assert "modified_by" not in related
+
+    def test_related_fields_no_lazy_load_queries(self, system_user):
+        """related_fields() should build URLs from raw FK IDs without loading
+        related objects, even when the object was fetched without select_related."""
+        org = Organization.objects.create(name="test-org")
+        team = Team.objects.create(name="test-team", organization=org)
+
+        # Re-fetch without select_related so this exercises the raw FK-id path.
+        # With select_related, the old implementation would also be query-free.
+        team = Team.objects.get(pk=team.pk)
+
+        with CaptureQueriesContext(connection) as ctx:
+            team.related_fields(None)
+
+        # With the optimization, related_fields uses raw FK IDs for URL
+        # construction and should not trigger any FK lazy-load queries.
+        assert len(ctx.captured_queries) == 0, f"related_fields() triggered {len(ctx.captured_queries)} queries: " f"{[q['sql'] for q in ctx.captured_queries]}"
+
+
+@pytest.mark.django_db
+class TestGetSummaryFieldsOptimization:
+    """Verify that get_summary_fields() skips null FKs without loading them."""
+
+    def test_summary_fields_null_fk_no_query(self, system_user):
+        """get_summary_fields() should not trigger any query for a null FK
+        even without select_related."""
+        org = Organization.objects.create(name="test-org")
+        # Force null FKs to exercise the null-FK short-circuit path
+        Organization.objects.filter(pk=org.pk).update(modified_by=None, created_by=None)
+        # Pre-load resource (non-null) so only null-FK behavior is tested
+        org = Organization.objects.select_related("resource").get(pk=org.pk)
+        assert org.modified_by_id is None
+        assert org.created_by_id is None
+
+        with CaptureQueriesContext(connection) as ctx:
+            summary = org.get_summary_fields()
+
+        assert isinstance(summary, dict)
+        assert "modified_by" not in summary
+        assert "created_by" not in summary
+        # Null FKs should be skipped without triggering any queries
+        assert len(ctx.captured_queries) == 0, (
+            f"get_summary_fields() triggered {len(ctx.captured_queries)} queries for null FKs: " f"{[q['sql'] for q in ctx.captured_queries]}"
+        )
+
+    def test_summary_fields_no_lazy_load_with_select_related(self, system_user):
+        """get_summary_fields() should not trigger per-FK lazy-load queries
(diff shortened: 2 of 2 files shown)


ansible_base/lib/abstract_models/common.py
```diff
@@ -209,12 +209,14 @@ def get_summary_fields(self):
...
-                # ignore relations on inherited django models
-                if field.name.endswith("_ptr"):
-                    continue
-                if hasattr(getattr(self, field.name), 'summary_fields'):
-                    response[field.name] = getattr(self, field.name).summary_fields()
+            if not isinstance(field, models.ForeignObject) or field.name.endswith("_ptr"):
+                continue
+            # Check the raw FK id to skip null FKs without loading the related object
+            if hasattr(field, 'attname') and not getattr(self, field.attname, None):
+                continue
```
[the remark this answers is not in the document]
huffmanca: Updated to use `is None`.

ansible_base/lib/abstract_models/common.py
```diff
@@ -228,8 +230,24 @@ def related_fields(self, request):
...
+                related_model = field.related_model
+                basename = get_cls_view_basename(related_model)
+                try:
+                    related_url = get_relative_url(f'{basename}-detail', kwargs={'pk': fk_id})
+                except NoReverseMatch:
+                    # Fall back to loading the object for models with get_absolute_url
+                    if obj := getattr(self, field.name):
+                        related_url = get_url_for_object(obj)
+                    else:
+                        continue
```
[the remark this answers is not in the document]
huffmanca: Updated comment.

john-westcott-iv (commented):
## AI Code Review: 9.3/10 — NEEDS_CHANGES

| Lens | Score | Findings |
|---|---|---|
| Functionality | 8.5/10 | 1 Major, 1 Minor, 2 Nit |
| Security | 10/10 | None |
| Quality | 9.5/10 | 4 Nit |
| **Overall** | **9.3/10** | |

The optimization is well-conceived and correctly eliminates N+1 FK queries in `related_fields()`. The main issue is unnecessary fallback complexity — see inline comments.

**Needs Human Judgment:** The codebase search confirmed no django-ansible-base model has a `get_absolute_url()` that diverges from the router URL (and the docs mark it deprecated). However, downstream projects (e.g., AWX) that extend CommonModel could theoretically rely on the fallback. Please confirm before simplifying.

<sub>Reviewed with `aap-sdlc-harness:code-review`</sub>

ansible_base/lib/abstract_models/common.py
```diff
@@ -228,8 +230,26 @@ def related_fields(self, request):
...
+                if related_url:
```
john-westcott-iv: **Major [Functionality]:** The `NoReverseMatch` fallback and 7-line comment are unnecessary complexity.

The only model in this repo with `get_absolute_url()` is `OAuth2Application`, and its implementation just calls `get_relative_url(f'{self.router_basename}-detail', ...)` — the same router URL this code already tries. Its own comment says: *"This is kind of annoying... better would be to not have this method."* The docs (`docs/lib/default_models.md`) mark `get_absolute_url` as deprecated.

When `NoReverseMatch` fires, the fallback:
1. Loads the FK object from DB (wasteful query)
2. Calls `get_url_for_object()` which, without `get_absolute_url`, retries the **same** `get_relative_url()` reverse — fails again
3. Returns `''`

So the fallback only helps models with `get_absolute_url()`, and the only such model delegates to the same URL anyway.

```suggestion
            fk_id = getattr(self, field.attname)
            if fk_id is not None:
                related_model = field.related_model
                basename = get_cls_view_basename(related_model)
                try:
                    response[field.name] = get_relative_url(f'{basename}-detail', kwargs={'pk': fk_id})
                except NoReverseMatch:
                    pass
```

ansible_base/lib/abstract_models/common.py
```diff
@@ -209,12 +209,14 @@ def get_summary_fields(self):
...
-            if isinstance(field, models.ForeignObject) and getattr(self, field.name):
-                # ignore relations on inherited django models
-                if field.name.endswith("_ptr"):
-                    continue
-                if hasattr(getattr(self, field.name), 'summary_fields'):
-                    response[field.name] = getattr(self, field.name).summary_fields()
+            if not isinstance(field, models.ForeignObject) or field.name.endswith("_ptr"):
+                continue
+            # Check the raw FK id to skip null FKs without loading the related object
+            if hasattr(field, 'attname') and getattr(self, field.attname, None) is None:
```
john-westcott-iv: **Nit [Functionality]:** `hasattr(field, 'attname')` is always `True` — every Django `Field` subclass has `attname` set during `contribute_to_class()`. This is a dead guard.

Also note the inconsistency: `get_summary_fields()` guards with `hasattr` but `related_fields()` accesses `field.attname` directly (line 240). Both are safe, but pick one style.

```suggestion
            if getattr(self, field.attname, None) is None:
```

ansible_base/lib/abstract_models/common.py
```diff
@@ -209,12 +209,14 @@ def get_summary_fields(self):
...
+                continue
+            obj = getattr(self, field.name)
```
john-westcott-iv: **Nit [Quality]:** `obj` is ambiguous in a scope that already has `field` (also an object) and `self`. A reader has to trace the `getattr(self, field.name)` descriptor to know this is the related model instance. Consider `related_obj`.

```suggestion
            related_obj = getattr(self, field.name)
            if related_obj and hasattr(related_obj, 'summary_fields'):
                response[field.name] = related_obj.summary_fields()
```

test_app/tests/lib/abstract_models/test_common_query_optimization.py
```diff
@@ -0,0 +1,93 @@
...
+        team = Team.objects.select_related("organization", "modified_by", "created_by").get(pk=team.pk)
+
+        connection.queries_log.clear()
+        team.related_fields(None)
+
+        # With the optimization, related_fields should not need to load FK objects.
+        # Some queries may still occur for reverse relation URL resolution,
+        # but no FK lazy-load queries (query.get calls) should happen.
+        fk_get_queries = [q for q in connection.queries_log if "FROM" in q.get("sql", "") and "LIMIT" in q.get("sql", "")]
+        assert len(fk_get_queries) == 0, f"related_fields() triggered {len(fk_get_queries)} FK lazy-load queries"
```
john-westcott-iv: **Minor [Functionality]:** Heuristic query detection is fragile — matching `FROM` + `LIMIT` in SQL strings couples the test to Django's current query shape. If Django changes how single-object gets are structured (e.g., drops `LIMIT`), this test would pass even if lazy loads occur.

Consider `assertNumQueries` or `django.test.utils.CaptureQueriesContext` for a more robust assertion. The same pattern appears on line 80.

huffmanca: Updated the test to use this approach.

ansible_base/lib/abstract_models/common.py
huffmanca: Updated to `related_obj`

ansible_base/lib/abstract_models/common.py
huffmanca: Kept this but changed to `getattr` here, per recommendation.

ansible_base/lib/abstract_models/common.py
huffmanca: AWX, Galaxy, and EDA were all checked:
- AWX has a separate `CommonModel` that does not extend from the `AbstractCommonModel` in DAB.
- EDA doesn't seem to use `get_absolute_url`.
- Galaxy has `get_absolute_url`, but not from here.

So I think we're safe here. I updated it to use the recommended approach.

ansible_base/lib/abstract_models/common.py
```diff
@@ -209,12 +209,14 @@ def get_summary_fields(self):
         for field in self._meta.fields:
             if field.name in self.ignore_relations:
                 continue  # This check is beneficial before getattr to prevent some error cases
-            if isinstance(field, models.ForeignObject) and getattr(self, field.name):
-                # ignore relations on inherited django models
-                if field.name.endswith("_ptr"):
-                    continue
-                if hasattr(getattr(self, field.name), 'summary_fields'):
-                    response[field.name] = getattr(self, field.name).summary_fields()
+            if not isinstance(field, models.ForeignObject) or field.name.endswith("_ptr"):
```
tznamena, line 213: A comment would be helpful to explain why we are skipping this case.

huffmanca: Comment has been added here!

john-westcott-iv (approved):
## 📋 Code Review Summary (Re-review)

This PR optimizes `related_fields()` and `get_summary_fields()` in `AbstractCommonModel` to eliminate N+1 FK queries. The implementation is clean and all prior review feedback has been addressed.

**Files Reviewed**: 2 files
**Comments Posted**: 0 inline comments (all prior issues resolved)

### ✅ Prior Review Issues — All Addressed

1. **`is None` guard in `get_summary_fields`** — Fixed. Now uses explicit `getattr(self, field.attname, None) is None` instead of truthiness check.
2. **Unnecessary `NoReverseMatch` fallback complexity** — Fixed. The `get_url_for_object(obj)` fallback was removed; confirmed AWX/EDA/Galaxy don't depend on it from DAB.
3. **`hasattr(field, 'attname')` dead guard** — Fixed. Removed, now uses `getattr` directly.
4. **Variable name `obj` → `related_obj`** — Fixed.
5. **Fragile query detection (SQL string matching)** — Fixed. Tests now use `CaptureQueriesContext`.
6. **Tests vacuously true without DEBUG=True** — Fixed. `CaptureQueriesContext` works regardless of DEBUG.
7. **`select_related` masking regression test** — Fixed. Test now uses plain `Team.objects.get(pk=team.pk)`.
8. **Comment explaining null FK skip** — Fixed. Added `# Check the raw FK id to skip null FKs without loading the related object`.

### 🔍 New Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 0 code quality suggestions

### Overall Assessment
**APPROVED** — The optimization is well-implemented, properly tested (100% coverage on new code per SonarCloud), and all CI checks pass. The approach of using raw FK IDs (`field.attname`) to construct URLs via `get_relative_url()` is the correct pattern to avoid lazy-loading related objects.

Reviewed with `aap-sdlc-harness:code-review`

test_app/tests/lib/abstract_models/test_common_query_optimization.py
```diff
@@ -0,0 +1,90 @@
...
+        org = Organization.objects.create(name="test-org")
+
+        related = org.related_fields(None)
+
+        # Null FKs should not be in the response
+        for field in org._meta.concrete_fields:
+            if isinstance(field, models.ForeignKey) and field.name not in org.ignore_relations:
+                fk_id = getattr(org, field.attname)
+                if fk_id is None:
+                    assert field.name not in related
```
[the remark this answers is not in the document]
huffmanca: Accepted proposed fix in latest commit.

test_app/tests/lib/abstract_models/test_common_query_optimization.py
```diff
@@ -0,0 +1,89 @@
...
+        # Pre-load all FK relations so get_summary_fields() doesn't need to query
+        user = User.objects.select_related("modified_by", "created_by", "resource").get(pk=user.pk)
+
+        with CaptureQueriesContext(connection) as ctx:
+            summary = user.get_summary_fields()
+
+        assert isinstance(summary, dict)
+        assert len(ctx.captured_queries) == 0, (
+            f"get_summary_fields() triggered {len(ctx.captured_queries)} queries: " f"{[q['sql'] for q in ctx.captured_queries]}"
+        )
```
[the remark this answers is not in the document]
huffmanca: Test gap acknowledged, even if the suggested fix wasn't directly used.

bhavenst (approved):

Issue #226, opened 2024-03-18
[RBAC] Make examples in rbac.md for client usage fully complete


AlanCoding:
followup from

https://github.com/ansible/django-ansible-base/pull/45#discussion_r1525428016

https://github.com/ansible/django-ansible-base/pull/45#discussion_r1525432743

AlanCoding:
also maybe https://github.com/ansible/django-ansible-base/pull/45#discussion_r1525444792

AlanCoding:
pasting from links

POST `/api/v1/role_definition/` with payload

```json
{
    "permissions": ["view_inventory"],
    "content_type": "shared.organization",
    "name": "View all org inventories",
    "description": "Custom Org Role"
}
```

Then do a second POST to `/api/v1/role_team_assignments/` (or role_user_assignments) with a payload of:

```json
{
    "role_definition": <id of the role created>,
    "object_id": <id of the org>,
    "team": 2
}
```

> Maybe add a link in here reminding that you defined rd as pointing to job templates to guide users as to why this allows all JTs in an organization.

Reminder that `rd` is a org-type role (not job template) and lists JT permissions.

Issue #220, opened 2024-03-15
[RBAC] Remove `ANSIBLE_BASE_BYPASS_ACTION_FLAGS` setting/feature


AlanCoding:
This was used for handling `is_system_auditor` as a user flag, which we are not supporting anymore. And its implementation complicates a lot.

AlanCoding:
Confirming:

no hits here

https://github.com/search?q=repo%3Aansible%2Fawx%20ANSIBLE_BASE_BYPASS_ACTION_FLAGS&type=code

https://github.com/search?q=repo%3Aansible%2Fgalaxy_ng%20ANSIBLE_BASE_BYPASS_ACTION_FLAGS&type=code

PR #993, opened 2026-05-06
Fix O(n) correlated subquery in get_object_by_ansible_id()

djulich:
## Summary

- Rewrites `get_object_by_ansible_id()` to use two constant-time indexed lookups instead of a correlated subquery that seq-scans all `auth_user` rows
- Removes unused `annotate_as` parameter (accepted via `**kwargs` for backward compatibility)
- Removes test assertion for the dropped annotation behavior

## Problem

Every JWT-authenticated API request calls `get_user_by_ansible_id()` → `get_object_by_ansible_id()`, which annotated the entire User queryset with a correlated subquery to `dab_resource_registry_resource`, then filtered for a match. PostgreSQL had to seq-scan all users and run the subquery for each one — O(n) per request.

**Scale Lab impact (81K users):**
- Idle: 43ms per request (65,693 buffer accesses)
- 100 concurrent users: 3,093ms mean
- 1,000 concurrent users: 21,731ms mean

## Fix

Instead of inside-out (annotate all users, filter), do outside-in:
1. Index scan on `dab_resource_registry_resource` by `ansible_id` (unique index) → get `object_id`
2. PK lookup on `auth_user` by `id` → get user

## Verification

Tested on a kind cluster running AAP 2.7:

| Users | Old query | New query | Speedup |
|------:|----------:|----------:|--------:|
| 1,988 | 1.6ms / 3,890 buffers | 0.1ms / 7 buffers | **16x** |
| 80,001 | 79.7ms / 241,343 buffers | 0.1ms / 7 buffers | **857x** |

Django ORM query capture confirms the fix generates exactly two indexed queries with no correlated subquery. End-to-end API requests through the gateway return in ~150ms with negligible auth overhead.

Fixes: AAP-72968

## Test plan

- [x] Existing tests pass (`test_app/tests/lib/utils/test_auth.py`)
- [x] Verified with EXPLAIN ANALYZE at 2K and 80K users
- [x] Verified Django ORM generates expected SQL (no correlated subquery)
- [x] End-to-end API request through gateway returns HTTP 200 with correct `ansible_id`

---

> [!NOTE]
> **Medium Risk**
> Changes the core `ansible_id`→model resolution path used during authenticated requests; while behavior is similar, it alters query patterns and error handling in a performance-critical area.
> 
> **Overview**
> Reworks `get_object_by_ansible_id()` to avoid a correlated subquery/annotation and instead do a direct lookup in `dab_resource_registry.Resource` by (`ansible_id`, content type) followed by a PK fetch on the target queryset.
> 
> Drops the `annotate_as` parameter/annotation behavior from `get_object_by_ansible_id()` and `get_user_by_ansible_id()`, and updates tests to assert only correct object resolution (including the deleted-resource case) rather than presence of an annotated field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 55c0ba84dca3e75e8076de4270402eee44a31f21. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Refactor**
  * Simplified object resolution to use the centralized resource registry, improving identifier-based lookups and producing clearer errors when an identifier is not found.
* **Tests**
  * Updated tests to validate lookup behavior for deleted resources and ensure returned objects retain their original identifier metadata.

### DIFF
--- a/ansible_base/lib/utils/auth.py
+++ b/ansible_base/lib/utils/auth.py
@@ -36,7 +36,7 @@ def get_organization_model() -> Type[AbstractOrganization]:
     return get_model_from_settings('ANSIBLE_BASE_ORGANIZATION_MODEL')
 
 
-def get_object_by_ansible_id(qs: QuerySet, ansible_id: Union[str, UUID], **kwargs) -> Model:
+def get_object_by_ansible_id(qs: QuerySet, ansible_id: Union[str, UUID]) -> Model:
     resource_cls = django_apps.get_model('dab_resource_registry', 'Resource')
     content_type_cls = django_apps.get_model('contenttypes', 'ContentType')
     cls = qs.model
@@ -44,9 +44,9 @@ def get_object_by_ansible_id(qs: QuerySet, ansible_id: Union[str, UUID], **kwarg
     try:
         object_id = resource_cls.objects.values_list('object_id', flat=True).get(ansible_id=ansible_id, content_type=ct)
     except resource_cls.DoesNotExist:
-        raise cls.DoesNotExist(f"{cls.__name__} with ansible_id {ansible_id} does not exist.")
+        raise cls.DoesNotExist(f"{cls.__name__} with ansible_id {ansible_id} does not exist.") from None
     return qs.get(pk=object_id)
 
 
-def get_user_by_ansible_id(ansible_id: Union[str, UUID], **kwargs) -> Model:
+def get_user_by_ansible_id(ansible_id: Union[str, UUID]) -> Model:
     return get_object_by_ansible_id(get_user_model().objects.all(), ansible_id)
--- a/ansible_base/lib/utils/auth.py
+++ b/ansible_base/lib/utils/auth.py
@@ -5,8 +5,7 @@ from django.apps import apps as django_apps
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ImproperlyConfigured
-from django.db.models import CharField, Model, OuterRef
-from django.db.models.functions import Cast
+from django.db.models import Model
 from django.db.models.query import QuerySet
 
 from ansible_base.lib.abstract_models.organization import AbstractOrganization
@@ -37,16 +36,17 @@ def get_organization_model() -> Type[AbstractOrganization]:
     return get_model_from_settings('ANSIBLE_BASE_ORGANIZATION_MODEL')
 
 
-def get_object_by_ansible_id(qs: QuerySet, ansible_id: Union[str, UUID], annotate_as: str = 'ansible_id_for_filter') -> Model:
+def get_object_by_ansible_id(qs: QuerySet, ansible_id: Union[str, UUID], **kwargs) -> Model:
     resource_cls = django_apps.get_model('dab_resource_registry', 'Resource')
     content_type_cls = django_apps.get_model('contenttypes', 'ContentType')
     cls = qs.model
     ct = content_type_cls.objects.get_for_model(cls)
-    pk_field_name = cls._meta.pk.name
-    pk_reference = Cast(OuterRef(pk_field_name), output_field=CharField())
-    resource_qs = resource_cls.objects.filter(object_id=pk_reference, content_type=ct).values('ansible_id')
-    return qs.annotate(**{annotate_as: resource_qs}).get(**{annotate_as: ansible_id})
+    try:
+        object_id = resource_cls.objects.values_list('object_id', flat=True).get(ansible_id=ansible_id, content_type=ct)
+    except resource_cls.DoesNotExist:
+        raise cls.DoesNotExist(f"{cls.__name__} with ansible_id {ansible_id} does not exist.")
+    return qs.get(pk=object_id)
 
 
-def get_user_by_ansible_id(ansible_id: Union[str, UUID], annotate_as: str = 'ansible_id_for_filter') -> Model:
-    return get_object_by_ansible_id(get_user_model().objects.all(), ansible_id, annotate_as=annotate_as)
+def get_user_by_ansible_id(ansible_id: Union[str, UUID], **kwargs) -> Model:
+    return get_object_by_ansible_id(get_user_model().objects.all(), ansible_id)
--- a/test_app/tests/lib/utils/test_auth.py
+++ b/test_app/tests/lib/utils/test_auth.py
@@ -50,7 +50,9 @@ def test_get_user_by_ansible_id_deleted_resource():
 
     # Even though user 4 can no longer be referenced by ansible_id, the others still work
     for i in range(3):
-        assert get_user_by_ansible_id(users[i].resource.ansible_id) == users[i]
+        found_user = get_user_by_ansible_id(users[i].resource.ansible_id)
+        assert found_user == users[i]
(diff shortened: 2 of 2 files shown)


test_app/tests/lib/utils/test_auth.py
```diff
@@ -51,8 +51,6 @@ def test_get_user_by_ansible_id_deleted_resource():
     # Even though user 4 can no longer be referenced by ansible_id, the others still work
     for i in range(3):
         assert get_user_by_ansible_id(users[i].resource.ansible_id) == users[i]
-        found_user = get_user_by_ansible_id(users[i].resource.ansible_id, annotate_as='a_id')
-        assert found_user.a_id == users[i].resource.ansible_id
```
AlanCoding, line 55: I think you should still assert that the user has the intended ansible_id.

ansible_base/lib/utils/auth.py
```diff
@@ -37,16 +36,17 @@ def get_organization_model() -> Type[AbstractOrganization]:
...
+    try:
+        object_id = resource_cls.objects.values_list('object_id', flat=True).get(ansible_id=ansible_id, content_type=ct)
+    except resource_cls.DoesNotExist:
+        raise cls.DoesNotExist(f"{cls.__name__} with ansible_id {ansible_id} does not exist.")
+    return qs.get(pk=object_id)
 
 
-def get_user_by_ansible_id(ansible_id: Union[str, UUID], annotate_as: str = 'ansible_id_for_filter') -> Model:
-    return get_object_by_ansible_id(get_user_model().objects.all(), ansible_id, annotate_as=annotate_as)
+def get_user_by_ansible_id(ansible_id: Union[str, UUID], **kwargs) -> Model:
```
AlanCoding: These `**kwargs` are a really bad anti-pattern.

If you're going to get rid of the annotation, then ok, there are important tradeoffs there. But you can't do this still pretending that some stray caller can pass the old kwargs to annotate it and still have it work.

Best thing to do is remove `kwargs`, getting rid of any promises that it'll add extra data to the returned `User`. This will _just_ give a user. If that causes problems, then we need to dive deeper into what code is hitting those problems.

AlanCoding (changes requested):
Should remove kwargs

djulich:
## YOLO Integration Test Results

Ran the full AAPQA Containerized Baby Yolo suite ([build #220](https://jenkins-csb-aap-main.dno.corp.redhat.com/job/AAPQA/job/AAPQA%20Provisioner/job/Baby%20Yolo/job/AAPQA-Containerized-Baby-Yolo/220)) with this DAB branch. **3,346 passed, 31 failed** — all failures are unrelated to this change.

| Category | Failures | Details |
|----------|:--------:|---------|
| EDA event streams/activations | 17 | Webhook integrations (HMAC, ECDSA, ServiceNow, Dynatrace, etc.), credential injectors |
| Kafka event persistence | 4 | Happy path, disabled, multi-activation, bad credential |
| Gateway basic auth config | 3 | Disabling/enabling basic authenticator combinations |
| User email change prevention | 2 | Cannot change own email (controller + gateway) |
| Workflow approval notification | 1 | Slack notification |
| RBAC change request | 1 | Job template change request 403 |
| Hub e2e | 1 | Collection repo + job template |

**Why none are related:** This PR only changes the code path that resolves a user by `ansible_id` during JWT authentication (`get_object_by_ansible_id` → two indexed lookups instead of a correlated subquery). If this change caused failures, we'd see widespread auth failures across all test categories. Instead, the failures are isolated to EDA webhooks, Kafka, notification integrations, and unrelated auth config — none of which touch the JWT user lookup path.

test_app/tests/lib/utils/test_auth.py
djulich: You're right - done.

ansible_base/lib/utils/auth.py
djulich: That's really true, thanks for pointing that out. I removed the kwargs.

AlanCoding (approved):

djulich:
## YOLO Integration Test Results (post-review update)

Ran the full AAPQA Containerized Baby Yolo suite ([build #226](https://jenkins-csb-aap-main.dno.corp.redhat.com/job/AAPQA/job/AAPQA%20Provisioner/job/Baby%20Yolo/job/AAPQA-Containerized-Baby-Yolo/226/)) with the updated code (removed `**kwargs`, added `ansible_id` assertion). **43 failures** — all unrelated to this change.

| Category | Failures | Details |
|----------|:--------:|---------|
| EDA event routing/persistence | 20 | Same webhook and Kafka failures as build #220 |
| Lightspeed codegen/explanation | 13 | AI assistant API tests (role explanation, content matches, chatbot, codegen) |
| Gateway proxy | 2 | gRPC thread safety, Lightspeed route timeout |
| RBAC | 2 | Job template change request 403, job launch with role |
| Hub e2e | 1 | Collection repo + job template |

**Why none are related:** This PR only changes the JWT user lookup path (`get_object_by_ansible_id`). A regression here would cause widespread authentication failures across all test categories. Instead, failures are isolated to EDA webhooks (same as before the review update), Lightspeed AI features, and gateway proxy routing — none of which touch the auth user lookup.

ansible_base/lib/utils/auth.py
```diff
@@ -37,16 +36,17 @@ def get_organization_model() -> Type[AbstractOrganization]:
...
     resource_cls = django_apps.get_model('dab_resource_registry', 'Resource')
     content_type_cls = django_apps.get_model('contenttypes', 'ContentType')
     cls = qs.model
     ct = content_type_cls.objects.get_for_model(cls)
-    pk_field_name = cls._meta.pk.name
-    pk_reference = Cast(OuterRef(pk_field_name), output_field=CharField())
-    resource_qs = resource_cls.objects.filter(object_id=pk_reference, content_type=ct).values('ansible_id')
-    return qs.annotate(**{annotate_as: resource_qs}).get(**{annotate_as: ansible_id})
+    try:
+        object_id = resource_cls.objects.values_list('object_id', flat=True).get(ansible_id=ansible_id, content_type=ct)
```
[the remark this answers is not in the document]
djulich: Not a valid concern for this PR. Here's why:                                                                                                        
                                                                                                                              
  1. No multi-database routing exists in DAB or its consumers. There are no DATABASE_ROUTERS configured, and no caller passes a routed queryset.      
  2. The only production caller is get_user_by_ansible_id() which calls get_object_by_ansible_id(get_user_model().objects.all(), ...) — that's        
  User.objects.all(), which uses the default database.                                                                                                
  3. The old code had the same limitation — the original implementation used content_type_cls.objects.get_for_model(cls) and                          
  resource_cls.objects.filter(...) without .using(qs.db) either. So even if this were a real concern, it's not a regression introduced by our fix.    
  4. Adding .using(qs.db) and .db_manager(qs.db) for a hypothetical multi-database scenario that doesn't exist would be designing for a future        
  requirement that may never come — exactly the kind of speculative code the bot tends to suggest.                                                    
                                                                                                                                                      
This is a classic AI code reviewer false positive — technically correct in theory but not applicable to the actual codebase. I'd recommend not making the change.

PR #995, opened 2026-05-06
[AAP-45394 AAP-45047] Fix: allow maps can now recover access after an earlier deny

bhavenst:
## Summary

Authenticator maps of `map_type=allow` could not recover `access_allowed` once it had been set to `False` by an earlier deny-all rule. The `create_claims()` function in `ansible_base/authentication/utils/claims.py` guarded the `allow` branch with `and not has_permission`, which silently dropped the case where a later allow-map fires affirmatively (`has_permission=True`). As a result, once `access_allowed=False` was set by the first map, no subsequent allow map — regardless of order — could ever restore access.

## Root Cause

At line 123 of `ansible_base/authentication/utils/claims.py`, the condition:

```python
if auth_map.map_type == 'allow' and not has_permission:
    access_allowed = False
    understood_map = True
```

only executed when `has_permission=False`. When a later allow-map evaluated to `ALLOW` (`has_permission=True`), the condition was `False`, control fell through to the `if not understood_map` catch-all, and the error `"Map type allow of rule ... does not know how to be processed"` was logged. The `access_allowed` variable was never reset to `True`.

## Changes

- **`ansible_base/authentication/utils/claims.py`** — Remove the `and not has_permission` guard and replace `access_allowed = False` with `access_allowed = has_permission`. This makes `access_allowed` always track the most recently evaluated allow-map result, letting later (higher-order) rules override earlier ones as the ordering mechanism is designed to support.
- **`test_app/tests/authentication/utils/test_claims.py`** — Add one parametrized case (`allow` + `always` trigger) to the existing `test_create_claims_single_map_acl` suite, and three new standalone regression tests:
  - `test_create_claims_deny_all_then_allow_override` — deny at order=1, allow-always at order=2 → access granted
  - `test_create_claims_deny_all_not_overridden_without_match` — deny at order=1, allow-group at order=2, user NOT in group → access denied
  - `test_create_claims_deny_all_overridden_with_group_match` — deny at order=1, allow-group at order=2, user IS in group → access granted

## Risk Assessment

- **Confidence:** HIGH (95%)
- **Risk:** Low — the change only affects `map_type == 'allow'` maps. The deny path (`has_permission=False`) produces the same result as before (`access_allowed = False`). The only behavior change is that the grant path (`has_permission=True`) now sets `access_allowed = True` instead of being silently dropped.
- **Scope:** 2 files changed

## Testing

- [ ] Unit tests pass
- [ ] Regression tests added (4 new test cases covering the fixed code path)
- [ ] Lint checks pass

## JIRA

- Ticket: [AAP-45394](https://redhat.atlassian.net/browse/AAP-45394)

## See Also

See also: https://github.com/ansible/django-ansible-base/pull/959 — an open PR with an identical fix (also covering AAP-45047 and AAP-45394). Reviewers may choose to close this PR in favor of #959 or merge whichever is reviewed first.

---
*Debuggernaut autonomous fix -- review carefully before merging.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication rule evaluation updated so later allow rules can override earlier denials, ensuring rule ordering is respected.

* **Tests**
  * Added regression tests covering rule override scenarios: deny-then-allow, non-matching allow, group-based matches, allow-then-deny precedence, revoke-deny overridden, and multi-rule "last writer wins" behavior.

### DIFF
--- a/ansible_base/authentication/utils/claims.py
+++ b/ansible_base/authentication/utils/claims.py
@@ -120,9 +120,8 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
         rule_responses.append({mpk: has_permission, 'enabled': auth_map.enabled})
 
         understood_map = False
-        if auth_map.map_type == 'allow' and not has_permission:
-            # If any rule does not allow we don't want to return this to true
-            access_allowed = False
+        if auth_map.map_type == 'allow':
+            access_allowed = has_permission
             understood_map = True
         elif auth_map.map_type == 'is_superuser':
             is_superuser = has_permission
--- a/test_app/tests/authentication/utils/test_claims.py
+++ b/test_app/tests/authentication/utils/test_claims.py
@@ -2740,23 +2740,21 @@ def test_create_claims_deny_all_then_allow_override(
     access remained denied.
     """
     # order=1: deny-all rule (trigger 'never' -> has_permission=False)
-    local_authenticator_map.map_type = "allow"
-    local_authenticator_map.triggers = {"never": {}}
+    local_authenticator_map.map_type = 'allow'
+    local_authenticator_map.triggers = {'never': {}}
     local_authenticator_map.order = 1
     local_authenticator_map.save()
 
     # order=2: allow-always rule (trigger 'always' -> has_permission=True)
-    local_authenticator_map_1.map_type = "allow"
-    local_authenticator_map_1.triggers = {"always": {}}
+    local_authenticator_map_1.map_type = 'allow'
+    local_authenticator_map_1.triggers = {'always': {}}
     local_authenticator_map_1.order = 2
     local_authenticator_map_1.save()
 
     authenticator = local_authenticator_map.authenticator
-    res = claims.create_claims(authenticator, "username", {}, [])
+    res = claims.create_claims(authenticator, 'username', {}, [])
 
-    assert res["access_allowed"] is True, (
-        "An allow-always map at order=2 must override a deny-all map at order=1 (AAP-45394)"
-    )
+    assert res['access_allowed'] is True, 'An allow-always map at order=2 must override a deny-all map at order=1 (AAP-45394)'
 
 
 def test_create_claims_deny_all_not_overridden_without_match(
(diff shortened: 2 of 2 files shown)


test_app/tests/authentication/utils/test_claims.py
```diff
@@ -2710,3 +2722,192 @@ def test_cache_existing_early_return_pattern(
...
+
+# --- AAP-45394 regression tests ---
+
+
+@mock.patch("ansible_base.authentication.utils.claims.logger")
+def test_create_claims_allow_grant_no_error_logged(
+    logger,
+    local_authenticator_map,
+    shut_up_logging,
+):
```
john-westcott-iv, line 2735: Nit: Docstring references AAP-45047, but this test was added as part of the AAP-45394 fix. Consider updating to AAP-45394 for consistency with the rest of the regression tests in this file.

ansible_base/authentication/utils/claims.py
```diff
@@ -120,9 +120,8 @@ def create_claims(authenticator: Authenticator, username: str, attrs: dict, grou
         rule_responses.append({mpk: has_permission, 'enabled': auth_map.enabled})
 
         understood_map = False
-        if auth_map.map_type == 'allow' and not has_permission:
-            # If any rule does not allow we don't want to return this to true
-            access_allowed = False
+        if auth_map.map_type == 'allow':
```
john-westcott-iv, line 123: Nit: With last-writer-wins semantics, `auth_map.order` is now the sole determinant of access decisions when multiple allow maps exist. Worth confirming the queryset feeding this loop orders deterministically (no ties) and that admin-facing docs reflect the ordering significance.

john-westcott-iv (approved):
**Code Review: 10.0/10 — READY_FOR_HUMAN_REVIEW**

Clean, correct bug fix with thorough regression tests. The core change correctly implements last-writer-wins semantics for allow-type authenticator maps. Two nits noted inline. No critical, major, or minor findings.

test_app/tests/authentication/utils/test_claims.py
bhavenst: I think it's OK because both Jiras will get closed w/ this PR..

ansible_base/authentication/utils/claims.py
bhavenst: I did check that and the docs are accurate (after this PR).

PR #998, opened 2026-05-07
[AAP-55298] Fix OAuth2 scope mutation causing spurious activity stream entries

bhavenst:
## Summary

OAuth2ScopePermission.has_permission() was mutating the token's `scope` attribute in-place by appending `' read'` to write-scoped tokens during permission checking. This left the model instance in a corrupted state (`scope="write read"` instead of `"write"`), which caused spurious activity stream entries when the token was subsequently saved by any code path that does not restrict `update_fields`.

This manifested as 192 seemingly identical activity stream entries for the `aap_operator_service_account`, each recording a scope change from `"write"` to `"write read"` on the same OAuth2 access token.

## Root Cause

In `ansible_base/oauth2_provider/permissions.py`, the line `request.auth.scope += ' read'` permanently mutated the `OAuth2AccessToken` model instance's `scope` field. The "write implies read" expansion was intended only for the `TokenHasReadWriteScope` permission check, but the original scope was never restored afterward. Any subsequent `.save()` on the token (without `update_fields` limiting the persisted fields) would persist the mutated scope to the database, and the activity stream would record the change.

## Changes

- **`ansible_base/oauth2_provider/permissions.py`**: Wrapped the scope expansion in a `try/finally` block that saves the original scope before mutation and restores it unconditionally after the permission check completes. The "write implies read" behavior is preserved for the permission check itself.
- **`test_app/tests/oauth2_provider/test_authentication.py`**: Added two regression tests:
  - `test_oauth2_scope_not_mutated_after_permission_check`: Verifies that the token's scope remains `"write"` after both GET and POST requests through the permission check.
  - `test_oauth2_scope_no_activitystream_for_scope_field`: Verifies that no activity stream entries are created for scope changes during authenticated requests with write-scoped tokens.

## Risk Assessment

- **Confidence:** Medium -- The scope mutation is clearly a bug and this fix addresses it. The full reproduction requires the operator's reconciliation loop, which is outside this codebase.
- **Risk:** Low -- The fix preserves the existing permission logic ("write implies read") while preventing model mutation. The `try/finally` pattern ensures the scope is always restored, even if the permission check raises an exception.
- **Scope:** 2 file(s) changed

## Testing

- [x] Unit tests pass (35/35 in test_authentication.py)
- [x] Regression tests added (2 new tests)
- [x] Lint checks pass (ruff check + ruff format)

## JIRA

- Ticket: [AAP-55298](https://redhat.atlassian.net/browse/AAP-55298)

---
*Assisted-by: claude-opus-4-6 | Review carefully before merging.*

[AAP-55298]: https://redhat.atlassian.net/browse/AAP-55298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed OAuth2 permission checks that were permanently modifying access token scopes in memory and the database.

* **Tests**
  * Added regression tests to verify token scopes remain unchanged during permission checks and that activity stream entries are not incorrectly created for scope changes.

### DIFF
--- a/ansible_base/oauth2_provider/permissions.py
+++ b/ansible_base/oauth2_provider/permissions.py
@@ -20,11 +20,7 @@ class OAuth2ScopePermission(BasePermission):
         is_authenticated = IsAuthenticated().has_permission(request, view)
         is_oauth = False
         has_oauth_permission = False
-        if (
-            is_authenticated
-            and request.auth
-            and isinstance(request.auth, oauth2_models.AbstractAccessToken)
-        ):
+        if is_authenticated and request.auth and isinstance(request.auth, oauth2_models.AbstractAccessToken):
             is_oauth = True
             scopes = request.auth.scope.split()
             if "write" in scopes and "read" not in scopes:
@@ -35,9 +31,7 @@ class OAuth2ScopePermission(BasePermission):
                 request.auth.scope = original_scope + " read"  # write implies read
                 try:
                     token_permission = TokenHasReadWriteScope()
-                    has_oauth_permission = token_permission.has_permission(
-                        request, view
-                    )
+                    has_oauth_permission = token_permission.has_permission(request, view)
                 finally:
                     request.auth.scope = original_scope
             else:
--- a/ansible_base/oauth2_provider/permissions.py
+++ b/ansible_base/oauth2_provider/permissions.py
@@ -20,11 +20,27 @@ class OAuth2ScopePermission(BasePermission):
         is_authenticated = IsAuthenticated().has_permission(request, view)
         is_oauth = False
         has_oauth_permission = False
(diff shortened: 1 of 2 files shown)


huffmanca (approved):
I've examined this locally, and while I can't reproduce the actual issue of the external entity repeatedly running `save()` on the token, we can easily reproduce the bug by fetching the token and running `.save()` while it has just the `write` permission on it. (Note that if you run the test repeatedly you'll need to change it back to the `write` scope).

With these changes, the activitystream entry disappears.

AlanCoding (approved):
yeah, `if "write" in scopes and "read" not in scopes:` is clearly the weird/anomalous case. Seeing this fixed in the auth class reads as intuitively good form.

PR #999, opened 2026-05-08
[AAP-69617] Add covering index on Resource for object_ansible_id lookups

huffmanca:
## Description

The service API assignment list endpoints annotate each row with `_object_ansible_id_annotation` via a correlated subquery against `dab_resource_registry_resource`. This subquery executed per-row index lookups that hit the heap to retrieve `ansible_id`, creating unnecessary I/O at scale.

Replace the `unique_together` + explicit index on `Resource(content_type, object_id)` with a `UniqueConstraint` that includes `ansible_id` as a covering column. This enables index-only scans for the existing correlated subquery, eliminating heap lookups without changing the annotation or serializer code.

This is the first step in an iterative approach to resolving the query performance issues on these endpoints. Follow-up work will address the annotation code itself.

Based on AlanCoding's review suggestion in the prior approach (batch lookup).

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Apply the migration: `python manage.py migrate dab_resource_registry`
2. Verify the covering index exists: `\d dab_resource_registry_resource` in psql should show the unique constraint including `ansible_id`
3. Run `EXPLAIN ANALYZE` on the assignment list query to confirm index-only scans on the resource table
4. Hit the service user/team assignment list endpoints and verify `object_ansible_id` is populated correctly

### Expected Results
- The `unique_resource_content_type_object_id` constraint exists with `ansible_id` as a covering column
- The correlated subquery resolves via index-only scan (no heap fetches)
- `object_ansible_id` appears correctly in list API responses

### Migration Details
- Removes the old `unique_together` constraint and explicit index
- Adds a `UniqueConstraint` with `include=["ansible_id"]` (requires PostgreSQL 11+)
- This is a non-breaking schema change — the uniqueness guarantee is identical

### Required Actions
- [ ] Blocked by PR/MR: #XXX

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for role assignment list responses to ensure all required data fields are consistently included in API responses for organization-level assignments.

* **Chores**
  * Refined database constraint definitions for improved data consistency and integrity across resource management operations.


ansible_base/rbac/service_api/views.py
```diff
@@ -111,26 +113,44 @@ def perform_destroy(self, instance):
             with transaction.atomic():
                 instance.role_definition.remove_global_permission(instance.actor)
 
+    def _batch_annotate_ansible_ids(self, instances):
+        """Batch-fetch Resource ansible_ids and set _object_ansible_id_annotation on each instance."""
+        by_ct = defaultdict(set)
```
john-westcott-iv: **nit:** A brief comment here would help future developers understand the two-loop pattern — why instances are iterated once to collect IDs and again to set attributes.

```suggestion
        # Group object_ids by content_type so the query uses efficient IN clauses
        # (e.g., WHERE ct=1 AND oid IN (5,6,7)) rather than per-instance ORs.
        # A second pass below applies the resolved ansible_ids back to each instance.
        by_ct = defaultdict(set)
```

ansible_base/rbac/service_api/views.py
```diff
@@ -111,26 +113,44 @@ def perform_destroy(self, instance):
...
+            return
+
+        q_filter = Q()
+        for ct_id, obj_ids in by_ct.items():
+            q_filter |= Q(content_type_id=ct_id, object_id__in=obj_ids)
+
+        lookup = {
+            (r['content_type_id'], r['object_id']): r['ansible_id']
+            for r in Resource.objects.filter(q_filter).values('content_type_id', 'object_id', 'ansible_id')
+        }
```
john-westcott-iv: **nit:** The tuple-keyed dict comprehension is a bit dense to parse — a plain loop is more readable:

```suggestion
        lookup = {}
        for r in Resource.objects.filter(q_filter).values('content_type_id', 'object_id', 'ansible_id'):
            lookup[(r['content_type_id'], r['object_id'])] = r['ansible_id']
```

ansible_base/rbac/service_api/views.py
```diff
@@ -111,26 +113,44 @@ def perform_destroy(self, instance):
...
-            content_type_id=OuterRef(ct_field),
-            object_id=OuterRef(oid_field),
-        ).values(
-            'ansible_id'
-        )[:1]
-    )
+        for inst in instances:
+            inst._object_ansible_id_annotation = lookup.get((inst.content_type_id, inst.object_id))
+
+    def list(self, request, *args, **kwargs):
```
john-westcott-iv: **discussion:** This `list()` override duplicates DRF's `ListModelMixin.list()` pagination logic. If DRF ever changes that method, this override would silently drift.

Have you considered extracting this into a small mixin with a hook? Something like:

```python
class PostPaginationHookMixin:
    def post_pagination_hook(self, instances):
        pass

    def list(self, request, *args, **kwargs):
        queryset = self.filter_queryset(self.get_queryset())
        page = self.paginate_queryset(queryset)
        instances = page if page is not None else list(queryset)
        self.post_pagination_hook(instances)
        serializer = self.get_serializer(instances, many=True)
        if page is not None:
            return self.get_paginated_response(serializer.data)
        return Response(serializer.data)
```

Then `_batch_annotate_ansible_ids` just becomes the hook implementation, and any other view that needs post-pagination annotation can reuse the pattern. Several DAB views already combine `AnsibleBaseDjangoAppApiView` with `ListModelMixin`, so this could have broader utility.

Not a blocker — the current approach works fine for this use case. Just thinking about maintainability if more views need similar batch optimizations.

ansible_base/rbac/service_api/views.py
```diff
@@ -111,26 +113,44 @@ def perform_destroy(self, instance):
...
                 instance.role_definition.remove_global_permission(instance.actor)
 
+    def _batch_annotate_ansible_ids(self, instances):
+        """Batch-fetch Resource ansible_ids and set _object_ansible_id_annotation on each instance."""
+        by_ct = defaultdict(set)
+        for inst in instances:
+            if inst.content_type_id and inst.object_id:
+                by_ct[inst.content_type_id].add(inst.object_id)
+        if not by_ct:
+            return
```
john-westcott-iv: **nit:** When all instances are global role assignments (no `content_type_id`/`object_id`), this returns early without setting `_object_ansible_id_annotation`. The serializer handles this gracefully via its `hasattr()` fallback, so it's not a bug — but explicitly setting `None` would make the intent clearer and avoid relying on the serializer's defensive check:

```suggestion
        if not by_ct:
            for inst in instances:
                inst._object_ansible_id_annotation = None
            return
```

john-westcott-iv (commented):
## Code Review: [AAP-69617] Replace correlated subquery with batch lookup for object_ansible_id

**Score**: 9.8/10 (Functionality 10, Security 10, Quality 9.5)

Clean PR — the optimization is sound. The batch `Q()` lookup correctly replaces the correlated subquery, the serializer's `hasattr()` fallback handles all edge cases, and the filter backends (`AnsibleIdAliasFilterBackend`) do their own Resource lookups independent of the annotation. No security concerns.

**Inline comments:** 2 readability suggestions, 1 defensive-code nit, 1 discussion on extracting the `list()` override into a reusable mixin.

**Needs human verification:** `resource_ansible_id_expr` was a module-level function that was deleted — verify no downstream consumers import it.

ansible_base/rbac/service_api/views.py
```diff
@@ -111,26 +113,44 @@ def perform_destroy(self, instance):
...
+
+    def list(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+        page = self.paginate_queryset(queryset)
+        instances = page if page is not None else list(queryset)
+        self._batch_annotate_ansible_ids(instances)
+        serializer = self.get_serializer(instances, many=True)
+        if page is not None:
+            return self.get_paginated_response(serializer.data)
+        return Response(serializer.data)
```
AlanCoding: I've got a bad feeling about both the override of `list` and the type of batch process that `_batch_annotate_ansible_ids` does.

I am not even a little convinced that this problem will be limited to these views. I believe that I did the same thing in other cases and it's just not yet obvious that the thing was done.

This is kind of like a "custom prefetch_related", but not on the model level, done after the _materialization_ of the queryset. This is why the awkward `list` override, and I only recognize because I've hacked this way before, and it does not age well.

It also just defies reason that the annotation _would_ hit the aforementioned DB performance problem but this custom stuff _would not_. The only way that happens is if it annotates objects _not fetched_ or something like that, which needs scrutiny on its own.

What I'd see as being better as a solution:
 - modify the annotate code to not do whatever it is doing now that is screwing up performance
 - add an index on the relevant models
 - develop a reusable GFK-to-GFK relation, so that `Assignment` models can point to `Resource` models without blowing everything up

AlanCoding:
I want to test the waters of this idea.

```diff
diff --git a/ansible_base/resource_registry/models/resource.py b/ansible_base/resource_registry/models/resource.py
index 3d769bf..e5d532d 100644
--- a/ansible_base/resource_registry/models/resource.py
+++ b/ansible_base/resource_registry/models/resource.py
@@ -105,9 +105,12 @@ class Resource(models.Model):
         return resource_type_cache(self.content_type_id)
 
     class Meta:
-        unique_together = ('content_type', 'object_id')
-        indexes = [
-            models.Index(fields=["content_type", "object_id"]),
+        constraints = [
+            models.UniqueConstraint(
+                fields=["content_type", "object_id"],
+                include=["ansible_id"],
+                name="unique_resource_content_type_object_id",
+            ),
         ]
 
     def update_from_content_object(self):
```

This requires a migration, which would be in the resource_registry app. The intent would be to backport it fairly widely, which includes the migration, which changes the indexing.

The act of fetching the `ansible_id` from a `Resource` entry is expected to be fairly common, coming from the object identifier fields, which is this.

huffmanca:
CodeRabbit appears to be correct about SQLite, and it seems the CI uses it for some of its tests. Meaning we're not going to be able to merge that in cleanly.

We could add the UniqueConstraint without the `include`, but that defeats the point. Alternatively, we could add in an explicit SQL statement for postgres, but that feels ugly.

I'm going to close this one out for now, and I think the best way to move forward with it is to work this in:
> develop a reusable GFK-to-GFK relation, so that Assignment models can point to Resource models without blowing everything up

PR #1000, opened 2026-05-08
[AAP-74442] Fix CI failures and code issues from CVE-2026-6266 backport

john-westcott-iv:
## Summary

Fixes CI failures and code correctness issues introduced by the CVE-2026-6266 email change enforcement backport (PR #994).

**Changes in `triggers.py`:**
- Add `_UNSET` sentinel value so deferred `.only()` loads trigger a DB lookup instead of silently skipping enforcement
- Add `rbac_post_save_refresh_email_stash` handler to prevent stale stash false-positives on subsequent saves
- Add early return when `email` is not in `instance.__dict__` to avoid blocking non-email save operations (fixes Hub CI test failures)

**Changes in `permission_registry.py`:**
- Connect the new `post_save` signal handler with dispatch UID `permission-registry-refresh-email`

**Changes in `test_policies.py`:**
- Fix `test_org_admin_cannot_change_email_when_manage_org_auth_disabled` by adding `org_member_rd` fixture and granting org membership to the member user

**Changes in `test_triggers.py`:**
- Add tests for post-save stash refresh, deferred load enforcement, deferred load without email assignment, and non-email save pass-through
- Verify `permission-registry-refresh-email` signal is registered

## Test plan
- [x] All 16 email policy signal tests pass locally
- [x] All 19 policy tests pass locally
- [ ] Hub CI tests pass (test_me_delete, test_me_update, test_superuser_can_not_be_deleted)
- [ ] SonarCloud quality gate passes
- [ ] Full CI green

Fixes: https://redhat.atlassian.net/browse/AAP-74442

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved email policy enforcement to correctly handle deferred database fields and multiple successive saves
  * Added post-save state refresh to prevent stale comparisons

* **Tests**
  * Expanded test coverage for post-save refresh behavior and deferred field scenarios

### DIFF
--- a/ansible_base/rbac/permission_registry.py
+++ b/ansible_base/rbac/permission_registry.py
@@ -6,7 +6,7 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db.models import Model
 from django.db.models.base import ModelBase  # post_migrate may call with phony objects
-from django.db.models.signals import post_delete, post_init, post_migrate, pre_save
+from django.db.models.signals import post_delete, post_init, post_migrate, post_save, pre_save
 from django.utils.functional import cached_property
 
 from ansible_base.rbac.managed import ManagedRoleConstructor, get_managed_role_constructors
@@ -151,6 +151,7 @@ class PermissionRegistry:
         if not getattr(self.user_model, 'EMAIL_ENFORCEMENT_VIA_SERIALIZER', False):
             post_init.connect(triggers.rbac_post_init_stash_email, sender=self.user_model, dispatch_uid='permission-registry-stash-email')
             pre_save.connect(triggers.rbac_pre_save_enforce_email_policy, sender=self.user_model, dispatch_uid='permission-registry-enforce-email')
+            post_save.connect(triggers.rbac_post_save_refresh_email_stash, sender=self.user_model, dispatch_uid='permission-registry-refresh-email')
 
         for cls in self._registry:
             triggers.connect_rbac_signals(cls)
--- a/ansible_base/rbac/triggers.py
+++ b/ansible_base/rbac/triggers.py
@@ -14,6 +14,13 @@ from ansible_base.rbac.validators import validate_team_assignment_enabled
 
 logger = logging.getLogger('ansible_base.rbac.triggers')
 
+# Sentinel used to distinguish "we never stashed the email" from "the email was
+# stashed as None".  When a User instance is loaded via .only() without the
+# email field, post_init cannot record the original value.  Using a sentinel
+# instead of None lets pre_save detect this and fall back to a DB lookup rather
+# than silently skipping enforcement.
+_UNSET = object()
+
 
 """
 As the caching module will fill in cached data,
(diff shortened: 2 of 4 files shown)


bhavenst (changes requested):
# PR Review Summary

**Files reviewed**: 4 changed (+107/-8) — `triggers.py`, `permission_registry.py`, `test_policies.py`, `test_triggers.py`

---

## Critical Issues (1 found)

### `rbac_post_save_refresh_email_stash` missing `update_fields` guard — security bypass
**`ansible_base/rbac/triggers.py:357-364`**

The `post_save` handler refreshes `_rbac_original_email` unconditionally when `email` is in `instance.__dict__`, but does **not** check `update_fields`. This enables a two-step bypass:

1. Load user (stash = `alice@example.com`)
2. Set `instance.email = 'hacked@evil.com'` in memory
3. Call `instance.save(update_fields=['first_name'])` — pre_save skips (email not being persisted), but post_save **refreshes stash** to `'hacked@evil.com'`
4. Call `instance.save()` — pre_save sees `original == instance.email`, skips enforcement
5. Email change persists without authorization

**Suggested fix:**
```python
def rbac_post_save_refresh_email_stash(instance, update_fields=None, **kwargs):
    if update_fields is not None and 'email' not in update_fields:
        return
    if 'email' in instance.__dict__:
        instance._rbac_original_email = instance.email
```

A regression test for this attack vector (partial save followed by full save) is needed.

---

## Important Issues (4 found)

### 1. `DoesNotExist` catch returns silently with no logging
**`triggers.py:337-339`**

The DB lookup fallback catches `DoesNotExist` and silently returns. For a security enforcement path, a `logger.debug` call would make enforcement skips auditable.

### 2. Zero logging in the entire email enforcement path
**`triggers.py:292-364`**

None of the three email enforcement functions emit any log messages, despite a `logger` being available at module level. For a CVE fix, logging when enforcement blocks a change, when it's skipped, and when the DB fallback is used would significantly improve operability and debugging.

### 3. No test for deferred-load + superuser-allowed path
**`test_triggers.py`**

`test_deferred_load_enforces_via_db_lookup` only tests the blocked (non-privileged) path. A complementary test confirming a superuser can successfully change email on a deferred-loaded instance via the new DB lookup would increase confidence in the happy path.

### 4. Hub CI still failing
The PR description lists Hub CI tests (`test_me_delete`, `test_me_update`, `test_superuser_can_not_be_deleted`) as to-be-fixed, but CI shows them still returning HTTP 400. The `'email' not in instance.__dict__` guard does not appear to resolve these failures.

---

## Suggestions

- Add a test for `None`/empty email values — the sentinel was introduced specifically to distinguish from `None`
- Add a test for the `DoesNotExist` race condition path (user deleted between deferred load and save)
- Add a test confirming `EMAIL_ENFORCEMENT_VIA_SERIALIZER=True` disables the new `post_save` signal
- Minor comment refinement: `rbac_post_init_stash_email` docstring says "same pattern as" parent stash but the email version uses a sentinel while parent doesn't — "similar pattern" would be more precise

---

## Strengths

- Well-structured sentinel pattern with correct identity comparison (`is _UNSET`)
- Deferred field handling is architecturally sound (two-part `__dict__` check + sentinel fallback)
- Test docstrings are specification-quality — describe scenarios, not implementation
- Tests use real Django `.only()` behavior rather than mocking
- Post-save stash refresh correctly addresses false-positive `ValidationError` on re-save
- Inline guard comments explain "why" not "what"

---

**Recommended action**: Fix the `update_fields` guard on `rbac_post_save_refresh_email_stash` (critical security bypass), add the corresponding regression test, and add logging to the enforcement path. The rest are improvements worth considering.

🤖 Review performed with [Claude Code](https://claude.com/claude-code)

john-westcott-iv:
Closing this ticket, will try again.

PR #1006, opened 2026-05-18
[AAP-73651] Fix cross-service RBAC sync for non-existent remote objects

bhavenst:
## Summary

- **Service API serializer**: When a synced assignment references an `object_id` that does not exist locally, fall back to `RemoteObject` instead of raising `ValidationError`. This unblocks cross-service RBAC sync when object sync ordering differs from assignment sync ordering.
- **JWT auth `map_user_fields()`**: Wrap `user.save()` with `crum.impersonate(None)` so the RBAC email policy pre_save signal treats JWT field sync as a system operation, preventing 400 errors on downstream services.

## Root Cause

The ATF test `test_org_admin_cross_org_team_role_visibility_on_remote_objects` creates a `role_team_assignment` with a synthetic `object_id="999999"`. Gateway creates the local assignment (as a `RemoteObject`) but then syncs to Controller, where the service API serializer rejects it:

```
{"object_id":"JobTemplate matching query does not exist."}
```

The service API is an internal endpoint for cross-service sync — it should accept assignments even when the referenced object doesn't exist locally, since the object may live on a different service or may not have been synced yet.

## Test plan

- [x] Existing test `test_object_role_without_valid_object_error` updated to verify RemoteObject fallback (201 instead of 400)
- [x] New test `test_map_user_fields_email_change_bypasses_rbac_policy` verifies JWT email sync bypasses RBAC policy
- [x] All 38 service API tests pass
- [x] All 84 JWT auth tests pass (1 pre-existing SQLite-only failure excluded)
- [x] All 12 email policy signal tests pass

Fixes: https://redhat.atlassian.net/browse/AAP-73651

## Summary by CodeRabbit

* **Bug Fixes**
  * User email and field updates now persist correctly even when pre-save enforcement policies would previously block them
  * Role assignments now proceed and synchronize remotely when the referenced local object is missing

* **Tests**
  * Added test covering user field updates bypassing enforcement for JWT-driven changes
  * Updated remote role-assignment tests to reflect acceptance and creation of remote assignments for nonexistent local objects

* **Chores**
  * Added informational logging when falling back to a remote lookup for missing local objects

### DIFF
--- a/ansible_base/rbac/service_api/serializers.py
+++ b/ansible_base/rbac/service_api/serializers.py
@@ -1,6 +1,7 @@
 import logging
 
 from crum import impersonate
+from django.core.exceptions import ValidationError as DjangoValidationError
 from django.db import transaction
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
@@ -137,7 +138,7 @@ class BaseAssignmentSerializer(serializers.ModelSerializer):
                 else:
                     try:
                         obj = model.objects.get(pk=object_id)
-                    except (model.DoesNotExist, ValueError, TypeError):
+                    except (model.DoesNotExist, ValueError, TypeError, DjangoValidationError):
                         logger.info("Object pk=%s not found locally for %s, using RemoteObject fallback", object_id, model.__name__)
                         obj = RemoteObject(content_type=rd.content_type, object_id=object_id)
 
--- a/ansible_base/rbac/service_api/serializers.py
+++ b/ansible_base/rbac/service_api/serializers.py
@@ -133,7 +133,7 @@ class BaseAssignmentSerializer(serializers.ModelSerializer):
                 else:
                     try:
                         obj = model.objects.get(pk=object_id)
-                    except model.DoesNotExist:
+                    except (model.DoesNotExist, ValueError, TypeError):
                         obj = RemoteObject(content_type=rd.content_type, object_id=object_id)
 
             # Validators not ran, because this should be an internal action
--- a/ansible_base/rbac/service_api/serializers.py
+++ b/ansible_base/rbac/service_api/serializers.py
@@ -1,3 +1,5 @@
+import logging
+
 from crum import impersonate
 from django.db import transaction
 from django.utils.translation import gettext_lazy as _
@@ -7,6 +9,8 @@ from ..api.fields import ActorAnsibleIdField
 from ..models import DABContentType, DABPermission, RoleDefinition, RoleTeamAssignment, RoleUserAssignment
 from ..policies import check_content_obj_permission
 from ..remote import RemoteObject
+
+logger = logging.getLogger('ansible_base.rbac.service_api.serializers')
 
 
 class ObjectAnsibleIdField(serializers.Field):
@@ -134,6 +138,7 @@ class BaseAssignmentSerializer(serializers.ModelSerializer):
                     try:
                         obj = model.objects.get(pk=object_id)
                     except (model.DoesNotExist, ValueError, TypeError):
+                        logger.info("Object pk=%s not found locally for %s, using RemoteObject fallback", object_id, model.__name__)
                         obj = RemoteObject(content_type=rd.content_type, object_id=object_id)
 
             # Validators not ran, because this should be an internal action
--- a/ansible_base/rbac/service_api/serializers.py
+++ b/ansible_base/rbac/service_api/serializers.py
@@ -133,8 +133,8 @@ class BaseAssignmentSerializer(serializers.ModelSerializer):
                 else:
                     try:
                         obj = model.objects.get(pk=object_id)
-                    except model.DoesNotExist as exc:
-                        raise serializers.ValidationError({'object_id': str(exc)})
+                    except model.DoesNotExist:
+                        obj = RemoteObject(content_type=rd.content_type, object_id=object_id)
 
             # Validators not ran, because this should be an internal action
 
--- a/test_app/tests/jwt_consumer/common/test_auth.py
+++ b/test_app/tests/jwt_consumer/common/test_auth.py
@@ -142,6 +142,43 @@ class TestJWTCommonAuth:
                 assert user.save.called
 
     @pytest.mark.django_db
+    def test_map_user_fields_email_change_bypasses_rbac_policy(self, django_user_model):
+        """JWT auth field sync must update email even when the RBAC email
+        policy signal would block it for a regular requesting user.
+
+        The pre_save signal rbac_pre_save_enforce_email_policy blocks email
+        changes from non-privileged CRUM users.  map_user_fields() runs during
+        JWT authentication (a trusted system operation), so it must bypass
+        the signal by clearing the CRUM user via impersonate(None).
+        """
+        from crum import impersonate
+
+        regular_user = django_user_model.objects.create_user(
+            username='regular', password='password',
+        )
+        target_user = django_user_model.objects.create_user(
+            username='jwt-synced', password='password', email='old@example.com',
+        )
+
+        common_auth = JWTCommonAuth()
+        common_auth.user = target_user
+        common_auth.token = {
+            'user_data': {
+                'username': 'jwt-synced',
+                'first_name': '',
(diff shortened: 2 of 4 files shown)


bhavenst:
## RBAC / ORG_ADMINS_CAN_SEE_ALL_USERS Impact Analysis

Confirming that the service API serializer change (RemoteObject fallback) has **no impact** on RBAC evaluation or `ORG_ADMINS_CAN_SEE_ALL_USERS` visibility behavior.

### What changed

In `service_api/serializers.py`, the `create()` method's `model.DoesNotExist` handler now creates `RemoteObject(content_type, object_id)` instead of raising `ValidationError`. This is the **only** code path affected — every downstream consumer already handles `RemoteObject` identically to a local object.

### Trace through the RBAC stack

| Step | Code path | Behavior with RemoteObject |
|---|---|---|
| **Permission check** | `check_content_obj_permission(user, obj)` → `isinstance(obj, RemoteObject)` branch in `policies.py:91` | Superuser passes via `has_super_permission()`. Non-superuser checked against `ANSIBLE_BASE_ENFORCE_REMOTE_OBJECT_PERMISSIONS`. Same as any other remote assignment. |
| **ObjectRole creation** | `give_or_remove_permission()` in `role.py:286` | Extracts `content_type` and `object_id` from the `RemoteObject`. Resulting `ObjectRole` row is `{role_definition, content_type, object_id}` — structurally identical to what a local-object path produces. |
| **RoleEvaluation cache** | `compute_object_role_permissions()` → `expected_direct_permissions()` in `role.py:626` | Entries are `(codename, content_type_id, object_id)` tuples derived from the `ObjectRole`, not from the source object. Identical output. |
| **`visible_items` filtering** | `_visible_items()` in `role.py:402` | Filters by whether the **requesting user** has `RoleEvaluation` entries for the assignment's `(content_type_id, object_id)`. An org admin of org1 has no evaluations for `awx.jobtemplate:999999`, so the assignment is excluded. Unchanged. |
| **Gateway `filter_queryset`** | `GatewayRoleTeamAssignmentViewSet.filter_queryset()` in `aap_gateway_api/views/api/v1/role.py:174` | When `ORG_ADMINS_CAN_SEE_ALL_USERS=True` → `can_view_all_users()` returns True → `visible_items` **skipped** → assignment visible. When False → `visible_items` applied → assignment hidden. This logic is entirely in the Gateway viewset and does not touch the serializer. |

### ATF test assertion verification

| Setting | Expected | Mechanism | Result |
|---|---|---|---|
| `ORG_ADMINS_CAN_SEE_ALL_USERS=True` | Assignment **visible** | `visible_items` skipped by `filter_queryset` override | ✅ |
| `ORG_ADMINS_CAN_SEE_ALL_USERS=False` | Assignment **not visible** | `visible_items` applied; org admin has no `RoleEvaluation` for the remote object | ✅ |

### Design rationale: accepting synthetic object_id

The service API is an **internal** endpoint called by Gateway during cross-service RBAC sync. The public API (in `BaseAssignmentSerializer.get_object_from_data()`) already accepts synthetic object_ids when `issubclass(model, RemoteObject)` (line 142-143 of `rbac/api/serializers.py`). The service API fallback extends this to the case where the model is local but the specific row doesn't exist — the object may not have been synced yet, or may only exist on the originating service. This is option 1 from the Jira ticket's suggested triage: "gateway accepts a synthetic remote object_id for this scenario."

test_app/tests/rbac/remote/test_service_api.py
```diff
@@ -665,20 +665,19 @@ def test_system_role_with_object_ansible_id_error(self, admin_api_client, rando,
...
-            "object_id": "99999",  # Non-existent inventory ID
+            "object_id": "99999",
         }
 
         response = admin_api_client.post(url, data=data)
-        assert response.status_code == 400, response.data
-        # Check if the error is about object not existing
-        error_msg = str(response.data)
-        assert "does not exist" in error_msg.lower()
+        assert response.status_code == 201, response.data
```
john-westcott-iv, line 680: Minor: The test validates the status code but does not verify the RemoteObject assignment was actually created with the expected `object_id`. Consider adding an assertion on the response data:

```suggestion
        response = admin_api_client.post(url, data=data)
        assert response.status_code == 201, response.data
        assert str(response.data["object_id"]) == "99999"
```

This ensures the fallback path produced the correct assignment, not just a 201 from some other code path.

test_app/tests/jwt_consumer/common/test_auth.py
```diff
@@ -141,6 +141,46 @@ def test_map_user_fields(self, user_fields, token, should_save, caplog, shut_up_
...
+                'first_name': '',
+                'last_name': '',
+                'email': 'new@example.com',
+                'is_superuser': False,
+            }
+        }
+
+        with impersonate(regular_user):
+            common_auth.map_user_fields()
+
```
john-westcott-iv, line 190: Minor: This test could pass vacuously if the `rbac_pre_save_enforce_email_policy` signal is not connected in the test environment. Consider adding a precondition that proves the signal *would* block a direct `user.save()` under `impersonate(regular_user)` — that way the test is a true regression test, not just confirming `map_user_fields` updates the email (which the existing parametrized test above already covers).

ansible_base/rbac/service_api/serializers.py
```diff
@@ -133,8 +133,8 @@ def create(self, validated_data):
                 else:
                     try:
                         obj = model.objects.get(pk=object_id)
-                    except model.DoesNotExist as exc:
-                        raise serializers.ValidationError({'object_id': str(exc)})
+                    except (model.DoesNotExist, ValueError, TypeError):
```
john-westcott-iv: Minor: The RemoteObject fallback is silent. Adding a log line here would help with debugging sync ordering issues and data integrity investigations in production:

```suggestion
                    except (model.DoesNotExist, ValueError, TypeError):
                        logger.info(
                            "Object pk=%s not found locally for %s, using RemoteObject fallback",
                            object_id, model.__name__,
                        )
                        obj = RemoteObject(content_type=rd.content_type, object_id=object_id)
```

john-westcott-iv (approved):
LGTM with three minor suggestions. Approving conditionally — please address the inline comments (test assertion strengthening, test precondition, logging) before merge. No re-review needed.

**9.5/10** — Functionality 9.5, Security 10, Quality 9.0

The RemoteObject fallback and `impersonate(None)` fix are both well-scoped and correct. The RBAC impact analysis in the PR description is thorough. SonarQube clean.

ansible_base/rbac/service_api/serializers.py
bhavenst: Fixed in f4a6bb6. Added `logger.info` with the object PK and model name.

test_app/tests/rbac/remote/test_service_api.py
bhavenst: Fixed in f4a6bb6. Added `assert str(response.data["object_id"]) == "99999"`.

test_app/tests/jwt_consumer/common/test_auth.py
bhavenst: Fixed in f4a6bb6. Added a precondition block that proves a direct `user.save()` under `impersonate(regular_user)` raises `ValidationError`, then verifies email is still the old value — before the actual `map_user_fields()` call.

bhavenst: Wow, vacuously?  WTF Claude..

ansible_base/rbac/service_api/serializers.py
```diff
@@ -133,8 +137,9 @@ def create(self, validated_data):
...
+                        logger.info("Object pk=%s not found locally for %s, using RemoteObject fallback", object_id, model.__name__)
+                        obj = RemoteObject(content_type=rd.content_type, object_id=object_id)
```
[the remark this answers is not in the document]
bhavenst: Fixed in f9c056a. Added `DjangoValidationError` to the except clause — Django's `UUIDField.to_python()` raises this instead of `ValueError` for malformed UUID strings.

PR #1008, opened 2026-05-20
[AAP-56519] Fix: Move capture_oauth_email_pipeline after associate_user

bhavenst:
## Summary

Fixes [AAP-56519](https://redhat.atlassian.net/browse/AAP-56519) — GitHub initial login fails with `IntegrityError: duplicate key value violates unique constraint "dab_authentication_authe_provider_id_uid_*_uniq"`.

(bhavenst) - I have tested this fix with a github authenticator and it works great.  It also solves the problem of the github ID becoming the username (numeric instead of actual username).

## Root Cause

The `capture_oauth_email_pipeline` function was running **before** `associate_user` in the social auth pipeline (`SOCIAL_AUTH_PIPELINE` in `settings_logic.py`). This caused it to prematurely create an `AuthenticatorUser` record via `get_or_create_authenticator_user()`, which independently determines a Django username from the numeric GitHub uid (e.g., `"49794041"`), creating a phantom user different from the pipeline's user (e.g., `"octocat"`). The subsequent `associate_user` step hit an `IntegrityError` on the `(provider, uid)` unique constraint, and its recovery logic detected a user mismatch, raising `AuthAlreadyAssociated` and failing the login. The bug was most reliably triggered when the GitHub user has a private email.

## Changes

- **`ansible_base/lib/dynamic_config/settings_logic.py`**: Reordered `SOCIAL_AUTH_PIPELINE` to place `capture_oauth_email_pipeline` **after** `associate_user`
- **`ansible_base/authentication/social_auth.py`**: Rewrote `capture_oauth_email_pipeline` to **update** an existing `AuthenticatorUser`'s email field (from `kwargs['social']` set by `associate_user`) instead of calling `get_or_create_authenticator_user()` which could create duplicate records
- **`test_app/tests/authentication/test_social_auth.py`**: Updated all existing tests for the new behavior and added regression tests (pipeline ordering, fallback lookup, missing record warning)

## Type of Change

## Risk Assessment
- **Confidence:** High — the root cause is clear and the fix is narrowly scoped to pipeline ordering and removing the premature `get_or_create` call
- **Risk:** Low — only affects the OAuth email capture step; `associate_user` already creates the `AuthenticatorUser` record, so this change simply stops duplicating that work

## Test Plan
- [x] Existing `capture_oauth_email_pipeline` tests updated for new behavior
- [x] New regression test: pipeline ordering asserts `associate_user` precedes `capture_oauth_email_pipeline`
- [x] New test: fallback DB lookup when `social` key is missing from kwargs
- [x] New test: warning logged when no `AuthenticatorUser` record exists

Fixes: [AAP-56519](https://redhat.atlassian.net/browse/AAP-56519)

## Summary by CodeRabbit

* **Bug Fixes**
  * OAuth email sync now updates existing accounts only when the incoming email differs; prefers an existing social object and logs a warning if no matching account is found.
  * Social auth pipeline ordering changed so account association runs before email capture, improving reliability.

* **Tests**
  * Refactored and expanded social auth tests covering edge cases, fallback DB lookup behavior, exception logging, and pipeline-ordering assertions.

[AAP-56519]: https://redhat.atlassian.net/browse/AAP-56519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AAP-56519]: https://redhat.atlassian.net/browse/AAP-56519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ansible_base/authentication/social_auth.py
```diff
@@ -224,14 +228,20 @@ def capture_oauth_email_pipeline(*args, backend, details, **kwargs):
...
+            social = AuthenticatorUser.objects.get(uid=uid, provider=backend.database_instance)
+
+        if email and social.email != email:
+            social.email = email
+            social.save(update_fields=['email'])
+            logger.info(f"Stored OAuth email for user {user.username} from {backend.database_instance.name}")
+    except AuthenticatorUser.DoesNotExist:
+        logger.warning(f"No AuthenticatorUser found for uid={uid}, cannot store OAuth email for user {user.username}")
     except Exception as e:
         logger.warning(f"Failed to store OAuth email for user {user.username}: {e}")
```
john-westcott-iv, line 246: Nit: The `except AuthenticatorUser.DoesNotExist` on line 243 can only be raised by the `.get()` on line 237, but because the `try` block wraps everything (the `.get()`, the email comparison, and the `.save()`), you have to mentally trace which line can actually raise it. Consider tightening the scope:

```suggestion
    social = kwargs.get('social')
    if social is None:
        logger.warning(
            f"'social' key missing from pipeline kwargs for user {user.username}; falling back to DB lookup. Check SOCIAL_AUTH_PIPELINE ordering."
        )
        try:
            social = AuthenticatorUser.objects.get(uid=uid, provider=backend.database_instance)
        except AuthenticatorUser.DoesNotExist:
            logger.warning(f"No AuthenticatorUser found for uid={uid}, cannot store OAuth email for user {user.username}")
            return

    try:
        if email and social.email != email:
            social.email = email
            social.save(update_fields=['email'])
            logger.info(f"Stored OAuth email for user {user.username} from {backend.database_instance.name}")
```

This way `DoesNotExist` sits right next to the `.get()` that causes it, and the broad `except Exception` only wraps the save path.

john-westcott-iv (approved):
Clean fix — root cause is clear, pipeline reorder is correct, tests are thorough (especially the structural ordering test and the real-DB integration tests). One readability nit inline about tightening the try/except scope.

PR #1019, opened 2026-06-04
AAP-78182 Skip unregistered ResourceType entries during post_migrate

### LINKED ISSUE #1018: initialize_resources crashes with KeyError on stale ResourceType entries after upgrade
## Summary

`initialize_resources` in `ansible_base/resource_registry/apps.py` crashes with a `KeyError` when the `dab_resource_registry_resourcetype` table contains entries for models that are not in the current `RESOURCE_LIST`.

## Steps to Reproduce

1. Deploy a version of an application (e.g. AAP controller 4.7.8) where a model (e.g. `JobTemplate`) is registered in `RESOURCE_LIST` — this creates a `ResourceType` row (e.g. `awx.job_template`)
2. Upgrade to a version (e.g. AAP controller 4.8.0) where that model has been removed from `RESOURCE_LIST`
3. Run `awx-manage migrate --noinput`

## Expected Result

Migration completes successfully. Unregistered `ResourceType` entries are skipped gracefully.

## Actual Result

The migration crashes during the `post_migrate` signal handler with:

```
  File "ansible_base/resource_registry/apps.py", line 80, in initialize_resources
    resource_config = registry.get_config_for_model(resource_model)
  File "ansible_base/resource_registry/registry.py", line 105, in get_config_for_model
    return self.registry[model._meta.label]
KeyError: 'main.JobTemplate'
```

## Root Cause

`initialize_resources` has two phases:

1. **Phase 1** ("Create resource types") — iterates `registry.get_resources()` and creates/updates `ResourceType` rows via `update_or_create`. This only touches models currently in `RESOURCE_LIST` — it never removes entries for models that have been deregistered.

2. **Phase 2** ("Create resources") — iterates `ResourceType.objects.all()` and calls `registry.get_config_for_model()` for each entry. This performs a bare dict lookup (`self.registry[model._meta.label]`) which raises `KeyError` for any `ResourceType` whose model is not in the current registry.

This can also affect multi-service deployments where one service creates `ResourceType` entries in a shared database that are not in another service's `RESOURCE_LIST`.

## Environment

- AAP Controller upgrade from 4.7.8 to 4.8.0 (AAP 2.7) on OpenShift
- django-ansible-base version shipped with AAP 2.7
- PostgreSQL 16

## Workaround

Manually delete the stale `ResourceType` row from the database before running the migration:

```sql
DELETE FROM dab_resource_registry_resourcetype WHERE name = 'awx.job_template';
```

## Proposed Fix

See #1019 — catch `KeyError` in Phase 2 and skip with a warning log, rather than crashing.

snecklifter:
## Summary

Fixes #1018

Skip `ResourceType` entries whose model is not in the current `RESOURCE_LIST` during the "Create resources" phase of `initialize_resources`, instead of crashing with a `KeyError`.

## Problem

When a model is removed from `RESOURCE_LIST` between upgrades, its `ResourceType` row persists in the database. Phase 2 of `initialize_resources` iterates `ResourceType.objects.all()` and calls `registry.get_config_for_model()`, which raises `KeyError` for models no longer in the registry.

This was observed during AAP controller upgrades (4.7.8 → 4.8.0) where `JobTemplate` was removed from `RESOURCE_LIST` but its `ResourceType` row remained, causing the migration job to fail repeatedly with:

```
KeyError: 'main.JobTemplate'
```

## Approach

Catch `KeyError` in Phase 2 and `continue` with a warning log, rather than deleting stale `ResourceType` rows. Deletion was considered but rejected because in multi-service deployments (e.g., AAP gateway + controller sharing a database), other services may have legitimately created `ResourceType` entries that are not in the current service's `RESOURCE_LIST`.

## Changes

- `ansible_base/resource_registry/apps.py`: Wrap `registry.get_config_for_model()` in a try/except `KeyError` — log a warning and skip the entry.
- `test_app/tests/resource_registry/test_migration.py`: Add test that creates an unregistered `ResourceType` entry and verifies `initialize_resources` completes without raising `KeyError`.

## Test plan

- [ ] Existing `test_existing_resources_created_in_post_migration` still passes
- [ ] New `test_unregistered_resource_type_skipped_on_post_migration` passes
- [ ] Manually verified fix resolves the `KeyError: 'main.JobTemplate'` on AAP 2.7 upgrade

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved post-migration resource initialization to gracefully handle missing or unregistered model references by logging a warning and skipping the affected resource type instead of failing the migration.
  * Added a test that verifies unregistered resource types are skipped and that existing stale entries remain after post-migration initialization.

### DIFF
--- a/ansible_base/resource_registry/apps.py
+++ b/ansible_base/resource_registry/apps.py
@@ -76,8 +76,12 @@ def initialize_resources(sender, **kwargs):
 
         # Create resources
         for r_type in ResourceType.objects.all():
-            resource_model = apps.get_model(r_type.content_type.app_label, r_type.content_type.model)
-            resource_config = registry.get_config_for_model(resource_model)
+            try:
+                resource_model = apps.get_model(r_type.content_type.app_label, r_type.content_type.model)
+                resource_config = registry.get_config_for_model(resource_model)
+            except (KeyError, LookupError):
+                logger.warning(f"skipping resource type '{r_type.name}': model is not in the resource registry")
+                continue
 
             logger.info(f"adding unmigrated resources for {r_type.name}")
 
--- a/test_app/tests/resource_registry/test_migration.py
+++ b/test_app/tests/resource_registry/test_migration.py
@@ -1,6 +1,8 @@
 import pytest
+from django.contrib.contenttypes.models import ContentType
 
-from ansible_base.resource_registry.models import Resource
+from ansible_base.resource_registry.apps import initialize_resources
+from ansible_base.resource_registry.models import Resource, ResourceType
 
 
 @pytest.mark.django_db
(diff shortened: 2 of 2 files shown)


snecklifter:
No interest, closing.

PR #1021, opened 2026-06-05
[AAP-76815] Add extra_fields query param and page_size support to resource list endpoint

huffmanca:
## Summary
- Add `extra_fields` query parameter to `ResourceListSerializer` with an `ALLOWED_EXTRA_FIELDS` whitelist — when `?extra_fields=resource_data` is passed, `resource_data` is included in list responses. Unknown field names are silently ignored. Default behavior unchanged (remains `write_only`).
- Enable `page_size` override on `ResourcesPagination` (capped at `DEFAULT_MAX_PAGE_SIZE`)
- Add `rest_filters_reserved_names` to `ResourceAPIMixin` so `FieldLookupBackend` doesn't reject `extra_fields` as an unknown field
- Add OpenAPI schema description for the `extra_fields` parameter via `extend_schema_if_available`

## Motivation
`migrate_service_data` currently makes an individual GET for every resource because the list endpoint doesn't return `resource_data`. This causes ~71K API calls at production scale. With this change, Gateway can fetch `resource_data` in bulk via the list endpoint.

## Test plan
- [x] `test_resources_list_excludes_resource_data_by_default` — default list still omits `resource_data`
- [x] `test_resources_list_includes_resource_data_with_extra_fields` — `?extra_fields=resource_data` includes it
- [x] `test_resources_list_ignores_unknown_extra_fields` — unknown values silently ignored
- [x] `test_resources_list_extra_fields_mixed_valid_and_invalid` — valid fields included, unknown ignored in same request
- [x] `test_resources_list_page_size` — `page_size` query param is respected
- [x] Validated end-to-end in aap-dev cluster: `register-services` completed successfully with all 3 services

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Resource list responses exclude detailed resource data by default and allow including it via an optional query parameter (e.g., extra_fields=resource_data).
  * Pagination now supports a page_size query parameter with an enforced maximum to control results per page.
  * API documentation/schema updated to describe the optional extra_fields parameter.

* **Tests**
  * Added API tests covering extra_fields behavior and pagination responses.

### DIFF
--- a/ansible_base/resource_registry/serializers.py
+++ b/ansible_base/resource_registry/serializers.py
@@ -27,10 +27,20 @@ class ResourceDataField(serializers.JSONField):
 
 
 class ResourceListSerializer(serializers.ModelSerializer):
+    ALLOWED_EXTRA_FIELDS = {"resource_data"}
+
     has_serializer = serializers.SerializerMethodField()
     url = serializers.SerializerMethodField()
     resource_type = serializers.CharField(required=False)
     resource_data = ResourceDataField(source="*", write_only=True, required=False, default={})
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        request = self.context.get("request")
+        if request:
+            requested = set(request.query_params.get("extra_fields", "").split(",")) - {""}
+            for field_name in requested & self.ALLOWED_EXTRA_FIELDS:
+                self.fields[field_name].write_only = False
 
     class Meta:
         model = Resource
--- a/ansible_base/resource_registry/serializers.py
+++ b/ansible_base/resource_registry/serializers.py
@@ -27,7 +27,7 @@ class ResourceDataField(serializers.JSONField):
 
 
 class ResourceListSerializer(serializers.ModelSerializer):
-    ALLOWED_EXTRA_FIELDS = {"resource_data"}
+    ALLOWED_EXTRA_FIELDS = frozenset({"resource_data"})
 
     has_serializer = serializers.SerializerMethodField()
     url = serializers.SerializerMethodField()
@@ -37,10 +37,12 @@ class ResourceListSerializer(serializers.ModelSerializer):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         request = self.context.get("request")
-        if request:
-            requested = set(request.query_params.get("extra_fields", "").split(",")) - {""}
-            for field_name in requested & self.ALLOWED_EXTRA_FIELDS:
-                self.fields[field_name].write_only = False
+        extra_fields = request.query_params.get("extra_fields", "") if request else ""
+        if extra_fields:
+            requested = set(extra_fields.split(","))
+            for field_name in requested.intersection(self.ALLOWED_EXTRA_FIELDS):
+                if field_name in self.fields:
+                    self.fields[field_name].write_only = False
 
     class Meta:
         model = Resource
--- a/ansible_base/resource_registry/views.py
+++ b/ansible_base/resource_registry/views.py
@@ -14,6 +14,7 @@ from rest_framework.viewsets import GenericViewSet, mixins
 
 from ansible_base.lib.utils.response import CSVStreamResponse, get_relative_url
 from ansible_base.lib.utils.schema import extend_schema_if_available
+from ansible_base.lib.utils.settings import get_setting
 from ansible_base.lib.utils.views.django_app_api import AnsibleBaseDjangoAppApiView
 from ansible_base.lib.utils.views.permissions import try_add_oauth2_scope_permission
 from ansible_base.resource_registry.constants import SHARED_USER_RESOURCE_TYPE
@@ -54,7 +55,10 @@ class ResourcesPagination(PageNumberPagination):
     # isn't set, the default is no pagination.
     page_size = 50
     page_size_query_param = "page_size"
-    max_page_size = DEFAULT_MAX_PAGE_SIZE
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.max_page_size = get_setting('RESOURCE_LIST_MAX_PAGE_SIZE', DEFAULT_MAX_PAGE_SIZE)
 
 
 class ResourceAPIMixin:
--- a/ansible_base/resource_registry/views.py
+++ b/ansible_base/resource_registry/views.py
@@ -13,6 +13,7 @@ from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet, mixins
 
 from ansible_base.lib.utils.response import CSVStreamResponse, get_relative_url
+from ansible_base.lib.utils.schema import extend_schema_if_available
 from ansible_base.lib.utils.views.django_app_api import AnsibleBaseDjangoAppApiView
 from ansible_base.lib.utils.views.permissions import try_add_oauth2_scope_permission
 from ansible_base.resource_registry.constants import SHARED_USER_RESOURCE_TYPE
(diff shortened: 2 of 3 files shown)


ansible_base/resource_registry/views.py
```diff
@@ -67,6 +71,7 @@ class ResourceAPIMixin:
     - Use Page/Number pagination
     """
 
+    rest_filters_reserved_names = ("extra_fields",)
```
AlanCoding, line 78: 👍  I was looking for this. Expected.

ansible_base/resource_registry/serializers.py
```diff
@@ -27,11 +27,21 @@ def to_internal_value(self, data):
 
 
 class ResourceListSerializer(serializers.ModelSerializer):
+    ALLOWED_EXTRA_FIELDS = {"resource_data"}
```
AlanCoding: I'm a bit confused. Is `resource_data` a direct field on `Resource`? If not, what exactly is happening?

AlanCoding: oh ok it's all under `ResourceDataField` which is already hooked up at the serializer level, and is required for anything passed here... which is just 1

AlanCoding (approved):
This is going to be dramatically in-efficient for a single request as we can not prefetch the needed data.

This is all just to say that we will need much more care on the gateway side of this. Doing this will dramatically increase performance, but I fear great risk of combining this field with page size of 200 or something. Because we will add a lot of server activity per-item when adding this. So whenever used, we need to not _also_ be maxing out page size, I think that could produce potentially timing-out response times.

huffmanca:
@AlanCoding ,

About this bit:
>This is going to be dramatically in-efficient for a single request as we can not prefetch the needed data.
Would it be possible to do something like this in the queryset?
```
def get_queryset(self):
    qs = super().get_queryset()
    request = getattr(self, "request", None)
    if request:
        extra = request.query_params.get("extra_fields", "")
        if "resource_data" in extra.split(","):
            qs = qs.prefetch_related("content_object")
    return qs
```
I definitely don't want to always do the prefetch, and this feels quite a bit like a targeted hack instead of a clean solution; however, I'm struggling to come up with ways to cleanly optimize these calls in `migrate_service_data` from within Gateway. 

> but I fear great risk of combining this field with page size of 200 or something. Because we will add a lot of server activity per-item when adding this. So whenever used, we need to not also be maxing out page size, I think that could produce potentially timing-out response times.

Fully understood. I think for now we can stick to the default page_size of 50 (which still reduces the HTTP overhead by 50x) and work on making it configurable.

AlanCoding:
It's more that you can't do the prefetch, because this uses data from the object in the other tables, via a generic-foreign-key. There are ways, but not likely to be worth it.

ansible_base/resource_registry/serializers.py
john-westcott-iv: **Nit**: `{"resource_data"}` is visually ambiguous with a dict literal for engineers less familiar with Python set syntax. `frozenset` removes the ambiguity and signals immutability:

```suggestion
    ALLOWED_EXTRA_FIELDS = frozenset({"resource_data"})
```

ansible_base/resource_registry/serializers.py
```diff
@@ -27,11 +27,21 @@ def to_internal_value(self, data):
...
     resource_type = serializers.CharField(required=False)
     resource_data = ResourceDataField(source="*", write_only=True, required=False, default={})
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        request = self.context.get("request")
+        if request:
+            requested = set(request.query_params.get("extra_fields", "").split(",")) - {""}
+            for field_name in requested & self.ALLOWED_EXTRA_FIELDS:
+                self.fields[field_name].write_only = False
```
john-westcott-iv: **Minor** (x3): A few readability and robustness improvements:

1. Guard on `extra_fields` being truthy instead of the `- {""}` set subtraction — clearer intent
2. Use `.intersection()` instead of `&` for readability
3. Add `if field_name in self.fields:` guard to prevent `KeyError` if `ALLOWED_EXTRA_FIELDS` ever drifts from declared fields

```suggestion
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        request = self.context.get("request")
        extra_fields = request.query_params.get("extra_fields", "") if request else ""
        if extra_fields:
            requested = set(extra_fields.split(","))
            for field_name in requested.intersection(self.ALLOWED_EXTRA_FIELDS):
                if field_name in self.fields:
                    self.fields[field_name].write_only = False
```

john-westcott-iv (commented):
## Code Review: [AAP-76815] extra_fields + page_size support

**Score**: 9.2/10 (Functionality 8.5 | Security 10 | Quality 9.0)

### Inline Findings
- **1 Major**: `ResourcesPagination.max_page_size` bypasses service-level settings — consider dedicated `RESOURCE_LIST_MAX_PAGE_SIZE`
- **3 Minor** (combined in one suggestion): Restructure `__init__`, use `.intersection()`, add `KeyError` guard
- **1 Nit**: `frozenset` for `ALLOWED_EXTRA_FIELDS`

### Missing Test Coverage
- `?extra_fields=` (empty string) — exercises the empty-guard path
- `page_size` exceeding max — verifies the cap is enforced

### Observation
The `extra_fields` pattern could become a reusable `ExtraFieldsMixin` in `ansible_base/lib/serializers/mixins.py` — wouldn't hold up this PR, but worth considering for follow-up.

See inline comments for code suggestions.

ansible_base/resource_registry/views.py
```diff
@@ -51,6 +53,8 @@ class ResourcesPagination(PageNumberPagination):
     # PageNumberPagination by itself doesn't work in some apps because when api_settings.PAGE_SIZE
     # isn't set, the default is no pagination.
     page_size = 50
+    page_size_query_param = "page_size"
+    max_page_size = DEFAULT_MAX_PAGE_SIZE
```
john-westcott-iv: **Major**: `DEFAULT_MAX_PAGE_SIZE` is hardcoded here, bypassing the service-level `MAX_PAGE_SIZE` setting that [`DefaultPaginator` respects via `get_setting`](https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/rest_pagination/default_paginator.py#L21).

Given @AlanCoding's concern that `resource_data` is expensive per-item and shouldn't be combined with large page sizes, consider introducing a dedicated `RESOURCE_LIST_MAX_PAGE_SIZE` setting. This decouples this endpoint's cap from the normal API pagination, letting teams tune it independently:

```python
class ResourcesPagination(PageNumberPagination):
    page_size = 50
    page_size_query_param = "page_size"

    def __init__(self, *args, **kwargs):
        self.max_page_size = get_setting(
            'RESOURCE_LIST_MAX_PAGE_SIZE', DEFAULT_MAX_PAGE_SIZE
        )
```

This follows the same pattern `DefaultPaginator` uses (settings in `__init__` to avoid load-order issues).

Also, `page_size_query_param = "page_size"` duplicates what [`DefaultPaginator` already defines](https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/rest_pagination/default_paginator.py#L16). Consider inheriting from `DefaultPaginator` or referencing its constant.

ansible_base/resource_registry/serializers.py
huffmanca: Updated to use a frozenset.

ansible_base/resource_registry/serializers.py
huffmanca: Updated to use precisely this snippet.

ansible_base/resource_registry/views.py
huffmanca: Updated to use the suggested snippet, but not extending DefaultPaginator just yet, since it appears to have other changes.

huffmanca:
All feedback is addressed, and it should be ready for @john-westcott-iv 's re-review.

john-westcott-iv (approved):

PR #1024, opened 2026-06-15
[AAP-78673] Extract shared role assignment utilities

tznamena:
## Description

Extract role assignment utilities (`AssignmentTuple`, `RemoteAssignmentFetcher`, `get_local_assignments()`, `create_local_assignment()`, `delete_local_assignment()`, and related helpers) from `resource_registry/tasks/sync.py` into a new shared module at `ansible_base/rbac/assignment_utils.py`.

- **What is being changed?** Nine assignment-related definitions are moved from `sync.py` into a standalone module. `get_local_assignments()` gains an optional `service` parameter and its N+1 query pattern is replaced with bulk resolution. An `AssignmentClient` Protocol is introduced so callers can provide their own HTTP client. All symbols are re-exported from `sync.py` for backward compatibility.
- **Why is this change needed?** Gateway's `migrate_service_data` command needs to reuse the same set-diff assignment utilities that `resource_sync` uses, but they are currently buried inside the resource_registry module. At 86K assignments, the per-row `Resource.get_resource_for_object()` call in `get_local_assignments()` issues O(N) queries.
- **How does this change address the issue?** The shared module makes the utilities importable by both `resource_sync` (existing consumer, no behavior change) and `migrate_service_data` (follow-up story AAP-78674). Bulk resolution reduces `get_local_assignments()` from O(N) queries to 4–5 total. The `service` parameter enables scoping to a single service's assignments.
- **Behavior change:** `get_local_assignments()` previously raised `RuntimeError` (via `get_ansible_id_or_pk()`) if any org/team object lacked a corresponding Resource entry, which crashed the entire assignment sync. It now logs an error and skips that individual assignment. This is intentional — one missing Resource entry should not poison whole set.

## Type of Change

- [ ] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites

A running PostgreSQL instance matching `test_app/settings.py` (default: port 55432).

### Steps to Test

1. Run the new tests: `pytest test_app/tests/rbac/test_assignment_utils.py -v`
2. Verify existing resource_sync tests still pass: `pytest test_app/tests/resource_registry/test_resource_sync.py -v`
3. Verify backward-compatible imports work: `python -c "from ansible_base.resource_registry.tasks.sync import AssignmentTuple, RemoteAssignmentFetcher, get_local_assignments, create_local_assignment, delete_local_assignment"`

### Expected Results

- All new tests pass (20 tests covering AssignmentTuple, RemoteAssignmentFetcher, get_local_assignments with service filter, bulk query count bounds, and backward-compatible imports).
- All existing `test_resource_sync.py` tests pass unchanged — re-exports in `sync.py` preserve import paths.
- The bounded-query-count test confirms `get_local_assignments()` stays under 15 queries for 10 assignments (would be 30+ with N+1).

## Release Notes

* **Bug Fixes**
  * Improved role assignment synchronization to gracefully handle missing assignment identifiers with fallback identification methods and optimized database query efficiency.

* **Tests**
  * Added comprehensive test coverage for role assignment synchronization, including network error handling, missing identifiers, and RBAC utility edge cases.


ansible_base/rbac/assignment_utils.py
```diff
@@ -0,0 +1,380 @@
...
+
+
+def get_remote_assignments(api_client: Any, page_size: int | None = None) -> RemoteAssignmentResult:
+    """Fetch remote assignments from the resource server and convert to tuples.
+
+    Returns a ``RemoteAssignmentResult`` so the caller can distinguish a
+    complete fetch from a partial one (e.g. due to HTTP errors or
+    timeouts mid-pagination).
+    """
+    return RemoteAssignmentFetcher(api_client, page_size=page_size).fetch()
```
AlanCoding: Everything above this line is working with the `api_client` and this doesn't really have any right to be in `ansible_base/rbac/`. It is made to work by passing it in, but similar to `ansible_base/rbac/claims.py`, that's not really good code organization.

I would say that anything dealing with the client is on the syncing side. It involves both the RBAC and resource_registry app. So it's the ugliest of all the things.

I think this logic would be better in `ansible_base/resource_registry/tasks/`, but if you absolutely want it in RBAC, I would want it quarantined in `ansible_base/rbac/client/` or something like that.

The logic after this appears to exclusively work with local models, which is 👍 perfectly good to keep around. But naming-wise, `assignment_utils` isn't something I love. I would associate it with syncing. Maybe `sync_utils.py` or something like that.

ansible_base/rbac/assignment_utils.py
```diff
@@ -0,0 +1,380 @@
...
+        return None
+
+    model_name = assignment.content_type.model
+    if model_name not in ('organization', 'team'):
+        return str(assignment.object_id)
+
+    key = (str(assignment.object_id), model_name)
+    resolved = object_map.get(key)
+    if resolved is None:
+        logger.warning(f"{model_name} {assignment.object_id} found without an associated Resource, skipping assignment.")
```
bhavenst: The old `get_ansible_id_or_pk()` raised `RuntimeError` here, which would abort the entire sync. This is more resilient (one bad record shouldn't kill an 86K-assignment sync), but a missing org/team Resource is a data integrity issue — consider `logger.error` instead of `logger.warning` so operators are more likely to notice it in monitoring.

ansible_base/rbac/assignment_utils.py
```diff
@@ -0,0 +1,380 @@
...
+                            assignment_type=assignment_type,
+                        )
+                    )
+
+                if not data.get('next'):
+                    return True
+
+                page += 1
+                logger.debug(f"Fetching next page {page} of {assignment_type} assignments")
+        except Exception:
```
bhavenst: Minor test gap: this `except Exception` path is never exercised by any test (only the HTTP-status-code failure path is tested). A one-liner like `client.list_user_assignments.side_effect = ConnectionError("timeout")` would cover it.

Similarly the `object_id` fallback on line 128 is untested — all test data uses `object_ansible_id: None`. A test confirming the `or assignment.get('object_id')` branch works would be nice.

bhavenst (approved):
Clean extraction with a nice bulk-resolution optimization. Two minor suggestions inline.

ansible_base/rbac/assignment_utils.py
```diff
@@ -0,0 +1,380 @@
...
+    is_complete: bool = False
+
+
+class RemoteAssignmentFetcher:
+    """Fetches role assignments from a remote resource server with pagination.
+
+    Collects user and team assignments into a single set.  If any page
+    request fails the fetcher stops early and marks the result as
+    incomplete so the caller can skip deletions safely.
+
```
john-westcott-iv: **[Nit — Code Organization]** Echoing Alan's feedback: `RemoteAssignmentFetcher` and `get_remote_assignments` are HTTP client code — they call the resource server's assignment API and process paginated responses. This is inherently a syncing/resource_registry concern, not an RBAC concern.

Placing this in `rbac/` creates a bidirectional dependency between the two apps: `rbac.assignment_utils` imports `resource_registry.models.Resource`, while `resource_registry.tasks.sync` imports from `rbac.assignment_utils`. Not a circular import (different sub-modules), but a coupling smell.

Consider splitting: local model code (`get_local_assignments`, bulk resolvers, `create/delete_local_assignment`) could stay in `rbac/`, while the remote-fetching code stays in `resource_registry/tasks/`.

ansible_base/rbac/role_sync_utils.py
```diff
@@ -0,0 +1,380 @@
...
+_SKIP = object()
+
+
+def _resolve_object_ansible_id(assignment, object_map: dict[tuple[str, str], str]):
+    """Resolve the ansible_id or pk for an assignment's target object.
+
+    Returns ``None`` for global assignments, the resolved ansible_id for
+    org/team objects, the raw ``object_id`` for other types, or the
+    sentinel ``_SKIP`` if an org/team resource is missing.
+    """
```
john-westcott-iv, line 136: **[Observation — Behavior Change]** This changes the error handling from the old code's `RuntimeError` (which would crash the entire sync) to a warning + skip pattern. This is a good resilience improvement — one missing Resource entry no longer poisons the entire assignment set. Worth calling out in the PR description so downstream consumers are aware.

ansible_base/rbac/role_sync_utils.py
```diff
@@ -0,0 +1,380 @@
...
+    return str(ansible_id_or_pk)
+
+
+def get_content_object(role_definition, assignment_tuple: AssignmentTuple) -> Any:
+    """Resolve the Django model instance for an assignment tuple's target object."""
+    if not is_rbac_installed():
+        raise RuntimeError("get_content_object requires ansible_base.rbac to be installed")
+    content_object = None
+    if role_definition.content_type.model in ('organization', 'team'):
+        object_resource = Resource.objects.get(ansible_id=assignment_tuple.ansible_id_or_pk)
```
john-westcott-iv, line 76: **[Nit — Pre-existing]** `role_definition.content_type.model` will raise `AttributeError` if `content_type` is `None` (global role definitions). Internal callers guard against this, but since this is now a public API (`__all__`), consider adding:

```suggestion
def get_content_object(role_definition, assignment_tuple: AssignmentTuple) -> Any:
    """Resolve the Django model instance for an assignment tuple's target object."""
    if not is_rbac_installed():
        raise RuntimeError("get_content_object requires ansible_base.rbac to be installed")
    if role_definition.content_type is None:
        raise ValueError("get_content_object requires a role_definition with a content_type")
    content_object = None
```

john-westcott-iv (commented):
## Code Review: [AAP-78673] Extract shared role assignment utilities

**Score**: 10.0/10 (Functionality 10, Security 10, Quality 10)

No Critical or Major findings. The extraction is clean, the N+1 query fix is well-designed (O(N) → 4-5 queries), backward compatibility is maintained via re-exports, and test coverage is comprehensive (20+ tests including bounded query count regression).

### Observations
- **Module placement** (see inline comment): agree with @AlanCoding that remote-fetching code belongs closer to resource_registry, not in rbac/
- **Behavior change**: `get_local_assignments` now skips (with warning) org/team objects lacking a Resource entry, instead of crashing with RuntimeError — a good resilience improvement, worth noting in the PR
- **Pre-existing**: `get_content_object` will AttributeError on global role definitions (content_type=None) — not introduced here, but worth a follow-up fix

### Limitations
API-only review — could not grep the codebase for stale references or verify callers of the changed functions.

ansible_base/rbac/assignment_utils.py
tznamena: All very good points, addressed them all in new commit.

ansible_base/rbac/assignment_utils.py
tznamena: Good catch, changed the log level.

ansible_base/rbac/assignment_utils.py
tznamena: Added tests.

ansible_base/rbac/assignment_utils.py
tznamena: Re-organized the code as per Alan's suggestion.

ansible_base/rbac/role_sync_utils.py
tznamena: Added.

ansible_base/rbac/role_sync_utils.py
tznamena: Added this to the description of the PR.

ansible_base/rbac/role_sync_utils.py
AlanCoding, line 1: I still don't like this module name. Like sync_utils.py would be better.

It's hard. This is kind of more-so about doing things with "remote" assignments. And we have a RemoteObject, so there is a potential kind of congruence between remote objects and remote assignments. But we don't need to get into that. I just find it really obscure to call it assignment_utils.

AlanCoding (approved):
only super minor nit pick remaining. this looks good.

ansible_base/rbac/role_sync_utils.py
tznamena: Fair point, I did not address the naming concern you raised initially. Will rename it.

tznamena: Renamed.

tznamena:
@AlanCoding Came up with a new name. Hope this better describes the module.

PR #1025, opened 2026-06-15
[AAP-68017] Fix compounding queries in role assignment list endpoints

AlexSCorey:
## Description

Switch assignment list views from page-number pagination to cursor-based pagination, annotate the queryset with object_ansible_id to avoid N+1 lookups, and update the remote sync fetcher to use cursor-based navigation. Tests updated to use len(results) instead of count since cursor pagination does not include a count field.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. 
2. 
3. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX
  

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved assignment listing reliability by enabling cursor pagination with consistent ordering and a fixed page size.
  * Enhanced filtering behavior by aligning RBAC list/query handling with cursor-based pagination.
* **API Changes**
  * Updated `object_ansible_id` validation to normalize UUID-like inputs and treat empty values as unset.
  * Updated remote role-assignment synchronization to request subsequent pages using the cursor from `next` links.
* **Tests**
  * Adjusted RBAC and sync tests to assert on paginated `results` length and cursor-based request parameters.

### DIFF
--- a/ansible_base/resource_registry/tasks/sync.py
+++ b/ansible_base/resource_registry/tasks/sync.py
@@ -8,7 +8,6 @@ from collections import defaultdict
 from dataclasses import dataclass, field
 from enum import Enum
 from io import StringIO, TextIOBase
-from typing import Any
 from urllib.parse import parse_qs, urlparse
 
 from asgiref.sync import sync_to_async
@@ -157,12 +156,12 @@ class RemoteAssignmentFetcher:
 
     def _fetch_page(self, list_fn, next_url):
         if next_url is None:
-            return list_fn()
+            return list_fn(filters={'page_size': self.page_size})
         cursor = parse_qs(urlparse(next_url).query).get('cursor', [None])[0]
         if cursor is None:
             logger.warning(f"Pagination URL missing cursor parameter: {next_url}")
             return None
-        return list_fn(filters={'cursor': cursor})
+        return list_fn(filters={'cursor': cursor, 'page_size': self.page_size})
 
     def _process_assignments(self, assignments_data, actor_id_key, assignment_type):
         for assignment in assignments_data:
--- a/ansible_base/resource_registry/tasks/sync.py
+++ b/ansible_base/resource_registry/tasks/sync.py
@@ -16,6 +16,7 @@ from django.db import transaction
 from django.db.models import QuerySet
 from django.db.utils import Error, IntegrityError
 from requests import HTTPError
+from urllib.parse import urlparse, parse_qs
 
 from ansible_base.lib.utils.apps import is_rbac_installed
 from ansible_base.rbac.role_sync_utils import (  # noqa: F401 — re-exported for backward compatibility
@@ -135,12 +136,17 @@ class RemoteAssignmentFetcher:
 
         Returns True if all pages were fetched successfully, False on any error.
         """
-        page = 1
+        next = None
         try:
             while True:
-                resp = list_fn(filters={'page': page, 'page_size': self.page_size})
+                if next is None:
+                    resp = list_fn()
+                else:
+                    next_query = parse_qs(urlparse(next).query)
+                    cursor = next_query.get('cursor', [None])[0]
+                    resp = list_fn(filters={'cursor': cursor})
                 if resp.status_code != 200:
-                    logger.warning(f"Failed to fetch {assignment_type} assignments page {page}: HTTP {resp.status_code}")
+                    logger.warning(f"Failed to fetch {assignment_type} assignments batch {next}: HTTP {resp.status_code}")
                     return False
 
                 data = resp.json()
@@ -162,8 +168,8 @@ class RemoteAssignmentFetcher:
                 if not data.get('next'):
                     return True
 
-                page += 1
-                logger.debug(f"Fetching next page {page} of {assignment_type} assignments")
+                next = data['next']
+                logger.debug(f"Fetching next batch {next} of {assignment_type} assignments")
         except Exception:
             logger.exception(f"Failed to fetch remote {assignment_type} assignments")
             return False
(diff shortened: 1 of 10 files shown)


test_app/tests/resource_registry/test_resource_sync.py
```diff
@@ -625,11 +625,11 @@ def test_create_api_client_default_jwt_expiration(mock_get_client):
...
 
 @pytest.mark.django_db
-def test_remote_assignment_fetcher_sends_page_size_on_all_pages():
-    """page_size should be included in filters on every page, not just the first."""
+def test_remote_assignment_fetcher_sends_cursor_on_subsequent_pages():
+    """Subsequent pages should extract the cursor from the next URL."""
     api_client = mock.Mock(spec=["list_user_assignments", "list_team_assignments"])
 
-    page1 = _mock_response(body={"results": [], "next": "http://example.com/page2"})
+    page1 = _mock_response(body={"results": [], "next": "http://example.com/api/assignments/?cursor=abc123"})
```
bhavenst, line 633: This test correctly uses a cursor-bearing URL — but the existing `test_get_remote_assignments_incomplete_on_failure` test (~line 459) still uses `"next": "http://example.com/page2"` with no `cursor` query parameter.

Under the new `_fetch_page`, that URL causes an early `return None` before the second mock (HTTP 500 or `ConnectionError`) is ever consumed. Both `"http_error"` and `"exception"` parametrize cases now test the "missing cursor" path, **not** the intended failure modes. The test still passes — but for the wrong reason.

**Fix:** update that test's `next` URL to include a cursor:
```python
"next": "http://example.com/api/assignments/?cursor=abc123"
```

ansible_base/resource_registry/tasks/sync.py
```diff
@@ -223,39 +224,49 @@ def _paginate(self, list_fn, actor_id_key: str, assignment_type: str) -> bool:
...
             return False
 
+    def _fetch_page(self, list_fn, next_url):
+        if next_url is None:
+            return list_fn()
+        cursor = parse_qs(urlparse(next_url).query).get('cursor', [None])[0]
+        if cursor is None:
+            logger.warning(f"Pagination URL missing cursor parameter: {next_url}")
+            return None
+        return list_fn(filters={'cursor': cursor})
```
bhavenst: `page_size` is now dead code. `RemoteAssignmentFetcher.__init__` (line 200) still accepts it, reads `RESOURCE_SYNC_PAGE_SIZE` from settings, and stores `self.page_size` — but `_fetch_page` never sends it to the server. Cursor pagination relies on the server's hardcoded `ServiceCursorPagination.page_size = 50`.

An operator who sets `RESOURCE_SYNC_PAGE_SIZE = 500` will believe they're controlling batch sizes, but the setting is silently ignored.

**Options:** either remove the dead `page_size` plumbing, or pass it alongside `cursor` (DRF `CursorPagination` honors a `page_size` query param by default).

bhavenst (commented):
Couple items worth addressing — nice fix overall, cursor pagination is the right call here.

ansible_base/resource_registry/tasks/sync.py
AlexSCorey: that's a good catch.  I'll change is to that the cursor class supports a page_size param.

test_app/tests/resource_registry/test_resource_sync.py
AlexSCorey: thank you

bhavenst (commented):
Previous findings addressed — thanks for the quick turnaround. The latest commit introduces a few stale test assertions though.

test_app/tests/resource_registry/test_resource_sync.py
```diff
@@ -552,16 +552,16 @@ def test_get_remote_assignments_filters_unknown_roles(static_api_client):
...
     api_client = mock.Mock(spec=["list_user_assignments", "list_team_assignments"])
     ok = _mock_response()
     api_client.list_user_assignments.return_value = ok
     api_client.list_team_assignments.return_value = ok
 
     RemoteAssignmentFetcher(api_client, page_size=100).fetch()
 
-    api_client.list_user_assignments.assert_called_with(filters={'page': 1, 'page_size': 100})
-    api_client.list_team_assignments.assert_called_with(filters={'page': 1, 'page_size': 100})
+    api_client.list_user_assignments.assert_called_with()
```
bhavenst: `_fetch_page` now calls `list_fn(filters={'page_size': self.page_size})` on the first page, but these assertions still expect no arguments. These will fail.

```python
api_client.list_user_assignments.assert_called_with(filters={'page_size': 100})
api_client.list_team_assignments.assert_called_with(filters={'page_size': 100})
```

The docstring on line 555 ("First page should be fetched with no filters") is also stale now.

test_app/tests/resource_registry/test_resource_sync.py
```diff
@@ -638,8 +638,8 @@ def test_remote_assignment_fetcher_sends_page_size_on_all_pages():
 
     user_calls = api_client.list_user_assignments.call_args_list
     assert len(user_calls) == 2
-    assert user_calls[0] == mock.call(filters={'page': 1, 'page_size': 100})
-    assert user_calls[1] == mock.call(filters={'page': 2, 'page_size': 100})
+    assert user_calls[0] == mock.call()
```
bhavenst: Same issue — `_fetch_page` now includes `page_size` in every call, but these assertions don't expect it:

```python
assert user_calls[0] == mock.call(filters={'page_size': 100})
assert user_calls[1] == mock.call(filters={'cursor': 'abc123', 'page_size': 100})
```

bhavenst (approved):
All previous findings addressed — page_size is no longer dead code, test assertions match the updated call signatures, and the stale pagination URL is fixed. LGTM, nice refactor.

PR #1022, opened 2026-06-08
[AAP-78190] fix: pass OIDC response as attrs for authenticator map evaluation

Dostonbek1:
## Description

- **What:** Adds an optional `attrs` parameter to `update_user_claims()` and passes the OIDC response from `create_user_claims_pipeline` so that identity provider claims are available for attribute-based authenticator map evaluation.
- **Why:** Commit 964366e (AAP-56519) refactored `capture_oauth_email_pipeline` to fix a GitHub login IntegrityError. The old version inadvertently stored the full OIDC response into `authenticator_user.extra_data` via `get_or_create_authenticator_user(extra_data=response)`. The new version only updates the email field, so `extra_data` no longer contains OIDC claims like `preferred_username`. Since `create_claims()` reads attributes from `extra_data`, attribute-based authenticator maps broke.
- **How:** Instead of relying on `extra_data` to contain the full identity provider response, the pipeline now explicitly passes the response as the attributes dict for trigger evaluation. Non-social-auth callers (LDAP/RADIUS/database) continue using `extra_data` as before.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Self-Review Checklist

- [x] I have performed a self-review of my code
- [x] I have added relevant comments to complex code sections
- [x] I have considered the security impact of these changes
- [x] I have considered performance implications
- [x] I have thought about error handling and edge cases

### Prerequisites
- Running environment with an OIDC authenticator configured (e.g., Keycloak)

### Steps to Test
1. Create an authenticator map with an attribute-based trigger, e.g.: `triggers: {attributes: {preferred_username: {contains: "some_user"}}}`
2. Log in via OIDC as a user whose `preferred_username` matches
3. Verify the user is mapped to the expected org/team/role

### Expected Results
- Attribute-based authenticator maps correctly match OIDC claims
- Group-based maps continue to work as before
- `test_create_user_claims_pipeline` passes with the new `attrs` assertion
- `test_update_user_claims_attrs_override` verifies the override path

---
**Note:** This PR was developed with assistance from Claude AI assistant.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced user claim attribute handling in the social authentication pipeline to improve claim-based user attribute matching during OAuth authentication flows.

* **Tests**
  * Added regression and integration test coverage for claim attribute evaluation and override behavior in authentication pipelines.

PR #1030, opened 2026-06-23
[AAP-79302] fix: normalize SAML response attrs in claims pipeline

Dostonbek1:
## Description

- **What:** Adds `get_claims_attrs(response)` to `SocialAuthMixin` and overrides it in the SAML backend to extract `response["attributes"]` before trigger evaluation.
- **Why:** PR #1022 introduced `attrs=response` to fix OIDC attribute-based map evaluation. OIDC responses carry claims at the top level, but SAML responses nest them under `response["attributes"]`. This caused all attribute-based authenticator maps to be **skipped** for SAML logins — triggers couldn't find `is_superuser`, role mappings, etc. at the top level.
- **How:** Rather than using a response-shape heuristic (checking for an `"attributes"` key), each backend now explicitly owns its response normalization via `get_claims_attrs()`. The default implementation returns the response unchanged (correct for OIDC and all other flat-response backends). The SAML backend overrides it to extract `response.get("attributes", response)`. The pipeline calls `backend.get_claims_attrs(response)` before passing attrs to `update_user_claims()`.

## Root Cause

The SAML plugin's `extra_data()` method already knows about the nesting:

```python
attrs = response["attributes"] if "attributes" in response else {}
```

But `create_user_claims_pipeline` was passing the **raw** `response` to `update_user_claims()`, so `create_claims()` evaluated triggers against the full SAML envelope instead of the flattened attributes. Result: every attribute-based map showed `"skipped"` in `last_login_map_results`.

## Type of Change

## Self-Review Checklist

### Unit Tests (no DB required)

```bash
DJANGO_SETTINGS_MODULE=test_app.settings pytest -xvs \
  test_app/tests/authentication/test_social_auth.py::test_create_user_claims_pipeline \
  test_app/tests/authentication/test_social_auth.py::test_get_claims_attrs_default_passthrough \
  test_app/tests/authentication/test_social_auth.py::test_create_user_claims_pipeline_calls_get_claims_attrs \
  test_app/tests/authentication/test_social_auth.py::test_create_user_claims_pipeline_saml_groups_extracted
```

### Integration Tests (requires PostgreSQL via tox)

```bash
tox -- test_app/tests/authentication/utils/test_claims.py::test_update_user_claims_saml_attrs_override
```

### Manual Verification

1. Configure a SAML authenticator with an attribute-based map trigger (e.g., `is_superuser: {contains: "true"}`)
2. Log in via SAML as a user whose IdP returns matching attributes
3. Verify the user is granted the expected superuser/role/org membership

---

## Release Notes

* **Bug Fixes**
  * Improved SAML authentication so attribute-based authenticators evaluate against the actual claims payload (rather than the full SAML envelope), fixing cases where attribute triggers didn’t match.

* **New Features**
  * Added a backend hook to customize how identity provider responses are normalized into claim attributes before updating user claims.

* **Tests**
  * Added unit/regression coverage for claim normalization, pipeline delegation, and SAML attribute-based trigger behavior.

### DIFF
--- a/ansible_base/authentication/authenticator_plugins/saml.py
+++ b/ansible_base/authentication/authenticator_plugins/saml.py
@@ -339,6 +339,12 @@ class AuthenticatorPlugin(SocialAuthMixin, SocialAuthValidateCallbackMixin, SAML
     def get_user_groups(self, extra_groups=[]):
         return extra_groups
 
+    def get_claims_attrs(self, response):
+        # SAML responses nest identity attributes under response["attributes"].
+        # Return the inner dict so that attribute-based authenticator map triggers
+        # evaluate against the flat claims dict rather than the full SAML envelope.
+        return response.get("attributes", response)
+
     def get_alternative_uid(self, **kwargs):
         if uid := kwargs.get("uid", None):
             if ":" in uid:
--- a/ansible_base/authentication/social_auth.py
+++ b/ansible_base/authentication/social_auth.py
@@ -175,6 +175,17 @@ class SocialAuthMixin:
         """
         return []
 
+    def get_claims_attrs(self, response):
+        """
+        Extract the attributes dict to use for authenticator map trigger evaluation.
+
+        The default implementation returns the response unchanged (suitable for OIDC
+        and other protocols where claims sit at the top level of the response).
+        Backends whose IdP wraps claims in a nested structure should override this
+        method to extract the appropriate sub-dict.
+        """
+        return response
+
     def ensure_strategy_in_args(self, args):
         if len(args) == 0:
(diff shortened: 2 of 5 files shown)


bhavenst (approved):
Clean, well-scoped fix. Template method pattern for response normalization is the right design — each backend owns its own response shape. Ordering is correct (groups extracted from raw response before `get_claims_attrs()` normalizes for trigger eval). Good test coverage across unit and integration. AWX CI failure is pre-existing/unrelated.

zkayyali812 (approved):

zkayyali812:
Merging. AWX failure is unrelated

PR #1031, opened 2026-06-23
[AAP-79004] Add covering index on Resource for ansible_id lookups

djulich:
## Summary

- Eliminates N+1 correlated subquery pattern in service API role assignment list endpoints (`role_user_assignments`, `role_team_assignments`)
- Replaces per-row `Subquery(OuterRef(...))` annotation with a single batch query after pagination using Q-object OR chain
- Adds test verifying `object_ansible_id` correctness and bounded query count (no correlated subqueries)

Resolves: https://redhat.atlassian.net/browse/AAP-79004

### Root cause

The `resource_ansible_id_expr()` helper produced a correlated subquery against `dab_resource_registry_resource` that PostgreSQL executed once per outer row. At customer scale (12,687 calls in a 4-hour capture, 9.51 min total DB time), this was a significant contributor to DB load.

### Approach

Override `list()` in `BaseSerivceRoleAssignmentViewSet` to batch-fetch `ansible_id` values after pagination:
1. Paginate normally
2. Collect unique `(content_type_id, object_id)` pairs from the page
3. Issue one `Resource.objects.filter(Q(...) | Q(...))` query
4. Set `_object_ansible_id_annotation` on each object (same attribute name the serializer already reads)

The serializer (`ObjectAnsibleIdField`) needs no changes — the annotation attribute name is preserved, and the fallback path still works for assign/unassign operations.

## Test plan

- [x] All 38 existing service API tests pass
- [x] All 16 public API compat tests pass
- [x] New `test_list_assignments_object_ansible_id_batch_fetch` verifies:
  - `object_ansible_id` values are correctly returned for multiple assignments
  - Exactly 1 batch query with OR chain (not N correlated subqueries)
- [x] Assign/unassign endpoints still work (serializer fallback path unchanged)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected service assignment listing at the organization scope so returned `object_ansible_id` values include all applicable organizations.
  * Improved resource uniqueness and indexing for `(content_type, object_id)` using a covering constraint that incorporates `ansible_id`.
* **Tests**
  * Added a test that verifies service assignment list results across multiple organizations.

### DIFF
--- a/ansible_base/resource_registry/migrations/0008_resource_covering_index_ansible_id.py
+++ b/ansible_base/resource_registry/migrations/0008_resource_covering_index_ansible_id.py
@@ -1,0 +1,27 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("dab_resource_registry", "0007_alter_resource_ansible_id_and_more"),
+    ]
+
+    operations = [
+        migrations.RemoveIndex(
+            model_name="resource",
+            name="dab_resourc_content_6d9d9c_idx",
+        ),
+        migrations.AlterUniqueTogether(
+            name="resource",
+            unique_together=set(),
+        ),
+        migrations.AddConstraint(
+            model_name="resource",
+            constraint=models.UniqueConstraint(
+                fields=("content_type", "object_id"),
+                include=("ansible_id",),
+                name="unique_resource_content_type_object_id",
+            ),
+        ),
+    ]
--- a/test_app/tests/rbac/remote/test_service_api.py
+++ b/test_app/tests/rbac/remote/test_service_api.py
@@ -153,7 +153,7 @@ def test_object_ansible_id_in_list_response(admin_api_client, rando, org_admin_r
     org_admin_rd.give_permission(rando, org2)
 
     url = get_relative_url('serviceuserassignment-list')
-    response = admin_api_client.get(url, format="json")
+    response = admin_api_client.get(url + '?page_size=200', format="json")
     assert response.status_code == 200, response.data
 
     org_assignments = [a for a in response.data['results'] if a['role_definition'] == org_admin_rd.name]
--- a/test_app/tests/rbac/remote/test_service_api.py
+++ b/test_app/tests/rbac/remote/test_service_api.py
@@ -5,7 +5,7 @@ from rest_framework.test import APIClient
 
 from ansible_base.lib.utils.response import get_relative_url
 from ansible_base.rbac.models import DABContentType, DABPermission, RoleDefinition, RoleTeamAssignment, RoleUserAssignment
-from test_app.models import Team, User
+from test_app.models import Organization, Team, User
 
 
 @pytest.mark.django_db
@@ -146,8 +146,6 @@ def test_list_role_team_assignments_includes_id(admin_api_client, inv_rd, invent
 @pytest.mark.django_db
 def test_object_ansible_id_in_list_response(admin_api_client, rando, org_admin_rd, organization):
     """Verify object_ansible_id is correctly returned for organization-level assignments."""
-    from test_app.models import Organization
-
     org2 = Organization.objects.create(name='Covering Index Test Org')
     org_admin_rd.give_permission(rando, organization)
     org_admin_rd.give_permission(rando, org2)
--- a/test_app/tests/rbac/remote/test_service_api.py
+++ b/test_app/tests/rbac/remote/test_service_api.py
@@ -144,6 +144,27 @@ def test_list_role_team_assignments_includes_id(admin_api_client, inv_rd, invent
 
 
 @pytest.mark.django_db
+def test_object_ansible_id_in_list_response(admin_api_client, rando, org_admin_rd, organization):
+    """Verify object_ansible_id is correctly returned for organization-level assignments."""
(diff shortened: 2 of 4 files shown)


AlanCoding:
There is a strange deja vu going on, where the same basic review played out in a prior PR here.

https://github.com/ansible/django-ansible-base/pull/999

Where basically I was saying the same thing I would say here, that batch annotations are a really ugly pattern, and we should really try to figure out another way.

AlanCoding:
Like, I was asking if we could look into the `UniqueConstraint` approach.

djulich:
@AlanCoding, I read your comments in PR #999, and I understand that you prefer a general solution to just fixing one particular view (understandable). But the `UniqueConstraint` approach didn't work because our CI uses SQLite which doesn't support `include` in `UniqueConstraint`. So this approach seems to be a dead end.

The only promising alternative seems to be (I quote you here) "develop a reusable GFK-to-GFK relation, so that Assignment models can point to Resource models without blowing everything up". I just think that this is a larger architectural change and the batch-fetch would provide a (tactical) fix for the active customer escalation.

AlanCoding:
> because our CI uses SQLite which doesn't support include in UniqueConstraint. So this approach seems to be a dead end.

The CI uses both postgres and sqlite3. You might have seen a sqlite3-specific error, but if those are in tests, those can be ignored because you know they will run in the postgres tests. But there's also the matter of not being able to migrate sqlite3. @chrismeyersfsu has worked around that in the past, because sqlite3 is not production-impacting, so even a hacky solution is sufficient, because in this case it is _just_ for CI and development.

Going this route, there are going to be more problems with overriding methods like `list`, like the OpenAPI schema getting messed up. You don't want that. Even if you are going to do batch annotation (which is a bad idea), it can be done without overriding `list`.

I also don't think you'll actually solve anything with this. You're doing it for the /service-index/ because that's what someone reported, but there are likely other views that do the same thing, or will at some point in the future do the same thing.

If the unique constraint doesn't work, then the ansible_id can be cached on the assignment records. This is a bad approach. But it's better that the other options with the batch in-python annotation.

ansible_base/resource_registry/migrations/0008_resource_covering_index_ansible_id.py
```diff
@@ -0,0 +1,27 @@
...
+            unique_together=set(),
+        ),
+        migrations.AddConstraint(
+            model_name="resource",
+            constraint=models.UniqueConstraint(
+                fields=("content_type", "object_id"),
+                include=("ansible_id",),
+                name="unique_resource_content_type_object_id",
+            ),
+        ),
```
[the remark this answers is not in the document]
djulich: Migration not online-safe (blocking writes) — Not actionable.

CodeRabbit warns that AddConstraint holds a write lock during index creation. In theory true, but the dab_resource_registry_resource table is small (one row per registered resource — at customer's scale, ~10K rows). Index creation takes milliseconds at that size. Making it CONCURRENTLY would add significant complexity (SeparateDatabaseAndState, raw SQL, SQLite branching) for no practical benefit. No other migration in this codebase uses CONCURRENTLY either.

djulich:
### Note on SQLite behavior

The `UniqueConstraint` with `include=["ansible_id"]` relies on PostgreSQL's covering index support. On SQLite (used in CI and local dev), Django's `_unique_supported()` checks `connection.features.supports_covering_indexes`, which is `False` for SQLite. This causes `_create_unique_sql()` to return `None`, and `add_constraint()` silently skips the constraint creation.

This means **on SQLite, there is no uniqueness enforcement on `(content_type, object_id)` after this migration runs**. The `RemoveIndex` and `AlterUniqueTogether` operations execute normally, but the replacement `AddConstraint` is a no-op.

This is acceptable because SQLite is not used in production — AAP deployments use PostgreSQL, where the covering index is fully functional.

djulich:
The failing CI checks (`py311-in-files`, `py312-in-files`) are unrelated to this PR. The only failing test is:

```
test_app/tests/lib/dynamic_config/test_dynaconf_settings.py::test_load_python_file_with_injected_scope
AttributeError: _post_hooks
```

This is a dynaconf settings test — no connection to the migration or model changes in this PR. All RBAC, service API, and SQLite tests pass.

test_app/tests/rbac/remote/test_service_api.py
```diff
@@ -143,6 +143,27 @@ def test_list_role_team_assignments_includes_id(admin_api_client, inv_rd, invent
     assert from_api['team_ansible_id'] == str(team.resource.ansible_id)
 
 
+@pytest.mark.django_db
+def test_object_ansible_id_in_list_response(admin_api_client, rando, org_admin_rd, organization):
+    """Verify object_ansible_id is correctly returned for organization-level assignments."""
+    from test_app.models import Organization
```
AlanCoding: ...could move to top of file

AlanCoding (approved):
yeah looks good.

There are some nuances to backporting this, but it should be something that can be backported, just need to check the earlier numbers are there when doing it.

test_app/tests/rbac/remote/test_service_api.py
djulich: Done

PR #1032, opened 2026-06-24
[AAP-78674] Include id in service-index assignment serializer response

tznamena:
## Description

- What is being changed?
  - Adding `id` to `assignment_common_fields` in the service-index assignment serializer so that `ServiceRoleUserAssignmentSerializer` and `ServiceRoleTeamAssignmentSerializer` include the assignment primary key in their API responses.
- Why is this change needed?
  - The gateway's `migrate_service_data` command implements PK-cursor-based pagination (`id__gt=<last_pk>&order_by=id`) for role assignment migration. Without `id` in the response, the cursor cannot advance and every run reprocesses all assignments — negating the performance benefit of cursor-based resume after crashes and skip-on-reinstall.
  - With `id` in the response, the gateway can track the last-processed assignment PK per service and skip already-processed items on re-runs. On a reinstall where all assignments are synced, the API returns zero results and migration completes immediately instead of reprocessing 100K+ assignments.

## Type of Change

- [ ] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites

Standard DAB test environment with PostgreSQL.

### Steps to Test
1. Run the service API tests: `pytest test_app/tests/rbac/remote/test_service_api.py -v`
2. Verify `test_list_role_user_assignments` passes — asserts `id` is present and matches the DB record PK
3. Verify `test_list_role_team_assignments_includes_id` passes — same for team assignments
4. Optionally, start a test service and query `GET /api/v1/service-index/role-user-assignments/` — confirm `id` appears in each result

### Expected Results

- All existing tests pass (no regressions — no test previously asserted `id` was absent)
- Two new/updated test assertions confirm `id` is present and correct in both user and team service-index assignment responses

### Why `id` was missing

The service-index `BaseAssignmentSerializer` inherits from `serializers.ModelSerializer` directly, not from `AbstractCommonModelSerializer` (which includes `id` automatically). The field list in `assignment_common_fields` was hand-curated for cross-service data exchange and `id` was simply not included. The RBAC public API already exposes `id` via `ImmutableCommonModelSerializer`.

### Security

No concerns. The `id` is a monotonically increasing integer PK — no sensitive information. The service-index API is restricted to service tokens (`HasResourceRegistryPermissions`), not accessible to regular users.

## Summary by CodeRabbit

* **New Features**
  * Assignment listings now include an `id` field in API responses.
  * Team assignment listings now return the assignment identifier alongside existing fields.

* **Bug Fixes**
  * Improved support for cursor-based pagination and recovery after interruptions by ensuring assignment records can be consistently resumed.

### DIFF
--- a/ansible_base/rbac/service_api/serializers.py
+++ b/ansible_base/rbac/service_api/serializers.py
@@ -73,7 +73,10 @@ class DABPermissionSerializer(serializers.ModelSerializer):
         fields = ['api_slug', 'codename', 'content_type', 'name']
 
 
-assignment_common_fields = ('created', 'created_by_ansible_id', 'object_id', 'object_ansible_id', 'content_type', 'role_definition')
+# 'id' is included so consumers can implement cursor-based pagination
+# (e.g. id__gt=<last_pk>&order_by=id) for efficient resume after
+# crashes and skip-on-reinstall in migrate_service_data.
+assignment_common_fields = ('id', 'created', 'created_by_ansible_id', 'object_id', 'object_ansible_id', 'content_type', 'role_definition')
 
 
 class BaseAssignmentSerializer(serializers.ModelSerializer):
--- a/test_app/tests/rbac/remote/test_service_api.py
+++ b/test_app/tests/rbac/remote/test_service_api.py
@@ -4,7 +4,7 @@ import pytest
 from rest_framework.test import APIClient
 
 from ansible_base.lib.utils.response import get_relative_url
-from ansible_base.rbac.models import DABContentType, DABPermission, RoleDefinition
+from ansible_base.rbac.models import DABContentType, DABPermission, RoleDefinition, RoleTeamAssignment, RoleUserAssignment
 from test_app.models import Team, User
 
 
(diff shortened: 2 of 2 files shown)


AlexSCorey (commented):
LGTM

AlanCoding:
I'm a little bit concerned about this. Because `id` is a local construct, and this view is very highly specifically tailor-made to handle synchronization between services, where the id should not factor in at all. It should be completely irrelevant information.

Yes, you need a means of durable pagination. However for `migrate_service_data` specifically, it seems like it would be sufficient to have these ordered by id (probably should be doing this to begin with) and iterate over the pages.

Tracking the last-seen id seems like it would exist to handle deleted (revoked) assignments within the iteration. Firstly, that shouldn't happen. Secondly, it shouldn't be possible to insert (grant) assignments into a previously-seen page. So we're only talking about the deletion case, in which case, you would only have to contend with seeing duplicates which is a smaller issue to deal with and shouldn't really need contract changes, which this is doing.

AlanCoding (approved):
might be the simplest solution. It's a special case so I can see it.

PR #1033, opened 2026-06-24
[AAP-79732] fix: claim mapping speed

zkayyali812:
## Description

[AAP-79732](https://redhat.atlassian.net/browse/AAP-79732) — SAML login takes 60-120+ seconds with 339 authenticator maps, frequently exceeding gateway timeouts and preventing users from logging in.

The root cause is in `_process_user_value()`. When evaluating the `"in"` operator (used by virtually all SAML attribute maps), the code iterates every user attribute value against every map's trigger in a nested loop. For a user with 33 SAML groups and 342 maps, this produces **11,286 iterations per login** — each one constructing an f-string and emitting a DEBUG log line. In production, the per-line log I/O costs ~5ms, so the logging alone accounts for ~56 seconds.

This PR fixes the performance with three targeted changes to `_process_user_value()`:

1. **Set-based `"in"` operator** — Replaces the per-value loop with a set intersection. Both sides are converted to sets and intersected in one operation. One summary log line per map instead of one per user value.

2. **Early exit for other operators** — For `equals`, `contains`, `ends_with`, and `matches`, the loop now breaks on first match (`or` join) or first mismatch (`and` join).

3. **Queryset materialization** — `maps.count()` issued a separate SQL query before iterating. `list()` evaluates the query once, `len()` is free.

Additionally: log construction gated behind `logger.isEnabledFor(DEBUG)`, trigger values defensively casefolded in the set path, matched values logged with `sorted()` for deterministic output, dead `_evaluate_in` function removed.

## Type of Change

- [x] Test update

## Self-Review Checklist

### Prerequisites

- A running django-ansible-base test environment with PostgreSQL

### Steps to Test

1. Run the existing claims test suite to verify no regressions:
   ```
   pytest test_app/tests/authentication/utils/test_claims.py -x -v
   ```
2. Run the new performance tests specifically:
   ```
   pytest test_app/tests/authentication/utils/test_claims.py -x -v -k "Performance or Correctness or EarlyExit"
   ```
3. For production validation, deploy to a staging environment with 339+ SAML authenticator maps and verify login completes in under 20 seconds.

### Expected Results

- All existing tests pass with no changes
- New performance tests assert:
  - 342 maps x 33 groups completes in < 100ms (was 60-120s in production)
  - Debug log output is 1 line per map, not 1 per user value (342, not 11,286)
  - `create_claims` with 342 DB-backed maps completes in < 1 second
  - Map loading uses exactly 1 SQL query (no N+1)
- Early exit tests verify iteration stops at the correct value via log line counting

### Benchmarks

Measured with a counting log handler (formatting enabled, no disk I/O) simulating 342 maps x 33 user groups:

| Metric | Before | After | Improvement |
|---|---|---|---|
| `_process_user_value` wall clock | 64.2ms | 2.7ms | 24x faster |
| Debug log lines emitted | 11,286 | 342 | 97% reduction |
| `process_user_attributes` (342 calls) | 68.4ms | 5.0ms | 14x faster |
| SQL queries for map loading | 2 | 1 | -1 query |

In production where per-line log I/O is ~5ms, estimated login time drops from **60-120 seconds to under 2 seconds**.

### Behavioral notes

- **All maps are still evaluated.** The outer loop in `create_claims()` is untouched.
- **Log format changed for the `"in"` operator.** Previously: one line per user value (`Attr [groups] value [foo] is not in [['admin']], skipping`). Now: one summary line per map (`Attr [groups] has no matching values in ['admin'], skipping`).
- **No functional regression.** All existing parametrized test cases produce identical results.

### Required Actions

### Screenshots/Logs

Production gateway log before fix — a single login request generating 12,000+ log lines over 72 seconds:
```
16:07:26 Processing 342 map(s) for Authenticator ID [2]
16:07:26 Map [1] Attr [groups] value [team-alpha] is not in [['org_platform_admins']], skipping
16:07:26 Map [1] Attr [groups] value [team-beta-members] is not in [['org_platform_admins']], skipping
...   (11,284 more lines like this)
16:08:38 POST /api/gateway/social/complete/saml/ => 0 bytes in 72049 msecs (HTTP/1.1 302)
```

After fix, the same login would produce:
```
16:07:26 Processing 342 map(s) for Authenticator ID [2]
16:07:26 Map [1] Attr [groups] has no matching values in ['org_platform_admins'], skipping
16:07:26 Map [337] Attr [groups] has no matching values in ['org_deploy_test_admins'], skipping
...   (340 more lines, one per map)
16:07:26 POST /api/gateway/social/complete/saml/ => 0 bytes in ~50 msecs (HTTP/1.1 302)
```

[AAP-79732]: https://redhat.atlassian.net/browse/AAP-79732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Performance Improvements**
  * Optimized claims creation by materializing authenticator mappings once and reducing repeated query work.
  * Improved attribute evaluation efficiency for the **“in”** operator by evaluating with set-based logic to avoid per-user-value checks.

* **Bug Fixes**
  * Fixed **“in”** matching semantics, including case-insensitive behavior and correct **or/and** join handling.
  * Improved behavior for empty user values and preserved existing access outcomes for non-matching/unknown operators.

* **Tests**
  * Added log-volume, correctness, early-exit, and bounded query tests to prevent performance regressions.

### DIFF
--- a/ansible_base/authentication/utils/claims.py
+++ b/ansible_base/authentication/utils/claims.py
@@ -520,6 +520,63 @@ def _get_operator_messages(operator: str, result: bool) -> str:
     return true_msg if result else false_msg
 
 
+def _process_in_operator(
+    has_access: Optional[bool], trigger_value, user_value: List[str], join_condition: str, attribute: str, map_id: int, tracking_id: str
+) -> Optional[bool]:
+    """Fast path for 'in' operator: set intersection instead of per-value loop.
+
+    Trigger values are casefolded defensively here even though the caller
+    (_prepare_case_insensitive_data) already lowercases them, so that direct
+    callers of _process_user_value also get correct case-insensitive behavior.
+    """
+    trigger_set = {f"{v}".casefold() for v in trigger_value}
+    user_set = {f"{v}".casefold() for v in user_value}
+    matched = user_set & trigger_set
+
+    result = bool(matched) if join_condition == 'or' else user_set.issubset(trigger_set)
+    has_access = has_access_with_join(has_access, result, join_condition)
+
+    if logger.isEnabledFor(logging.DEBUG):
+        if matched:
+            _prefixed_debug(map_id, tracking_id, f"Attr [{attribute}] matched value(s) {sorted(matched)} in {trigger_value}, {_result_suffix(result)}")
+        else:
+            _prefixed_debug(map_id, tracking_id, f"Attr [{attribute}] has no matching values in {trigger_value}, {_result_suffix(result)}")
+
+    return has_access
+
+
+_OPERATOR_DISPATCH = {
+    "equals": _evaluate_equals,
+    "matches": _evaluate_matches,
+    "contains": _evaluate_contains,
+    "ends_with": _evaluate_ends_with,
+}
+
+
+def _process_scalar_operator(
+    has_access: Optional[bool], operator: str, trigger_value, user_value: List[str], join_condition: str, attribute: str, map_id: int, tracking_id: str
+) -> Optional[bool]:
+    """Per-value loop with early exit for equals/matches/contains/ends_with."""
+    evaluate_fn = _OPERATOR_DISPATCH[operator]
+
+    for a_user_value in user_value:
+        user_str = f"{a_user_value}".casefold()
+        result = evaluate_fn(user_str, trigger_value)
+        has_access = has_access_with_join(has_access, result, join_condition)
+
+        if logger.isEnabledFor(logging.DEBUG):
+            header = f"Attr [{attribute}] value [{user_str}]"
+            message = _get_operator_messages(operator, result)
+            _prefixed_debug(map_id, tracking_id, f"{header} {message} [{trigger_value}], {_result_suffix(result)}")
+
+        if result and join_condition == 'or':
+            break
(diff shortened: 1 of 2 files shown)


bhavenst (approved):
Code is correct and well-tested. The set-based `"in"` optimization (intersection + `issubset`) is semantically equivalent to the old per-value loop for all join conditions and `has_access` states. Early exit logic for other operators is sound. No silent failures introduced. Nice work on the 24x speedup.

A few optional suggestions for test hardening:

1. **`django_assert_num_queries(1)` fragility** — Consider adding a comment explaining which query is expected, so future readers know why it's 1 and not 2.
2. **Empty `user_value` with non-`"in"` operators** — The `not user_value` early return now applies to all operators but only the `"in"` path has an empty-value test. One test for e.g. `equals` with `[]` would document the behavior.
3. **`has_access=False` + match + `or` join on `"in"` path** — `test_in_preserves_existing_has_access_or` passes `True`, which stays `True` regardless. A test with `False` + matching value would verify the set logic flips it correctly.

PR #1037, opened 2026-06-26
[AAP-80761] Catch ParseError in OAuth2 extract_body

dragid10:
## Description
- Large file uploads (e.g. collection publishes) can fail with a 502 because `extract_body()` tries to parse the multipart body looking for an OAuth2 token. If the body contains base64-encoded content that is malformed or truncated, Django raises a `ParseError` that was not being caught.
- The token is always in the `Authorization` header for these requests, so catching `ParseError` and returning empty is the correct behavior.

## Type of Change

## Self-Review Checklist

### Prerequisites
- PostgreSQL running (`make postgres`)

### Steps to Test
1. `pytest test_app/tests/oauth2_provider/test_authentication.py -v`
2. All 37 tests should pass, including the two new regression tests.

### Expected Results
- `test_oauth2_extract_body_handles_parse_error` confirms `ParseError` is caught in `extract_body()`
- `test_oauth2_multipart_parse_error_full_request` confirms end-to-end OAuth2 auth succeeds when the multipart body triggers a `ParseError`

## Summary by CodeRabbit

* **Bug Fixes**
  * OAuth2 authentication now continues to work when request body parsing fails, falling back to the `Authorization` header.
  * Improved handling for requests with unsupported content types or malformed multipart data.

* **Tests**
  * Added regression coverage for parsing errors during OAuth2 body extraction.
  * Added an end-to-end check confirming valid Bearer token requests still succeed after multipart parsing failures.

### DIFF
--- a/ansible_base/oauth2_provider/authentication.py
+++ b/ansible_base/oauth2_provider/authentication.py
@@ -4,7 +4,7 @@ import logging
 from django.utils.encoding import smart_str
 from oauth2_provider.contrib.rest_framework import OAuth2Authentication
 from oauth2_provider.oauth2_backends import OAuthLibCore as _OAuthLibCore
-from rest_framework.exceptions import UnsupportedMediaType
+from rest_framework.exceptions import ParseError, UnsupportedMediaType
 
 from ansible_base.lib.logging import log_auth_event
 from ansible_base.lib.utils.hashing import hash_string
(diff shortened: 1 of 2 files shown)


bhavenst (approved):
Clean, focused fix. The `ParseError` catch is the right approach — `extract_body()` only exists to check for an OAuth2 token in the POST body, and when that parse fails, falling back to the `Authorization` header is correct behavior.

Both regression tests cover the fix well (unit + integration). CI failures (`test_load_python_file_with_injected_scope`) are pre-existing and unrelated.

dragid10:
Note: the two `in-files` CI failures (`py311-in-files`, `py312-in-files`) are pre-existing and blocked until #1034 is merged. All other CI checks pass, including all pinned-dependency test environments and downstream consumers (AWX, EDA, Hub).

PR #1043, opened 2026-07-01
Add defer_rbac_cache context manager for batched RBAC updates

pb82:
## Description

This PR defines a new annotation: `defer_rbac_cache`. This annotation allows to defer calling `compute_object_role_permissions`.

The annotation keeps track of the modified roles and teams and only calls compute_object_role_permissions in the end (when the surrounded function finishes).

**Why is this needed?** Because we have cases (migrate_service_data) that repeatedly call  update_after_assignment and this leads to repeated calls to compute_object_role_permissions with only one role each time.

**Why a new annotation needed? No simpler way?** We want to keep the original behavior of update_after_assignment and make the new behavior optional. Otherwise every user of functions like `give_permission` (which call update_after_assignment) would have to keep track of the modified objects and call compute_object_role_permissions manually.

**Observed performance increase**

In an environment with 1k Organizations and 10k Teams, `give_permissions` was called 2000 times to simulate the behavior of migrate_service_data.

With the new annotation it took ~30 seconds, without it ~80 seconds. Speedup: 2.6 times. This should further increase with more object roles in the database.

Alternatively, a full migration could be performed.

### Steps to Test

See above, or run a full migration in an environment with a large number of object roles.

### Expected Results

The migration should complete faster.

## Summary by CodeRabbit

* **New Features**
  * Added a deferred RBAC cache update flow, allowing multiple permission/role changes to be batched and applied once when updates complete.
  * Supports combining multiple object permission changes in a single batch, with correct results after the batch ends.
* **Bug Fixes**
  * Ensured RBAC cache recomputation and permission evaluations occur correctly after deferred batches, including nested batches.
  * Verified that behavior remains immediate when batching is not used.
* **Tests**
  * Expanded RBAC trigger test coverage for deferred behavior, including multi-change, nested, and no-op scenarios.

### DIFF
--- a/ansible_base/rbac/triggers.py
+++ b/ansible_base/rbac/triggers.py
@@ -1,4 +1,6 @@
 import logging
+import threading
+from contextlib import contextmanager
 from typing import Union
 from uuid import UUID
 
@@ -108,8 +110,49 @@ def needed_updates_on_assignment(role_definition, actor, object_role, created=Fa
     return (recompute_team_ids, to_update)
 
 
+# stores state for the defer_rbac_cache annotation so that it can be accessed by function in the call chain
+# of the annotation
+class _DeferRBACCache(threading.local):
+    def __init__(self):
+        self.active = False
+        self.team_ids = set()
+        self.object_roles = set()
+
+
+_defer_rbac_cache = _DeferRBACCache()
+
+
+# allows deferring the rbac computation in cases where many object roles are updated in short order
+@contextmanager
+def defer_rbac_cache():
+    if _defer_rbac_cache.active:
+        raise RuntimeError("defer_rbac_cache cannot be nested")
+    _defer_rbac_cache.active = True
+    try:
+        yield
+    finally:
+        team_ids = _defer_rbac_cache.team_ids
+        object_roles = _defer_rbac_cache.object_roles
+        _defer_rbac_cache.active = False
+        _defer_rbac_cache.team_ids = set()
+        _defer_rbac_cache.object_roles = set()
+
+        if team_ids:
+            compute_team_member_roles(team_ids=team_ids)
+
+        if object_roles:
+            compute_object_role_permissions(object_roles=object_roles)
+
+
 def update_after_assignment(recompute_team_ids, to_update):
     "Call this with the output of needed_updates_on_assignment"
+    if _defer_rbac_cache.active:
+        if recompute_team_ids is not None:
+            _defer_rbac_cache.team_ids.update(recompute_team_ids)
+        if to_update is not None:
+            _defer_rbac_cache.object_roles.update(to_update)
+        return
+
     if recompute_team_ids is not None:
         compute_team_member_roles(team_ids=recompute_team_ids)
 
--- a/test_app/tests/rbac/test_triggers.py
+++ b/test_app/tests/rbac/test_triggers.py
@@ -242,6 +242,16 @@ def test_defer_rbac_cache_without_context_manager(organization, inventory, rando
     assert rando.has_obj_perm(inventory, 'change')
 
 
+@pytest.mark.django_db
+def test_defer_rbac_cache_with_team_assignment(organization, team, rando, org_team_member_rd):
+    """defer_rbac_cache should also defer and flush team membership
+    recomputation (the team_ids path)."""
+    with defer_rbac_cache():
+        org_team_member_rd.give_permission(rando, organization)
+
+    assert rando.has_obj_perm(team, 'member_team')
+
+
 class TestEmailPolicySignal:
     """Tests for the pre_save signal that prevents unauthorized email
     changes across all services."""
--- a/test_app/tests/rbac/test_triggers.py
+++ b/test_app/tests/rbac/test_triggers.py
@@ -7,7 +7,7 @@ from rest_framework.exceptions import ValidationError
 
 from ansible_base.rbac.models import ObjectRole, RoleDefinition, RoleEvaluation, RoleTeamAssignment, RoleUserAssignment
 from ansible_base.rbac.permission_registry import permission_registry
-from ansible_base.rbac.triggers import dab_post_migrate, post_migration_rbac_setup
+from ansible_base.rbac.triggers import dab_post_migrate, defer_rbac_cache, post_migration_rbac_setup
 from test_app.models import Inventory, Organization, User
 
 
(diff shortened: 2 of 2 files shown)


djyasin:
1. compute_team_member_roles does not accept a team_ids parameter — this will crash at runtime

  The context manager calls compute_team_member_roles(team_ids=team_ids) on line 143 of the diff, but the actual function signature at
  ansible_base/rbac/caching.py:176 is:

`  def compute_team_member_roles():`

  It accepts no arguments. This will raise a TypeError at runtime whenever a deferred block includes team-related role changes. Either the PR is targeting a
  branch that has a different version of this function, or this is a bug.

  2. Semantic mismatch: recompute_team_ids is a bool, not a set of IDs

  In needed_updates_on_assignment (line 80 of the existing code):

`  recompute_teams = bool(has_team_perm and (created or deleted or changes_team_owners))`

  The first element of the return tuple is a boolean, not a set of team IDs. But defer_rbac_cache treats it as an iterable of IDs:

```
  if recompute_team_ids is not None:
      _defer_rbac_cache.team_ids.update(recompute_team_ids)

```
  Calling set.update(True) will iterate over a bool — which raises TypeError: 'bool' object is not iterable. The existing non-deferred code just uses it as a
  truthiness check (if update_teams: compute_team_member_roles()), but the deferred path incorrectly treats it as data.

ansible_base/rbac/triggers.py
```diff
@@ -108,8 +110,51 @@ def needed_updates_on_assignment(role_definition, actor, object_role, created=Fa
...
+# Derives from threading.local to ensure that each instance is a thread-independent storage and calls to
+# update_after_assignment from other threads don't affect the context manager.
+class _DeferRBACCache(threading.local):
+    def __init__(self):
+        self.active = False
+        self.team_ids = set()
+        self.object_roles = set()
+
+
+_defer_rbac_cache = _DeferRBACCache()
```
AlanCoding, line 122: I know this is a pattern borrowed from other places, but that doesn't mean it's good.

I would suggest that this line should actually be moved to inside of `defer_rbac_cache`. So we create `_defer_rbac_cache` inside that method, and it's a local there instead of a global here.

Because if you muck with a global, then you have to track the prior state of the variable(s). This is not done. This creates a bug if you ever nest the context manager, where the outer context manager will completely "forget" what variables it was supposed to have. In this function, that does mean that it appends & updates to the rolling data structures. But then that is still probably a bug, where things assigned after the inner context manager, but inside the outer context manager, do not update in the intended batch.

For this, I think we can/should be more a stickler about the exact context manager scope. We don't need to think about nested context managers if the variable is initialized _in_ the context manager.

ansible_base/rbac/triggers.py
```diff
@@ -108,8 +110,51 @@ def needed_updates_on_assignment(role_definition, actor, object_role, created=Fa
...
+        team_ids = _defer_rbac_cache.team_ids
+        object_roles = _defer_rbac_cache.object_roles
+        _defer_rbac_cache.active = False
+        _defer_rbac_cache.team_ids = set()
+        _defer_rbac_cache.object_roles = set()
+
+        if team_ids:
+            compute_team_member_roles(team_ids=team_ids)
+
+        compute_object_role_permissions(object_roles=object_roles if object_roles else None)
```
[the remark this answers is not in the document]
AlanCoding: > The boolean active also lets an inner defer_rbac_cache() flush/reset the outer context early.

Kind of same comment I left.

> Line 144 conflates “no object-role updates” with “global recompute”, while Line 154 drops the existing to_update=None global-recompute contract once any scoped roles are collected.

Hmm... it's probably true that this triggers a global re-compute, whereas the prior signals behavior would have been no-op. That's a pretty big difference... maybe valid.

djulich:
PR title could be changed to "Add defer_rbac_cache context manager for batched RBAC updates" or similar.

djulich:
The core optimization concept is solid — batching compute_object_role_permissions calls during bulk operations like migrate_service_data is the right approach, and the 2.6x speedup is meaningful.

There are two substantive issues that need to be addressed before merge, both already well-identified in existing review comments:

1. Nesting safety

The active boolean flag means nested defer_rbac_cache() calls will cause the inner exit to flush and reset state prematurely, breaking the outer context. This was flagged independently by both @AlanCoding and CodeRabbit. A depth counter that only flushes at depth 0 is the right fix. Note that the thread-local global is still needed — update_after_assignment has to check deferral state without it being passed explicitly — so moving everything into a local variable (as @AlanCoding suggested) wouldn't fully work here, but the nesting problem it identifies is real.

2. Empty set → global recompute escalation

The finally block does object_roles=object_roles if object_roles else None, which means an empty deferred set (no scoped updates queued) passes None to compute_object_role_permissions. Since None means "recompute ALL object roles" (caching.py:292-293), this silently escalates a no-op into a full global recompute — the opposite of the performance goal. Flagged by CodeRabbit, validated by @AlanCoding.

Additional notes

- Tests should be added for empty deferred blocks and nested usage (per CodeRabbit's nitpick).
- @djyasin's two comments (re: compute_team_member_roles signature and recompute_team_ids type) appear to be based on an older version of the code and do not apply to the current devel branch — compute_team_member_roles does accept team_ids (caching.py:197), and recompute_team_ids is a set or None, not a bool (triggers.py:98-108).

ansible_base/rbac/triggers.py
pb82: @AlanCoding i've added nesting protection. Regarding the global variable, unfortunately I don't see a better option in this case. The problem is that a function called by the annotated function needs to access the state of the context manager.

djyasin (approved):

ansible_base/rbac/triggers.py
AlanCoding: Acknowledged. You are right. We can't get rid of the global.

However, I am going still going to be pig-headed about the fact that we can not possibly coherently manage multiple layers of nested context here. This is a new context manager. It is better to not support something than to do it badly with unclear expectations. So this is what I would suggest.

```diff
@@ -122,33 +120,31 @@ class _DeferRBACCache(threading.local):
 _defer_rbac_cache = _DeferRBACCache()
 
 
-# allows deferring the rbac computation in cases where many object roles are updated in short order
 @contextmanager
 def defer_rbac_cache():
+    if _defer_rbac_cache.depth > 0:
+        raise RuntimeError("Nested use of defer_rbac_cache is not supported")
     _defer_rbac_cache.depth += 1
     try:
         yield
     finally:
         _defer_rbac_cache.depth -= 1
```

This results in deleting a _lot_ of other code you have, because much of that is concerned with the nesting management. In this world-view, there is either 1 cache store, or there is 0.

The problem with nesting is that you can't coherently set expectations for what gets re-computed on the exit from which context manager, and I am convinced there are bugs in there.

pb82: That makes sense and simplifies the code. I'll update the PR.

AlanCoding (approved):

PR #1045, opened 2026-07-02
[AAP-75091] Make related_fields() namespace-aware for URL resolution

lucasc017:
## Summary

- Fix `related_fields()` in `AbstractCommonModel` to retry URL resolution with the request's namespace prefix when bare `reverse('{basename}-detail')` fails with `NoReverseMatch`
- Fixes missing `user`, `role_definition`, and `created_by` fields in the `related` dict of the `role_user_assignments` API when DAB views are registered under Django URL namespaces (e.g. galaxy_ng's `namespace="ui_v2"`)
- Add 4 tests covering namespace retry, no-request, no-namespace, and bare-preferred scenarios

## Problem

When a consuming app registers DAB views under a Django URL namespace — e.g. `path("_ui/v2/", include(urls, namespace="ui_v2"))` — the bare `reverse('user-detail')` call in `related_fields()` fails with `NoReverseMatch` because the actual URL name is `galaxy:ui_v2:user-detail`. The exception is silently caught and the field is dropped from the response.

This causes the `_ui/v2/role_user_assignments/` API to return:
```json
{"related": {"content_object": "/api/galaxy/v3/namespaces/..."}}
```
instead of:
```json
{"related": {"content_object": "...", "user": "...", "role_definition": "...", "created_by": "..."}}
```

`content_object` works because Pulp models define `get_absolute_url()` which bypasses the `reverse()` path entirely.

## Fix

Extract `request.resolver_match.namespace` at the top of `related_fields()` and use it as a fallback when the bare reverse fails:

1. Try `reverse('{basename}-detail')` (bare, existing behavior)
2. On `NoReverseMatch`, if a namespace is available, retry with `reverse('{namespace}:{basename}-detail')`

Applied to both the FK loop and the reverse-relations loop.

## Impact

- **AWX/Controller**: No impact — AWX includes DAB URLs without namespaces, so bare reverse always succeeds and the retry is never triggered
- **galaxy_ng**: Fixes the bug — the namespace retry correctly resolves `galaxy:ui_v2:user-detail` etc.
- **Other consumers**: No impact for non-namespaced deployments; beneficial for any future namespaced deployments

## Test plan

- [x] `test_related_fields_namespace_retry` — namespace fallback works when bare reverse fails
- [x] `test_related_fields_no_namespace_retry_without_request` — existing behavior preserved when request is None
- [x] `test_related_fields_no_namespace_retry_without_namespace` — existing behavior preserved when namespace is empty
- [x] `test_related_fields_bare_reverse_preferred_over_namespace` — bare URLs used when they resolve
- [x] Full `test_common.py` suite (35 tests) passes
- [x] Serializer `test_common.py` suite (17 tests) passes — no regressions
- [x] flake8, black, isort clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved related link URL resolution when a page is using a view namespace.
  * Added a retry that uses the namespaced route when the initial (non-namespaced) lookup fails.
  * Avoided prematurely treating reverse links as missing when a namespaced match is available.
* **Tests**
  * Added/extended test coverage for namespace-aware URL resolution, including scenarios with no request, an empty namespace, and successful direct lookups.

### DIFF
--- a/ansible_base/lib/abstract_models/common.py
+++ b/ansible_base/lib/abstract_models/common.py
@@ -222,6 +222,11 @@ class AbstractCommonModel(models.Model):
 
     def related_fields(self, request):
         response = {}
+
+        namespace = None
+        if request and hasattr(request, 'resolver_match') and request.resolver_match:
+            namespace = request.resolver_match.namespace
+
         # See docs/lib/default_models.md
         # Automatically add all of the ForeignKeys for the model as related fields
         for field in self._meta.concrete_fields:
@@ -237,10 +242,15 @@ class AbstractCommonModel(models.Model):
             if fk_id is not None:
                 related_model = field.related_model
                 basename = get_cls_view_basename(related_model)
+                view_name = f'{basename}-detail'
                 try:
-                    response[field.name] = get_relative_url(f'{basename}-detail', kwargs={'pk': fk_id})
+                    response[field.name] = get_relative_url(view_name, kwargs={'pk': fk_id})
                 except NoReverseMatch:
-                    pass
+                    if namespace:
+                        try:
+                            response[field.name] = get_relative_url(f'{namespace}:{view_name}', kwargs={'pk': fk_id})
+                        except NoReverseMatch:
+                            pass
 
         basename = get_cls_view_basename(self.__class__)
 
@@ -255,7 +265,13 @@ class AbstractCommonModel(models.Model):
             try:
                 response[field_name] = get_relative_url(reverse_view, kwargs={'pk': self.pk})
             except NoReverseMatch:
-                missing_relations.append(reverse_view)
+                if namespace:
+                    try:
+                        response[field_name] = get_relative_url(f'{namespace}:{reverse_view}', kwargs={'pk': self.pk})
+                    except NoReverseMatch:
+                        missing_relations.append(reverse_view)
+                else:
+                    missing_relations.append(reverse_view)
 
         if missing_relations and settings.DEBUG:
             logger.error(f"Wanted to add {', '.join(missing_relations)} for {self.__class__} but view was missing")
--- a/test_app/tests/lib/abstract_models/test_common.py
+++ b/test_app/tests/lib/abstract_models/test_common.py
@@ -110,7 +110,7 @@ def test_related_fields_view_resolution(shut_up_logging):
 
 
 @pytest.mark.django_db
-def test_related_fields_namespace_retry(shut_up_logging):
+def test_related_fields_namespace_retry(shut_up_logging, system_user):
     """When bare reverse fails but namespaced reverse succeeds, the field should be included."""
     from types import SimpleNamespace
 
@@ -129,6 +129,7 @@ def test_related_fields_namespace_retry(shut_up_logging):
         result = model.related_fields(request)
 
     assert len(result) > 0, "Namespaced retry should have resolved at least one URL"
+    assert 'created_by' in result, "FK namespace retry should have resolved created_by"
 
 
 @pytest.mark.django_db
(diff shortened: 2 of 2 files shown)


ansible_base/lib/abstract_models/common.py
```diff
@@ -237,10 +242,15 @@ def related_fields(self, request):
...
                 basename = get_cls_view_basename(related_model)
+                view_name = f'{basename}-detail'
                 try:
-                    response[field.name] = get_relative_url(f'{basename}-detail', kwargs={'pk': fk_id})
+                    response[field.name] = get_relative_url(view_name, kwargs={'pk': fk_id})
                 except NoReverseMatch:
-                    pass
+                    if namespace:
+                        try:
+                            response[field.name] = get_relative_url(f'{namespace}:{view_name}', kwargs={'pk': fk_id})
```
fosterseth, line 251: if namespace is defined should we just try this to begin with? or are there instances we still should do the first "try"

get_relative_url is lightweight so I think performance wise what we have here is fine either way

test_app/tests/lib/abstract_models/test_common.py
```diff
@@ -109,6 +109,102 @@ def test_related_fields_view_resolution(shut_up_logging):
     assert 'related_fields_test_model-more_teams-list' in [call[0][0] for call in reverse.call_args_list]
 
 
+@pytest.mark.django_db
+def test_related_fields_namespace_retry(shut_up_logging, system_user):
+    """When bare reverse fails but namespaced reverse succeeds, the field should be included."""
+    from types import SimpleNamespace
```
fosterseth: these imports probably should be at top of file, no risk of circular imports

fosterseth (approved):
minor points, but looks good

ansible_base/lib/abstract_models/common.py
lucasc017: Good question — we keep bare-first because the view could be registered at multiple URL paths (some namespaced, some not). If bare reverse succeeds, that URL is canonical and avoids baking in a specific namespace prefix. The test `test_related_fields_bare_reverse_preferred_over_namespace` validates this: when both work, bare wins.

test_app/tests/lib/abstract_models/test_common.py
lucasc017: Good catch — moved `SimpleNamespace` and `NoReverseMatch` to the module top level in b13f562.

ansible_base/lib/abstract_models/common.py
lucasc017: Keeping as-is — bare-first is the safer default since it matches the existing behavior for non-namespaced consumers like AWX.

fosterseth, line 251: also it appears `reverse` (which get_relative_url calls) can take in a `current_app` keyword argument.

`get_relative_url(f'{basename}-detail', current_app=current_app)`

would doing that allow us to just have the single try: except without the nested on? might be worth looking into

so current_app = None -> bare
current_app = namespace from request -> namespaced url

lucasc017: Looked into it — `current_app` only disambiguates between multiple *instances* of the same application namespace (e.g. two `include('polls.urls')` under different instance names). It doesn't help when the view name itself is bare.

With a bare name like `reverse('user-detail', current_app='galaxy:ui_v2')`, Django still searches for `user-detail` in the root URL conf and ignores `current_app` since there's no namespace prefix in the view name to trigger the namespace resolution loop. The namespace must be in the view name: `reverse('galaxy:ui_v2:user-detail')`.

So the nested try/except is the correct approach here.

PR #1035, opened 2026-06-26
[AAP-78675] feat: add service-scoped filtering to role assignment endpoints

jay-steurer:
## What

Add `content_type__service` filter support to the service-index role assignment
endpoints (`ServiceRoleUserAssignmentViewSet` and `ServiceRoleTeamAssignmentViewSet`)
so that each service can request only its own role assignments. Global assignments
(where `content_type` is NULL, e.g. Platform Auditor) are always included in
filtered responses.

## Why

With 3 services each fetching 86K assignments every 15 minutes, gateway serves
258K assignments per sync cycle. With service-scoped filtering, each service fetches
only its own assignments, reducing API load by ~67%. This also reduces network
transfer and deserialization overhead on both gateway and services.

Jira: [AAP-78675](https://redhat.atlassian.net/browse/AAP-78675)

## Intent Adherence: 90/100

- Solves the problem described in the story: 40/40
- No deviations from stated intent: 20/20
- No over-engineering or scope creep: 15/20
  (-5) Added rest_filters_reserved_names coupling between custom filter and FieldLookupBackend
- A reviewer would agree the code matches the story: 15/20
  (-5) AC #8 (scale perf) is architectural, AC #9 (backport) is post-merge

## Acceptance Criteria

1. [PASS] Endpoints accept content_type__service filter parameter
2. [PASS] Filtered results only contain matching service assignments
3. [PASS] Global assignments (content_type=None) included in all filtered responses
4. [PASS] No filter = all assignments (backward compatible)
5. [PASS] RemoteAssignmentFetcher passes service filter when configured
6. [PASS] resource_sync continues to work (104 tests pass)
7. [PASS] Unit tests cover filtered response, global inclusion, backward compat
8. [N/A] Performance measurable at scale (architectural improvement)
9. [N/A] Backport to 2.6/2.7 (post-merge)

## Changeset

| File | Action | Description |
|------|--------|-------------|
| ansible_base/rest_filters/.../ansible_id_backend.py | Modified | New ServiceFilterBackend class |
| ansible_base/rbac/service_api/views.py | Modified | Register filter, reserve param name |
| ansible_base/resource_registry/tasks/sync.py | Modified | Add service_filter to fetcher/executor |
| test_app/tests/rbac/remote/test_service_api.py | Modified | 7 new tests in TestServiceFilter |
| test_app/tests/resource_registry/test_resource_sync.py | Modified | 1 new test + updated assertion |

## Test plan

- [x] Unit tests added (8 new tests)
[...]

PR #1046, opened 2026-07-02
Revert 1042 1038 and 1025

AlexSCorey:
Revert #1042 and #1038 and #1025

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved service listing pagination to traverse results using descending IDs for more consistent ordering.
  * Updated RBAC pagination behavior and related responses to rely on `count` for accurate filtering totals.
  * Improved remote resource assignment synchronization by switching from cursor-based pagination to explicit page-based retrieval.
  * Simplified `object_ansible_id` input handling to use standard UUID field parsing/validation.
* **Documentation**
  * Updated the resource registry REST response example to include the `count` field.

### DIFF
--- a/ansible_base/rbac/api/serializers.py
+++ b/ansible_base/rbac/api/serializers.py
@@ -16,7 +16,6 @@ from ansible_base.lib.utils.response import get_relative_url
 from ansible_base.rbac.models import RoleDefinition, RoleTeamAssignment, RoleUserAssignment
 from ansible_base.rbac.permission_registry import permission_registry  # careful for circular imports
 from ansible_base.rbac.policies import check_content_obj_permission, visible_users
-from ansible_base.rbac.service_api.serializers import ObjectAnsibleIdField
 from ansible_base.rbac.validators import check_locally_managed, validate_permissions_for_model
 
 from ..models import DABContentType, DABPermission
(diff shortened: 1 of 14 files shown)


john-westcott-iv (changes requested):
The revert of _paginate and removal of _build_filters drops the service_filter support that was added by PR #1035 (AAP-78675). That PR used _build_filters() to pass content_type__service when service_filter is set, but since that helper is removed by this revert, the remote fetch in _paginate now silently ignores service_filter — it only passes {'page': page, 'page_size': self.page_size}.

The local side still filters by service (get_local_assignments(service=self.service_filter)), so a sync with service_filter set would compare a service-scoped local set against an unscoped remote set, producing incorrect diffs.

Since #1035 landed after #1025, this can't be a direct revert — _paginate needs to include the service_filter logic. Something like:

```python
def _paginate(self, list_fn, actor_id_key, assignment_type):
    page = 1
    try:
        while True:
            filters = {'page': page, 'page_size': self.page_size}
            if self.service_filter:
                filters['content_type__service'] = self.service_filter
            resp = list_fn(filters=filters)
            ...
```

Everything else in the revert looks correct and CI is green (SonarCloud failure is a false positive on coverage for removed tests+code).

john-westcott-iv (approved):
Looks good. The service_filter issue from the previous review is addressed — _paginate now applies content_type__service when service_filter is set, with new test coverage for both the set and None cases. CI is fully green.

PR #1040, opened 2026-07-01
[AAP-80508] feat: Enhance OIDC discovery with revocation and PKCE metadata

ldjebran:
## Description
Add DABConnectDiscoveryInfoView extending DOT's ConnectDiscoveryInfoView to include revocation_endpoint and code_challenge_methods_supported in the OIDC discovery response. These fields are not provided by django-oauth-toolkit 2.3.0 but are needed by MCP/OAuth2.1 clients.

Add WellKnownNotFoundView to return JSON 404 for unhandled /.well-known/ paths instead of 200 and falling through to the login page.

issue: https://issues.redhat.com/browse/AAP-80741
issue: https://issues.redhat.com/browse/AAP-80508

Related: https://github.com/ansible/jewel/pull/144

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test update
- [ ] Refactoring (no functional changes)
- [ ] Development environment change
- [ ] Configuration change

## Self-Review Checklist

- [X] I have performed a self-review of my code
- [X] I have added relevant comments to complex code sections
- [ ] I have updated documentation where needed
- [ ] I have considered the security impact of these changes
- [ ] I have considered performance implications
- [ ] I have thought about error handling and edge cases
- [X] I have tested the changes in my local environment

# Tests

 1. Verifies the OIDC discovery document includes a revocation_endpoint pointing to the token revocation URL
  2. Verifies the discovery document advertises PKCE support with code_challenge_methods_supported (S256 and plain)
  3. Verifies that unhandled /.well-known/ paths return a JSON 404 instead of falling through to the login page
  4. Verifies that when the upstream view returns a non-200 response (e.g. OIDC disabled), our view passes it through unchanged without crashing on JSON
  parsing

# Testing with gateway

## Test 1
1. start gateway
2. in terminal run
curl -k -v https://localhost/.well-known/oauth-authorization-server/o

Expected result:
1. should receive http status 404 with body {"error": "not_found"}

## Test 2
1. in terminal run:
curl -k https://localhost/o/.well-known/openid-configuration/

Expected result:
1. json content with included fields:
    "revocation_endpoint": "https://aap.foo.redhat.com/o/revoke_token/"
    "code_challenge_methods_supported": ["S256", "plain"] 

### Required Actions

- [ ] Requires documentation updates
  
[...]

PR #1047, opened 2026-07-06
[AAP-82091] Add Allow header to NotFoundView .well-known catch-all

john-westcott-iv:
## Summary
- Adds `Allow` header to `NotFoundView` response so OPTIONS requests return proper HTTP headers

## Problem
The `NotFoundView` added in PR #1040 for the `.well-known/` catch-all is a plain Django `View` that returns `JsonResponse({"error": "not_found"}, status=404)` without an `Allow` header. This breaks jewel's `test_docs.py` API schema tests which send OPTIONS to every discovered URL pattern and expect an `Allow` header.

## Fix
Add `response["Allow"] = "GET, HEAD, OPTIONS"` to the response. This keeps the view as a plain Django View (not DRF), so it stays out of the OpenAPI schema — which is the correct behavior for a catch-all 404 route.

## Test plan
- [ ] `GATEWAY_TEST_DIRS="" tox -e py312 -- aap_gateway_api/tests/views/api/v1/docs/test_docs.py -v` passes in jewel

## References
- JIRA: https://issues.redhat.com/browse/AAP-82091
- Introduced by: #1040
- Failing CI: https://github.com/ansible/jewel/actions/runs/28816255495

## Summary by CodeRabbit

* **Bug Fixes**
  * Unhandled paths now return a proper `Allow` header alongside the JSON 404 response, improving HTTP compliance and making supported methods clearer to clients.

### DIFF
--- a/ansible_base/oauth2_provider/views/not_found.py
+++ b/ansible_base/oauth2_provider/views/not_found.py
@@ -6,4 +6,6 @@ class NotFoundView(View):
     """Return a JSON 404 for unhandled paths."""
 
     def dispatch(self, request, *args, **kwargs):
-        return JsonResponse({"error": "not_found"}, status=404)
+        response = JsonResponse({"error": "not_found"}, status=404)
+        response["Allow"] = "GET, HEAD, OPTIONS"
+        return response


ldjebran (approved):
LGTM

AlanCoding (approved):

PR #1049, opened 2026-07-13
Fix role_team_access 500: Team model lacks encrypted_fields

djulich:
## Summary

- `TeamAccessListMixin` inherited from `AbstractCommonModelSerializer`, whose `to_representation()` accesses `obj.encrypted_fields` unconditionally
- The Team model (in AWX and other consumers) does not define `encrypted_fields`, causing `AttributeError: 'Team' object has no attribute 'encrypted_fields'` on every `role_team_access` request
- Switch to `serializers.ModelSerializer`, matching the pattern already used by `UserAccessListMixin` for the same reason — the actor model is not a DAB `AbstractCommonModel`

## Test plan

- [x] All 6 existing access list tests pass (`test_app/tests/rbac/api/test_access_lists.py`)
- [ ] Verify `role_team_access/<model_name>/<pk>/` returns 200 instead of 500 on a running controller
- [ ] Scale Lab validation: `ta JT org_admin` endpoint should go from 100% fail to passing

Closes: AAP-82540

### DIFF
--- a/ansible_base/rbac/api/serializers.py
+++ b/ansible_base/rbac/api/serializers.py
@@ -10,7 +10,7 @@ from rest_framework.exceptions import PermissionDenied
 from rest_framework.serializers import ValidationError
 
 from ansible_base.lib.abstract_models.common import get_url_for_object
-from ansible_base.lib.serializers.common import AbstractCommonModelSerializer, CommonModelSerializer, ImmutableCommonModelSerializer
+from ansible_base.lib.serializers.common import CommonModelSerializer, ImmutableCommonModelSerializer
 from ansible_base.lib.utils.auth import get_team_model
 from ansible_base.lib.utils.response import get_relative_url
 from ansible_base.rbac.models import RoleDefinition, RoleTeamAssignment, RoleUserAssignment
@@ -331,7 +331,7 @@ class UserAccessListMixin(AccessListMixin, serializers.ModelSerializer):
     _expected_fields = ['id', 'url', 'related', 'username', 'is_superuser', 'first_name', 'last_name', 'object_role_assignments']
 
 
-class TeamAccessListMixin(AccessListMixin, AbstractCommonModelSerializer):
+class TeamAccessListMixin(AccessListMixin, serializers.ModelSerializer):
     object_role_assignments = serializers.SerializerMethodField()
     url = serializers.SerializerMethodField()
     related = serializers.SerializerMethodField('_get_related')


AlanCoding (changes requested):
I think this is a stealth breaking API change

djulich:
> I think this is a stealth breaking API change

Hi @AlanCoding, I guess you're meaning that before the PR change, the `role_team_access` response included a `summary_fields` key (inherited from `AbstractCommonModelSerializer`). After our change, that field disappears. Any client parsing `summary_fields` from this endpoint would break.

The irony is that the endpoint was broken anyway (500 error), so no client was successfully consuming it. But your point is still valid from a design perspective - once the 500 error is fixed, the response shape should match what was intended, not silently drop fields. The question is: what is intended here?

The clean fix would be to either add `summary_fields` as an explicit SerializerMethodField on TeamAccessListMixin (if it should be there), or confirm it was never part of the contract (since `_expected_fields` doesn't include it and the endpoint never worked).

djulich:
**Re: "stealth breaking API change" concern**

I diffed the actual resolved serializer fields before and after the change using test_app (as @AlanCoding suggested). The response schema is identical - no fields are added or removed.

Before (with `AbstractCommonModelSerializer`):
```
['id', 'name', 'object_role_assignments', 'organization', 'related', 'url']
```

After (with `serializers.ModelSerializer`):
```
['id', 'name', 'object_role_assignments', 'organization', 'related', 'url']
```

The reason: `AbstractCommonModelSerializer` declares `summary_fields` as a `SerializerMethodField`, but the dynamically-created `DynamicActorSerializer` gets its `Meta.fields` from `TeamAccessListMixin._expected_fields`, which is `['id', 'url', 'related', 'name', 'organization', 'object_role_assignments']`. Since DRF treats an explicit `Meta.fields` list as an allowlist, `summary_fields` was always filtered out and never appeared in the API response — even on Gateway and EDA where the endpoint doesn't crash.

What `AbstractCommonModelSerializer` does contribute beyond the field declarations:

1. `to_representation()` override — iterates `obj.encrypted_fields` to mask values with `$encrypted$`. This is exactly what causes the crash on Controller (AWX's `Team` model doesn't have `encrypted_fields`). On Gateway/EDA it runs but is a no-op since `Team` has `encrypted_fields = []` (no fields to mask). Removing it has zero effect on the response.
2. `ValidationSerializerMixin` — adds validation logic in the `__init__` and form default population. This only matters for write operations. The `role_team_access` endpoint is read-only (it lists teams with access to a resource), so this has no effect.
3. `_get_summary_fields()` method — declared but never called because `summary_fields` is excluded by `Meta.fields`.
4. `_get_related()` method — overridden by `AccessListMixin._get_related()` which takes precedence in the MRO, so the `AbstractCommonModelSerializer` version was never called.
5. `get_url()` method — also overridden by `AccessListMixin.get_url()`.

In short: every behavior contributed by `AbstractCommonModelSerializer` was either (a) already overridden by `AccessListMixin`, (b) filtered out by `Meta.fields`, or (c) the cause of the crash. The change is safe across all consumers.

This matches the design choice already made for `UserAccessListMixin`, which uses plain `ModelSerializer` with the comment: "controller uses auth.User model so this needs to be as compatible as possible, thus ModelSerializer". The same reasoning applies to `Team` — it's an actor model owned by the consuming application, not a DAB `AbstractCommonModel`.

The test verification was done by instantiating the serializer through `TeamAccessViewSet().get_serializer_class()` (which goes through the same `DynamicActorSerializer` factory the view uses at runtime) and comparing `.fields.keys()` before and after.

AlanCoding (approved):

fosterseth (approved):

PR #1051, opened 2026-07-14
Add per-application PKCE enforcement for OAuth2

mabashian:
## Description
Adds a pkce_required boolean field to OAuth2Application enabling per-application PKCE enforcement at the authorize endpoint. New apps default to pkce_required=True; existing apps receive False via migration to preserve current behavior. The global PKCE_REQUIRED setting acts as a floor — when True, PKCE is enforced regardless of the per-app setting.

This is a requirement for ANSTRAT-2250 (MCP Server OAuth 2.1 Integration) to move forward.

## Type of Change

- [ ] Documentation update
- [x] Test update
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites
A running AAP instance with the gateway accessible.
A client compatible with OAuth 2.1 or a script that simulates a client

### Steps to Test
1. Create an OAuth2 application with pkce_required=true via the API
2. Hit /o/authorize/ without code_challenge — should get 400
3. Hit /o/authorize/ with code_challenge and code_challenge_method=S256 — should get 200
4. Create an application with pkce_required=false, repeat steps 2-3 — both should succeed

### Expected Results
PKCE is enforced only for apps with pkce_required=true or when global PKCE_REQUIRED=True. Existing apps are unaffected. All tests pass.

### Required Actions

- [x] Requires documentation updates
  The new application-level field will need to be documented.  Documentation on the global pkce_enabled setting will also need to be updated
  
  UI and gateway
- [ ] Blocked by PR/MR: #XXX
  

## Summary by CodeRabbit

* **New Features**
  * Added a `pkce_required` option to OAuth2 applications (defaulting to enabled) to require `code_challenge` for authorization-code requests.
  * Authorization endpoint now enforces PKCE on both GET and POST; missing or empty `code_challenge` is rejected with `invalid_request` while preserving `state`.
  * Added support for global PKCE enforcement that applies to all applications.
* **Bug Fixes**
  * Updated authorization endpoint handling to ensure the PKCE-enforcing implementation is used.
* **Tests**
  * Added coverage for application create/update defaults, PKCE-required vs not-required behavior, and global override behavior.

### DIFF
--- a/ansible_base/oauth2_provider/migrations/0010_oauth2application_pkce_required.py
+++ b/ansible_base/oauth2_provider/migrations/0010_oauth2application_pkce_required.py
@@ -7,6 +7,8 @@ class Migration(migrations.Migration):
         ('dab_oauth2_provider', '0009_add_openid_roles_scopes'),
     ]
 
+    # Two-step operation: AddField(default=False) backfills existing rows with False,
+    # then AlterField(default=True) sets the default for new records only.
     operations = [
         migrations.AddField(
             model_name='oauth2application',
--- a/ansible_base/oauth2_provider/migrations/0010_oauth2application_pkce_required.py
+++ b/ansible_base/oauth2_provider/migrations/0010_oauth2application_pkce_required.py
@@ -1,0 +1,21 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dab_oauth2_provider', '0009_add_openid_roles_scopes'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='oauth2application',
+            name='pkce_required',
+            field=models.BooleanField(default=False, help_text='When True, clients must use PKCE (send code_challenge) when requesting authorization codes for this application.'),
+        ),
+        migrations.AlterField(
+            model_name='oauth2application',
+            name='pkce_required',
+            field=models.BooleanField(default=True, help_text='When True, clients must use PKCE (send code_challenge) when requesting authorization codes for this application.'),
+        ),
+    ]
--- a/ansible_base/oauth2_provider/migrations/0010_oauth2application_pkce_required.py
+++ b/ansible_base/oauth2_provider/migrations/0010_oauth2application_pkce_required.py
@@ -16,6 +16,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='oauth2application',
             name='pkce_required',
-            field=models.BooleanField(default=True, help_text='When True, clients must use PKCE (send code_challenge) when requesting authorization codes for this application.'),
+            field=models.BooleanField(default=True, verbose_name='PKCE Required', help_text='When True, clients must use PKCE (send code_challenge) when requesting authorization codes for this application.'),
         ),
     ]
--- a/ansible_base/oauth2_provider/views/authorization.py
+++ b/ansible_base/oauth2_provider/views/authorization.py
@@ -29,10 +29,4 @@ class AuthorizationView(oauth_views.AuthorizationView):
             error.redirect_uri = redirect_uri
             return self.error_response(OAuthToolkitError(error=error), application=application)
 
-        kwargs["scopes"] = scopes
-        kwargs["credentials"] = credentials
-        kwargs.update(credentials)
-        self.oauth2_data = kwargs
-        kwargs["application"] = application
-
-        return self.render_to_response(self.get_context_data(**kwargs))
+        return super().get(request, *args, **kwargs)
--- a/ansible_base/oauth2_provider/views/authorization.py
+++ b/ansible_base/oauth2_provider/views/authorization.py
@@ -1,0 +1,34 @@
+from oauth2_provider import views as oauth_views
+from oauth2_provider.exceptions import OAuthToolkitError
+from oauth2_provider.models import get_application_model
+from oauthlib.oauth2.rfc6749.errors import InvalidRequestError
+
+from ansible_base.lib.utils.settings import get_setting
+
+
+class AuthorizationView(oauth_views.AuthorizationView):
+    def get(self, request, *args, **kwargs):
+        try:
+            scopes, credentials = self.validate_authorization_request(request)
+        except OAuthToolkitError as error:
+            return self.error_response(error, application=None)
+
+        app_model = get_application_model()
+        try:
+            application = app_model.objects.get(client_id=credentials["client_id"])
+        except app_model.DoesNotExist:
+            error = InvalidRequestError(description="Invalid client_id.")
+            return self.error_response(OAuthToolkitError(error=error), application=None)
+
+        pkce_required_globally = get_setting('OAUTH2_PROVIDER', {}).get('PKCE_REQUIRED', False)
+        if (application.pkce_required or pkce_required_globally) and "code_challenge" not in credentials:
+            error = InvalidRequestError(description="This application requires PKCE. Include a code_challenge parameter.")
+            return self.error_response(OAuthToolkitError(error=error), application=application)
+
+        kwargs["scopes"] = scopes
+        kwargs["credentials"] = credentials
+        kwargs.update(credentials)
+        self.oauth2_data = kwargs
+        kwargs["application"] = application
+
+        return self.render_to_response(self.get_context_data(**kwargs))
--- a/ansible_base/oauth2_provider/views/authorization.py
+++ b/ansible_base/oauth2_provider/views/authorization.py
@@ -9,7 +9,7 @@ from ansible_base.lib.utils.settings import get_setting
 class AuthorizationView(oauth_views.AuthorizationView):
     def get(self, request, *args, **kwargs):
         try:
-            scopes, credentials = self.validate_authorization_request(request)
+            _, credentials = self.validate_authorization_request(request)
         except OAuthToolkitError as error:
             return self.error_response(error, application=None)
 
--- a/ansible_base/oauth2_provider/views/authorization.py
+++ b/ansible_base/oauth2_provider/views/authorization.py
@@ -13,16 +13,20 @@ class AuthorizationView(oauth_views.AuthorizationView):
         except OAuthToolkitError as error:
             return self.error_response(error, application=None)
 
+        redirect_uri = credentials.get("redirect_uri")
+
         app_model = get_application_model()
         try:
             application = app_model.objects.get(client_id=credentials["client_id"])
         except app_model.DoesNotExist:
             error = InvalidRequestError(description="Invalid client_id.")
+            error.redirect_uri = redirect_uri
             return self.error_response(OAuthToolkitError(error=error), application=None)
 
         pkce_required_globally = get_setting('OAUTH2_PROVIDER', {}).get('PKCE_REQUIRED', False)
         if (application.pkce_required or pkce_required_globally) and "code_challenge" not in credentials:
             error = InvalidRequestError(description="This application requires PKCE. Include a code_challenge parameter.")
+            error.redirect_uri = redirect_uri
             return self.error_response(OAuthToolkitError(error=error), application=application)
 
         kwargs["scopes"] = scopes
--- a/ansible_base/oauth2_provider/views/authorization.py
+++ b/ansible_base/oauth2_provider/views/authorization.py
@@ -13,20 +13,32 @@ class AuthorizationView(oauth_views.AuthorizationView):
         except OAuthToolkitError as error:
             return self.error_response(error, application=None)
 
-        redirect_uri = credentials.get("redirect_uri")
+        error_response = self._check_pkce_required(credentials["client_id"], credentials)
+        if error_response is not None:
+            return error_response
 
+        return super().get(request, *args, **kwargs)
+
+    def _check_pkce_required(self, client_id, credentials):
         app_model = get_application_model()
         try:
-            application = app_model.objects.get(client_id=credentials["client_id"])
+            application = app_model.objects.get(client_id=client_id)
         except app_model.DoesNotExist:
-            error = InvalidRequestError(description="Invalid client_id.")
-            error.redirect_uri = redirect_uri
-            return self.error_response(OAuthToolkitError(error=error), application=None)
+            return None
 
         pkce_required_globally = get_setting('OAUTH2_PROVIDER', {}).get('PKCE_REQUIRED', False)
         if (application.pkce_required or pkce_required_globally) and "code_challenge" not in credentials:
+            redirect_uri = credentials.get("redirect_uri")
             error = InvalidRequestError(description="This application requires PKCE. Include a code_challenge parameter.")
             error.redirect_uri = redirect_uri
             return self.error_response(OAuthToolkitError(error=error), application=application)
 
-        return super().get(request, *args, **kwargs)
+        return None
+
+    def form_valid(self, form):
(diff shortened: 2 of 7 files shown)


ansible_base/oauth2_provider/views/authorization.py
```diff
@@ -0,0 +1,34 @@
+from oauth2_provider import views as oauth_views
+from oauth2_provider.exceptions import OAuthToolkitError
+from oauth2_provider.models import get_application_model
+from oauthlib.oauth2.rfc6749.errors import InvalidRequestError
+
+from ansible_base.lib.utils.settings import get_setting
+
+
+class AuthorizationView(oauth_views.AuthorizationView):
+    def get(self, request, *args, **kwargs):
```
ldjebran, line 27: by inheriting directly we are loosing the functionality parent get method, maybe we should not inherit get method but only validate_authorization_request method 

if we need to inherit , maybe check the the PKCE and then call super().get()

ansible_base/oauth2_provider/views/authorization.py
```diff
@@ -0,0 +1,34 @@
+from oauth2_provider import views as oauth_views
+from oauth2_provider.exceptions import OAuthToolkitError
+from oauth2_provider.models import get_application_model
+from oauthlib.oauth2.rfc6749.errors import InvalidRequestError
+
+from ansible_base.lib.utils.settings import get_setting
+
+
+class AuthorizationView(oauth_views.AuthorizationView):
```
ldjebran, line 9: AuthorizationView also support POST with using method form_valid may be we should also inherit method form_valid to check the PKCE.

ansible_base/oauth2_provider/views/authorization.py
mabashian: 👍 I refactored the view to do the PKCE check first and then delegate to super().get() so we preserve all the upstream AuthorizationView behavior (prompt handling, skip_authorization, approval_prompt auto-reuse, form setup, etc).  I also extracted a shared _check_pkce_required helper and added a form_valid override so the PKCE check covers both GET and POST authorization flows.  See commit 5f82d545fa49920b9da5e6358b54086858cc537d.

ansible_base/oauth2_provider/views/authorization.py
mabashian: 👍 I added a form_valid override that runs the same PKCE check on the POST path. Both get() and form_valid() now use a shared _check_pkce_required helper before delegating to super().  See commit 847d46bffb66374c4c7cb90a801ed1a409b75a16.

ansible_base/oauth2_provider/views/authorization.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def _check_pkce_required(self, client_id, credentials):
+        app_model = get_application_model()
+        try:
+            application = app_model.objects.get(client_id=client_id)
+        except app_model.DoesNotExist:
+            return None
+
+        pkce_required_globally = get_setting('OAUTH2_PROVIDER', {}).get('PKCE_REQUIRED', False)
+        if (application.pkce_required or pkce_required_globally) and "code_challenge" not in credentials:
```
ldjebran: do we need to check that code_challenge has a value e.g. not an empty string, what do you think ?, what the rfc says about it ?

ansible_base/oauth2_provider/views/authorization.py
```diff
@@ -0,0 +1,44 @@
...
+            return error_response
+
+        return super().get(request, *args, **kwargs)
+
+    def _check_pkce_required(self, client_id, credentials):
+        app_model = get_application_model()
+        try:
+            application = app_model.objects.get(client_id=client_id)
+        except app_model.DoesNotExist:
+            return None
```
ldjebran: so if the application does not exist we return None, this needs some reflections

ansible_base/oauth2_provider/views/authorization.py
mabashian: Good question. The [RFC](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1#name-code_challenge-syntax) requires code_verifier to be 43-128 characters, and the S256 challenge is always a 43-character Base64URL string, so an empty string is not valid. Oauthlib only checks code_challenge is None (not truthiness), so an empty string would slip through our check. It would ultimately fail at token exchange when the verifier doesn't match, but better to reject early. I updated the check to use not credentials.get("code_challenge") so empty strings are rejected, and added a test for it. See commit a70074a484d05d46f24c8823a9043cddf31c732b.

ansible_base/oauth2_provider/views/authorization.py
```diff
@@ -0,0 +1,44 @@
...
+from oauth2_provider.models import get_application_model
+from oauthlib.oauth2.rfc6749.errors import InvalidRequestError
+
+from ansible_base.lib.utils.settings import get_setting
+
+
+class AuthorizationView(oauth_views.AuthorizationView):
+    def get(self, request, *args, **kwargs):
+        try:
+            scopes, credentials = self.validate_authorization_request(request)
```
ldjebran: maybe self.validate_authorization_request should not be called as its will be called in uper().get(request, *args, **kwargs)

ansible_base/oauth2_provider/views/authorization.py
```diff
@@ -0,0 +1,44 @@
...
+        app_model = get_application_model()
+        try:
+            application = app_model.objects.get(client_id=client_id)
+        except app_model.DoesNotExist:
+            return None
+
+        pkce_required_globally = get_setting('OAUTH2_PROVIDER', {}).get('PKCE_REQUIRED', False)
+        if (application.pkce_required or pkce_required_globally) and not credentials.get("code_challenge"):
+            redirect_uri = credentials.get("redirect_uri")
+            error = InvalidRequestError(description="This application requires PKCE. Include a code_challenge parameter.")
```
ldjebran: we are missing to return the state , as per https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1#section-4.1.2.1

its mandatory, the client may use the state to reference the current request

this may apply to other errors returned

ansible_base/oauth2_provider/views/authorization.py
mabashian: @ldjebran per our conversation yesterday I removed the DoesNotExist handler because django-oauth-toolkit's own AuthorizationView.get() does a bare .get() with no try/except since validate_authorization_request() already verifies the client_id exists.  Our custom check now matches that behavior. See commit 2a146289b913f57ed7a2bcd9f6ab44005b9f194c.

ansible_base/oauth2_provider/views/authorization.py
mabashian: Good catch, I'm now passing state from the credentials to InvalidRequestError so it's included in the error redirect per [RFC 6749 §4.1.2.1.](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1#section-4.1.2.1) Also updated the form_valid path to forward state through the credentials dict. See commit 0ea12f6d2741ef9b3ac789b2270dff0c7f5f707c.

ansible_base/oauth2_provider/views/authorization.py
mabashian: @ldjebran Here's what claude has to say about it:

> The reviewer is right that validate_authorization_request() gets called twice — once by us and once by super().get(). The question is whether we can avoid it.
> 
> The problem is we need the credentials dict (specifically client_id and code_challenge) to do our PKCE check before super().get() runs. And super().get() doesn't expose any hook between validation and rendering where we could inject our check.
> 
> The alternatives would be:
> 
> 1. Accept the double call (current approach) — validate_authorization_request() is idempotent, so calling it twice is wasteful but harmless. Keeps our code simple and doesn't duplicate DOT's logic.
> 2. Cache the result — Override validate_authorization_request() to store the result in self, so the second call returns the cached value. Works but feels hacky.
> 3. Don't call super().get() — Copy DOT's full get() body into our override. This is what the reviewer originally flagged us away from because it loses upstream behavior on DOT upgrades.
> 
> I think the double call is the best tradeoff — it's the price we pay for not duplicating DOT's get() logic. The alternative is essentially going back to what we had before. What do you think?

ansible_base/oauth2_provider/views/authorization.py
ldjebran: resolved

ansible_base/oauth2_provider/views/authorization.py
ldjebran: resolved

ansible_base/oauth2_provider/views/authorization.py
ldjebran: resolved, returned back with unitest

ansible_base/oauth2_provider/views/authorization.py
ldjebran: resolved

ansible_base/oauth2_provider/views/authorization.py
ldjebran: resolved

ansible_base/oauth2_provider/views/authorization.py
ldjebran: resolved

ansible_base/oauth2_provider/views/authorization.py
```diff
@@ -0,0 +1,43 @@
+from oauth2_provider import views as oauth_views
+from oauth2_provider.exceptions import OAuthToolkitError
+from oauth2_provider.models import get_application_model
+from oauthlib.oauth2.rfc6749.errors import InvalidRequestError
+
+from ansible_base.lib.utils.settings import get_setting
+
+
+class AuthorizationView(oauth_views.AuthorizationView):
```
BrennanPaciorek, line 9: Can we have some comments on the overrides describing why we're overriding what we are in this class?

ansible_base/oauth2_provider/views/authorization.py
```diff
@@ -0,0 +1,43 @@
...
+    def get(self, request, *args, **kwargs):
+        try:
+            _, credentials = self.validate_authorization_request(request)
+        except OAuthToolkitError as error:
+            return self.error_response(error, application=None)
```
BrennanPaciorek, line 35: https://github.com/django-oauth/django-oauth-toolkit/blob/master/oauth2_provider/views/base.py#L180-L184

This ends up being done twice, correct? Is there any reasonable way to avoid this?

ansible_base/oauth2_provider/views/authorization.py
```diff
@@ -0,0 +1,43 @@
...
+        if error_response is not None:
+            return error_response
+
+        return super().get(request, *args, **kwargs)
+
+    def _check_pkce_required(self, client_id, credentials):
+        application = get_application_model().objects.get(client_id=client_id)
+
+        pkce_required_globally = get_setting('OAUTH2_PROVIDER', {}).get('PKCE_REQUIRED', False)
+        if (application.pkce_required or pkce_required_globally) and not credentials.get("code_challenge"):
```
BrennanPaciorek: This helper method appears to be our own, unless I just have not dug down enough super-classes (I checked immediate superclass, but did not check the underlying library mixins). Since we are not interfering with the PKCE code from the base class, do we need to include the PKCE_REQUIRED global setting handling here?

If not, are there benefits to keeping it aside from consistent error handling?

BrennanPaciorek (commented):
Have a few questions about the implementation, which I've left in inline comments.

ansible_base/oauth2_provider/views/authorization.py
ldjebran: Logically we should process in our function only when global PKCE_REQUIRED is False and application.pkce_required  is True

if PKCE_REQUIRED is True we need to let the base code deal with the logic.

As @mabashian is on PTO, 
@BrennanPaciorek I will update with this logic , is that OK ?

ansible_base/oauth2_provider/views/authorization.py
ldjebran: Yes, `validate_authorization_request()` gets called twice — once by us to extract credentials for the per-app PKCE check, then again inside `super().get()`. We considered two alternatives:

1. **Cache the result** — override `validate_authorization_request()` to store the result in `self` so the second call returns the cached value. Works but couples us to DOT's internal call sequence.
2. **Skip `super().get()`** — copy DOT's full `get()` body into our override. Avoids the double call but means we'd need to track upstream changes on every DOT upgrade.

The double call is the cleanest tradeoff: `validate_authorization_request()` is idempotent (read-only validation, no side effects), and we keep full delegation to `super().get()` so we automatically inherit any upstream behavior changes.

BrennanPaciorek: Agreed. Resolving now that we've documented this tradeoff.

BrennanPaciorek: Or I guess I cannot resolve my own comments, but I'm good to resolve this one.

BrennanPaciorek: If possible though, I would like us to document this in code comments so it presents as an intentional tradeoff to someone reading the code.

ansible_base/oauth2_provider/views/authorization.py
ldjebran: updated

ansible_base/oauth2_provider/views/authorization.py
ldjebran: updated with comment before that lines

ansible_base/oauth2_provider/views/authorization.py
ldjebran: added a doc string

ansible_base/oauth2_provider/views/authorization.py
ldjebran: and yes this is our own function we added

BrennanPaciorek (approved):

ansible_base/oauth2_provider/migrations/0010_oauth2application_pkce_required.py
```diff
@@ -0,0 +1,21 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='oauth2application',
+            name='pkce_required',
+            field=models.BooleanField(default=False, help_text='When True, clients must use PKCE (send code_challenge) when requesting authorization codes for this application.'),
+        ),
+        migrations.AlterField(
+            model_name='oauth2application',
+            name='pkce_required',
+            field=models.BooleanField(default=True, verbose_name='PKCE Required', help_text='When True, clients must use PKCE (send code_challenge) when requesting authorization codes for this application.'),
```
AlanCoding, line 21: You should squash these 2 into 1 step.

ldjebran: done

AlanCoding (approved):

ldjebran:
@AlanCoding Reverted the squash — the two operations are intentional. `AddField(default=False)` backfills existing rows with `False`, then `AlterField(default=True)` changes the model default so new records get `True`. Django has no single operation that applies different defaults to existing vs new rows, so both steps are required.

ldjebran:
rebased

ldjebran:
rebased

AlanCoding (approved):
Might be good to leave a comment on the migration saying what it's doing. Because the migration behavior for old/new rows is very important high-level.

AlanCoding (approved):

PR #1054, opened 2026-07-20
[AAP-82119] Add bulk-update endpoint for resource registry

Dostonbek1:
## Description

- **What:** Add `POST /service-index/resources/bulk-update/` endpoint to the resource registry that accepts a list of resource updates and processes them in a single database transaction.
- **Why:** During `migrate_service_data`, each resource item triggers an individual HTTP PATCH to set its `service_id`. For 3,962 users, this means 3,962 sequential round-trips taking 203s. The bulk endpoint eliminates this per-item overhead.
- **How:** A new `@action` on `ResourceViewSet` that validates a list payload, fetches all resources in a single query, and applies updates atomically. Also adds `bulk_update_resources()` to `ResourceAPIClient` for callers.

## Type of Change

## Self-Review Checklist

### Prerequisites

- SQLite test environment (or PostgreSQL with docker)

### Steps to Test

1. Run: `DJANGO_SETTINGS_MODULE=test_app.sqlite3settings python -m pytest test_app/tests/resource_registry/test_bulk_update.py -v`
2. All 11 tests should pass covering: service_id updates, ansible_id renames, partially_migrated flag, mixed fields, not-found errors, empty/invalid/oversized payloads, permission checks, and resource_data updates.

### Expected Results

- `POST /service-index/resources/bulk-update/` accepts a JSON list (max 1000 items)
- Each item requires `ansible_id`, optional: `service_id`, `new_ansible_id`, `is_partially_migrated`, `resource_data`
- Returns `{"updated": N, "errors": [...]}`
- All operations are atomic (single transaction)
- Permission check: requires `HasResourceRegistryPermissions` (superuser or resource_api_actions)

### Required Actions

  - aap-gateway: `AAP-82119/batch-resource-migration` branch uses this endpoint

### Performance Impact

At Kyndryl scale (3,962 users, 2,288 teams, 130 orgs), this endpoint reduces resource migration HTTP calls from ~6,400 to ~130 (one per API page instead of one per resource item).

---

## Summary by CodeRabbit

* **New Features**
  * Added a bulk-update API (`bulk-update`) to update multiple resources in one request (up to 1,000 items).
  * Supports changing service links, identifiers (including `new_ansible_id`), migration status, and optionally resource content.
  * Returns an `updated` count plus per-item errors, allowing partial success.
* **Bug Fixes**
  * Improved REST client behavior to always send valid request bodies, even when the payload is empty.
* **Validation**
  * Rejects malformed payloads, missing/invalid `items`, oversized batches, items with no update fields, and duplicate identifiers within a request.
* **Tests**
  * Added comprehensive end-to-end coverage for success, partial failure, authorization, and content rules.

### DIFF
--- a/ansible_base/resource_registry/rest_client.py
+++ b/ansible_base/resource_registry/rest_client.py
@@ -107,7 +107,7 @@ class ResourceAPIClient(BaseServiceClient):
         Bulk-update multiple resources in a single HTTP request.
 
         Each item must contain 'ansible_id' and one or more fields to update:
-        service_id, new_ansible_id, is_partially_migrated, resource_data.
+        new_service_id, new_ansible_id, is_partially_migrated, resource_data.
 
         Returns the response from POST /resources/bulk-update/.
         """
--- a/ansible_base/resource_registry/rest_client.py
+++ b/ansible_base/resource_registry/rest_client.py
@@ -102,6 +102,17 @@ class ResourceAPIClient(BaseServiceClient):
         action = "patch" if partial else "put"
         return self._make_request(action, f"resources/{ansible_id}/", self._get_request_dict(data))
 
+    def bulk_update_resources(self, items: list[dict]):
+        """
+        Bulk-update multiple resources in a single HTTP request.
+
+        Each item must contain 'ansible_id' and one or more fields to update:
+        service_id, new_ansible_id, is_partially_migrated, resource_data.
+
+        Returns the response from POST /resources/bulk-update/.
+        """
+        return self._make_request("post", "resources/bulk-update/", data={"items": items})
+
     def delete_resource(self, ansible_id):
         return self._make_request("delete", f"resources/{ansible_id}/")
 
--- a/ansible_base/resource_registry/serializers.py
+++ b/ansible_base/resource_registry/serializers.py
@@ -147,17 +147,19 @@ class ResourceTypeSerializer(serializers.ModelSerializer):
 class BulkResourceUpdateItemSerializer(serializers.Serializer):
     """Serializer for a single item in a bulk resource update request."""
 
-    UPDATE_FIELDS = ("service_id", "new_ansible_id", "is_partially_migrated", "resource_data")
+    UPDATE_FIELDS = ("new_service_id", "new_ansible_id", "is_partially_migrated", "resource_data")
 
     ansible_id = serializers.UUIDField(help_text="The ansible_id of the resource to update.")
-    service_id = serializers.UUIDField(required=False, help_text="New service_id value.")
+    new_service_id = serializers.UUIDField(required=False, help_text="New service_id to assign.")
     new_ansible_id = serializers.UUIDField(required=False, help_text="New ansible_id to assign (renames the resource identifier).")
     is_partially_migrated = serializers.BooleanField(required=False, help_text="Partially migrated flag.")
     resource_data = serializers.JSONField(required=False, help_text="Resource data to update on the content object.")
 
     def validate(self, attrs):
         if not any(field in attrs for field in self.UPDATE_FIELDS):
-            raise serializers.ValidationError("At least one update field is required (service_id, new_ansible_id, is_partially_migrated, or resource_data).")
+            raise serializers.ValidationError(
+                "At least one update field is required (new_service_id, new_ansible_id, is_partially_migrated, or resource_data)."
+            )
         return attrs
 
 
--- a/ansible_base/resource_registry/serializers.py
+++ b/ansible_base/resource_registry/serializers.py
@@ -144,6 +144,23 @@ class ResourceTypeSerializer(serializers.ModelSerializer):
         return get_relative_url('resourcetype-detail', kwargs={"name": obj.name})
 
 
+class BulkResourceUpdateItemSerializer(serializers.Serializer):
+    """Serializer for a single item in a bulk resource update request."""
+
+    UPDATE_FIELDS = ("service_id", "new_ansible_id", "is_partially_migrated", "resource_data")
+
+    ansible_id = serializers.UUIDField(help_text="The ansible_id of the resource to update.")
+    service_id = serializers.UUIDField(required=False, help_text="New service_id value.")
+    new_ansible_id = serializers.UUIDField(required=False, help_text="New ansible_id to assign (renames the resource identifier).")
+    is_partially_migrated = serializers.BooleanField(required=False, help_text="Partially migrated flag.")
+    resource_data = serializers.JSONField(required=False, help_text="Resource data to update on the content object.")
+
+    def validate(self, attrs):
+        if not any(field in attrs for field in self.UPDATE_FIELDS):
+            raise serializers.ValidationError("At least one update field is required (service_id, new_ansible_id, is_partially_migrated, or resource_data).")
+        return attrs
+
+
 class UserAuthenticationSerializer(serializers.Serializer):
     username = serializers.CharField()
     password = serializers.CharField()
--- a/ansible_base/resource_registry/service_client.py
+++ b/ansible_base/resource_registry/service_client.py
@@ -136,7 +136,7 @@ class BaseServiceClient:
         if hasattr(self, 'timeout'):
             kwargs["timeout"] = self.timeout
 
-        if data:
+        if data is not None:
             kwargs["json"] = data
         if params:
             kwargs["params"] = params
--- a/ansible_base/resource_registry/views.py
+++ b/ansible_base/resource_registry/views.py
@@ -2,11 +2,12 @@ import logging
 from collections import OrderedDict
 
 from django.conf import settings
+from django.db import IntegrityError, transaction
 from django.db.models import Q
 from django.http import HttpResponseNotFound
 from django.shortcuts import get_object_or_404
 from django.urls.exceptions import NoReverseMatch
-from rest_framework import permissions
+from rest_framework import permissions, status
 from rest_framework.decorators import action
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
@@ -20,7 +21,7 @@ from ansible_base.lib.utils.views.permissions import try_add_oauth2_scope_permis
 from ansible_base.resource_registry.constants import SHARED_USER_RESOURCE_TYPE
 from ansible_base.resource_registry.models import Resource, ResourceType, service_id
 from ansible_base.resource_registry.registry import get_registry
-from ansible_base.resource_registry.serializers import ResourceListSerializer, ResourceSerializer, ResourceTypeSerializer
+from ansible_base.resource_registry.serializers import BulkResourceUpdateItemSerializer, ResourceListSerializer, ResourceSerializer, ResourceTypeSerializer
 from ansible_base.rest_filters.rest_framework.field_lookup_backend import FieldLookupBackend
 from ansible_base.rest_filters.rest_framework.order_backend import OrderByBackend
 from ansible_base.rest_filters.rest_framework.type_filter_backend import TypeFilterBackend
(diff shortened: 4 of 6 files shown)


Dostonbek1:
## Testing Results: Production-Scale Performance Benchmark

Tested against a local AAP dev instance (AAP 2.7, controller 4.8.3) with production-scale data:
- **3,962 users**
- **2,288 teams**
- **130 organizations**
- **57 role definitions**
- **6,428 total resources**

### Bulk-Update Endpoint Performance

| Approach | Time | Rate | Speedup |
|----------|------|------|---------|
| **Bulk (50/batch, `POST /resources/bulk-update/`)** | **9.2s** | 700 items/sec | **52x** |
| Per-item PATCH (`PATCH /resources/{id}/`) | 481.2s (8.0 min) | 13.4 items/sec | 1x (baseline) |

### Functional Verification

| Test | Result |
|------|--------|
| Bulk service_id update | ✅ |
| Bulk ansible_id rename (new_ansible_id) | ✅ |
| Bulk is_partially_migrated flag | ✅ |
| Mixed field updates in single batch | ✅ |
| Per-item error isolation (savepoints) | ✅ |
| ansible_id collision handling | ✅ |
| Non-existent resource error reporting | ✅ |
| Payload validation (missing items key, non-list) | ✅ |
| MAX_BULK_SIZE limit enforcement | ✅ |
| Permission denied for non-admin | ✅ |
| resource_data on non-manageable type error | ✅ |

### Note on Payload Format

The endpoint accepts `{"items": [...]}` (object wrapper) rather than a bare JSON list. This is required for compatibility with AWX's custom JSON parser (`awx/api/parsers.py`) which [rejects non-object JSON](https://github.com/ansible/awx/blob/devel/awx/api/parsers.py#L30-L31) at the parser level.

### Acceptance Criteria

> Resource migration at production-scale should complete in under 2 minutes.

**Result: 9.2 seconds — well under target. ✅**

jay-steurer:
# PR Review Summary — [AAP-82119] Bulk-Update Endpoint for Resource Registry

**Files**: 5 changed (335 additions, 3 deletions)
- `ansible_base/resource_registry/views.py` — new `bulk_update` action
- `ansible_base/resource_registry/serializers.py` — new `BulkResourceUpdateItemSerializer`
- `ansible_base/resource_registry/rest_client.py` — new `bulk_update_resources` client method
- `ansible_base/resource_registry/service_client.py` — `if data` → `if data is not None` fix
- `test_app/tests/resource_registry/test_bulk_update.py` — 14 tests

---

## Critical Issues (3 found)

### 1. Uncaught `ValidationError` from `update_resource()` aborts the batch mid-way with partial commits
`views.py` — the `except (ValueError, IntegrityError)` block

`resource.update_resource()` calls `serializer.is_valid(raise_exception=True)`, which raises `rest_framework.serializers.ValidationError` — **not** a subclass of `ValueError`. It escapes the catch entirely.

**Impact:** If one item has invalid `resource_data`, the uncaught exception propagates out of the loop. All items already processed are committed, all remaining items are skipped, and the caller gets a generic 400 with no information about which items succeeded. This is a silent partial-commit scenario.

**Fix:**
```python
from rest_framework.serializers import ValidationError as DRFValidationError

try:
    with transaction.atomic():
        self._apply_resource_update(resource, item)
except (ValueError, IntegrityError) as e:
    errors.append({"ansible_id": ansible_id_str, "error": str(e)})
    continue
except DRFValidationError as e:
    errors.append({"ansible_id": ansible_id_str, "error": e.detail})
    continue
```

### 2. `str(IntegrityError)` leaks database schema details to API consumers
`views.py` — `errors.append({"ansible_id": ansible_id_str, "error": str(e)})`

PostgreSQL `IntegrityError` messages contain table names, column names, constraint names, and SQL fragments (e.g., `duplicate key value violates unique constraint "resource_registry_resource_ansible_id_key"`). While the endpoint is admin-only, service tokens can be compromised, and sanitizing DB exceptions is a security best practice.

**Fix:**
```python
except IntegrityError:
    errors.append({"ansible_id": ansible_id_str, "error": "Update violates a uniqueness or integrity constraint."})
except ValueError as e:
    errors.append({"ansible_id": ansible_id_str, "error": str(e)})
```

### 3. Duplicate `ansible_id` values in a batch cause silent state corruption
`views.py` — `resources_by_id` dict maps one object per ID

If the same `ansible_id` appears twice, both items operate on the **same Python object**. Consequences:
- Last-write-wins silently — no indication the first update was overwritten
- `"updated"` counter is inflated (reports 2, but only 1 net update)
- If item 1 renames via `new_ansible_id`, item 2 operates on stale state

**Fix:** Add a uniqueness check after serializer validation:
```python
seen_ids = set()
for item in items:
    aid = str(item["ansible_id"])
    if aid in seen_ids:
        return Response(
            {"detail": f"Duplicate ansible_id in request: {aid}"},
            status=status.HTTP_400_BAD_REQUEST,
        )
    seen_ids.add(aid)
```

---

## Important Issues (4 found)

### 4. Metadata save bypasses `no_reverse_sync()` — risk of sync feedback loops
`views.py` — `_apply_resource_update`, the `resource.save(update_fields=update_fields)` call

The existing `Resource.update_resource()` model method wraps all saves in `no_reverse_sync()` to prevent signal-based reverse synchronization loops. The bulk endpoint's bare `save()` for metadata fields (`service_id`, `ansible_id`, `is_partially_migrated`) skips this, potentially triggering cascading sync calls back to the calling service.

**Fix:** Either wrap in `no_reverse_sync()`, or better, delegate to `resource.update_resource()` for all fields.

### 5. `_apply_resource_update` duplicates logic from `Resource.update_resource()`
`views.py` — the entire `_apply_resource_update` static method

The model method already accepts `ansible_id`, `service_id`, `is_partially_migrated`, and `resource_data` as parameters and handles them atomically with `no_reverse_sync()`. Reimplementing this in the view creates divergence risk — if the model method is updated (audit logging, new validation, side effects), the bulk path silently diverges.

**Fix:** Delegate to the model method instead of reimplementing:
```python
@staticmethod
def _apply_resource_update(resource, item):
    resource_data = item.get("resource_data", {})
    if resource_data and not resource.content_type.resource_type.can_be_managed:
        raise ValueError(f"Resource type '{resource.content_type.resource_type.name}' cannot be managed.")
    resource.update_resource(
        resource_data=resource_data,
        ansible_id=item.get("new_ansible_id"),
        service_id=item.get("service_id"),
        is_partially_migrated=item.get("is_partially_migrated"),
        partial=True,
    )
```
This also fixes issue #4 automatically.

### 6. Zero logging in a bulk mutation endpoint
`views.py` — entire `bulk_update` method

No `logger` calls anywhere. For an endpoint that can modify up to 1000 resources per request, this is an operational blindspot. No trace of batch size, success/failure counts, caller identity, or per-item errors.

**Fix:** Add logging at entry, on errors, and at completion:
```python
logger.info("Bulk update requested: %d items", len(items_data))
# ... on per-item error:
logger.warning("Bulk update item %s failed: %s", ansible_id_str, e)
# ... at end:
logger.info("Bulk update completed: %d updated, %d errors", updated, len(errors))
```

### 7. Partial-failure tests don't verify DB state (savepoint correctness untested)
`test_bulk_update.py` — `test_bulk_update_not_found`, `test_bulk_update_ansible_id_collision`

These tests check `resp.data["updated"]` and `resp.data["errors"]` counts but never call `refresh_from_db()` on the *successful* items to verify their changes persisted. If the savepoint implementation were broken and rolled back everything, the response could still report success while the database reverted.

**Fix:** Add `refresh_from_db()` assertions for successfully-updated resources in both tests.

---

## Suggestions (5 found)

### 8. No test for the `if data is not None` fix in `service_client.py`
This behavioral change affects all callers of `_make_request`. A test sending `data={}` (empty dict, previously falsy and silently dropped) would prevent regression.

### 9. No test for duplicate `ansible_id` in the same batch
Whatever behavior is chosen (reject or last-write-wins), it should be explicitly tested.

### 10. No test for unauthenticated access
`test_bulk_update_permission_denied` covers non-admin (403) but not unauthenticated (401). These exercise different code paths in the permission class.

### 11. No test combining `resource_data` with metadata fields in one item
If `_apply_resource_update` has an early-return after handling `resource_data`, metadata fields would be silently dropped. A test with both would catch this.

### 12. `bulk_update_resources` client method has no response validation
Unlike some other client methods, it returns raw `requests.Response` with no status checking. Callers that don't inspect the response will miss partial failures. Consider at minimum logging non-200 responses.

---

## Strengths

- **Well-designed serializer** — `BulkResourceUpdateItemSerializer` with `UPDATE_FIELDS` validation is clean and extensible
- **Per-item savepoints** — correct approach for partial-failure resilience; one bad item doesn't roll back the batch
- **Comprehensive test suite** — 14 tests covering happy paths, validation, permissions, partial failures, and edge cases
- **The `if data is not None` fix** — correct and safe; fixes a real bug where empty-but-valid payloads were silently dropped
- **Good batch size limit** — 1000-item cap prevents abuse
- **`select_related("content_type__resource_type")`** — avoids N+1 queries in the processing loop

---

## Recommended Action

1. **Fix** the uncaught `ValidationError` (issue #1) — this is the highest-risk bug
2. **Sanitize** `IntegrityError` messages (issue #2) — information disclosure
3. **Reject** duplicate `ansible_id` values (issue #3) — silent corruption
4. **Delegate** to `Resource.update_resource()` (issue #5) — fixes both the `no_reverse_sync` gap (#4) and the logic duplication
5. **Add** logging (issue #6)
6. **Strengthen** partial-failure tests with DB assertions (issue #7)

---
*Assisted-by: Claude Code*

[AAP-82119]: https://redhat.atlassian.net/browse/AAP-82119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Dostonbek1:
Thanks for the thorough review @jay-steurer! All issues addressed in the latest push:

### Critical Issues — Fixed

1. **Uncaught ValidationError** — Added `DRFValidationError` to the except clause. `resource.update_resource()` raises `rest_framework.exceptions.ValidationError` which is now properly caught and reported as a per-item error.

2. **IntegrityError leaks schema details** — Sanitized: `IntegrityError` now always returns `"Update violates a uniqueness or integrity constraint."` instead of the raw database message.

3. **Duplicate ansible_id in batch** — Added a pre-processing uniqueness check after serializer validation. Duplicate `ansible_id` values now return HTTP 400 with a clear message identifying the duplicate.

### Important Issues — Fixed

4+5. **no_reverse_sync() / Logic duplication** — Refactored `_apply_resource_update` to delegate entirely to `Resource.update_resource()`. This eliminates the duplicated logic, automatically gets `no_reverse_sync()` coverage for content object saves, and ensures future model method changes are inherited.

6. **Zero logging** — Added `logger.info` at entry (item count + user) and completion (updated + error counts), plus `logger.warning` for each per-item failure.

7. **Partial-failure tests missing DB assertions** — `test_bulk_update_ansible_id_collision` and `test_bulk_update_resource_data_not_manageable` now verify both that the successful item persisted and the failed item was not changed. Also added an assertion that IntegrityError messages are sanitized in the response.

### Suggestions — Addressed

8–11. Added 3 new tests:
- `test_bulk_update_duplicate_ansible_id` — verifies the duplicate rejection
- `test_bulk_update_unauthenticated` — verifies 401 for unauthenticated requests
- `test_bulk_update_resource_data_with_metadata` — verifies both `resource_data` AND metadata fields are applied together in one item

12. `bulk_update_resources` client response validation — noted; the Gateway caller in the jewel PR already inspects the response and handles non-200 gracefully (see `_send_bulk_update` in jewel#167).

All 18 tests passing.

john-westcott-iv:
## Code Review

The endpoint solves a real need — N individual PATCHes for resource sync are slow. The test coverage is solid (15 cases). But the implementation is a loop-of-singles behind a single HTTP endpoint rather than using Django's bulk operations. Here are my findings, from architectural to specific.

### Architecture: Loop-of-singles vs true bulk

**Current flow for 100 items:**
```
1 SELECT to fetch all Resources (good — batched)
Then for EACH item:
  1 SAVEPOINT
  1 UPDATE (resource.save)
  1+ queries from update_resource()
  1 RELEASE SAVEPOINT
= ~400+ queries for 100 items
```

**Optimized flow:**
```
1 SELECT — fetch all Resources
1 bulk_update — metadata fields (service_id, ansible_id, is_partially_migrated)
Per content type (group items by type):
  1 SELECT — fetch content objects
  1 bulk_update — content object fields
= ~2 + (2 × num_types) queries total
```

For 1000 items across 5 types: ~12 queries instead of ~3000+. The approach:
1. Separate metadata updates from content updates
2. Batch metadata with `Resource.objects.bulk_update()`
3. Group content updates by `content_type` — all Users in one batch, all Teams in another, etc.
4. Validate all items per group before applying, collect errors, bulk-apply only valid ones
5. Per-item savepoints become unnecessary for metadata-only updates

This also makes the 1000 limit about payload size rather than query count — you could comfortably handle 10k items.

### Specific findings

**1. `{"items": [...]}` wrapper (Minor)**

The `{"items": [...]}` wrapper adds 15 lines of manual type-checking validation. DRF handles bare lists natively with `Serializer(data=request.data, many=True)`. Unless top-level parameters are planned (e.g. `dry_run`, `on_conflict`), a bare list is simpler and more idiomatic.

**2. Exception handling too narrow (Major)**

`views.py:170-172` catches only `ValueError` and `IntegrityError`. If `update_resource()` raises a DRF `ValidationError`, `PermissionDenied`, or anything else, it propagates uncaught and crashes the entire request with a 500. The caller loses the response body and can't tell which items succeeded. Either broaden to `except Exception` for the per-item loop, or add the specific exceptions `update_resource()` can raise.

**3. `service_id` naming inconsistency (Nit)**

`new_ansible_id` uses the `new_` prefix because it's setting a new value. `service_id` is doing the same thing but lacks the prefix. For consistency: either `new_service_id` or drop the `new_` from `new_ansible_id`.

**4. `can_be_managed` check is too late (Minor)**

Line 210: the check runs after metadata fields are already saved to DB. If it fails, the savepoint rolls back both — so it's correct, but wasteful. Move the check to the top of `_apply_resource_update` (or into the serializer's `validate`) to fail fast before doing any DB work.

**5. Partial success within an item (Design question)**

Currently if `resource_data` fails, the savepoint rolls back the metadata save too (all-or-nothing per item). Should the metadata succeed and the content failure be reported as a warning? This would let the caller know "metadata now matches, content update needs retry." The response could become `{"updated": N, "errors": [...], "warnings": [...]}`.

**6. Duplicate ansible_ids in batch (Minor)**

No deduplication check. Same `ansible_id` twice means both items operate on the same in-memory object. The second may conflict with or undo the first. At minimum, detect and reject duplicates; or document last-write-wins behavior.

**7. Not-found behavior — reject vs upsert (Design question)**

Currently, a missing `ansible_id` is an error. If the use case is resource sync between services, upsert (create if missing) might be more useful — the caller shouldn't need to know whether the resource exists yet. Bigger design decision though, since creation requires content_type and object_id.

**8. `update_fields` on save vs `update_resource` (Clarification)**

`resource.save(update_fields=...)` saves the Resource model's own columns (different table). `resource.update_resource()` saves the content object (User/Team/Org — different table). These can't be combined into one query, but the metadata saves across items CAN be batched with `bulk_update()`.

*Assisted-by: Claude Code / Opus 4.6 (Anthropic)*

Dostonbek1:
Thanks for the thorough review @john-westcott-iv! Really appreciate the detailed analysis. Here's how we addressed each point:

### 1. Architecture: Loop-of-singles vs true bulk

Valid observation. We've added a **Performance note** in the docstring acknowledging this trade-off. The short answer: `ResourceTypeProcessor.save()` is a per-service plugin point with custom logic (e.g. `RoleDefinitionProcessor` handles M2M permissions) that can't be expressed as a single `bulk_update()` call. The primary performance gain of this endpoint is eliminating N HTTP round-trips — the DB query overhead is negligible on a local connection for typical batch sizes. True DB-level batching for metadata-only fields is noted as a future optimization.

### 2. Exception handling too narrow ✅ Fixed

Added a broad `except Exception` fallback that logs the full traceback at ERROR level and returns a generic "Internal error processing this item." message to the caller. This prevents one unexpected failure from crashing the entire batch.

### 3. `service_id` naming inconsistency ✅ Fixed

Renamed to `new_service_id` for consistency with `new_ansible_id`. Both fields assign new values, so the `new_` prefix makes sense for both. The corresponding gateway client has been updated in the jewel PR.

### 4. `can_be_managed` check too late ✅ Fixed

Moved the check **before** the `transaction.atomic()` block so we fail-fast without ever opening a savepoint for items that will obviously be rejected.

### 5. Partial success within an item

Keeping **all-or-nothing per item** as an intentional design choice. If metadata says `service_id = gateway` but content is stale, the resource is in an inconsistent state. All-or-nothing simplifies retry semantics (the caller just retries the whole item) and matches `Resource.update_resource()`'s existing single-resource contract. Added an inline comment documenting this decision.

### 6. Duplicate ansible_ids in batch

Already implemented! Lines 178-188 detect and reject duplicates with a `seen_ids` set, and there's a test (`test_bulk_update_duplicate_ansible_id`) covering it.

### 7. Not-found behavior — reject vs upsert

Keeping **reject** as the correct behavior for this use case. Resource creation requires `content_type` + `object_id` + the content object to exist — context the bulk-update caller doesn't carry. In the migration flow, creation happens on the gateway side (local DB), then this endpoint is called to tell upstream "update your metadata." If the upstream resource doesn't exist, something is fundamentally wrong and we want to surface that error, not silently create orphans.

### 8. `update_fields` vs `update_resource`

Same as #1 — acknowledged in the Performance note as a future optimization target for metadata-only fields.

---

All code changes pushed in the latest commit.

john-westcott-iv (approved):
All prior review findings addressed:

- Exception handling broadened (IntegrityError, ValueError, DRFValidationError, catch-all Exception)
- `service_id` renamed to `new_service_id` for consistency with `new_ansible_id`
- `can_be_managed` check moved before the transaction
- Duplicate ansible_id detection added with test
- `_apply_resource_update` simplified to single `update_resource()` call
- Performance note explaining loop-of-singles trade-off
- IntegrityError messages sanitized (no schema leaks)
- Logging at start/completion

CI linting/checks passing, test suites still running (just pushed). LGTM pending CI green.

*Assisted-by: Claude Code / Opus 4.6 (Anthropic)*

jay-steurer (approved):

john-westcott-iv (approved):
Refactoring looks good — same logic decomposed into focused helpers to satisfy SonarQube cognitive complexity. No behavior changes, no new issues.

New tests (`test_bulk_update_unexpected_exception`, `test_bulk_update_resources` integration test) are good additions. All 19 core checks passing (eda-server failure is unrelated).

*Assisted-by: Claude Code / Opus 4.6 (Anthropic)*

PR #1053, opened 2026-07-20
Batch RBAC cleanup for org deletion scaling (AAP-82668)

AlanCoding:
## Summary

Organization deletion at scale (20+ teams) times out because Django's signal cascade fires per-object RBAC recomputation queries. This PR introduces composable context managers that batch the cleanup work, reducing query count by **95%** (762 → 36 at 20 teams, 2 users/team).

The approach avoids monkeypatching or re-entrant nesting — each DAB app owns its own context manager, independently composable by the caller:

| Context manager | App | What it batches |
|---|---|---|
| `defer_rbac_cache()` | `rbac` | ObjectRole/RoleEvaluation cleanup, skips per-team M2M queries |
| `no_activity_stream()` | `activitystream` | Suppresses audit logging signals |
| `no_reverse_sync_for(Team)` | `resource_registry` | Suppresses reverse sync HTTP calls for Team |
| `defer_resource_cleanup()` | `resource_registry` | Batches Resource row deletion |

### Caller usage (what AWX should do)

```python
from ansible_base.activitystream import no_activity_stream
from ansible_base.rbac.triggers import defer_rbac_cache
from ansible_base.resource_registry.signals.handlers import (
    defer_resource_cleanup, no_reverse_sync_for,
)

with no_activity_stream():
    with no_reverse_sync_for(Team):
        with defer_resource_cleanup():
            with defer_rbac_cache():
                org.delete()
```

### Query reduction breakdown (20 teams, 2 users/team)

| Configuration | Queries | Per team | Reduction |
|---|---|---|---|
| No context managers | 762 | 38.1 | baseline |
| `defer_rbac_cache` only | 383 | 19.1 | 50% |
| All 4 context managers | 36 | 1.8 | **95%** |

### Key findings

- `team_pre_delete` was querying `member_roles` and `provides_teams` M2M per team, but both consumers (post_delete and deferred flush) run after cascade deletes those M2M entries — the stashed data was never used. Skipping these in deferred mode saves 2 queries/team.
- `remove_resource` was doing individual `Resource.get_resource_for_object` lookups per cascade-deleted object. Batching into a single bulk delete per content type eliminated ~40 queries.
- Correctness in deferred mode comes from `_bulk_ancestor_roles` (queries via `RoleEvaluation.permission_partials`, not M2M) and orphan ObjectRole cleanup.
[...]

PR #1055, opened 2026-07-20
[AAP-82668] Add deferred_activity_stream context manager for batched audit logging

john-westcott-iv:
## Summary

- Adds `deferred_activity_stream()` context manager that accumulates `Entry` objects and audit log lines during bulk operations
- On successful exit, entries are `bulk_create`d and audit lines emitted via `transaction.on_commit`
- On exception, accumulated state is discarded (no false audit trail for rolled-back transactions)
- Re-entrant: inner calls are no-ops (outermost caller owns the flush)
- Populates `created_by` on deferred entries before `bulk_create` since `save()` is bypassed
- Extracts `_flush_deferred_activity_stream` and `_format_audit_lines` helpers to reduce cognitive complexity

## Context

Organization deletion at scale triggers per-object activity stream signals during the Django cascade. This context manager enables callers to batch that work:

```python
with deferred_activity_stream():
    org.delete()  # entries accumulate instead of individual INSERTs
# flush: bulk_create + on_commit for audit lines
```

Used by both PR #1050 (raw SQL RBAC cascade optimization) and PR #1053 (batch RBAC rebuilding) to defer activity stream work during org deletion.

Companion AWX PR: https://github.com/ansible/awx/pull/16553

## Test plan

- [x] 71 activity stream tests pass (existing + new)
- [x] `TestFlushDeferredActivityStream` — bulk create, user fill-in, on_commit scheduling
- [x] `TestFormatAuditLines` — create/delete/update formatting, M2M, edge cases

## Summary by CodeRabbit

- **New Features**
  - Added a public context manager to defer activity stream side effects, batching entry persistence and emitting related audit events after the surrounding transaction commits.
  - Exposed the deferred activity stream capability for external use.

- **Bug Fixes**
  - Improved audit-log message formatting across create, update, delete, and many-to-many activity changes, including clearer per-field update details and consistent “with/from” wording.

- **Tests**
  - Expanded unit and integration coverage for audit formatting, deferred flushing via commit, independence of buffered entries vs audit lines, nested behavior, and discard-on-error handling.

### DIFF
--- a/ansible_base/activitystream/signals.py
+++ b/ansible_base/activitystream/signals.py
@@ -25,6 +25,16 @@ class ActivityStreamEnabled(threading.local):
 activitystream_enabled = ActivityStreamEnabled()
 
 
+class _DeferredActivityStream(threading.local):
+    def __init__(self) -> None:
+        self.active: bool = False
+        self.entries: list = []
+        self.audit_lines: list[str] = []
+
+
+_deferred_activity_stream = _DeferredActivityStream()
+
+
 @contextmanager
 def no_activity_stream() -> Generator[None, None, None]:
     previous_value = activitystream_enabled.enabled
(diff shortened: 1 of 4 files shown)


ansible_base/activitystream/signals.py
```diff
@@ -35,6 +45,66 @@ def no_activity_stream() -> Generator[None, None, None]:
...
+    While active, _store_activitystream_entry accumulates Entry objects and
+    audit log lines in thread-local lists instead of writing them immediately.
+    On successful exit, Entry objects are bulk-created and audit lines are
+    emitted via transaction.on_commit (so they only fire after the enclosing
+    transaction commits — a rollback silently discards them).
+    Re-entrant: inner calls are no-ops (outermost caller owns the flush).
+    """
+    if _deferred_activity_stream.active:
+        yield
+        return
```
AlanCoding, line 90: I have a personal _preference_ that this raise an error / not be allowed. But this is probably fine, at least for this case being so simple.

test_app/tests/activitystream/test_signals.py
```diff
@@ -942,3 +1058,88 @@ def test_audit_log_m2m_preposition(user, operation, method_name, expected_prepos
...
+
+    @pytest.mark.django_db
+    @pytest.mark.parametrize(
+        'entries_count,lines_count',
+        [(0, 2), (1, 0), (2, 3)],
+        ids=['only_lines', 'only_entries', 'both'],
+    )
+    def test_handles_entries_and_lines_independently(self, system_user, entries_count, lines_count):
+        """Each path (entries and audit_lines) operates independently."""
+        from django.contrib.contenttypes.models import ContentType
```
AlanCoding: ...could put imports at top

AlanCoding (changes requested):
I think I have a question, why not have a test... even one test, that does `with deferred_activity_stream():`?

Then should be good to go.

ansible_base/activitystream/signals.py
john-westcott-iv: Agreed it could go either way. Kept it as a no-op for consistency with how `defer_rbac_cache` works in triggers.py — both context managers are designed to be composable and potentially nested by callers who may not know if an outer layer already activated deferral. Raising would make composition harder (callers would need to track state).

test_app/tests/activitystream/test_signals.py
john-westcott-iv: Fixed — moved `ContentType` import to the top of the file.

john-westcott-iv:
Added `TestDeferredActivityStream` with 4 integration tests that use the context manager directly:

1. **test_entries_are_deferred_and_bulk_created** — creates objects inside the context, asserts no entries exist during deferral, confirms they appear after exit
2. **test_entries_discarded_on_exception** — raises inside the context, confirms no entries are created
3. **test_reentrant_inner_is_noop** — nested context managers, confirms only outermost flushes
4. **test_delete_entries_are_deferred** — deletes an object inside the context, confirms the delete entry is deferred and appears after exit

john-westcott-iv:
@coderabbitai review

AlanCoding:
## `diff()` FK lazy-loads survive deferral — dominant cost is unchanged

`deferred_activity_stream` successfully batches `Entry.objects.create` → `bulk_create` and defers `log_auth_event` → `on_commit`. But the dominant query cost in bulk-delete scenarios comes from `diff()` in `_store_activitystream_entry`, and that still runs eagerly per-object.

### Root cause

In `_store_activitystream_entry` (line 234 in the PR diff), `diff()` is called before any deferral check:

```python
delta = diff(old, new, exclude_fields=excluded, limit_fields=limit, all_values_as_strings=True)
```

Inside `diff()`, line 302 checks each field:

```python
if all_values_as_strings:
    if getattr(obj, field) is None:   # <-- triggers FK lazy-load
        value = None
    else:
        value = field_obj.value_to_string(obj)
```

For FK fields, `getattr(obj, field)` uses the descriptor name (e.g., `owner`, `created_by`) instead of the attname (e.g., `owner_id`, `created_by_id`). This triggers a database SELECT to load the related object — even though `value_to_string()` only returns the raw PK as a string anyway.

### Reproducer

Using the test_app `Animal` model (AuditableModel with 4 FKs — `owner`, `created_by`, `modified_by`, `resource`):

```python
from django.test.utils import CaptureQueriesContext
from django.db import connection
from test_app.models import Animal
from ansible_base.activitystream.signals import no_activity_stream, deferred_activity_stream

owner = User.objects.get(username='admin')
N = 50

# setup
with no_activity_stream():
    for i in range(N):
        Animal.objects.create(name=f'repro-{i}', owner=owner)

# Baseline
with CaptureQueriesContext(connection) as ctx:
    Animal.objects.filter(name__startswith='repro-').delete()
# => 256 queries (202 SELECTs)

# With deferred_activity_stream
with CaptureQueriesContext(connection) as ctx:
    with deferred_activity_stream():
        Animal.objects.filter(name__startswith='repro-').delete()
# => 159 queries (152 SELECTs — 151 are User FK lazy-loads from diff())

# With no_activity_stream
with CaptureQueriesContext(connection) as ctx:
    with no_activity_stream():
        Animal.objects.filter(name__startswith='repro-').delete()
# => 5 queries
```

The deferral saves ~100 queries (bulk_create + deferred audit lines), but 151 of the remaining 159 are User FK lazy-loads from `diff()` — 3 per Animal (owner, created_by, modified_by).

### Suggested regression test

A test that asserts the query count of `deferred_activity_stream` during bulk deletion stays within a bound proportional to the number of objects, not the number of FK fields × objects:

```python
@pytest.mark.django_db
def test_deferred_delete_query_scaling(system_user):
    """deferred_activity_stream should not scale queries with FK field count."""
    N = 20
    with no_activity_stream():
        for i in range(N):
            Animal.objects.create(name=f'scale-{i}', owner=system_user)

    with CaptureQueriesContext(connection) as ctx:
        with deferred_activity_stream():
            Animal.objects.filter(name__startswith='scale-').delete()

    # Without the FK lazy-load fix, this is ~3*N SELECTs from diff().
    # With the fix, diff() uses attname (owner_id) instead of descriptor
    # (owner), avoiding the lazy-load entirely.
    selects = sum(1 for q in ctx if q['sql'].strip().startswith('SELECT'))
    assert selects < N, (
        f"Expected fewer than {N} SELECTs for {N} deletes, got {selects}. "
        f"diff() is likely lazy-loading FK fields via getattr(obj, field) "
        f"instead of getattr(obj, field_obj.attname)."
    )
```

### Suggested fix direction

In `diff()`, when `all_values_as_strings=True` and the field is a FK, use `getattr(obj, field_obj.attname)` (the `_id` column) for the None check and skip loading the related object. The `value_to_string()` call already returns just the raw PK — the related object is never needed.

john-westcott-iv:
Great analysis. Fixed — the `diff()` utility now uses `field_obj.attname` (e.g. `owner_id`) instead of the descriptor name (e.g. `owner`) for the None check when `all_values_as_strings=True`. This avoids triggering FK lazy-load SELECTs since `value_to_string()` only ever needs the raw PK via `value_from_object()` → `getattr(obj, self.attname)`.

Confirmed that FK `value_to_string` always returns the PK (e.g. `"42"`), never the related object's display field (e.g. username). The related object is never needed.

Added the regression test you suggested — asserts fewer than N SELECTs for N deletes under `deferred_activity_stream`.

AlanCoding (approved):

Issue #300, opened 2024-04-10
[RBAC] Optimize `compute_object_role_permissions` as iterable with `prefetch_related`


AlanCoding:
If you use Django debug toolbar, you can run the following to see how performant DAB RBAC is for rebuilding the entire RoleEvaluation table

```python
from ansible_base.rbac.caching import compute_team_member_roles, compute_object_role_permissions

compute_object_role_permissions()
```

You find that it is limited by a particular constraint, which is that we need to some simple-looking prefetches added:

```diff
diff --git a/ansible_base/rbac/caching.py b/ansible_base/rbac/caching.py
index a90bf00..24584ab 100644
--- a/ansible_base/rbac/caching.py
+++ b/ansible_base/rbac/caching.py
@@ -168,7 +168,7 @@ def compute_object_role_permissions(object_roles=None, types_prefetch=None):
     if types_prefetch is None:
         types_prefetch = TypesPrefetch.from_database(RoleDefinition)
     if object_roles is None:
-        object_roles = ObjectRole.objects.iterator()
+        object_roles = ObjectRole.objects.prefetch_related('permission_partials', 'permission_partials_uuid', 'provides_teams__has_roles')
 
     for object_role in object_roles:
         role_to_delete, role_to_add = object_role.needed_cache_updates(types_prefetch=types_prefetch)
```

However, this loses `.iterator()` which is probably unacceptable, because this is the one big memory-intensive table involved in querysets.

There is a very good proposed solution at https://djangosnippets.org/snippets/1949/ but it is not a standard Django util.

```python
import gc

def queryset_iterator(queryset, chunksize=1000):
    '''''
    Iterate over a Django Queryset ordered by the primary key

    This method loads a maximum of chunksize (default: 1000) rows in it's
    memory at the same time while django normally would load all rows in it's
    memory. Using the iterator() method only causes it to not preload all the
    classes.

    Note that the implementation of the iterator does not support ordered query sets.
    '''
    pk = 0
    last_pk = queryset.order_by('-pk')[0].pk
    queryset = queryset.order_by('pk')
    while pk < last_pk:
        for row in queryset.filter(pk__gt=pk)[:chunksize]:
            pk = row.pk
            yield row
        gc.collect()
```

This will give 1 order-of-magnitude improvement in the above scenario.

The ToS on the site provides license for the code snippets: https://djangosnippets.org/about/tos/

[...]

PR #1064, opened 2026-07-22
AAP-83963 — Skip RBAC recompute in post_migrate when no migrations were applied

kdelee:
## Summary
Skip `compute_team_member_roles()` and `compute_object_role_permissions()` when no migrations were actually applied.

**Measured on ScaleLab** (42K ObjectRoles, 418K RoleEvaluations):
- `compute_team_member_roles()`: 1.2s
- `compute_object_role_permissions()`: **1,128s (18.8 minutes)**
- Per pod: ~19 minutes in init container
- 3 replicas rolling sequentially: **~57 minutes per rollout**

Checks the `plan` kwarg from Django's `post_migrate` signal — empty list means no migrations ran.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved post-migration access-control setup by skipping unnecessary permission recomputation when no migrations are applied.
  * Preserved existing role and permission updates when migrations require recomputation.

### DIFF
--- a/ansible_base/rbac/triggers.py
+++ b/ansible_base/rbac/triggers.py
@@ -448,6 +448,11 @@ def post_migration_rbac_setup(sender, *args, **kwargs):
 
     dab_post_migrate.send(sender=sender)
 
+    plan = kwargs.get('plan', None)
+    if plan is not None and len(plan) == 0:
+        logger.info('Skipping RBAC recompute — no migrations were applied')
+        return
+
     compute_team_member_roles()
     compute_object_role_permissions()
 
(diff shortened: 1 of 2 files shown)


john-westcott-iv (approved):

AlanCoding:
I'm not entirely convinced there isn't a gotcha with this. I worry that, for example, aap-gateway might run `post_migrate` logic that assumes that this will run after it.

I see that 18 min computation time. That's not good. However, that was introduced really early, really hastily, with known gains left on the table, with https://github.com/ansible/django-ansible-base/issues/300 being really obvious and cheap.

I don't know exactly how much that will claw back, but these days it's really extremely easy to just take that issue and get a patch.

AlanCoding (approved):

PR #1066, opened 2026-07-22
[AAP-82668] Replace Redis cache with thread-local for get_system_user()

john-westcott-iv:
## Summary

Replaces the Redis-based `get_system_user()` cache (#1056) with a
thread-local in-memory cache, and clears it at the start of each HTTP
request via `request_started` signal.

## Why

Profiling at 50K scale showed `get_system_user()` called ~49,800 times
during org deletion cascade. The Redis cache from #1056 eliminates the
DB query but **49,800 Redis round-trips (~0.35ms each) still account
for 17 seconds** of the 41-second delete time.

Thread-local cache eliminates all network round-trips. The cache is:
- **Thread-safe** — `threading.local()` isolates per-thread
- **Per-request fresh** — `request_started` signal clears at the start of each HTTP request
- **Management-command safe** — cache lives for the command's duration (no `request_started` fires)
- **Invalidated on username change** — cache is keyed on `system_username`
- **Invalidated on delete** — `AbstractDABUser.delete()` clears the cache when the system user is deleted

## Cache invalidation

| Scenario | Mechanism |
|----------|-----------|
| New HTTP request | `request_started` signal clears thread-local |
| System username setting changes | Cache key mismatch → automatic miss |
| System user deleted via `user.delete()` | `AbstractDABUser.delete()` override clears cache |
| System user deleted via queryset `.filter().delete()` | Not caught by model `delete()` — next `request_started` clears it |
| Management command | Cache lives for command duration, cleared on next request |

Depends on #1065 (revert of #1056).

## Test plan

- [x] `test_second_call_returns_cached_result` — zero DB queries on second call
- [x] `test_cache_invalidated_when_username_changes` — cache miss on setting change
- [x] `test_cache_not_set_for_none_result` — None results not cached
- [x] `test_clear_system_user_cache` — explicit clear works
- [x] `test_request_started_clears_cache` — signal clears thread-local

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented stale system-user data by moving to request-scoped caching that resets at the start of each request.
  - Cache is now cleared only when deleting the system user, not on every user save.
- **Refactor**
  - Updated system-user caching to use per-thread local storage instead of shared caching.
- **Tests**
  - Expanded/updated coverage to validate thread-local reuse, cache invalidation on system-username changes, and cache clearing on request start and explicit clears.

### DIFF
--- a/ansible_base/lib/abstract_models/user.py
+++ b/ansible_base/lib/abstract_models/user.py
@@ -21,16 +21,3 @@ class AbstractDABUser(AbstractUser):
 
     all_objects = UserManager()
     objects = UserManager()
-
-    def save(self, *args, **kwargs):
-        super().save(*args, **kwargs)
-        from ansible_base.lib.utils.models import clear_system_user_cache
-
-        clear_system_user_cache()
-
-    def delete(self, *args, **kwargs):
-        result = super().delete(*args, **kwargs)
-        from ansible_base.lib.utils.models import clear_system_user_cache
-
-        clear_system_user_cache()
-        return result
--- a/ansible_base/lib/abstract_models/user.py
+++ b/ansible_base/lib/abstract_models/user.py
@@ -1,5 +1,7 @@
 from django.contrib.auth.models import AbstractUser, UserManager
 
+from ansible_base.lib.utils.create_system_user import get_system_username
+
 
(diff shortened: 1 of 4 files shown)


john-westcott-iv:
closing in favor of #1067

PR #1072, opened 2026-07-30
[AAP-85068] fix(rbac): resolve UUID ansible_ids via Resource table in get_content_type

AlexSCorey:
When syncing role assignments from the Gateway, object_ansible_id is a
  UUID. For organization and team types this was already resolved through
  the Resource table, but all other content types passed the UUID directly
  as a PK, causing "Field id expected a number" errors on models with
  integer primary keys.

  Try the Resource table lookup first for all content types, falling back
  to direct PK lookup for backward compatibility.

  Fixes: AAP-71217

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role synchronization target resolution by matching existing resource records, including organization and team targets.
  * Added more resilient fallback handling when resource matching fails or validation issues occur.
  * Improved consistency when updating permissions, team memberships, and roles after changes or deletions.

* **Tests**
  * Added database coverage for UUID-based resource resolution and primary-key fallback behavior during role synchronization.

### DIFF
--- a/ansible_base/rbac/role_sync_utils.py
+++ b/ansible_base/rbac/role_sync_utils.py
@@ -5,6 +5,7 @@ from dataclasses import dataclass
 from typing import Any
 
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ValidationError
 from django.db.models import Q
 
 from ansible_base.lib.utils.apps import is_rbac_installed
@@ -75,6 +76,17 @@ def get_content_object(role_definition, assignment_tuple: AssignmentTuple) -> An
     if role_definition.content_type.model in ('organization', 'team'):
         object_resource = Resource.objects.get(ansible_id=assignment_tuple.ansible_id_or_pk)
         return object_resource.content_object
+    try:
+        ct = ContentType.objects.get_for_model(
+            role_definition.content_type.model_class()
+        )
+        object_resource = Resource.objects.get(
+            ansible_id=assignment_tuple.ansible_id_or_pk,
+            content_type=ct,
+        )
+        return object_resource.content_object
+    except (Resource.DoesNotExist, ValidationError):
+        pass
     model = role_definition.content_type.model_class()
     return model.objects.get(pk=assignment_tuple.ansible_id_or_pk)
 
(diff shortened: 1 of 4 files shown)


AlanCoding (changes requested):
To me this seems like ContentType thrashing. The criteria you are grasping for in the new `try:` is a duplicated criteria, which is de-facto addressed by the previous determination.

```python
if role_definition.content_type.model in ('organization', 'team'):
```

This is ugly, but this is how the code answers the question "is the assignment tuple field a Resource UUID or a UUID primary key?"

This patch further confuses what we are objectively doing. We need to fix up the existing criteria, instead of taking the wrong path, and then adding more criteria.

It looks like the _original cause_ of this might be that Authenticator was added as a new RBAC resource, _and_ has a resource registry entry. The easy solution might be to add its model name to `('organization', 'team')`.

Yeah, you could figure out a way to do _better_, but that is the direction of what we should want for this. That place should be how we fall into the right condition.

ansible_base/rbac/role_sync_utils.py
```diff
@@ -66,16 +66,24 @@ def get_ansible_id_or_pk(assignment) -> str:
     return str(ansible_id_or_pk)
 
 
+def _is_resource_registered(model) -> bool:
+    """Check if a model is registered in the resource registry."""
+    from ansible_base.resource_registry.models import ResourceType
+
+    ct = ContentType.objects.get_for_model(model)
+    return ResourceType.objects.filter(content_type=ct).exists()
```
AlanCoding: This makes queries, and there is no need for it to. Because you can use the pre-loaded resource_registry registry once you are starting from a model class.

```python
def _is_resource_registered(model) -> bool:
    """Check if a model is registered in the resource registry."""
    from ansible_base.resource_registry.registry import get_registry

    registry = get_registry()
    if not registry:
        return False
    return model._meta.label in registry.get_resources()
```

ansible_base/rbac/triggers.py
AlanCoding, line 1: I'm relatively unclear whether any changes to this file are intended.

I have been merging thick refactors to this code path recently. And I'm sure that you are likely to hit conflicts if you managed anything around this. However, some of the changes look like a response to the SonarCloud method cognitive complexity rule, which is very much not necessary now, and the basic intent remains unclear.

AlanCoding (approved):
looks great

PR #1078, opened 2026-08-04
AAP-83436 Add unique constraint for global role assignments

lallen92:
## Description
On AAP 2.7, Gateway→Controller resource synchronization can create duplicate global `RoleUserAssignment` records for the same user and role definition. When the JWT consumer processes claims for an affected user, `give_global_permission` calls `get_or_create`, which internally calls get() and raises `MultipleObjectsReturned`. The broad except Exception in `process_rbac_permissions` converts this to an HTTP 401, rejecting a valid JWT.

The root cause is that `unique_together = ('user', 'object_role')` does not prevent duplicates when object_role is NULL (global assignments), because SQL treats NULL != NULL in unique constraints.

## Solution
- Add a partial UniqueConstraint on (user, role_definition) where object_role IS NULL for RoleUserAssignment, and (team, role_definition) where object_role IS NULL for RoleTeamAssignment.
- Migration 0009 deduplicates existing data (batched deletes).
- Migration 0010 auto-generated AddConstraint migration that actually enforces the constraint.
- Add validators = [] to assignment serializer Meta classes to suppress DRF auto-generated UniqueTogetherValidator that conflicts with the user_ansible_id/team_ansible_id alias fields (uniqueness is enforced at the database and get_or_create layers)

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

## Test Plan
E2E Reproduction on AAP 2.7

# 1. User works before duplicates
curl -u testuser-aap83436:TestPass123! .../organizations/
→ HTTP 200, count=1

# 2. Inject duplicate global RoleUserAssignment on Controller
Before: 1 assignment(s)
After:  2 assignment(s)

# 3. User gets 401
curl -u testuser-aap83436:TestPass123! .../organizations/
→ HTTP 401: "Authentication credentials were not provided."

# Controller log:
```
ERROR ansible_base.jwt_consumer.common.auth Unable to validate user permissions - gateway claims fetch or processing failed for user 6d550ddb-...:
  get() returned more than one RoleUserAssignment -- it returned 2!
```
# 4. Apply fix (deduplicate)
Deleted 1 duplicate(s)
After fix: 1 assignment(s)

# 5. User works again
curl -u testuser-aap83436:TestPass123! .../organizations/
→ HTTP 200, count=1

### Required Actions

- [ ] Blocked by PR/MR: #XXX
  

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate global role assignments for the same user or team and role.
  * Existing duplicate assignments are cleaned up automatically during database updates.
  * Repeated assignment requests now behave consistently without creating duplicates.
  * Distinct object-level role assignments continue to be supported.

* **Tests**
  * Added coverage for duplicate prevention, idempotent assignment creation, database enforcement, and API behavior.


ansible_base/rbac/migrations/0009_unique_global_assignments.py
AlanCoding, line 1: This migration file is super custom. I do want to ask the question if this is necessary. I have encountered many many problems similar to this, and solved it by doing multiple migration steps (sometimes as many as 4 or more!). Each step is still atomic, but the intent is still pretty self-obvious and the risk low (go back and then forward again).

My current vibe is that would be much lower risk than what you're doing here. And rolling out a migration to a DAB app is a higher-risk thing, regarding data integrity.

lallen92: Yeah thats fair, tbh it did feel a bit off but was concerned about data getting created inside the small window of deduplication and creation of the constraint, i'll make a note of this going forward, thanks!

ansible_base/rbac/migrations/0009_unique_global_assignments.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dab_rbac', '0008_remote_permissions_cleanup'),
+    ]
+
+    operations = [
+        migrations.RunPython(deduplicate_global_assignments, migrations.RunPython.noop),
```
AlanCoding, line 56: yeah great 👍 totally safe to run forward and backward, changes data but no change to functionality

AlanCoding (approved):

PR #1079, opened 2026-08-04
fix: enforce permissions for changing role assignments

pb82:
## Description

Currently there are two problems with assigning and removing role assignments. 

**First**, check_content_obj_permission only required change permission to assign or remove any role on an object, regardless of what permissions that role contained.

**Second**, the service API's _unassign had no permission check at all.

With this change, we fix both: deletions are now permission-checked, and assigning or removing roles requires the user to hold every object-level permission (change, delete, view, plus any model-specific ones like execute) except add (add must be enforced at an organization level).

For example, a user with only change permissions on a Job Template (but no execute) can assign themselves JT Admin, gaining execute rights they never had.

The test needed to be updated, because the old behavior 

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened authorization checks for managing remote objects, requiring all applicable permissions except “add.”
  * Validated removal permissions before deleting role assignments or synchronizing remote unassignments.
  * Ensured users with only change or view access cannot manage roles.
  * Preserved superuser access while improving permission-denied handling.

* **Tests**
  * Expanded coverage for object-management permissions and role-assignment removal scenarios.

### DIFF
--- a/ansible_base/rbac/service_api/views.py
+++ b/ansible_base/rbac/service_api/views.py
@@ -14,6 +14,7 @@ from ansible_base.rest_filters.rest_framework import ansible_id_backend
 from ansible_base.rest_filters.rest_framework.ansible_id_backend import ServiceFilterBackend
 
 from ..models import DABContentType, DABPermission, RoleTeamAssignment, RoleUserAssignment
+from ..policies import check_can_remove_assignment
 from . import serializers as service_serializers
 
 
@@ -95,6 +96,11 @@ class BaseSerivceRoleAssignmentViewSet(
         if not existing:
             output_serializer = self.get_serializer(existing)
             return Response(output_serializer.data, status=status.HTTP_200_OK)
+
+        # check permissions before removing a role assignment. Previously it was possible
+        # to remove a role assignment for a remote object in Gateway with neither side (Gateway
+        # and the remote side, e.g. Controller) checking if this was allowed.
+        check_can_remove_assignment(request.user, existing)
 
         # Save properties for sync after it is done locally (at which point assignment will not exist)
         role_definition = existing.role_definition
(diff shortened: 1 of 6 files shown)


AlanCoding:
Overlapping with https://github.com/ansible/django-ansible-base/pull/1039

ansible_base/rbac/service_api/views.py
```diff
@@ -96,6 +97,11 @@ def _unassign(self, request):
             output_serializer = self.get_serializer(existing)
             return Response(output_serializer.data, status=status.HTTP_200_OK)
 
+        # check permissions before removing a role assignment. Previously it was possible
+        # to remove a role assignment for a remote object in Gateway with neither side (Gateway
+        # and the remote side, e.g. Controller) checking if this was allowed.
+        check_can_remove_assignment(request.user, existing)
```
AlanCoding, line 103: did this get split into another PR?

AlanCoding (changes requested):
I don't think we should do this.

The criteria expressed in this PR (that you must hold all permissions for the object) is too draconian.

I think the contract of `check_content_obj_permission(request_user, obj)` doesn't really work, we should consider what permissions are being assigned for determining if the assignment is allowed to happen.

I do think that https://github.com/ansible/django-ansible-base/pull/1039 is the right way here, expanding the contract, requiring all _assigned permission_ plus a management permission (the change permission) for that object. The reason that's _not_ complete is that it requires much more thoughtful app-specific reasoning about what permissions a creator should automatically get. That actually still has a high-level feature loophole that still needs to be worked out via real discussion. But the change, I think, is still the right now.

AlanCoding:
> Second, the service API's _unassign had no permission check at all.

that was covered in https://github.com/ansible/django-ansible-base/pull/1080

PR #1084, opened 2026-08-05
[AAP-85471] Handle reverse-sync errors in role assignment/unassignment

djulich:
## Summary
- `maybe_reverse_sync_assignment` and `maybe_reverse_sync_unassignment` let exceptions propagate unhandled, causing the caller's API to return a raw HTML 500 page.
- This wraps both functions in try/except to log a warning instead, matching the pattern already used by `maybe_reverse_sync_object_deletion` in the same file.
- The root cause (a settings mismatch between Controller and Gateway for `ANSIBLE_BASE_ALLOW_TEAM_ORG_MEMBER`) is addressed in a companion Gateway PR — this is the defensive fix ensuring reverse-sync failures never produce unhandled 500s.

Reolves https://redhat.atlassian.net/browse/AAP-85471 (together with https://github.com/ansible/jewel/pull/193)

## Test plan
- [ ] Verify that a reverse-sync failure (e.g., Gateway rejects with 400) results in a logged warning, not an unhandled 500
- [ ] Verify existing `test_creator_permission_does_reverse_sync` still passes
- [ ] Verify role assignment and unassignment still work when reverse sync succeeds

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience during assignment and unassignment synchronization by handling client and synchronization failures gracefully.
  * Added warning logging when synchronization cannot be completed, preventing these errors from interrupting other operations.

### DIFF
--- a/ansible_base/rbac/sync.py
+++ b/ansible_base/rbac/sync.py
@@ -16,6 +16,8 @@ This module handles RBAC-specific reverse-sync scenarios:
 
 import logging
 
+from ansible_base.resource_registry.utils.sync_to_resource_server import get_current_user_resource_client
+
 logger = logging.getLogger('ansible_base.rbac.sync')
 
 
@@ -47,13 +49,10 @@ def maybe_reverse_sync_assignment(assignment):
         return
 
     try:
-        from ansible_base.resource_registry.utils.sync_to_resource_server import get_current_user_resource_client
-
         client = get_current_user_resource_client()
         client.sync_assignment(assignment)
     except Exception as e:
         logger.warning(f"Failed to reverse-sync assignment {assignment}: {e}")
-        return
 
 
 def maybe_reverse_sync_unassignment(role_definition, actor, content_object):
@@ -61,13 +60,10 @@ def maybe_reverse_sync_unassignment(role_definition, actor, content_object):
         return
 
     try:
-        from ansible_base.resource_registry.utils.sync_to_resource_server import get_current_user_resource_client
-
         client = get_current_user_resource_client()
         client.sync_unassignment(role_definition, actor, content_object)
     except Exception as e:
         logger.warning(f"Failed to reverse-sync unassignment for {role_definition}: {e}")
-        return
 
 
 def maybe_reverse_sync_object_deletion(content_object):
--- a/ansible_base/rbac/sync.py
+++ b/ansible_base/rbac/sync.py
@@ -46,20 +46,28 @@ def maybe_reverse_sync_assignment(assignment):
     if not reverse_sync_enabled_all_conditions(assignment):
         return
 
-    from ansible_base.resource_registry.utils.sync_to_resource_server import get_current_user_resource_client
+    try:
+        from ansible_base.resource_registry.utils.sync_to_resource_server import get_current_user_resource_client
 
-    client = get_current_user_resource_client()
-    client.sync_assignment(assignment)
+        client = get_current_user_resource_client()
+        client.sync_assignment(assignment)
+    except Exception as e:
+        logger.warning(f"Failed to reverse-sync assignment {assignment}: {e}")
+        return
 
 
 def maybe_reverse_sync_unassignment(role_definition, actor, content_object):
     if not reverse_sync_enabled_all_conditions(role_definition):
         return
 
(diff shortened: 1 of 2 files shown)


test_app/tests/rbac/test_rbac_cross_service_sync.py
```diff
@@ -271,3 +271,30 @@ def test_sync_url_construction(inventory, enable_reverse_sync):  # noqa: F811
...
+def test_sync_assignment_handles_errors_gracefully(rando, inventory, inv_rd, enable_reverse_sync, exception_cls):  # noqa: F811
+    with enable_reverse_sync():
+        with override_settings(RESOURCE_SERVER={'URL': 'http://gateway.example.com', 'SECRET_KEY': 'test-secret'}):
+            assignment = inv_rd.give_permission(rando, inventory)
+            with patch('ansible_base.resource_registry.utils.sync_to_resource_server.get_current_user_resource_client') as mock_get_client:
+                mock_get_client.return_value.sync_assignment.side_effect = exception_cls("sync failed")
+
+                from ansible_base.rbac.sync import maybe_reverse_sync_assignment
+
+                maybe_reverse_sync_assignment(assignment)
```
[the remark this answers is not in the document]
djulich: Pushed changes which address both points from the coderabbit review:

1. Call assertions — each test now asserts mock_get_client.assert_called_once() and sync_assignment/sync_unassignment.assert_called_once_with(...) to prove the failure path ran, not just that no exception escaped.
2. Client acquisition failure — added two new tests (test_sync_assignment_handles_client_acquisition_error and test_sync_unassignment_handles_client_acquisition_error) where get_current_user_resource_client itself raises, covering the other branch in the try/except.

ansible_base/rbac/sync.py
```diff
@@ -46,20 +46,28 @@ def maybe_reverse_sync_assignment(assignment):
     if not reverse_sync_enabled_all_conditions(assignment):
         return
 
-    from ansible_base.resource_registry.utils.sync_to_resource_server import get_current_user_resource_client
+    try:
+        from ansible_base.resource_registry.utils.sync_to_resource_server import get_current_user_resource_client
```
AlanCoding: Doesn't seem like the import should go in the `try`. Even Sonar is likely to yell at you for this.

ansible_base/rbac/sync.py
```diff
@@ -46,20 +46,28 @@ def maybe_reverse_sync_assignment(assignment):
...
 
-    client = get_current_user_resource_client()
-    client.sync_assignment(assignment)
+        client = get_current_user_resource_client()
```
AlanCoding, line 52: again, need to back out of the `try`. Or else add a new try-except to articulate what error you're looking for

AlanCoding:
> maybe_reverse_sync_assignment and maybe_reverse_sync_unassignment let exceptions propagate unhandled, causing the caller's API to return a raw HTML 500 page.

...isn't returning a 500 in this case and safely rolling back the point?

ansible_base/rbac/sync.py
djulich: Of course you're right - fixed.

ansible_base/rbac/sync.py
djulich: Of course you're right - fixed.

djulich:
> > maybe_reverse_sync_assignment and maybe_reverse_sync_unassignment let exceptions propagate unhandled, causing the caller's API to return a raw HTML 500 page.
> 
> ...isn't returning a 500 in this case and safely rolling back the point?

So the current behavior (given AWX's `ATOMIC_REQUESTS = True`) is:
- 500 + full rollback — neither controller nor gateway has the assignment. Consistent, but the user gets a confusing raw HTML error page.

With the new try/except:
- 201 + gateway out of sync — controller has the assignment, gateway doesn't. The user thinks it succeeded, but the systems are inconsistent.

@AlanCoding, you are right that this is not improving the code!

I would propose to either revoke the PR altogether, or to catch the error and re-raise a proper DRF `APIException` so the user gets a clean JSON error instead of raw HTML.

djulich:
I'm converting this PR to draft since we need to fix the architectural direction first. See also https://redhat-internal.slack.com/archives/C06CM09FQSV/p1785946739478459?thread_ts=1785941475.398889&cid=C06CM09FQSV

PR #1088, opened 2026-08-06
Fix namespace-scoped role assignment sync failing with UUID as integer PK

Funi1234:
## Description

**What is being changed?**

Three functions in \`role_sync_utils.py\` and the pagination loop in \`sync.py\` are updated to correctly handle \`object_ansible_id\` UUIDs for non-organization/team content types.

When Gateway returns \`object_ansible_id\` as a UUID for namespace-scoped role assignments (e.g. \`galaxy.collection_namespace_owner\`), Hub's sync passes it directly to \`model.objects.get(pk=<uuid>)\`. \`Namespace\` uses an integer PK so this raises \`ValueError: Field 'id' expected a number\`, the local \`RoleUserAssignment\` is never created, and the user gets a 403 on collection upload.

A second defect in the same path caused perpetual spurious sync cycles: \`get_ansible_id_or_pk()\` and the bulk resolution helpers only resolved UUIDs via the Resource registry for \`organization\` and \`team\` types, so local tuples carried integer PKs while remote tuples carried UUIDs — the set comparison never matched.

There is also a deeper issue: \`galaxy.Namespace\` is not registered in galaxy_ng's resource registry, so Hub has no Resource rows for namespaces. The Resource registry fixes in \`role_sync_utils.py\` alone are not sufficient — the UUID lookup returns None and we still fall through to the broken path. Confirmed on a live AAP 2.6 environment.

- \`get_content_object()\`: attempt Resource registry lookup for all content types (UUID-validated to avoid UUIDField errors); fall back to integer PK for types with no registry entry
- \`get_ansible_id_or_pk()\`: resolve via Resource for all types, preserving the mandatory-entry guard for org/team
- \`_bulk_resolve_object_ansible_ids()\`: extend to all content types, keyed by \`(object_id, app_label, model)\` to avoid collisions across services sharing a model name
- \`_resolve_object_ansible_id()\`: use the extended map for all types; fall back to raw PK when no registry entry exists
- \`_paginate()\`: batch-resolve which \`object_ansible_id\` UUIDs Hub has Resource rows for per page; fall back to integer \`object_id\` for any UUID Hub doesn't recognise — ensures both sides of the set comparison carry the same identifier regardless of whether the content type is registered

Fixes: [AAP-77392](https://redhat.atlassian.net/browse/AAP-77392)

## Type of Change

## Self-Review Checklist

- [x] I have considered performance implications (batch UUID lookup per page, not per assignment)

### Prerequisites

A running AAP 2.6 environment with Gateway connected to Hub (\`RESOURCE_SERVER\` configured).

### Steps to Test

1. Create a non-superuser and a namespace in Hub
2. Assign \`galaxy.collection_namespace_owner\` to the user on that namespace via Gateway \`POST /api/gateway/v1/role_user_assignments/\`
3. Trigger Hub resource sync: \`pulpcore-manager shell -c "from galaxy_ng.app.tasks.resource_sync import run; run()"\`
4. Check logs for \`assignment_errors\`
5. Attempt a collection upload as the assigned user

### Expected Results

\`assignments_created: 1\`, \`assignment_errors: 0\`. Collection upload succeeds (200, not 403).

### Required Actions

- [x] Requires downstream repository changes — a separate galaxy_ng PR is needed to add \`ResourceConfig(models.Namespace)\` to \`resource_api.py\`. Without it Hub has no Resource rows for namespaces so the UUID path in \`get_content_object\` always misses. The \`_paginate\` fallback covers this gap in the meantime.

### Screenshots/Logs

Before fix:
\`\`\`
ERROR: Failed to create assignment ... galaxy.collection_namespace_owner
ValueError: Field 'id' expected a number but got '<uuid>'
assignments_created: 0 | assignment_errors: 1
\`\`\`

After fix (verified on live AAP 2.6):
\`\`\`
assignments_created: 1 | assignment_errors: 0
\`\`\`

[AAP-77392]: https://redhat.atlassian.net/browse/AAP-77392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role assignment synchronization across registered resource types.
  * Prevented conflicts when different services use the same model names.
  * Added reliable fallback handling for unrecognized UUIDs and legacy numeric identifiers.
  * Improved organization and team assignment handling when resources are missing.
  * Ensured remote assignment pagination consistently resolves resource identifiers.

### DIFF
--- a/ansible_base/rbac/role_sync_utils.py
+++ b/ansible_base/rbac/role_sync_utils.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import uuid as _uuid_module
 from dataclasses import dataclass
 from typing import Any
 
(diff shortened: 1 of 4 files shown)


Funi1234:
Fixed in https://github.com/ansible/django-ansible-base/pull/1072

PR #1089, opened 2026-08-06
[AAP-86325] Pin pytest-django<4.13 in django4 tox environment

tznamena:
## Description

- **What is being changed?**
  Add `pytest-django<4.13` as a dep in the `[testenv:py311-django4,py312-django4]` section of `pyproject.toml`.

- **Why is this change needed?**
  pytest-django 4.13.0 introduced `_pre_setup_ran_eagerly` which relies on Django 5.x internals. The django4 tox environment downgrades to Django 4.2.x after installing the latest pytest-django, causing `AttributeError` at test runtime.

- **How does this change address the issue?**
  Pinning `pytest-django<4.13` only in the django4 environment ensures a compatible version is resolved, while Django 5.x environments remain unconstrained.

## Type of Change

JIRA: AAP-86325

## Summary by CodeRabbit

* **Chores**
  * Updated test environment dependencies to support compatibility with Django 4 testing setups.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -138,6 +138,7 @@ legacy_tox_ini = """
     deps =
         -r{toxinidir}/requirements/requirements_all.txt
         -r{toxinidir}/requirements/requirements_dev.txt
+        pytest-django<4.13
     commands_pre =
         python -m pip install --upgrade "Django>=4.2.21,<4.3.0"
     docker = db


ptoscano:
Why fixing it in the tox configuration? That dependency is specified in requirement files:
```
requirements/requirements_dev.txt:21:pytest-django
requirements/requirements_testing.txt:5:pytest-django
```
So I think it needs to be pinned in these files, otherwise there will be a duplication and disconnection of requirements.

AlexSCorey (approved):

tznamena:
> Why fixing it in the tox configuration? That dependency is specified in requirement files

The cap is scoped to the `[testenv:py311-django4,py312-django4]` section in `pyproject.toml` rather than added to `requirements_dev.txt` or `requirements_testing.txt` because those files are shared across all tox environments. Adding a `pytest-django<4.13` cap there would unnecessarily constrain the Django 5.x test environments (`py311`, `py312`, etc.) which are fully compatible with pytest-django 4.13+.

By placing the pin only in the django4 tox environment's `deps`, we ensure:
- Django 4.2 tests get a compatible pytest-django version
- Django 5.x tests continue using the latest pytest-django
- No shared requirements files need to carry a version-specific workaround

PR #1091, opened 2026-08-06
AAP-85286: Add initialize_resource_registry management command

jay-steurer:
## Summary

- Adds a new `manage.py initialize_resource_registry` management command that populates `ResourceType` and `Resource` records for all registered models
- Fixes a bug where the `post_migrate` backfill (`initialize_resources`) is silently skipped during OCP multi-pass upgrades because `migrations_are_complete()` returns `False` when other Django apps still have pending migrations
- The command is idempotent — safe to run multiple times via `update_or_create` and `bulk_create(ignore_conflicts=True)`

## Problem

On OCP, `manage.py migrate` often runs in multiple passes (init containers). When `dab_resource_registry` migrations complete in the first pass but other apps are still pending, the `post_migrate` signal handler skips the backfill. On subsequent passes, Django doesn't fire `post_migrate` for `dab_resource_registry` (no new migrations), so the backfill never runs. This leaves the resource registry empty, which breaks gateway's `migrate_service_data`.

## Changes

- **New file:** `ansible_base/resource_registry/management/commands/initialize_resource_registry.py` — calls existing `initialize_resources(sender=None)`
- **New file:** `test_app/tests/resource_registry/test_initialize_resource_registry.py` — tests for empty registry population and idempotency

## Test plan

- [x] `test_initialize_resource_registry_populates_empty_registry` — verifies ResourceType and Resource records are created from empty state
- [x] `test_initialize_resource_registry_is_idempotent` — verifies running twice produces same result
- [x] Full `resource_registry` test suite passes (263 tests)

Assisted-by: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Added a command to initialize all registered resource types and records.
  * Displays progress messages when initialization starts and finishes.
  * Initialization can run even when migrations are not marked complete.
  * RBAC assignment fields are added automatically when RBAC is enabled.

* **Bug Fixes**
  * Ensured repeated initialization runs do not create duplicate registry entries.

* **Tests**
  * Added coverage for empty registries, repeated execution, and initialization before migrations are marked complete.

### DIFF
--- a/ansible_base/resource_registry/apps.py
+++ b/ansible_base/resource_registry/apps.py
@@ -10,10 +10,79 @@ import ansible_base.lib.checks  # noqa: F401 - register checks
 from ansible_base.lib.utils.db import ensure_transaction, migrations_are_complete
 from ansible_base.resource_registry.utils.settings import resource_server_defined
 
-logger = logging.getLogger('ansible_base.resource_registry.apps')
+logger = logging.getLogger("ansible_base.resource_registry.apps")
 
 
-def initialize_resources(sender, **kwargs):
+def _sync_resource_types(registry, resource_type_cls, content_type_cls):
+    """Create or update ResourceType rows for every resource in the registry."""
+    for key, resource_config in registry.get_resources().items():
+        content = content_type_cls.objects.get_for_model(resource_config.model)
+
+        if serializer := resource_config.managed_serializer:
+            resource_type = f"shared.{serializer.RESOURCE_TYPE}"
+        else:
+            resource_type = f"{registry.api_config.service_type}.{content.model}"
+        defaults = {
+            "externally_managed": resource_config.externally_managed,
+            "name": resource_type,
+        }
+
+        try:
+            resource_type_cls.objects.update_or_create(content_type=content, defaults=defaults)
+        except IntegrityError as e:
+            # if previous DAB migrations used the wrong content type id, we need to correct that now
+            # to eliminate integrity errors at the end of the migration process when this function
+            # gets called.
+            if not resource_type_cls.objects.filter(name=resource_type).exists():
+                raise e
+            rt = resource_type_cls.objects.get(name=resource_type)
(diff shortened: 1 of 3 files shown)


ansible_base/resource_registry/management/commands/initialize_resource_registry.py
```diff
@@ -0,0 +1,16 @@
+import logging
+
+from django.core.management.base import BaseCommand
+
+from ansible_base.resource_registry.apps import initialize_resources
+
+logger = logging.getLogger("ansible_base.resource_registry.apps")
```
AlanCoding: don't really using the logger.

ansible_base/resource_registry/management/commands/initialize_resource_registry.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class Command(BaseCommand):
+    help = "Populate the resource registry (ResourceType and Resource records) for all registered models. Safe to run multiple times."
```
AlanCoding: nit: it would be good to say here that you shouldn't _ordinarily_ need to run this because within the normal migrations it is expected to run, but this could be useful for repair.

ansible_base/resource_registry/apps.py
```diff
@@ -101,12 +123,23 @@ def proxies_of_model(cls):
...
+        enabled
+        and resource_server_defined()
+        and (
+            "SECRET_KEY" not in settings.RESOURCE_SERVER
+            or not settings.RESOURCE_SERVER["SECRET_KEY"]
+        )
+    ):
+        logger.error(
+            "RESOURCE_SERVER['SECRET_KEY'] is not configured. Reverse sync will not be enabled."
+        )
```
AlanCoding: I don't think any of this is necessary... just a reformatter ran. Hard to tell what's relevant here.

AlanCoding (approved):
This is a fairly humble request. Sure, it's not a bad idea to provide repair commands that will re-run the critical setup logic for various DAB apps. I'm fully on board with that. Left super minor nits, take them or leave them.

PR #1098, opened 2026-08-13
AAP-87884 Explicitly delete role assignments in deferred team cleanup

jay-steurer:
## Summary
- `cleanup_deleted_team_roles` relied solely on CASCADE from `ObjectRole` deletion to remove `RoleUserAssignment` and `RoleTeamAssignment` records
- When team deletions are deferred across a transaction boundary (via `defer_rbac_computations`), the CASCADE chain can miss records, leaving orphaned assignment rows
- Now explicitly deletes `RoleUserAssignment` and `RoleTeamAssignment` after removing ObjectRoles so cleanup is reliable regardless of transaction timing
- Adds a test (`test_defer_rbac_computations_delete_org_cleans_team_assignments`) that verifies `RoleUserAssignment` records for cascade-deleted teams are cleaned up when an org is deleted within `defer_rbac_computations()`

## Context
This is the upstream root-cause fix for a CI failure in aap-gateway stable-2.6 (PR #1584, Views A-R job). The gateway-side defensive fix is in aap-gateway PR #1585.

## Test plan
- [ ] New test `test_defer_rbac_computations_delete_org_cleans_team_assignments` passes
- [ ] Existing deferred RBAC tests still pass (`test_defer_rbac_computations_delete_team_direct_role`, `test_defer_rbac_computations_delete_team_via_org_role`, `test_defer_rbac_computations_delete_org_cleans_team_role`)

Assisted-by: Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of deleted teams and their associated role assignments.
  * Ensured deferred cleanup correctly removes related access records.

* **Tests**
  * Expanded coverage for role-based access, deferred cleanup, bulk permissions, remote permissions, audit isolation, and email enforcement.
  * Added validation for post-migration permission recomputation.

### DIFF
--- a/ansible_base/rbac/caching.py
+++ b/ansible_base/rbac/caching.py
@@ -75,12 +75,24 @@ def bulk_ancestor_roles(team_pks: Iterable[int]) -> set['ObjectRole']:
 
 
 def cleanup_deleted_team_roles(team_pks: set[int]) -> tuple[set['ObjectRole'], set[int]]:
-    """Remove ObjectRoles and evaluations for deleted teams. Called by defer_rbac_computations flush."""
+    """Remove ObjectRoles, evaluations, and role assignments for deleted teams.
+
+    Called by defer_rbac_computations flush. Explicitly deletes
+    RoleUserAssignment/RoleTeamAssignment records rather than relying
+    solely on CASCADE from ObjectRole deletion, because the CASCADE
+    chain can miss records when deletions are deferred across a
+    transaction boundary.
+    """
+    from ansible_base.rbac.models import RoleTeamAssignment, RoleUserAssignment
+
     ancestor_roles = bulk_ancestor_roles(team_pks)
     team_ct_id = permission_registry.team_ct_id
+    str_pks = [str(pk) for pk in team_pks]
     RoleEvaluation.objects.filter(content_type_id=team_ct_id, object_id__in=team_pks).delete()
     deleted_or_ids = set(ObjectRole.objects.filter(content_type_id=team_ct_id, object_id__in=team_pks).values_list('id', flat=True))
     ObjectRole.objects.filter(id__in=deleted_or_ids).delete()
+    RoleUserAssignment.objects.filter(content_type_id=team_ct_id, object_id__in=str_pks).delete()
+    RoleTeamAssignment.objects.filter(content_type_id=team_ct_id, object_id__in=str_pks).delete()
     eval_model = get_evaluation_model(permission_registry.team_model)
     eval_model.objects.filter(content_type_id=team_ct_id, object_id__in=team_pks).delete()
     return ancestor_roles, deleted_or_ids
--- a/test_app/tests/rbac/test_triggers.py
+++ b/test_app/tests/rbac/test_triggers.py
@@ -6,10 +6,20 @@ from django.test.utils import override_settings
 from rest_framework.exceptions import ValidationError
 
 from ansible_base.rbac.caching import compute_team_member_roles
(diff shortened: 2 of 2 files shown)


Dostonbek1 (commented):
## Code Review Summary

This is the **correct root-cause fix** for AAP-87884. The bug is real: when org/team deletes run inside `defer_rbac_computations()`, `cleanup_deleted_team_roles()` removes `ObjectRole` rows but `RoleUserAssignment`/`RoleTeamAssignment` rows for those teams can survive because the CASCADE chain does not reliably fire across the deferred flush boundary.

Explicitly deleting assignments by `(content_type_id, object_id)` after `ObjectRole` deletion is the right layer to fix this — it covers **all** deferred-delete paths (org cascade, direct team delete, bulk operations), not just gateway's org API.

**Files Reviewed**: 2 files (meaningful change in `caching.py`; test churn in `test_triggers.py`)

### Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 1 code quality suggestion

### Overall Assessment
LGTM — merge this **before** the gateway defensive PR ([jewel#201](https://github.com/ansible/jewel/pull/201)).

### General Feedback
- The new `test_defer_rbac_computations_delete_org_cleans_team_assignments` test directly reproduces the JIRA scenario (team-member assignment on cascade-deleted team within `defer_rbac_computations`) and complements the existing org-scoped-role test.
- Using `str_pks` for assignment `object_id` filters is correct (`AssignmentBase.object_id` is `TextField`).
- Deletion order (ObjectRoles first, then explicit assignment cleanup) is sound — the explicit pass catches orphans that CASCADE missed.
- The large formatting diff in `test_triggers.py` is noisy but harmless; consider keeping functional diffs separate from style-only changes in future commits.
- Companion gateway PR #201 becomes optional once consumers pick up this DAB release, but is still reasonable for stable-2.6 backports that cannot immediately bump DAB.

ansible_base/rbac/caching.py
```diff
@@ -75,12 +75,24 @@ def bulk_ancestor_roles(team_pks: Iterable[int]) -> set['ObjectRole']:
...
+    """
+    from ansible_base.rbac.models import RoleTeamAssignment, RoleUserAssignment
+
     ancestor_roles = bulk_ancestor_roles(team_pks)
     team_ct_id = permission_registry.team_ct_id
+    str_pks = [str(pk) for pk in team_pks]
     RoleEvaluation.objects.filter(content_type_id=team_ct_id, object_id__in=team_pks).delete()
     deleted_or_ids = set(ObjectRole.objects.filter(content_type_id=team_ct_id, object_id__in=team_pks).values_list('id', flat=True))
     ObjectRole.objects.filter(id__in=deleted_or_ids).delete()
+    RoleUserAssignment.objects.filter(content_type_id=team_ct_id, object_id__in=str_pks).delete()
```
Dostonbek1, line 94: Consider also deleting assignments **before** `ObjectRole` deletion as a belt-and-suspenders alternative — not required for this fix, but worth noting: if CASCADE partially fires and removes some assignments while leaving others, the current order still works because this filter is keyed on team `(content_type_id, object_id)`, not `object_role_id`.

One edge case to be aware of (acceptable): if `count > len(results)` pagination logic elsewhere skips migration, here we delete by the full `team_pks` set passed in, so we're not paginated — good.

Optional follow-up: a focused unit test calling `cleanup_deleted_team_roles()` directly (not via full `defer_rbac_computations` org delete) for direct team delete would tighten coverage, though existing deferred-delete team tests may already cover it.

PR #1102, opened 2026-08-17
Add query count scaling regression test for role-team-access

larrymou9:
Follows up on [AAP-78658](https://redhat.atlassian.net/browse/AAP-78658). `test_team_access_list_query_count` mirrors the existing `test_user_access_list_query_count` test, since both endpoints share `AccessListMixin`. Confirms 0 extra queries going from 2 to 20 teams in the result — no regression here.

## How to verify locally

```bash
pytest test_app/tests/rbac/api/test_access_lists.py::test_team_access_list_query_count -v
```

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify that expanding team access-list results does not cause excessive database queries.
  * Helps protect performance when displaying larger sets of teams.

[AAP-78658]: https://redhat.atlassian.net/browse/AAP-78658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test_app/tests/rbac/api/test_access_lists.py
+++ b/test_app/tests/rbac/api/test_access_lists.py
@@ -295,3 +295,40 @@ def test_user_access_list_query_count(admin_api_client, inv_rd, inventory):
         f"({queries_small} -> {queries_large}). "
         f"object_role_assignments may be doing per-user queries."
     )
+
+
+@pytest.mark.django_db
+def test_team_access_list_query_count(admin_api_client, inv_rd, inventory):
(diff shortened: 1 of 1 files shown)


AlanCoding (approved):
So I think making 20 items in a single test to count queries is probably fine for now. But if these continue to proliferate, at some point we probably need a different strategy so it doesn't blow up our runtime and we can continue to test with increasingly larger sets and more diverse data.

PR #1105, opened 2026-08-17
AAP-87084 — Add actor_role_ids method, deprecate _actor_role_filter

AlanCoding:
## Summary
- Add `actor_role_ids(actor)` to `RoleEvaluationFields` — returns a queryset of ObjectRole PKs, replacing the `_actor_role_filter` pattern that returned a dict of kwargs to be `**`-splatted
- All internal callers updated to use `role_id__in=cls.actor_role_ids(actor)` 
- `_actor_role_filter` kept as a deprecated wrapper for backwards compatibility (AWX calls it directly)

No external callers found in gateway, EDA, or galaxy_ng.

## Test plan
- [x] Full RBAC test suite passes (661 tests)
- [x] flake8 clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role-based permission evaluation for users and teams.
  * Corrected permission visibility, accessible-object checks, permission lookups, and object-permission validation.
  * Improved consistency when determining which objects and permissions are available based on assigned roles.
* **Compatibility**
  * Preserved support for existing role-filtering integrations and workflows.

### DIFF
--- a/ansible_base/rbac/models/role.py
+++ b/ansible_base/rbac/models/role.py
@@ -823,7 +823,7 @@ class RoleEvaluationFields(models.Model):
     @staticmethod
     def _actor_role_filter(actor):
         """Deprecated — use ``actor_role_ids`` instead."""
-        return {'role_id__in': RoleEvaluationFields.actor_role_ids(actor)}
+        return {'role_id__in': RoleEvaluationFields.actor_role_ids(actor)}  # pragma: no cover
 
     @classmethod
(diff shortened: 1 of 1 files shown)


AlanCoding:
I made this because I didn't like the proliferation of references to `_actor_role_filter` in AWX.

AlanCoding:
so the code coverage is because of the method we're only keeping for merge order compatibility. So it seems obviously the right thing to ignore that here.

PR #1103, opened 2026-08-17
Add failing query count regression test for resource-list N+1 (AAP-88287)

larrymou9:
Follows up on [AAP-78658](https://redhat.atlassian.net/browse/AAP-78658) by extending the query-count-scaling test pattern (introduced in #1096) to the Resources API.

`test_resources_list_extra_fields_query_count` covers `resource-list?extra_fields=resource_data`. **This test currently FAILS on purpose** — it documents a real N+1: `ResourceDataField.to_representation()` accesses `resource.content_object` per row (a GenericForeignKey not covered by `ResourceViewSet`'s `select_related`), costing one extra query per resource in the page (11 queries for 2 resources → 29 queries for 20 resources, a delta of +18 for +18 resources).

## Bug filed

Tracked in **[AAP-88287](https://redhat.atlassian.net/browse/AAP-88287)** ("Resources API: `resource-list?extra_fields=resource_data` does N+1 `content_object` lookups per row"), linked as Related to the [AAP-78658](https://redhat.atlassian.net/browse/AAP-78658) epic, consistent with sibling N+1 tickets under that epic (AAP-79004, AAP-72968, AAP-68016/17/18).

## How to verify locally

```bash
pytest test_app/tests/resource_registry/test_resources_api.py::test_resources_list_extra_fields_query_count -v
```

Expected to fail (`AssertionError: ... 18 < 5`) until [AAP-88287](https://redhat.atlassian.net/browse/AAP-88287) is fixed.

[AAP-88287]: https://redhat.atlassian.net/browse/AAP-88287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Tests**
  * Added coverage ensuring resource lists with extra fields maintain consistent query performance as result size increases.
  * Added safeguards against excessive per-resource database queries.

[AAP-78658]: https://redhat.atlassian.net/browse/AAP-78658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[...]

PR #1109, opened 2026-08-19
AAP-88287 Prefetch content_object in ResourceViewSet to fix N+1

AlanCoding:
## Summary
- Add `prefetch_related("content_object")` to `ResourceViewSet.queryset`
- Fixes N+1 queries when `?extra_fields=resource_data` is requested — each resource's `content_object` (GenericForeignKey) was loaded individually
- Django's GFK prefetch groups by content_type, reducing N extra queries to one per distinct type

## Test plan
- [ ] Verified with the regression test from #1103 — query count delta drops from 18 to within threshold
- [ ] CI passes (existing resource_registry tests unaffected)

## Summary by CodeRabbit

* **Performance**
  * Improved resource loading efficiency when displaying associated content.
  * Reduced unnecessary database requests for resource details.
  * Associated resource content is loaded more efficiently when requested through the resource details view.


ansible_base/resource_registry/views.py
```diff
@@ -102,7 +102,7 @@ class ResourceViewSet(
 
     resource_purpose = "resources indexed from connected AAP services for cross-service resource management"
 
-    queryset = Resource.objects.select_related("content_type__resource_type").all()
+    queryset = Resource.objects.select_related("content_type__resource_type").prefetch_related("content_object").all()
```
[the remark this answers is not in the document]
AlanCoding: ok I meant to do that but somehow it went wrong.

larrymou9:
Hello @AlanCoding once this is merge, the following test will pass as well https://github.com/ansible/django-ansible-base/pull/1103 and  https://github.com/ansible/django-ansible-base/pull/1110.

larrymou9 (approved):

ansible_base/resource_registry/views.py
```diff
@@ -112,6 +112,13 @@ def get_serializer_class(self):
 
         return super().get_serializer_class()
 
+    def get_queryset(self):
+        qs = super().get_queryset()
+        extra = self.request.query_params.get("extra_fields", "")
+        if "resource_data" in extra.split(","):
```
john-westcott-iv, line 118: Optional, non-blocking: this duplicates the `extra_fields`/`resource_data` parsing already in `ResourceListSerializer.__init__` (`serializers.py`). Not a bug today — both sides split on `,` and check membership the same way — but if a new value is ever added to `ALLOWED_EXTRA_FIELDS` without updating this check, the queryset-level prefetch could silently fall out of sync with what the serializer actually exposes. Consider checking membership against `ResourceListSerializer.ALLOWED_EXTRA_FIELDS` here instead of hardcoding `"resource_data"`, as a follow-up if you feel like it.

john-westcott-iv (approved):
Looks good — this is a minimal, correct fix for the N+1 (`prefetch_related` on a `GenericForeignKey` groups the follow-up query by content type, which is the right approach here). One optional, non-blocking suggestion inline — doesn't need to hold up this PR.

ansible_base/resource_registry/views.py
AlanCoding: I don't quite follow what the action would be here.

> Consider checking membership against ResourceListSerializer.ALLOWED_EXTRA_FIELDS here instead of hardcoding "resource_data"

That doesn't work in any obvious way. Even if `ALLOWED_EXTRA_FIELDS` changed, we would still be looking for "resource_data" here specifically. I mean, I can kind of see the sentiment to combine the two loops, but they are in quite different places and that's not obvious to me either.

PR #1110, opened 2026-08-21
AAP-88856 + AAP-88875: Add parallel CI job and non-transactional shared data-seeding for query-count-scale tests

larrymou9:
## Summary

Combines two phases of [AAP-88874](https://redhat.atlassian.net/browse/AAP-88874) into a single PR (previously split across this PR + a stacked fork PR, which was confusing to review as two separate, dependent PRs). Now combined here into a single PR.

### Phase 1 — [AAP-88856](https://redhat.atlassian.net/browse/AAP-88856): parallel CI job scaffold

- Adds a new `query-count-scale` GitHub Actions job that runs on a single Python version (3.12) in **parallel** with the existing `tox` job matrix, so it doesn't extend current PR check times.
- Adds a new `py312-query-count-scale` tox environment targeting a new, isolated test directory: `test_app/tests/query_count_scale/`.
- That tox env deliberately **omits `pytest-xdist` (`-n auto`)** — this directory seeds one shared dataset per run rather than using per-test DB transactions, so xdist workers would multiply seeding cost instead of parallelizing it.
- Excludes `test_app/tests/query_count_scale/` from the main `py312`/`py312-sqlite`/`py312-in-files` tox envs (`--ignore=...`), so this directory's non-transactional, committed data never leaks into the existing ~3,700 unit tests' workspace.

### Phase 2 — [AAP-88875](https://redhat.atlassian.net/browse/AAP-88875): shared, non-transactional data seeding

- Expands `create_large()` (the `LARGE=1` path of `create_demo_data`) per Alan Rominger's guidance: switches permission assignment from a nested per-call `give_permission()` loop to a single `bulk_give_permissions()` call, and adds richer assignment diversity (direct object-level, org-level, team-mediated, and mixed combinations) across a broader set of resource types (orgs/teams/users/inventories/credentials/role definitions). Locally seeds ~150 orgs/380 users/2,000 assignments in ~13-25s, well under the ~1 minute CI budget.
- Adds a session-scoped, autouse `conftest.py` fixture in `test_app/tests/query_count_scale/` that seeds this dataset once per test session, outside of any per-test transaction.
- **Deliberately avoids `@pytest.mark.django_db`/pytest-django's default per-test isolation** in this directory: wrapping each test in pytest-django's `atomic()`/savepoint isolation silently inflates query counts when the code under test opens its own nested `atomic()` (e.g. `ansible_base/rbac/caching.py`'s permission-recomputation batches used by `bulk_give_permissions()`). Instead, a session-scoped `_unblocked_db` fixture calls `setup_databases()`/`teardown_databases()` directly and keeps DB access open for the whole session, with local idempotent counterparts to `admin_user`/`local_authenticator`/`admin_api_client` so they don't pull pytest-django's `db` fixture back in.

### Benchmark test (proves the new pattern catches regressions)

Includes `test_resource_list_extra_fields_query_count_hard_cutoff`, a proof-of-concept for the new **hard-cutoff** pattern ("a request to this endpoint must not exceed N queries" + "list must be non-empty"), replacing the old 2-vs-20-object delta comparison used in #1103, for the same known bug (AAP-88287). Exercises this against the full seeded dataset (~700 resources instead of ~27).

**⚠️ This test is expected to FAIL on this PR** — same as #1103's equivalent old-pattern test — because it exercises a real, currently-unfixed N+1 bug (`ResourceDataField.to_representation()` doing a per-resource `content_object` query). Validated locally: fails on unfixed `devel` (53 queries for 683 resources, exceeds the 15-query cutoff), passes with #1109's fix applied locally. It'll go green once [#1109](https://github.com/ansible/django-ansible-base/pull/1109) merges to `devel` and this branch is rebased — no further code changes expected at that point.

The generalized, parameterized version of this coverage across more endpoints (AAP-88876) is tracked separately as a follow-on story under AAP-88874.

[AAP-88874]: https://redhat.atlassian.net/browse/AAP-88874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AAP-88856]: https://redhat.atlassian.net/browse/AAP-88856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AAP-88875]: https://redhat.atlassian.net/browse/AAP-88875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Testing**
  - Added large-scale query-count regression coverage for list endpoints.
  - Added seeded datasets and authentication fixtures to support performance testing.
  - Added coverage artifact collection and reporting for the query-count test suite.
  - Updated large demo-data tests to skip unsupported SQLite scenarios.

- **Chores**
  - Expanded demo data generation with inventory and credential records.
  - Added broader permission assignments to large demo datasets.
  - Added mypy to the lint configuration.


larrymou9:
@coderabbitai Re: the "not merge-ready" risk note about not asserting the seeded resource count — this is intentional and already addressed in the follow-on PR.

This PR (AAP-88856) is phase 1 of the new query-count-scale architecture: it just scaffolds the CI job/tox env and includes a small benchmark test seeding ~20 resources inline as a proof-of-concept for the new hard-cutoff pattern (see PR description).

Phase 2 (AAP-88875, stacked on this branch — see https://github.com/larrymou9/django-ansible-base/pull/1) replaces that inline seeding with a much larger shared, session-scoped dataset (~700+ resources) and deliberately changes this same test's assertion from "count == N seeded" to "count > 0". That's per explicit guidance from this session's design discussion: with a shared dataset that other tests may also add to, asserting a non-empty list (proof the seeding step worked) is the right check — an exact count would be brittle and isn't the point of the hard-cutoff pattern (the query-count assertion itself is what catches the N+1 regression, not the row count).

So no code change needed here; this will resolve naturally once AAP-88875 lands on top. Flagging for human reviewers in case it's useful context.

larrymou9:
@AlanCoding @john-westcott-iv — this is phase 1 (AAP-88856) of the new query-count-scale test architecture from our discussion with Elijah (see epic AAP-88874). CI is green other than the benchmark test, which is expected to fail until #1109 merges (documented in the PR description). Would appreciate a look when you have a moment — no rush given everything else going on. Follow-on phases (AAP-88875 shared data seeding, AAP-88876 generalized coverage) are ready to stack on top once this lands.

larrymou9:
Folded the AAP-88875 (shared, non-transactional data-seeding) work directly into this branch/PR instead of keeping it as a separate stacked PR — this now covers both AAP-88856 and AAP-88875 in one place. Closing the previously-separate fork PR accordingly.

.github/workflows/ci.yml
```diff
@@ -87,6 +87,51 @@ jobs:
...
+        run: pip3.12 install tox tox-docker
+
+      - name: Run tox
+        env:
+          TOX_DOCKER_GATEWAY: "127.0.0.1"
+        run: |
+          echo "::remove-matcher owner=python::"  # Disable annoying annotations from setup-python
+          tox --colored yes -e py312-query-count-scale
+
+      - name: Upload coverage as artifact
```
john-westcott-iv: On line 56 we inject the PR number into the coverage report. Not sure if we want to do the same here.

test_app/tests/query_count_scale/conftest.py
```diff
@@ -0,0 +1,249 @@
...
+        yield
+        if not django_db_keepdb:
+            try:
+                teardown_databases(db_cfg, verbosity=request.config.option.verbose)
+            except Exception as exc:  # pragma: no cover -- mirrors django_db_setup's own handling
+                request.node.warn(pytest.PytestWarning(f"Error when trying to teardown test databases: {exc!r}"))
+
+
+@pytest.fixture(scope='session')
+def _seed_large_dataset(_unblocked_db):
```
john-westcott-iv: Do you need this as a separate function? Can you not just call the command before the yield in the _unblocked_db fixture?

.github/workflows/ci.yml
john-westcott-iv, line 123: Missing the "Inject PR number into coverage.xml" step that the main tox job has (line 56). `sonar-pr.yml` downloads every `coverage-.*` artifact and pulls the PR number from whichever `coverage.xml` `find` returns first — if that's this job's report, SonarCloud PR analysis hard-fails (`exit 1`, "PR number not found"). `find`'s order across sibling artifact dirs isn't guaranteed alphabetical, so this is a live landmine, not a cosmetic gap. Recommend adding the same injection step here.

test_app/tests/query_count_scale/conftest.py
```diff
@@ -0,0 +1,249 @@
...
+
+    # Check if seeding already completed successfully by validating all expected counts
+    expected = settings.DEMO_DATA_COUNTS
+    actual_orgs = Organization.objects.filter(name__startswith='large_').count()
+    actual_users = get_user_model().objects.filter(username__startswith='large_user_').count()
+    actual_teams = Team.objects.filter(name__startswith='large_team_').count()
+    actual_rds = RoleDefinition.objects.filter(name__startswith='Large Role Definition').count()
+
+    # Also check inventory/credential if they're in DEMO_DATA_COUNTS
+    actual_inventories = Inventory.objects.filter(name__startswith='large_inventory_').count() if 'inventory' in expected else 0
```
john-westcott-iv: `_seed_large_dataset`'s idempotency check validates org/user/team/roledefinition/inventory counts before skipping reseed, but never checks `credential` count even though `DEMO_DATA_COUNTS` now includes `credential: 60`. A local `--reuse-db` database that predates this PR's credential addition would silently skip reseeding forever, leaving `large_credentials` permanently empty. CI is unaffected (fresh DB every run) — this only bites local dev with `--reuse-db`. Suggest adding an `actual_credentials` check mirroring `actual_inventories` above.

test_app/tests/query_count_scale/conftest.py
```diff
@@ -0,0 +1,249 @@
...
+        for i in range(30):
+            AuthenticatorMap.objects.create(
+                name=f'large_map_{i}',
+                authenticator=local_authenticator,
+                map_type='allow',
+            )
+
+
+@pytest.fixture(scope='session')
+def local_authenticator(_unblocked_db):
```
john-westcott-iv: The existing function-scoped `local_authenticator` (`ansible_base/lib/testing/fixtures.py:78`) and this new session-scoped one differ only in scope + how they access the DB — `get_or_create()` is safe for both (under per-test rollback it behaves identically to `.create()`, since no prior row survives to be found). Suggest extracting the shared creation logic into `ansible_base/lib/testing/fixtures.py` so this suite is not carrying its own private copy, and any future non-transactional suite can reuse it directly:

```python
def _get_or_create_local_authenticator():
    from ansible_base.authentication.models import Authenticator
    authenticator, _ = Authenticator.objects.get_or_create(
        name="Test Local Authenticator",
        defaults=dict(
            enabled=True,
            create_objects=True,
            remove_users=False,
            type="ansible_base.authentication.authenticator_plugins.local",
            configuration={},
        ),
    )
    return authenticator


@pytest.fixture
def local_authenticator(db):
    return _get_or_create_local_authenticator()


@pytest.fixture(scope="session")
def session_local_authenticator(_unblocked_db):
    return _get_or_create_local_authenticator()
```

Same pattern applies below to `admin_user`/`admin_api_client` — see those comments.

test_app/tests/query_count_scale/conftest.py
```diff
@@ -0,0 +1,249 @@
...
+            remove_users=False,
+            type='ansible_base.authentication.authenticator_plugins.local',
+            configuration={},
+        ),
+    )
+    return authenticator
+
+
+@pytest.fixture(scope='session')
+def admin_user(_unblocked_db):
```
john-westcott-iv: Same pattern as suggested on `local_authenticator` above: extract the get-or-create-superuser-named-"admin" logic into a `_get_or_create_admin_user()` helper in `ansible_base/lib/testing/fixtures.py`, then add a `session_admin_user(_unblocked_db)` fixture that just calls it. One difference from `local_authenticator`: the function-scoped counterpart here is pytest-django's built-in `admin_user`, not something defined in this repo — so there's no existing in-repo fixture to pair the helper with, this would just introduce the helper fresh for the session-scoped case.

test_app/tests/query_count_scale/conftest.py
```diff
@@ -0,0 +1,249 @@
...
+        user_data = {'password': 'password', username_field: username}
+        if 'email' in user_model.REQUIRED_FIELDS:
+            user_data['email'] = 'admin@example.com'
+        user = user_model._default_manager.create_superuser(**user_data)
+
+    return user
+
+
+@pytest.fixture(scope='session')
+def admin_api_client(_unblocked_db, admin_user, local_authenticator):
```
john-westcott-iv: Same pattern again, and this one does have a direct in-repo sibling to unify with (`ansible_base/lib/testing/fixtures.py:98`). Extract the login logic into a shared helper, e.g. `_login_admin_api_client(user, authenticator) -> (client, login_ok)`, have the existing function-scoped `admin_api_client` call it and discard `login_ok` (safe to ignore there — a stale user cannot survive per-test rollback), and add `session_admin_api_client(_unblocked_db, session_admin_user, session_local_authenticator)` that calls it and asserts `login_ok` (matching this PR's existing rationale for asserting under `--reuse-db`).

Separately: since this client is shared across every test in the session with no per-test reset, if any test ever calls `.logout()` on it (the original fixture's own docstring calls this out as a real scenario — testing the logout signal), every subsequent test in the session would silently run unauthenticated with nothing to catch it. Worth adding a one-line comment/docstring note here along the lines of "no test in this directory may call `.logout()` on this shared client — use a throwaway `APIClient()` for logout-testing instead" so a future contributor does not hit this the hard way.

john-westcott-iv (commented):
Reviewed together with @john-westcott-iv. Two concrete findings posted inline (SonarCloud PR-number injection gap, and a missing `credential` count in the seed idempotency check), plus three inline suggestions about extracting shared session-scoped fixture logic into `ansible_base/lib/testing/fixtures.py`, and a non-blocking design discussion below.

**Design discussion (non-blocking):** The current pattern needs a hand-tuned `max_queries` cutoff per endpoint (14+ magic numbers already), which has to be re-tuned any time `DEMO_DATA_COUNTS` changes, and conflates "cost of RBAC/permission-filtering" with "cost of serializing each row" into one opaque number.

For the pure serializer-level N+1 cases (most of these, including the currently-disabled ones), consider sweeping page size against the same endpoint instead — e.g. `page_size=1, 1, 5, 20, 50` (the doubled `1` gives a stable floor while sidestepping edge cases where an empty/near-empty queryset short-circuits differently) — and asserting query count stays flat as page size grows, rather than comparing to an absolute cutoff. That targets the N+1 signal directly and should let most of the current ~13-25s seed (dominated by the 2,000-assignment `bulk_give_permissions()` call) shrink substantially, since page-size-driven N+1s don't need a large backing table. The exception is the RBAC-assignment-list cases (`roleuserassignment-list`, `roleteamassignment-list`, disabled `service*assignment-list`), which specifically test permission-filtering cost against total assignment volume and would still need real scale there.

Bonus idea: if the sweep-and-assert-flat pattern were pulled into a small generic pytest fixture (e.g. `assert_query_count_flat(client, url, page_sizes=(1,1,5,20,50))`), it wouldn't be specific to django-ansible-base's own endpoints — services consuming dab (Controller, EDA, Hub, etc.) could reuse it to guard their own list endpoints against N+1s instead of reinventing this per-service.

One open question we couldn't resolve from the diff: whether `resource-list`'s view actually honors `page_size` the way the other endpoints do — the commit history implies it needed the full ~700-object dataset to expose the extra_fields N+1, which might mean it paginates differently. Worth checking before assuming the sweep transfers uniformly.

Not requesting changes to this PR for the discussion item — flagging for the epic (AAP-88874) / a possible follow-up story.

.github/workflows/ci.yml
larrymou9: Fixed in 488d293 — added the same `Inject PR number into coverage.xml` step used by the main job.

test_app/tests/query_count_scale/conftest.py
larrymou9: Fixed in 488d293 — the idempotency check now validates every model in `DEMO_DATA_COUNTS` (including `credential`) via a dict comprehension, so future keys added there are automatically covered too.

test_app/tests/query_count_scale/conftest.py
larrymou9: Done in af14581 — extracted `_get_or_create_local_authenticator()` into `ansible_base/lib/testing/fixtures.py`; both the existing function-scoped `local_authenticator` and this suite's renamed `session_local_authenticator` now call it.

test_app/tests/query_count_scale/conftest.py
larrymou9: Done in af14581 — added `_get_or_create_admin_user()` to `ansible_base/lib/testing/fixtures.py` as you noted, fresh for this case since there's no in-repo sibling to pair it with. `session_admin_user` calls it.

test_app/tests/query_count_scale/conftest.py
larrymou9: Done in af14581 — extracted `_login_admin_api_client(user, authenticator) -> (client, login_ok)`. The function-scoped `admin_api_client` calls it and ignores `login_ok`; the renamed `session_admin_api_client` calls it and asserts `login_ok`. Also added the docstring note about never calling `.logout()` on the shared session client, using a throwaway `APIClient()` instead for logout tests.

larrymou9:
@john-westcott-iv Thanks for the thorough review! Addressed all 5 inline points (replied on each thread with the specific commit):

**Concrete fixes:**
1. Added the missing PR-number injection step to `py312-query-count-scale` — 488d293
2. Idempotency check now validates every `DEMO_DATA_COUNTS` key (including `credential`) — 488d293

**Refactor (shared fixture logic):**
3-5. Extracted `_get_or_create_local_authenticator()`, `_get_or_create_admin_user()`, and `_login_admin_api_client()` into `ansible_base/lib/testing/fixtures.py`. Both the existing function-scoped fixtures and this suite's renamed `session_local_authenticator`/`session_admin_user`/`session_admin_api_client` now share the same logic. Also added the docstring warning about never calling `.logout()` on the shared session client.

Ran the full `query_count_scale` suite plus the broader `test_app/tests/authentication/` suite locally against fresh Postgres to confirm no regressions from the refactor.

Re: the design discussion on page-size sweeps vs. hard cutoffs — agreed that's a good idea, flagging it for a follow-up under AAP-88874 rather than scoping it into this PR.

test_app/tests/query_count_scale/conftest.py
```diff
@@ -0,0 +1,214 @@
...
+        'organization': Organization.objects.filter(name__startswith='large_').count(),
+        'user': get_user_model().objects.filter(username__startswith='large_user_').count(),
+        'team': Team.objects.filter(name__startswith='large_team_').count(),
+        'roledefinition': RoleDefinition.objects.filter(name__startswith='Large Role Definition').count(),
+        'inventory': Inventory.objects.filter(name__startswith='large_inventory_').count(),
+        'credential': Credential.objects.filter(name__startswith='large_credential_').count(),
+    }
+
+    # Skip seeding if every count present in DEMO_DATA_COUNTS already matches
+    if all(actual_counts.get(key, 0) == count for key, count in expected.items() if key in actual_counts):
```
john-westcott-iv: Nice generalization beyond what we suggested — but the `if key in actual_counts` filter means a future `DEMO_DATA_COUNTS` key with no matching `actual_counts` entry gets silently skipped rather than flagged, reproducing the same class of bug this fixes (just for a hypothetical future resource type instead of `credential`). Suggest a set-comparison guard so it fails loudly instead:

```python
missing = expected.keys() - actual_counts.keys()
assert not missing, (
    f"DEMO_DATA_COUNTS has keys with no corresponding count check in _seed_large_dataset: "
    f"{sorted(missing)}. Add a query for each new resource type above."
)

if all(actual_counts[key] == count for key, count in expected.items()):
    return
```

Once `missing` is asserted empty, the `.get(key, 0)` default and `if key in actual_counts` filter in the original `all(...)` are no longer needed — every expected key is guaranteed present.

test_app/tests/query_count_scale/test_list_query_count.py
```diff
@@ -0,0 +1,87 @@
...
+# AuthenticatorMaps. Page size is 50, so counts are per-page, not per-total-rows.
+#
+# Passing (regression guards, already well-optimized): resource-list,
+# organization-list, roleuserassignment-list, team-list, user-list,
+# inventory-list, roledefinition-list, roleteamassignment-list,
+# authenticator-list, resourcetype-list, dabcontenttype-list,
+# dabpermission-list, activitystream-list, aap_flags_states-list.
+#
+# Failing (real N+1s found this way -- disabled below until fixed, see TODOs).
+QUERY_COUNT_CASES = [
```
john-westcott-iv, line 34: These 6 disabled cases (the `extra_fields` one plus the 5 filed under AAP-92618/92626/92627/92628) have no mechanism ensuring they get re-enabled once their underlying bugs are fixed — we checked, and none of the linked JIRA tickets have "uncomment this test case" as acceptance criteria. A commented-out block is easy to forget forever; nothing signals when it's safe (or necessary) to bring back.

Suggest `pytest.mark.xfail(strict=True)` instead of commenting these out:

```python
pytest.param(
    'resource-list', {'extra_fields': 'resource_data'}, 15,
    marks=pytest.mark.xfail(reason="AAP-88287 N+1 in ResourceDataField.to_representation(); fixed by #1109", strict=True),
    id='resource_list_with_extra_fields',
),
```

With `strict=True`, the case keeps running and reports `xfail` (non-blocking) as long as it still fails — but the moment the underlying fix lands and the case starts *passing*, `strict=True` turns that into a hard CI failure on `devel`, forcing someone to notice and remove the marker. That's a self-enforcing signal instead of relying on someone remembering to revisit a comment or a JIRA ticket that doesn't even reference this test file.

john-westcott-iv (commented):
Re-reviewed after the update — all three findings from the previous pass (SonarCloud PR-number injection, credential idempotency check, fixture extraction) landed correctly, with the credential check actually generalized beyond what we suggested. No new issues in the diff itself. Two follow-up suggestions posted inline: tightening the new completeness check with a set-comparison guard, and converting the commented-out `QUERY_COUNT_CASES` entries to `strict=True` xfail so they can't be silently forgotten.

Separately, worth noting: SonarCloud's quality gate is currently failing (`new_coverage = 0%`), but this traces to a coverage.xml parse failure unrelated to this PR's diff (`ansible_base/authentication/utils/authenticator_map.py` line-count mismatch across all three coverage reports) — looks like a pipeline/checkout timing issue, not something introduced here.

john-westcott-iv (approved):
Approving with minor comments pending CI going green.

PR #1111, opened 2026-08-24
AAP-82996 Add bulk RBAC assignment signals

AlanCoding:
## Summary

Add two new Django signals to the RBAC pipeline to support bulk processing of assignment changes:

- **`dab_rbac_assignments_created`**: Fires after `bulk_give_permissions` creates assignments
- **`dab_rbac_assignments_pre_delete`**: Fires before `bulk_remove_permissions` deletes assignments

Both signals provide `assignment_data` as a list of `(assignment, content_object)` tuples, pairing each `RoleUserAssignment`/`RoleTeamAssignment` with its corresponding model instance. This allows consumers like AWX to:
- Batch-process activity stream entries without per-item DB fetches
- Access the content object that was passed to `bulk_give_permissions`/`bulk_remove_permissions`
- Perform efficient bulk sync operations

## Key Changes

1. **New signals** (`ansible_base/rbac/signals.py`):
   - `dab_rbac_assignments_created` - sent from `give_assignments` after creation
   - `dab_rbac_assignments_pre_delete` - sent from `remove_assignments` before deletion

2. **Pipeline modifications** (`ansible_base/rbac/pipeline.py`):
   - Extract content objects from `PermissionTriple` data in bulk functions
   - Build `(content_type_id, object_id) -> content_object` mapping to handle different types with same ID
   - Thread content objects through to signal send sites
   - Modified `_create_assignments` to return only newly created assignments (not pre-existing)

3. **Comprehensive tests** (`test_app/tests/rbac/test_signals.py`):
   - 12 test cases covering both signals
   - Tests for bulk operations, single operations, idempotency, and mixed user/team assignments
   - Verification of signal payload structure and content object pairing

## Signal Behavior

- **Created signal**: Only fires when new assignments are created (idempotent re-creates don't fire)
- **Pre-delete signal**: Fires before deletion so assignment data remains accessible
- **Independent of `fire_signals_on_create`**: Bulk signals fire regardless of the per-item signal flag
- **Content object keyed by `(content_type_id, object_id)`**: Handles different model types with same numeric ID

## Testing

Run tests with:
```bash
pytest test_app/tests/rbac/test_signals.py -v
```

Note: Some existing test fixtures have unrelated errors (`resourcewithadminperm` model missing), but the new signal tests that use simpler fixtures pass successfully.

## Summary by CodeRabbit

* **New Features**
  * Added notifications when role assignments are created or are about to be deleted.
  * Assignment notifications include assignment details and related content-object information when available.
  * Improved assignment processing to distinguish newly created assignments from existing ones.

* **Bug Fixes**
  * Prevented duplicate notifications when permissions are granted repeatedly.
  * Ensured notifications are not emitted when removing nonexistent permissions.
  * Standardized empty content-object data handling.


AlanCoding:
## Database Performance Impact

**No additional database queries are introduced by this PR.**

The signal implementation reuses existing queries that were already happening for `_fire_post_save` and `_audit_log_created`:

1. Query to get existing assignment PKs: `RoleUserAssignment.objects.filter(pair_q).values_list('pk', flat=True)`
2. Bulk create with conflict handling: `RoleUserAssignment.objects.bulk_create(..., ignore_conflicts=True)`
3. Re-fetch to get PKs: `RoleUserAssignment.objects.filter(pair_q)`

The only new cost is **in-memory filtering** to determine which assignments are newly created vs. pre-existing:
```python
new_users = [a for a in db_users if a.pk not in existing_user_pks]  # O(N) with O(1) lookups
```

This filtering enables the signal to fire only for actual changes:
- **Create signal**: fires only with newly created assignments (not idempotent re-creates)
- **Delete signal**: fires only with assignments that actually existed and will be deleted

This matches the existing behavior of `_fire_post_save` which already filtered to only fire for new assignments.

AlanCoding:
## Re: Concurrent call duplicate signals

**Analysis of the race condition:**

The scenario described (concurrent calls both seeing empty `existing_pks`, both firing signals for the same assignment) is theoretically possible, but:

1. **Existing behavior has the same issue**: The current `_fire_post_save` has identical logic - it fires for assignments where `pk not in existing_pks`. This PR doesn't introduce the race, it inherits it.

2. **Requires pathological caller behavior**: The race only happens if the caller invokes `bulk_give_permissions` concurrently with overlapping assignment data. That's already questionable usage - why would you try to assign the same role to the same user on the same object in parallel?

3. **Consumer responsibility**: If AWX connects to this signal and has concurrent calls with overlapping data, AWX's signal handler should be idempotent. DAB provides best-effort notification of changes; handling duplicate notifications is the consumer's responsibility.

4. **Actual damage is minimal**: Worst case is duplicate activity stream entries. Not ideal, but not data corruption - more of an audit log cosmetic issue.

**Recommendation:** Leave as-is. If concurrent overlapping calls are a real production scenario for AWX (which would be surprising), they should either:
- Serialize their calls at the application level
- Make their signal handler idempotent (check if activity stream entry exists before creating)
- Use a distributed lock if they need strict once-and-only-once semantics

Adding locking/tracking to DAB to protect against AWX's concurrent calls with duplicate data seems like solving the wrong problem at the wrong layer.

ansible_base/rbac/pipeline.py
```diff
@@ -319,15 +356,30 @@ def _recompute_after_remove(
...
     if not user_assignments and not team_assignments:
         return
 
+    # Always fire bulk signal BEFORE deletion with whatever content_objects we have (may be empty dict)
+    all_assignments = list(user_assignments) + list(team_assignments)
+    if all_assignments:
+        dab_rbac_assignments_pre_delete.send(
+            sender=None,
+            assignments=all_assignments,
+            content_objects=content_objects or {},
```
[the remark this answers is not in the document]
AlanCoding: reply in main thread

ansible_base/rbac/pipeline.py
```diff
@@ -348,19 +407,34 @@ def give_assignments(
...
         return []
 
     lookup = _ensure_object_roles(list(user_resolved) + list(team_resolved))
-    assignments = _create_assignments(list(user_resolved), list(team_resolved), lookup, fire_signals_on_create=fire_signals_on_create)
-    _recompute_after_give(lookup, assignments)
-    return assignments
+    all_assignments, new_assignments = _create_assignments(list(user_resolved), list(team_resolved), lookup, fire_signals_on_create=fire_signals_on_create)
+
+    # Fire bulk signal only for newly created assignments (not idempotent re-creates)
+    if new_assignments:
```
lallen92: The assignment rows are already bulk created at this point, but what if the receiver raises an exception here, `_recompute_after_give` wouldn't run and the exception propagates to the caller, would result in a data inconsistency. Worth using `send_robust` logging any returned exceptions, so a receiver failure is visible but isolated? 

Similar point for remove_assignments but just the reverse, signal fires before the deletion.

test_app/tests/rbac/test_custom_signals.py
```diff
@@ -0,0 +1,299 @@
...
+
+        # Second call should be idempotent (no new assignments created)
+        mck = MagicMock()
+        dab_rbac_assignments_created.connect(mck.signal_handler, dispatch_uid='test-idempotent')
+
+        try:
+            assignments = bulk_give_permissions(user_permissions=[(org_inv_rd, rando, organization)])
+
+            # Signal should not fire if no new assignments were created
+            if not assignments or all(a.pk is None for a in assignments):
```
lallen92, line 76: Would this check always be false? assignments is a DB re-fetch, so it's never empty and pks are never None.

ansible_base/rbac/pipeline.py
AlanCoding: ...generally, signals are not atomic.

The way you get atomicity is wrapping your _entire_ operation (probably a request) in the transaction.atomic block, and then if an exception is raised everything gets rolled back. This is less obvious, but that rollback includes both the assignments given and anything _done in signals_.

test_app/tests/rbac/test_custom_signals.py
AlanCoding: the methods are idempotent, so there's a real chance that it you bulk assign a list of assignments, and the user(s) already have all those assignments.

The real question is whether you go ahead and send the signal with an empty list. That has appeal.

ansible_base/rbac/pipeline.py
```diff
@@ -348,19 +407,34 @@ def give_assignments(
...
+        dab_rbac_assignments_created.send(
+            sender=None,
+            assignments=new_assignments,
+            content_objects=content_objects or {},
+        )
```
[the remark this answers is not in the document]
AlanCoding: > The stated public contract requires assignment_data with (assignment, content_object) tuples.

Where? Is the `assignments` kwarg not a list or set of assignments? Why would you think otherwise?

test_app/tests/rbac/test_custom_signals.py
AlanCoding: yeah, I'm going to push a commit to remove the conditional. Both more flexible and more simple.

lallen92 (approved):
LGTM

AlanCoding:
Closing in favor of https://github.com/ansible/django-ansible-base/pull/1116

PR #1113, opened 2026-08-26
AAP-87798 Skip corrupted Gateway assignments with UUID object_id

Funi1234:
## Summary

- Detects corrupted Gateway assignments where \`object_id\` contains a UUID instead of an integer PK — a symptom of Gateway data corruption where a RoleDefinition UUID ends up in the wrong field
- Skips those assignments with a warning log rather than crashing with \`ValueError: Field 'id' expected a number but got '<uuid>'\`
- Suppresses the deletion pass for that sync cycle so the matching valid local assignment is not lost (same guard already used for incomplete remote fetches)
- Valid assignments on the same page are still applied normally
- Hardens \`get_content_object\` to use \`filter().first()\` with a content-type check rather than \`get()\`, so a UUID for the wrong resource type falls through to the PK lookup rather than raising an unhandled \`DoesNotExist\`

Fixes: AAP-87798

## Test plan

- [ ] New test \`test_paginate_skips_corrupted_uuid_object_id\` — verifies the corrupted assignment is skipped, \`has_invalid_assignments\` is flagged, and a valid assignment on the same page still passes through
- [ ] New test \`test_sync_assignments_skips_deletions_when_has_invalid_assignments\` — verifies deletions are suppressed when the flag is set
- [ ] Full tox py312 suite passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization handling for corrupted assignment identifiers.
  * Prevented affected assignments from being deleted while allowing unrelated revoked assignments to sync normally.
  * Improved content lookup fallback for registered resources using primary-key identifiers.

* **Tests**
  * Added coverage for selective assignment protection and primary-key content resolution.


tools/test-runner/dab-tox
AlanCoding, line 1: ...did you mean to commit this file?

test_app/tests/resource_registry/test_resource_sync.py
```diff
@@ -1425,3 +1425,98 @@ def test_cleanup_orphans_continues_after_deletion_error(admin_api_client, static
...
     assert User.objects.filter(username__in=["orphan_one", "orphan_two"]).count() == 2
     assert executor.deleted_count == 0
+
+
+@pytest.mark.django_db
+def test_paginate_skips_corrupted_uuid_object_id():
+    """Assignments whose object_id is a UUID are skipped and flagged as invalid.
+
+    This guards against corrupted Gateway data (AAP-87798) where a RoleDefinition
+    UUID ends up as object_id for a namespace assignment.  The corrupted assignment
```
AlanCoding: 😱

AlanCoding:
> Detects corrupted Gateway assignments where `object_id` contains a UUID instead of an integer PK — a symptom of Gateway data corruption where a RoleDefinition UUID ends up in the wrong field

So did you also fix the root cause of this data being the way it was?

Funi1234:
> > Detects corrupted Gateway assignments where `object_id` contains a UUID instead of an integer PK — a symptom of Gateway data corruption where a RoleDefinition UUID ends up in the wrong field
> 
> So did you also fix the root cause of this data being the way it was?

I'm going to be honest, I did not investigate the `why` this is happening.  A brief poke around suggests it could be in gateway given that the data seems to be coming from dab_rbac_roleteamassignment

ansible_base/rbac/role_sync_utils.py
```diff
@@ -84,8 +85,19 @@ def get_content_object(role_definition, assignment_tuple: AssignmentTuple) -> An
         raise ValueError("get_content_object requires a role_definition with a content_type")
     model = role_definition.content_type.model_class()
     if _is_resource_registered(model):
-        object_resource = Resource.objects.get(ansible_id=assignment_tuple.ansible_id_or_pk)
-        return object_resource.content_object
+        expected_ct = role_definition.content_type
```
AlanCoding, line 88: A caller of this method

https://github.com/ansible/django-ansible-base/blob/e4d226b1238fe11fd524e97e412e9a1501cad08e/ansible_base/rbac/role_sync_utils.py#L267-L274

Does `RoleDefinition.objects.get`, without `.prefetch_related('content_type')`. It also should not do this. What you're doing here will incur an additional query, which is unlikely to be acceptable in this location.

AlanCoding:
This looks like the real bug might have been from https://github.com/ansible/django-ansible-base/pull/1093

Still unconfirmed. But that appears to be the real thing, the first thing, we need to be addressing.

PR #1114, opened 2026-08-27
AAP-89166 fix: honour _skip_reverse_resource_sync class flag in pre_save and pre_delete handlers

Funi1234:
Fixes the root cause of PATCH/PUT and DELETE returning 500 for Namespace in Hub (AAP-89166).

## What's going on

Hub recently registered `Namespace` in the resource registry so it can generate stable UUIDs for each namespace — these UUIDs are needed so the Gateway assignment sync can resolve `object_ansible_id` back to a local Namespace instance (AAP-77392). Namespaces are a Hub-only concept though, Gateway has no handler for them, so there's nothing to push upstream. That's why `_skip_reverse_resource_sync = True` is set on the model class — it's the intended opt-out.

The problem is that two signal handlers weren't checking the flag before doing things that crash for this kind of "lightweight" registration (registered for UUID generation only, no managed serializer):

- **`decide_to_sync_update` (pre_save)** — calls `serializer_class().get_fields()` before checking anything. No managed serializer means `serializer_class` is `None`. Calling `None()` raises `TypeError`. Every PATCH/PUT on a Namespace returned 500.
- **`sync_to_resource_server_pre_delete` (pre_delete)** — called `instance.resource.ansible_id` directly. `Resource` uses a `GenericForeignKey` so there's no automatic reverse accessor on the instance. `AttributeError`. Every DELETE on a Namespace returned 500.

## Changes

- `decide_to_sync_update`: check `_skip_reverse_resource_sync` at the class level before touching `serializer_class`
- `sync_to_resource_server_pre_delete`: same class-level check; also replaces `instance.resource.ansible_id` with `Resource.get_resource_for_object()` — `GenericForeignKey` never provided a reverse accessor so the original line was fragile for all models, not just Namespace
- Two new tests covering both handlers with the class-level flag set, with comments documenting the exact exception each path raised before the fix

## Backward compatibility

No behaviour change for any existing model — `getattr(type(instance), '_skip_reverse_resource_sync', False)` returns `False` for all currently registered models (User, Team, Organization, etc.) so both early returns are no-ops for them. The `get_resource_for_object()` swap is functionally equivalent for models where the lookup succeeds, and strictly better for models where it doesn't.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved synchronization handling for models configured to skip reverse resource synchronization.
  - Prevented errors when deleting objects without an associated resource.
  - Added warning handling when no matching resource is found.
  - Ensured deletions synchronize using the correct associated resource identifier when available.

- **Tests**
  - Added coverage for skipped synchronization, missing resources, and successful deletion synchronization.

### DIFF
--- a/ansible_base/resource_registry/signals/handlers.py
+++ b/ansible_base/resource_registry/signals/handlers.py
@@ -62,6 +62,12 @@ def decide_to_sync_update(sender, instance, raw, using, update_fields, **kwargs)
         # We only concern ourselves with updates
         return
 
+    # A class-level flag signals that this model is registered for UUID generation
+    # only and has no managed_serializer. For those models serializer_class is None,
+    # and calling it below raises TypeError. Skip the whole handler.
+    if getattr(type(instance), '_skip_reverse_resource_sync', False):
+        return
+
     try:
         resource = Resource.get_resource_for_object(instance)
     except Resource.DoesNotExist:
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):
The `type(instance)` threw me for a loop. Probably a mostly Hub thing. But of course I'm always in favor of less reverse syncing.

PR #1116, opened 2026-08-28
AAP-90162 Unify RBAC assignment integrations on the bulk signals

AlanCoding:
## Merge order & downstream: this DAB PR lands first, downstream stays green

This is a **DAB-first** merge: this PR goes in ahead of the consumer patches, and the
temporary `_fire_post_save` shim (below) keeps the *old* per-row `post_save`/`post_delete`
receivers working in the meantime. So **downstream integration tests are expected to pass**
both before and after the consumer patches — un-migrated AWX/Hub keep working via the shim,
and the already-written consumer patches (verified against the final contract here) work
once they repoint their DAB pin. DAB's own suite is green, including new ordering-guarantee
tests for the signals.

## Builds on

- **#1115** (AAP-90129, SQLite-depth fix) — **closed / folded into this branch.** Its
  PK-range `_insert_new` (constant-size query, no OR-of-pairs that SQLite rejects past ~1000
  terms) is the sole new-row path here.
- **#1111** (AAP-82996, bulk assignment signals) — **closed / folded into this branch.**, stacked. Purely additive addition of a new signal which will be used by AWX and galaxy_ng.

## What this branch is

The complete solution for epic **AAP-90162** — move external consumers off DAB's per-row
`post_save`/`post_delete` on `RoleUserAssignment`/`RoleTeamAssignment` and onto the bulk
signals (`dab_rbac_assignments_created` / `dab_rbac_assignments_pre_delete`), in one release.

1. **Merge the bulk signals onto the SQLite-depth fix.** Reconciles the two overlapping
   rewrites of `_create_assignments`: keeps #1115's PK-range `_insert_new` as the sole
   new-row path and layers on the bulk signals.

2. **Retire the `fire_signals_on_create` toggle.** The JWT grant path drops the kwarg; every
   bulk path is treated uniformly.

3. **Delete `AwxJWTAuthentication._sync_old_rbac`.** It existed only to compensate for the
   JWT bulk path suppressing `post_save`; with consumers on the bulk signals it's redundant.

4. **Fold global (singleton) assignments into the shared pipeline (AAP-90170).**
   `give_global_permission` / `remove_global_permission` previously created/deleted rows
   directly, so global roles emitted no bulk signal — a gap that would swallow global (incl.
   JWT-granted global) roles once consumers drop their per-row receivers. Rather than adding
   parallel global-only functions, global triples are now first-class in the *same* pipeline:
   a `(role_definition, actor, None)` triple (null content object) flows through
   `bulk_give_permissions` / `give_assignments` (and the removal equivalents) alongside
   object triples, freely mixed. Global rows carry a null `object_role`/`content_type`/
   `object_id` and don't populate `RoleEvaluation`, so recompute is skipped for them. The
   validation gates (`validate_global_assignment`) moved into the pipeline so a bulk caller
   passing a list of global triples can't bypass them. The JWT claim path (`save_user_claims`)
   now folds its global grants into the same bulk call.

## Signal timing: `created` fires after recompute, `pre_delete` before (intentional asymmetry)

`dab_rbac_assignments_created` fires **after** `_recompute_after_give`, so consumers see
fully-settled state — rows exist *and* `RoleEvaluation` reflects them, so `has_obj_perm` /
`accessible_objects` are accurate inline. This matches the past-tense "created" semantics.

`dab_rbac_assignments_pre_delete` fires **before** deletion (and therefore before
recomputation), by design: consumers must be able to read the soon-to-be-removed rows and
resolve their content objects — critical for cascade deletes where the content object is
itself going away.

The asymmetry is deliberate, not an oversight: creation is additive, so its data-rich
moment and its settled-evaluations moment coincide in one signal; deletion is destructive,
so those two moments conflict (rows readable only *before*, fresh evals only *after*), and
`pre_delete` picks the data-rich moment. We deliberately stopped at **two** signals:

- No `pre_create` — it would be redundant (validation is internal, signals don't veto) and
  couldn't carry the uniform `AssignmentBase` payload (no rows exist yet).
- A future `dab_rbac_assignments_deleted` (past tense, mirroring `created`) is the reserved,
  additive extension point if a consumer ever needs post-removal *settled* state — not an
  overload of `pre_delete`.

This is documented in `docs/apps/rbac/bulk_operations.md` (previously the signals were
undocumented), and locked in by ordering-guarantee tests (`created` sees fresh evaluations;
`pre_delete` sees the pre-removal state).

## Removal-path query cost

The bulk-unassign path (`_find_assignments`) and the JWT stale-removal path
(`save_user_claims`) fetch assignment rows fresh from the DB, so consumers dereferencing
`assignment.role_definition` / `assignment.user|team` in a signal handler incurred a per-row
query (2 extra queries flagged in AWX bulk-unassign measurements). We already hold the
materialized `role_definition` and actor — on the resolved triples for `_find_assignments`,
and the single known user for `save_user_claims` — so those are stitched back onto the loaded
rows in Python (no re-fetch, no `select_related` JOIN). The GFK `content_object` is not
attachable this way; it still travels via the `content_objects` dict.

## TEMPORARY: `post_save` is still re-fired (for clean merge order)

To let DAB merge **before** the consumer patches without breaking them, `_fire_post_save` is
re-added and marked TEMPORARY (see the copious comments in `pipeline.py`). It fires per-row
`post_save(created=True)` for every bulk-created assignment, so not-yet-migrated AWX/Hub
receivers keep working, and it drives DAB's own audit in the interim (hence the create path
calls `_fire_post_save` **instead of** `_audit_log_created` — never both, to avoid
double-auditing). This **knowingly regresses bulk performance** in AWX/Hub until they
migrate, at which point it self-corrects. A final DAB PR removes `_fire_post_save` and swaps
the create path back to `_audit_log_created` once both consumers are on the bulk signals.

Because `post_save` now fires on **every** bulk path — including the JWT/SSO claim path that
previously suppressed it — the Hub mirror bug below is already fixed by this branch alone
once Hub repoints its DAB pin.

## The bug this fixes

DAB's JWT/SSO claim path grants roles through `bulk_give_permissions`, and `bulk_create`
skips `post_save`. Consumers that mirror assignments via per-row `post_save` therefore never
saw JWT-granted roles. AWX papered over this with `_sync_old_rbac`; **Hub had no
compensation**, so JWT/SSO-granted roles were silently not mirrored into Pulp roles / Django
groups. Routing every path (including the bulk path) through the bulk signals fixes it.

## Signal contract

Both signals send `sender=None`, `assignments: list[AssignmentBase]` (mixed user/team), and
`content_objects: dict[(content_type_id, object_id) -> instance]`. `content_objects` is fully
populated on the `bulk_give_permissions` / `bulk_remove_permissions` paths (straight from the
input triples, no re-fetch); it may be empty/partial on the lower-level `give_assignments` /
`remove_assignments` paths, where a consumer falls back to `assignment.content_object`.
`created` fires only when rows were actually created (idempotent re-gives are silent);
`pre_delete` fires before deletion (rows still readable).

## Consumer work (separate repos)

Both consumer branches are written and verified against the final contract above (imports
resolve, payloads match, no references to the folded/deleted APIs):

- **AWX** — AAP-90164: remove multiple `post_save` connections and replace with the 2 new connection points via the custom signal https://github.com/ansible/awx/pull/16625, this is where the main optimization work happens
- **galaxy_ng/Hub** — AAP-90165: simple replace of `post_save` with the bulk signal, and keep  `rbac_signal_in_progress()` guard. https://github.com/ansible/galaxy_ng/pull/2699
- **jewel/aap-gateway** - needs a simple update for a `post_save` connection for clearing cache. Not a bug, just a blocker for removal of that code in DAB.

## Summary by CodeRabbit

* **New Features**
  * Added consistent handling for global and object-level role assignments.
  * Improved bulk permission grants and removals across users and teams.
  * Added assignment notifications for permission changes, including global assignments.

* **Bug Fixes**
  * Improved AWX role membership synchronization.
  * Prevented duplicate assignments when permissions are granted repeatedly.
  * Applied validation and audit behavior consistently.

* **Tests**
  * Expanded coverage for global assignments, bulk operations, notifications, idempotency, and audit logging.


ansible_base/rbac/pipeline.py
```diff
@@ -140,92 +174,93 @@ def _ensure_object_roles(requested_assignments: list[ResolvedAssignment]) -> Obj
...
+    needing a pre-existing row must fetch it themselves -- see RoleDefinition.give_permission.
+
+    Callers guard against the empty case (see _create_assignments), so ``assignments`` is
+    always non-empty here.
+    """
+    batch_pairs = {(getattr(a, actor_id_field), a.object_role_id) for a in assignments}
+    max_before = model.objects.aggregate(_m=models.Max('id'))['_m'] or 0
+    model.objects.bulk_create(assignments, ignore_conflicts=True)
+
+    return [row for row in model.objects.filter(id__gt=max_before) if (getattr(row, actor_id_field), row.object_role_id) in batch_pairs]
```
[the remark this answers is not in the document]
AlanCoding: I know, it's fine, very minor consequence.

ansible_base/rbac/pipeline.py
```diff
@@ -385,18 +541,26 @@ def bulk_remove_permissions(
...
+    content_objects = {**user_content_objects, **team_content_objects}
+
+    # Note: no early-out on an empty lookup -- global triples have no ObjectRole but still
+    # need to be found (via object_role IS NULL) and removed. _find_assignments handles both.
     lookup = _lookup_object_roles(user_resolved + team_resolved)
-    if not lookup:
-        return
 
     user_found = _find_assignments(user_resolved, lookup, RoleUserAssignment, 'user_id')
     team_found = _find_assignments(team_resolved, lookup, RoleTeamAssignment, 'team_id')
```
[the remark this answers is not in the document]
AlanCoding: eh, not new here, derailing the intent of this patch.

ansible_base/rbac/triggers.py
```diff
@@ -39,6 +39,45 @@
...
+# Content objects dict is populated when called from bulk_remove_permissions (which has the model instances).
+# Content objects dict is empty when remove_assignments is called directly without content_objects parameter.
+# Consumers can look up content objects as: content_objects.get((assignment.content_type_id, assignment.object_id))
+# Missing key means not provided. See the global-assignment note on the created signal above.
+#
+# REENTRANCY HAZARD: unlike the old post_delete, this fires while the rows STILL EXIST. A
+# handler that mirrors a removal into another system (e.g. editing an old-RBAC M2M) can
+# trigger a reverse-sync that calls remove_permission again, re-finds the not-yet-deleted
+# row, and recurses infinitely. Consumers doing such mirroring MUST guard against reentrancy.
+dab_rbac_assignments_pre_delete = Signal()
```
lallen92, line 79: Question about the final consumer contract: this signal appears to be sent only by `remove_assignments()`. Do assignments removed through an ordinary ORM/resource cascade emit an equivalent bulk notification? The motivation above and in the docs calls out the case where the content object itself is going away, but a Django cascade would not normally enter `remove_assignments()`. The temporary native `post_delete` behavior seems to protect unmigrated consumers during the merge window, but after AWX/Hub move exclusively to this signal they may no longer observe cascade-driven removals. If consumers deliberately do not need those events, clarifying that boundary would resolve this; otherwise, could we add a cascade regression test and retain or provide a notification path for it?

ansible_base/rbac/pipeline.py
```diff
@@ -140,92 +195,168 @@ def _ensure_object_roles(requested_assignments: list[ResolvedAssignment]) -> Obj
...
+    New rows are detected by PK range -- id greater than the max seen before the insert --
+    rather than a per-pair filter, so the query is constant-size regardless of batch shape
+    (skinny or fat) and rides the existing PK index. The freshly-fetched rows are matched back
+    to the in-memory objects we built (by (actor, role_definition, object_role) identity) and
+    those in-memory objects are returned with their PK backfilled. This means the returned
+    assignments keep the caller's related instances (role_definition, actor, object_role,
+    content_object) with no re-fetch, and a row a concurrent caller commits in the id gap is
+    ignored because it has no matching in-memory object. role_definition is part of the key so
+    global assignments (object_role is None) for different role definitions stay distinct.
+    (A genuine same-identity concurrent insert can still be double-counted as created -- the
```
lallen92, line 231: I think this acknowledged race is externally observable under the new created-only signal contract. For the same previously absent identity, caller A can read `max_before`, caller B can insert and commit, A's `ignore_conflicts` insert can then be skipped, and A can still find B's row in the `id__gt=max_before` scan and report it as its own creation. Both callers could consequently emit `post_save(created=True)` / `dab_rbac_assignments_created`, potentially duplicating audit or mirror side effects even though only one row was inserted. Is concurrent assignment of the same identity serialized elsewhere? If not, could inserted-row detection be made concurrency-safe (or the signal contract weakened explicitly) and covered with a two-connection test?

ansible_base/rbac/pipeline.py
```diff
@@ -280,15 +413,36 @@ def _find_assignments(
...
+        if ra.content_type is None:  # global assignment
+            q |= Q(role_definition_id=ra.role_definition.pk, object_role__isnull=True, **{actor_field: ra.actor.pk})
+        else:
+            obj_role = lookup.get((ra.role_definition.pk, ra.object_id))
+            if obj_role is not None:
+                q |= Q(object_role_id=obj_role.pk, **{actor_field: ra.actor.pk})
     if not q:
         return []
-    return list(model.objects.filter(q))
+    assignments = list(model.objects.filter(q))
```
lallen92, line 434: Could we add a removal counterpart to `test_skinny_shape_scale`? The creation path now avoids an OR branch per requested assignment, but this lookup still builds one `Q` node per removal triple. A skinny removal of roughly 1,100 assignments appears capable of reaching SQLite's expression-tree limit in the same way as the original creation query. The relationship reattachment work avoids N+1 dereferences after this query, but seems separate from its expression depth. If the large removal reproduces, grouping/chunking this lookup would keep the advertised bulk create/remove behavior symmetric.

lallen92 (commented):
AI-assisted design review focused on the final signal contract and bulk-operation edge cases. These are questions rather than asserted blockers; please push back where repository or downstream guarantees make the scenarios inapplicable.

ansible_base/rbac/triggers.py
AlanCoding: > Do assignments removed through an ordinary ORM/resource cascade emit an equivalent bulk notification?

Nope nope, not supported. Creating a `RoleUserAssignment` or the like outside of the designed API for DAB RBAC, which is historically `give_permission` is categorically unsupported. The point of DAB RBAC is that it delivers efficient queryset performance (same as AWX system that is its heritage) by maintaining `RoleEvaluation` entries. If you create the assignment on your own, you don't get the role evaluations, and evaluations will be wrong.

I don't know what to say about your point on deletion, other than it is expected to work. Yes, this is a signals-based library, so if you delete without firing Django signals that will leave both stale assignments & evaluations. Thankfully for us, this is a common problem in the ecosystem, and Django makes this hard to do. Even "bulk" deletions will fire signals. It's the `bulk_create` that is the _exception_ in that it does not fire signals. I can go into more detail, but that is how I got into this mess with the current bug to begin with.

ansible_base/rbac/pipeline.py
AlanCoding: Acknowledged. You have accurately described the race condition that still exists.

The whole thing is a workaround that really shouldn't exist, and it is from a pain point specifically with Django `bulk_create` and `ignore_conflicts`. In that case, it doesn't give a return value. Ordinarily, `bulk_create` does.

So we're left with these ugly tradeoffs. We want to be idempotent. But handling `bulk_create` _without_ `ignore_conflicts`... is actually a pattern I'm familiar with from the AWX callback receiver. I know what that looks like, and it would technically solve the problem. But there's a lot of thrashing. You have to try to create the list, and then if there's an error go one-by-one. For our JWT case, this will be the norm, not the exception. So I can't convince myself that's better, and I've concluded that the duplicate logs and duplicate signals are the least-bad option, because we have control of the other side of the signal and it's totally possible to solve there if it ever matters.

ansible_base/rbac/pipeline.py
AlanCoding: Yeah, there's just no way to do this universally. There are 3 fields that constitute identity and uniqueness. So you have a list of 1,100 `(x, y, z)` tuples. But for all the information we have, there can be 1,100 unique `x`, 1,100 unique `y`, and 1,100 unique `z` values. So you'd be sniffing for the special case where there is only 1 single `x`, which yes, will happen in many circumstances. But that's more logic. There's no real way to have a single code path for this.

This (permission removal) _is_ asymmetric with permission assignment. But the entry-by-entry passing isn't unique, `_create_assignments` also passes in a full list of `(x, y, z)`, it just doesn't _look_ like it does it because that uses `bulk_create` and this is exactly the intended use of that method, and it does batching for exactly this reason.

You really can't refactor that out, and I think the only answer is to do your own custom batching logic of the `Q`s.

ansible_base/rbac/pipeline.py
```diff
@@ -7,7 +7,7 @@
 from django.conf import settings
 from django.db import connection, models
 from django.db.models import Q
-from django.db.models.signals import post_save
+from django.db.models.signals import post_save  # TEMPORARY: see _fire_post_save / AAP-90162 merge-order note
```
Dostonbek1: Nit: the `# TEMPORARY` comment on the import is helpful, but consider adding a brief note in `give_assignments` near the `dab_rbac_assignments_created.send(...)` call that during the merge window both `post_save` (from `_fire_post_save` in `_create_assignments`) and `dab_rbac_assignments_created` fire for every new assignment. A consumer that accidentally connects to both would see duplicate processing. The PR description covers this, but a one-line inline comment would make the contract discoverable from the code alone.

Something like:
```suggestion
from django.db.models.signals import post_save  # TEMPORARY: see _fire_post_save / AAP-90162 merge-order note (consumers must connect to ONE signal, not both)
```

Dostonbek1 (commented):
## Code Review Summary

This is a large, architecturally significant PR that replaces the per-row `post_save`/`post_delete` pattern for RBAC assignment consumers with two new bulk signals (`dab_rbac_assignments_created` / `dab_rbac_assignments_pre_delete`), and folds global (singleton) assignments into the shared pipeline. The PR description is exceptionally thorough.

**Files Reviewed**: 12 files, 2132 lines of diff

### Issues Found
- 0 security/safety issues
- 2 correctness concerns (1 minor, 1 informational, both acknowledged by the author)
- 4 code quality / improvement suggestions

### Overall Assessment
LGTM with minor comments. The design is solid, the signal contract is well-thought-out, test coverage is comprehensive (100% on all changed files per Codecov), and the merge-order strategy is pragmatic. The acknowledged race condition in `_insert_new` and the removal-path SQLite limit are acceptable trade-offs that have been discussed.

### Highlights
- Signal timing contract (created after recompute, pre_delete before) is well-reasoned and locked in by ordering-guarantee tests.
- The `_fire_post_save` shim is a pragmatic bridge. The TEMPORARY markers and tracking ticket references make cleanup easy.
- Re-attaching materialized `role_definition` and actor on the removal path in `_find_assignments` to avoid per-row refetches is a concrete performance win.
- The AWX `_sync_old_rbac` deletion (67 lines of production code + ~250 lines of test mocks) is a satisfying cleanup.
- Documentation in `bulk_operations.md` is thorough and covers the key consumer questions.

### General Notes
- The `_find_assignments` OR-of-Q SQLite depth limit on the removal path (raised by lallen92) should be tracked as a follow-up. It won't bite in production (Postgres), but could fail in the SQLite CI job if a future test exercises large removal batches.
- The `_insert_new` race condition (duplicate signals for concurrent same-identity inserts) is the right trade-off given Django's `bulk_create(ignore_conflicts=True)` limitations. Agree with the author that it can be solved on the consumer side if it ever matters.

See inline comments for specific items.

ansible_base/rbac/pipeline.py
```diff
@@ -280,15 +416,36 @@ def _find_assignments(
...
+            obj_role = lookup.get((ra.role_definition.pk, ra.object_id))
+            if obj_role is not None:
+                q |= Q(object_role_id=obj_role.pk, **{actor_field: ra.actor.pk})
     if not q:
         return []
-    return list(model.objects.filter(q))
+    assignments = list(model.objects.filter(q))
+
+    # We already hold the materialized role_definition and actor on the resolved triples we
+    # filtered by, so re-fetching them (even via a select_related JOIN) is wasteful. Attach
```
Dostonbek1: Worth noting for a follow-up: `_find_assignments` still builds one `Q` node per removal triple. A skinny removal of ~1100 assignments would hit SQLite's expression-tree depth limit in the same way the old creation query did (as lallen92 pointed out). The creation path is now fixed via `_insert_new`, but removal is not.

Not a blocker, but maybe add a `# TODO(AAP-90129): chunk Q nodes for large removal batches (SQLite depth limit)` here so it doesn't get forgotten?

ansible_base/rbac/validators.py
```diff
@@ -264,21 +264,53 @@ def validate_team_assignment_enabled(content_type: Optional[Model], has_team_per
...
     if actor._meta.model_name not in ('user', 'team'):
         raise ValidationError(f'Cannot give permission to {actor}, must be a user or team')
 
+
+def validate_assignment(rd, obj) -> None:
+    """Validate the role-definition/object pairing for an object-scoped assignment.
+
+    Only the rd<->object content-type match is checked here; the actor is validated separately
+    (validate_assignment_actor) because it is per-actor, whereas this check depends solely on
+    (role_definition, object content type) and so can be safely validated once per such pair.
```
Dostonbek1: Minor: `validate_assignment` changed from `(rd, actor, obj)` to `(rd, obj)`. If any external code calls this directly with positional args, the old `actor` argument silently becomes `obj`. Since `validators.py` is a public module, it might be worth a note in release notes or a changelog entry. Probably low risk since this is mainly used internally by the pipeline, but worth being aware of.

ansible_base/rbac/models/role.py
```diff
@@ -224,52 +224,69 @@ def remove_global_permission(self, actor):
...
         return assignment
 
     def give_permission(self, actor, content_object):
-        from ansible_base.rbac.pipeline import bulk_give_permissions
+        from ansible_base.rbac.pipeline import _resolve_content_object, bulk_give_permissions
 
         is_user = actor._meta.model_name == 'user'
         perm = [(self, actor, content_object)]
         assignments = bulk_give_permissions(
             user_permissions=perm if is_user else [],
```
Dostonbek1: On an idempotent `give_permission` call, the old `get_or_create` returned the existing row directly. Now the full `bulk_give_permissions` pipeline runs (ObjectRole lookup/creation, `bulk_create`, PK-range scan), and then falls back to a `.get()` query when `created` is empty. For the single-row convenience API this is a minor cost regression. Same pattern in `give_or_remove_global_permission`.

Not a blocker since it's bounded and constant for single-assignment calls, and it keeps the pipeline unified. Just flagging the trade-off.

ansible_base/rbac/pipeline.py
```diff
@@ -385,18 +610,34 @@ def bulk_remove_permissions(
...
 
-    This is the bulk replacement for remove_permission. Deletes assignments,
-    cleans up orphaned ObjectRoles, and runs a single recomputation pass.
+    A None content object denotes a global (singleton) assignment (found by object_role IS
+    NULL); object-scoped and global triples may be freely mixed. This is the bulk replacement
+    for remove_permission. Deletes assignments, cleans up orphaned ObjectRoles, and runs a
+    single recomputation pass.
     """
     if not user_permissions and not team_permissions:
         return
```
Dostonbek1: The double cache clear for global user removals is subtle but correct: this clears on the caller's in-memory actor instance (which `remove_assignments` won't see because it gets DB-fetched rows), and `_clear_singleton_caches` inside `remove_assignments` clears on the DB-fetched instance. The existing comment explains the reasoning well. Just noting this is intentional for anyone else reading the code.

ansible_base/rbac/pipeline.py
AlanCoding: Yeah, agree, that's good to say in the comment for someone who sees it without prior context.

ansible_base/rbac/pipeline.py
AlanCoding: You know, I'm just not happy with not solving this right now. That's where I was going in the other conversation thread with @pb82 . I might just push a commit now that does it.

AlanCoding:
## Temporary Test Summary (Removed)

**Purpose**: Verified SQLite expression-tree depth limit fix for `bulk_remove_permissions` when removing 1000+ assignments.

**Test file**: `test_bulk_remove_sqlite_depth_limit.py` (created, then deleted)

**Before fix**:
```
bulk_remove_permissions with 2500 removal triples fails on SQLite:
  OperationalError: too many terms in compound SELECT
```
Root cause: `_find_assignments()` builds OR-of-pairs Q filter, which exceeded SQLite's ~1000 expression-tree depth limit.

**After fix**:
- Automatic chunking based on `connection.features.max_query_params`
- SQLite: chunks at 999 (safe batch size)
- PostgreSQL: doesn't chunk (unlimited)
- Result: Same behavior and performance characteristics as `bulk_create()`

**Why test was deleted**:
Permanent test coverage provided by `test_bulk_remove_chunking.py`, which exercises chunking logic on both PostgreSQL and SQLite without being database-specific.

---

**Implementation**:
- Added `_find_assignments_chunked()` helper that wraps `_find_assignments()`
- Uses database-aware batching: `connection.features.max_query_params`
- All 21 existing + 3 new tests pass
- Mechanical parity with Django's `bulk_create()` batching strategy

ansible_base/rbac/pipeline.py
AlanCoding: pushed

ansible_base/rbac/pipeline.py
AlanCoding: pushed

ansible_base/rbac/pipeline.py
AlanCoding: pushed batching logic.

ansible_base/rbac/validators.py
AlanCoding: true, but we should have relatively good coverage for downstream tests here.

Dostonbek1 (approved):

PR #1119, opened 2026-08-31
[Aap 85987] Add validation pattern info

daphnemaeve:
### Description
Jira: https://redhat.atlassian.net/browse/AAP-85987

Exposes CleanTextMixin validation patterns in DRF OPTIONS responses so frontends can perform client-side validation before submission.

When ENHANCED_INPUT_VALIDATION_ENABLED is True and a serializer uses CleanTextMixin, the OPTIONS response for each text field includes a JavaScript-compatible regex pattern, a human-readable description, and flags.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Step 1: Set up the aap-dev environment

This feature requires the `ENHANCED_INPUT_VALIDATION_ENABLED` setting to be enabled at deploy time in addition there are a few dependencies that need to be added as fragments. You can use/view this [branch](https://github.com/ansible/aap-dev/pull/913) to mirror the same setup on your aap-dev.

**1a. Add `ENHANCED_INPUT_VALIDATION_ENABLED` setting in aap-dev:**
You can look at the changes in this [branch](https://github.com/ansible/aap-dev/compare/main...prat98:aap-dev:ANSTRAT-1756-add-enhanced-validation-setting-to-deployment) to see how to add the setting

**1b. Configure sources to use the feature branches:**

Run `make configure-sources` and configure the following sources (see [configure-sources docs](docs/how-to-guides/configure-sources.md) for details):

| Source | Repo | Ref (branch) |
|--------|------|--------------|
| gateway | `git@github.com:daphnemaeve/jewel.git` | `ui-validation-proof-of-concept` |
| awx | `git@github.com:daphnemaeve/awx.git` | `awx-test-options-validation`
| eda | `git@github.com:daphnemaeve/eda-server.git` | `test-eda-validation-metadata`
| django-ansible-base | `git@github.com:daphnemaeve/django-ansible-base.git` | `AAP-85987-add-validation-pattern-info-to-metadata` |

gateway: https://github.com/ansible/jewel/compare/devel...daphnemaeve:jewel:ui-validation-proof-of-concept
eda: https://github.com/ansible/eda-server/commit/c8877c7ca66ab29c00157b6fdb0930e38bf1aca6
awx: https://github.com/ansible/awx/commit/b583d466a34067773311c1b67fcff69256746402

**1c. Deploy the environment:**
```bash
AAP_VERSION=devel make aap
```

The `ENHANCED_INPUT_VALIDATION_ENABLED = True` setting is already configured in the devel manifests for controller, gateway, hub, and EDA on this branch, so no additional configuration is needed.

### 1. Curl commands
`curl -s -X OPTIONS -u admin:adminpassword http://localhost:44927/api/gateway/v1/teams/ | python3 -m json.tool`

### 2. Set up AAP UI

Checkout https://github.com/ansible-automation-platform/aap-ui/pull/2879
From aap-ui repository:

`npm ci`
`cd platform`
`export PLATFORM_SERVER=http://localhost:44927`
`npm run start`
- Open `localhost:4100` (should open automatically after running `npm run start` to start platform ui.
- Navigate to Access Management > Teams.
- Inspect Network Tab.
- Note that the OPTIONS response should contain `pattern` and `pattern_desciption` fields for name and description fields.
- Click create / edit team.
- Enter valid input (e.g. `test`).
- Click away.
- Note no validation errors.
- Enter invalid input (e.g. `test<script>`).
- Click away.
- Note validation error.
---

### Required Actions

- [X] Blocked by PR/MR: #XXX
  

### Related PRs

- AAP UI: https://github.com/ansible-automation-platform/aap-ui/pull/2879
- Jewel/Gateway: https://github.com/ansible/jewel/pull/214

## Summary by CodeRabbit

* **New Features**
  * API field metadata now provides client-side validation patterns and descriptions for supported text fields.
  * Enhanced validation metadata is enabled by default for supported API configurations.
* **Bug Fixes**
  * Improved alignment between advertised client-side validation rules and server-side text validation.
  * Corrected field classification for aliased and model-backed serializer fields.
  * URL fields are now treated as free text, while format-specific fields remain excluded.
* **Documentation**
  * Clarified field coverage, JSON field exclusions, and client-side validation limitations.


ansible_base/lib/metadata.py
```diff
@@ -0,0 +1,157 @@
...
+    base = RESOURCE_NAME_PATTERN.replace(r'\Z', '$')
+    # _ZALGO_RE catches 5+ *consecutive* combining marks. _ZALGO_INTERLEAVED_APPROX
+    # additionally catches marks interleaved with base characters, approximating (but
+    # not exactly replicating) the server's _has_dense_combining_marks() sliding
+    # window check -- see the comment on _ZALGO_INTERLEAVED_APPROX in validation.py
+    # for the known gap. Both are best-effort client-side hints;
+    # validate_resource_name() is the actual security boundary.
+    deny_patterns = [_ZALGO_RE.pattern, _ZALGO_INTERLEAVED_APPROX]
+    lookaheads = ''.join(f'(?!.*{_wrap_alternation(p)})' for p in deny_patterns)
+    return f'^{lookaheads}{base[1:]}'
```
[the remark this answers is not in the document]
daphnemaeve: `build_tier1_frontend_pattern()` at [metadata.py:40](vscode-webview://0fj1q8uen75kpqrm90v5fig0k4o5cl4c4rgqhb6onu8b0of5n10u/ansible_base/lib/metadata.py#L40) replaces `\Z` with `$`, and the pattern is only ever shipped with flags: `'u'` ([metadata.py:116](vscode-webview://0fj1q8uen75kpqrm90v5fig0k4o5cl4c4rgqhb6onu8b0of5n10u/ansible_base/lib/metadata.py#L116)) — no m (multiline) flag anywhere in this codebase. Per the ECMAScript spec, $ without the m flag already asserts "position equals input length" — i.e. absolute end of string, with no exception for a trailing line terminator. That's exactly Python's \Z semantics (also no trailing-newline exception, unlike Python's default $). So $ here is already strictly equivalent to server-side \Z; appending (`?![\s\S]`) would be a no-op behaviorally.

vidyanambiar (approved):

ansible_base/lib/metadata.py
```diff
@@ -0,0 +1,155 @@
...
+        _HANDLER_URI_RE.pattern,
+        _INJECTION_RE.pattern,
+    ]
+    lookaheads = ''.join(f'(?!.*{_wrap_alternation(p)})' for p in deny_patterns)
+    return f'^{lookaheads}[\\s\\S]*$'
+
+
+def inject_clean_text_patterns(field, field_info):
+    from ansible_base.lib.serializers.mixins import CleanTextMixin
+    from ansible_base.lib.utils.settings import get_setting
```
AlanCoding: As a new module in `ansible_base/lib/` these imports should go to the top unless there is a reason that this can not go chronologically after. Some of these will import Django settings, and this is used in settings, but I don't think that is a blocker as these things were designed to work with that.

AlexSCorey:
`inject_clean_text_patterns()` decides whether to advertise a `pattern` by checking the DRF
  **serializer** field's Python class (`isinstance(field, serializers.CharField)` / `isinstance(field,
  (serializers.SlugField, serializers.IPAddressField))`), but `CleanTextMixin._classify_fields()` — the
  backend logic it's meant to mirror — decides by checking the **underlying model field's**
  `get_internal_type()` (via `field.source`, which defaults to `field.field_name`). These two signals
  usually agree for auto-generated `ModelSerializer` fields, but nothing keeps them in sync, and the
  field's *name* has no bearing on whether they diverge — only whether the declared serializer field type
  matches the actual model column type for that attribute.

  **Two independent failure directions, verified concretely:**

  - **False positive** (client told "validated" when it isn't): if a serializer field is declared as
  `serializers.CharField()` — explicitly, instead of letting DRF auto-generate the matching type — for an
  attribute whose model column is actually `models.SlugField`/`models.GenericIPAddressField`, the backend
  still correctly excludes it (model field type unchanged), but this code sees a `CharField` instance and
  falls through to advertising a `pattern` the backend never enforces on that field. This requires a
  deliberate serializer-authoring choice to trigger.
  - **False negative** (client told "no rule" when there is one): `models.CharField(choices=...)` is
  always mapped by DRF's `ModelSerializer.build_standard_field()` to `serializers.ChoiceField`, which does
  **not** subclass `CharField` — this happens automatically, with no unusual authoring choice involved.
  The backend still validates it (model field type is `CharField`), but this code's `isinstance(field,
  serializers.CharField)` check fails, so no pattern is advertised even though a real POST/PUT to that
  field would be checked. `CharField(choices=...)` is already common in this repo's own models
  (feature_flags, oauth2_provider, authentication, activitystream) — this direction is the more likely one
  to actually surface.

  Neither is triggered by an existing DAB serializer today, but `CleanTextMixin`/`CleanTextMetadata` are
  meant to be adopted by controller/EDA/gateway/hub, where both patterns are plausible.

  **Scope note:** both directions only affect what's advertised in the OPTIONS response / browsable-API
  form (`CleanTextMetadata.get_field_info()` is only reached from that pipeline). The actual write-time
  validation (`CleanTextMixin.validate()` → `_classify_fields()`) is untouched by this code either way — a
  real POST/PUT is still validated correctly regardless of what OPTIONS said. This is a
  false-guarantee-at-the-description-layer issue, not a bypass of real enforcement.

  **Suggested fix** — replace both `isinstance` checks with a single check against the actual model field,
  mirroring `_classify_fields()` exactly instead of duplicating a separate heuristic:

  ```python
  model = getattr(getattr(serializer, 'Meta', None), 'model', None)
  if model is None:
      return field_info  # not a ModelSerializer — nothing to classify against

  try:
      model_field = model._meta.get_field(field.source)
  except FieldDoesNotExist:
      return field_info  # e.g. a computed/method field with no real model column

  if not hasattr(model_field, 'get_internal_type') or model_field.get_internal_type() not in ('CharField',
  'TextField'):
      return field_info

  This removes isinstance(field, (serializers.SlugField, serializers.IPAddressField)) and
  isinstance(field, serializers.CharField) entirely, so there's one source of truth for "should this field
  be validated" instead of two that can drift apart.

  Two things this touches that are worth calling out up front:
  1. It requires field.parent.Meta.model to exist. That's consistent with CleanTextMixin.validate()
     itself, which already assumes self.Meta.model unconditionally — but the current integration test
     (_CleanNameDescSlugSerializer(CleanTextMixin, serializers.Serializer)) deliberately uses a plain
     Serializer with no Meta.model, specifically because today's implementation doesn't need one. That
     test would need to move to a real model-backed ModelSerializer to keep exercising the OPTIONS path
     (and would then also be able to cover the choices=/ChoiceField case directly, closing the
     test-coverage gap noted separately).
  2. The _make_field() helper and its many call sites in test_metadata.py mock the serializer field
     directly and never set up a model or model field; they'd need to configure
     field.parent.Meta.model._meta.get_field(...).get_internal_type.return_value explicitly per test case
     instead of relying on Mock(spec=field_cls)'s default behavior.

ansible_base/lib/metadata.py
```diff
@@ -0,0 +1,165 @@
...
+    # get_internal_type() to return their own name, so they are naturally
+    # excluded. URLField inherits CharField's 'CharField', so it IS included.
+    model = getattr(getattr(serializer, 'Meta', None), 'model', None)
+    if model is None:
+        return field_info
+
+    try:
+        source = field.source or field.field_name
+        model_field = model._meta.get_field(source)
+    except Exception:
```
ttuffin, line 112: Instead of a broad catch-all Exception here, I think we should narrow it down to `FieldDoesNotExist` which Django's `get_field` raises for unknown fields: https://github.com/django/django/blob/stable/6.1.x/django/db/models/options.py#L670

ttuffin (approved):

PR #1126, opened 2026-09-03
AAP-79229 Make candidates deterministic, aim for managed as has better support in legacy RBAC

lallen92:
## Description
This is predominantly an awx issue, custom roles with identical permissions to built-in roles cause HTTP 500/403 resource creation failures. A custom role with the same permission matrix as a managed role can be selected during legacy RBAC translation, and a multi-word custom name raises ValueError too many values to unpack (expected 2) in [get_role_from_object_role()](https://github.com/ansible/awx/blob/devel/awx/main/models/rbac.py#L589). The compatibility code infers legacy role fields by splitting a role definition's display name into exactly two words, although custom role names have no such constraint and duplicate permission sets make role selection ambiguous.

There were two main issues, the first here in DAB, there is no specified ordering and it also ignores `name`/`managed`/`content_type` when filtering, so if there is a custom or a manged role with the same set of permissions, its indeterministic which gets chosen. Ideally if we default to managed we'd reduce this
issue even occurring. However, this would only reduce the likely of it happening we still need to handle custom roles in legacy rbac translation.  

Related PR: https://github.com/ansible/awx/pull/16639

## Proposed Fix:
Order candidates by `managed` status descending and ID ascending.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Create a custom role whose permissions exactly match a built-in platform role, using a name with more than two words (for example, Test Custom 1).
2. Assign it to a non-admin user or group at organization scope, ensuring the user has the valid organization permission required to create the target resource.
3. Re-authenticate as that user.
4. Create a platform resource or copy an existing one through the API.
5. Depending on which duplicate role definition is selected, the request either succeeds or returns HTTP 500/403.
Note: potentially nondeterministic

Deterministic developer reproduction
Create an ObjectRole backed by the multi-word custom RoleDefinition and call get_role_from_object_role(object_role) directly. This reliably raises `ValueError: too many values to unpack (expected).

### Expected Results
The Automation Controller should gracefully handle and parse custom role names during evaluation without crashing.

If a custom role shares identical permissions with a built-in role, the system should either prioritize the built-in system role during queries or process the custom name without executing a destructive string split that induces a ValueError. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX
  

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role permission assignment and removal consistency.
  * Corrected visibility checks and role reuse behavior for managed roles with matching permissions.
  * Added safeguards to prevent duplicate global user and team role assignments.
  * Improved permission cache updates when evaluation data is prefetched.


pb82 (approved):
@lallen92 this needs to be merged before the respective AWX PR right? Also, this only seems to change the ordering, so LGTM

PR #1134, opened 2026-09-11
AAP-91663: Require organization on OAuth2 apps and lift DRF pin (CVE-2026-73228/73229)

SiddharthRajaraman:
## Description

**What is being changed?**
- Set `organization` to `required=True` in `OAuth2ApplicationSerializer` so OAuth2 application creation rejects missing organization explicitly.
- Remove the `djangorestframework<3.16` pin and require `djangorestframework>=3.17.2` to address CVE-2026-73228 and CVE-2026-73229.
- Regenerate `requirements/requirements_all.txt` via `./updater.sh upgrade`.

DRF 3.16+ ([encode/django-rest-framework#9531](https://github.com/encode/django-rest-framework/pull/9531)) no longer infers `required=True` for nullable `unique_together` fields. DAB relied on that inference for `OAuth2Application.organization`, so unpinning DRF caused `test_oauth2_provider_application_validator` to fail (POST without organization returned 201 instead of 400). The `<3.16` pin was added in #866 as a workaround.

Explicit serializer validation restores the intended API behavior independent of DRF version inference, allowing DRF to be upgraded to a patched version.

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [x] Development environment change
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. `cd requirements && ./updater.sh upgrade` (already run in this PR)
2. `pytest test_app/tests/oauth2_provider/views/test_application.py::test_oauth2_provider_application_validator -v`

### Expected Results
- POST to application-list without `organization` returns **400** with organization required
- `requirements_all.txt` resolves `djangorestframework` to **>= 3.17.2** (currently 3.18.1)

## Additional Context

Fixes [AAP-91663](https://redhat.atlassian.net/browse/AAP-91663).

Related: #866 (added the DRF pin and extended the validator test).

Per @alancoding's guidance: remove pin, run updater, ensure validator test passes.

Analysis assisted by AI.

[AAP-91663]: https://redhat.atlassian.net/browse/AAP-91663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Organization is now explicitly required when creating or updating OAuth2 applications, preventing incomplete application configurations.
  * Updated dependency versions to address security vulnerabilities and maintain compatibility with the required framework behavior.

### DIFF
--- a/requirements/requirements_all.txt
+++ b/requirements/requirements_all.txt
@@ -1,32 +1,32 @@
-asgiref==3.11.0
+asgiref==3.12.1
     # via
-    #   -r requirements/requirements_resource_registry.in
+    #   -r /privaterequirements/requirements_resource_registry.in
     #   channels
     #   django
-attrs==25.4.0
+attrs==26.1.0
     # via
     #   jsonschema
     #   referencing
-certifi==2025.11.12
+certifi==2026.7.22
     # via requests
-cffi==2.0.0
+cffi==2.1.1
     # via cryptography
 channels==4.3.2
-    # via -r requirements/requirements_channels.in
-charset-normalizer==3.4.4
+    # via -r /privaterequirements/requirements_channels.in
+charset-normalizer==3.5.1
     # via requests
-cryptography==46.0.5
+cryptography==50.0.1
     # via
-    #   -r requirements/requirements.in
+    #   -r /privaterequirements/requirements.in
     #   jwcrypto
     #   social-auth-core
 defusedxml==0.8.0rc2
     # via
     #   python3-openid
     #   social-auth-core
-django==5.2.9
+django==5.2.17
     # via
-    #   -r requirements/requirements.in
+    #   -r /privaterequirements/requirements.in
     #   channels
     #   django-auth-ldap
     #   django-crum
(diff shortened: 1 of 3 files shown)


requirements/requirements_all.txt
```diff
@@ -1,32 +1,32 @@
...
     #   social-auth-core
 defusedxml==0.8.0rc2
     # via
     #   python3-openid
     #   social-auth-core
-django==5.2.9
+django==5.2.17
     # via
-    #   -r requirements/requirements.in
+    #   -r /privaterequirements/requirements.in
```
AlanCoding, line 29: this is a "your machine" thing, hopefully get it scrubbed as it's finalized.

AlanCoding:
Great writeup! Love to see that documented, pleasure to read.

I have relatively good trust that we have existing test coverage for this, and the DAB tests, themselves, are demonstrated to be passing. This only adds a serializer field which is pretty minimal and un-objectionable. Honestly cleaner than hacking the extra_kwargs things. As far as the core fix goes :+1: I'm aligned.

The highest-concern downstream effect is jewel. We have a downstream test and that is passing.

As for AWX and Hub downstream tests, I think we have a sanity concern that we need to address before diving any deeper. That is, `djangorestframework>=3.17.2` probably _force upgrades_ those components in the downstream tests. So this might be because _those components_ need to upgrade DRF, and they _haven't_. So here, **you** may have an option to unblock the current work and kick the can to those components. Remove the `>=`, so we do not constrain it. Logically, is this expected to work with the old version? Yes, absolutely. From your rationale, this is setting the field properties instead of using the default, because the problem is that the default changed. Perfectly backward compatible.

So remove that, and see if checks will pass locally, and you may be good to go. _Absolutely_ file issues for those components to upgrade DRF.

PR #1136, opened 2026-09-15
fix: ensure visbility of teams across all organizations

ttuffin:
This change piggybacks off the existing ORG_ADMINS_CAN_SEE_ALL_USERS setting to ensure org admins and superusers have visibilty of teams across all organizations. When editing an organization, they will be able to give teams from other orgs the ability to administer the target organization.

https://redhat.atlassian.net/browse/AAP-83161
https://redhat.atlassian.net/browse/AAP-83164

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. 
2. 
3. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX
  

## Summary by CodeRabbit

* **Access Control**
  * Team listings now consistently respect user visibility permissions.
  * Unauthenticated users cannot view teams.
  * Superusers and authorized administrators can view teams according to configured visibility settings.

* **Role Assignment**
  * Cross-organization role assignments now follow team visibility rules.
  * Authorized administrators can assign roles when the target team is visible.
  * Unauthorized users remain blocked, and hidden-team requests are rejected.


AlexSCorey (approved):
This looks good to me.

wfealdel (approved):
LGTM!

ansible_base/rbac/policies.py
```diff
@@ -16,6 +16,30 @@
 from ansible_base.rbac.validators import permissions_allowed_for_role
 
 
+def visible_teams(request_user, queryset=None) -> QuerySet:
+    """Gives a queryset of teams that another user should be able to view.
+
+    Team visibility is governed by the same setting as user visibility
+    (ORG_ADMINS_CAN_SEE_ALL_USERS) because the UI exposes a single control
```
AlanCoding: If you're going to do this, it would be better to introduce a new setting like `ORG_ADMINS_CAN_SEE_ALL_TEAMS`

But also, there's a specific reason that setting exists for users, and this reason doesn't apply to teams.

The user model isn't tracked in the RBAC system, and this is mostly intentional as you wind up with a bunch of special rules anyway, because permission to modify a user auth gives you permission to _use that user account_.

For a team, the obvious alternative is to just give the desired permission. You can create a global role that gives `view_team` permission and then you just give that role to whomever needs it.

this also "breaks" other expectations of the RBAC system. Like now, it'll show a list of users with view permission to a team. But now... all org admins have view permission to that team, which won't be shown.

PR #1087, opened 2026-08-05
Add input validation utilities and CleanTextMixin

prat98:
## Description

- **What is being changed?**
  This PR introduces a comprehensive two-tier input validation framework for DAB serializers:
  1. **Tier 1 -- `validate_resource_name`** (allowlist): A strict Unicode-aware regex (via the `regex` library) for resource name fields. Uses Unicode general categories (`\p{L}`, `\p{N}`, `\p{M}`) to support all scripts -- Latin, CJK, Cyrillic, Arabic, Indic, etc. Permits letters, numbers, combining marks, spaces, dots, `@`, hyphens, and underscores, with a 512-character limit. Also blocks invisible characters (`\p{Default_Ignorable_Code_Point}`) and Zalgo text (5+ consecutive combining marks). Values are NFC-normalized before matching.
  2. **Tier 2 -- `validate_free_text`** (blocklist): A validator for general text fields that rejects dangerous patterns -- **all** HTML tags (detected via `nh3`, a real HTML parser based on html5ever), unsafe URI schemes (`javascript:`, `vbscript:`, and `data:` URIs with all WHATWG-defined script-capable MIME types), shell substitution (`$(...)`, `${...}`), template injection (`{{ }}`, `{% %}`), and control/invisible Unicode characters. Every input is checked in HTML-entity-decoded, percent-decoded, and NFKC-normalized form (up to three decode passes), so encoded payloads and fullwidth obfuscations cannot bypass the checks. URI scheme matching strips tab/LF/CR per the WHATWG URL parser spec.
  3. **`CleanTextMixin`**: A serializer mixin wired into the DRF `validate()` lifecycle that applies Tier 1 to name fields and Tier 2 to all other `CharField`/`TextField` columns. Supports:
     - **Grandfathering** on update -- unchanged fields are skipped even if they contain legacy violations
     - **JSONField recursive validation** -- traverses nested dicts and lists, validating every string value with Tier 2. Grandfathering works at every nesting depth, not just the top level
     - **`excluded_fields`** for opting out entire fields (e.g., Jinja2 templates, custom login HTML)
     - **`excluded_json_keys`** for skipping specific sub-keys within JSONFields (e.g., PEM keys, passwords)
     - **Audit logging** -- every validation violation is logged at WARNING level with the field name, resource type, user, and client IP (but never the raw input). Logging happens regardless of whether enforcement is on or off, giving operators visibility before enabling enforcement
     - Reports all validation errors at once in DRF-compatible format

- **Why is this change needed?**
[...]

PR #1137, opened 2026-09-15
[AAP-87588] Inject validation patterns into authenticator_plugins schema endpoint

vidyanambiar:
## Summary
- Injects tier 2 validation patterns into the `configuration_schema` returned by `GET /authenticator_plugins/`
- Non-encrypted `CharField` config fields (e.g. `KEY`, `BIND_DN`) get `pattern`, `patternDescription`, and `flags`
- Encrypted fields (e.g. `SECRET`, `BIND_PASSWORD`) and non-string fields (e.g. `START_TLS`) are excluded

JIRA: [AAP-87588](https://redhat.atlassian.net/browse/AAP-87588)

## Motivation
The Gateway UI uses `GET /authenticator_plugins/` to populate both the Create and Edit authenticator forms. Without pattern metadata in `configuration_schema`, the UI has no way to provide client-side validation hints for authenticator configuration sub-fields (AAP-87588).

### Set up the aap-dev environment

This feature requires the `ENHANCED_INPUT_VALIDATION_ENABLED` setting to be enabled at deploy time in addition there are a few dependencies that need to be added as fragments. You can use/view this [draft PR](https://github.com/ansible/aap-dev/pull/913) to mirror the same setup on your aap-dev.

**1. Add `ENHANCED_INPUT_VALIDATION_ENABLED` setting in aap-dev:**
You can look at the changes in this [draft PR](https://github.com/ansible/aap-dev/pull/913) to see how to add the setting

**2. Configure sources to use the feature branches:**

Run `make configure-sources` and configure the following sources (see [configure-sources docs](docs/how-to-guides/configure-sources.md) for details):

| Source | Repo | Ref (branch) |
|--------|------|--------------|
| django-ansible-base | `git@github.com:vidyanambiar/django-ansible-base.git` | `AAP-87588-config-validation-patterns` |

**3. Deploy the environment:**
```bash
AAP_VERSION=devel make aap
```

The `ENHANCED_INPUT_VALIDATION_ENABLED = True` setting is already configured in the devel manifests for controller, gateway, hub, and EDA on this branch, so no additional configuration is needed.

###  Curl command
```
curl -s -u $CREDS http://localhost:44928/api/gateway/v1/authenticator_plugins/ | python3 -m json.tool
```

Expected: Every configuration_schema entry with type: "CharField" that isn't in the plugin's `configuration_encrypted_fields` (e.g. `SECRET, BIND_PASSWORD`, `SP_PRIVATE_KEY`) should have `pattern`, `patternDescription`, and `flags: "i"`. Encrypted fields and non-string fields (e.g. `BooleanField`, `JSONField`, `ListField`) should have none of those keys.

image

## Test plan
- [x] `test_inject_validation_patterns_adds_patterns_to_charfields` — GitHub plugin: `KEY` gets pattern, `SECRET` does not
- [x] `test_inject_validation_patterns_noop_when_disabled` — no patterns when `ENHANCED_INPUT_VALIDATION_ENABLED=False`
- [x] `test_inject_validation_patterns_skips_non_charfields` — LDAP plugin: `START_TLS` skipped, `BIND_DN` gets pattern, `BIND_PASSWORD` skipped
- [x] `test_authenticator_plugins_endpoint_includes_patterns` — integration test through the full endpoint (requires DB)

## Dependencies
- Depends on #1087 (metadata_pr) for `validation_enabled()` and `get_tier2_pattern()` exports

[AAP-87588]: https://redhat.atlassian.net/browse/AAP-87588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Authenticator configuration fields now provide validation patterns and descriptions for supported text inputs.
  - Validation behavior can be enabled or disabled through the enhanced input validation setting.
  - Encrypted fields and non-text configuration fields remain unaffected.

- **Bug Fixes**
  - Authenticator plugin responses now consistently include applicable validation details, helping prevent invalid configuration values.


vidyanambiar:
### PR #1137 Review — [[AAP-87588] Inject validation patterns into authenticator_plugins schema endpoint](https://github.com/ansible/django-ansible-base/pull/1137)
Author: vidyanambiar • Target: devel • +84 / −1 across 2 files

**What it does**
Adds a helper function _inject_validation_patterns that enriches the authenticator plugin configuration schema endpoint with tier-2 input validation metadata (pattern, patternDescription, flags). Only CharField entries that are not in the encrypted-fields list get the pattern injected. The feature is gated behind the ENHANCED_INPUT_VALIDATION_ENABLED setting — when disabled, the function is a no-op.

**What looks good :white_check_mark:**
• Clean feature gating — early return when validation is disabled keeps the hot path untouched.
• Security-conscious — encrypted fields (e.g. SECRET, BIND_PASSWORD) are explicitly excluded from pattern injection.
• Lazy imports — get_tier2_pattern and validation_enabled are imported inside the function, avoiding circular-import or startup-cost issues.
• Focused scope — only CharField instances receive patterns; non-string fields like START_TLS (boolean) are correctly skipped.
• Good test coverage — 4 new tests cover the key cases: enabled + CharField, disabled (no-op), non-CharFields + encrypted fields, and a full integration test hitting the endpoint.

**Observations / minor suggestions :memo:**
1. In-place mutation — _inject_validation_patterns mutates config_schema entries in place. This is fine here because the list is freshly created per request, but a brief docstring noting the side-effect would improve readability (CodeRabbit also flagged docstring coverage at 14% vs the 80% threshold).
2. encrypted_fields type — the code does if field_name in encrypted_fields, which works for both lists and sets. If the encrypted-fields list ever grows large, converting to a set before the loop would be a micro-optimization, but this is unlikely to matter in practice.
3. No error path for missing tier2 pattern — if get_tier2_pattern() returned None or an incomplete dict, the function would raise a KeyError/TypeError. This is probably fine if that function is guaranteed well-behaved, but a defensive check or a brief comment noting the contract would be a nice touch.

**CI Status :traffic_light:**
10 of 16 checks passed (linting, type checks, AWX downstream, CodeRabbit). 6 are still in progress (main test suite py312 / py312-sqlite, downstream consumers eda-server, hub, jewel, and the canary run). No failures so far.

**Overall impression**
This is a clean, well-scoped feature addition. The logic is straightforward, the test cases cover the important branches, and the security-sensitive exclusion of encrypted fields is handled correctly. The suggestions above are all minor polish — nothing blocking.

_[Ansible Engineering] AI-generated | effort: high. Review for accuracy._


[AAP-87588]: https://redhat.atlassian.net/browse/AAP-87588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

vidyanambiar:
The `py312-sqlite` CI check failure is unrelated to this PR. The only failing test is `test_app/tests/lib/logging/test_runtime.py::test_no_log_needed`, which asserts that `caplog` captures no log output:
```
AssertionError: assert 'ERROR    opentelemetry.exporter.otlp.proto.grpc.exporter:exporter.py:466 Failed to export traces to localhost:4317, error code: StatusCode.UNAVAILABLE\n' == ''
```
This PR only touches `ansible_base/authentication/views/authenticator_plugins.py` and its test file (injecting Tier 2 validation patterns into the configuration_schema response). Neither file references logging, OpenTelemetry, or tracing. `test_runtime.py` hasn't been modified in this branch, and its last change (#660) predates this work by a wide margin.

ansible_base/authentication/views/authenticator_plugins.py
```diff
@@ -14,6 +35,7 @@ def get(self, request, format=None):
                 klass = get_authenticator_class(p)
                 config = klass.configuration_class()
                 config_schema = config.get_configuration_schema()
+                _inject_validation_patterns(config_schema, config, klass.configuration_encrypted_fields)
```
AlexSCorey, line 38: It seems that we do not exclude encrypted fields from the validation, for LDAP and TACACS because those encrypted fields are declared on instances of those plugins, not on class.

AlexSCorey (commented):
Just a question about encrypted fields.

ansible_base/authentication/views/authenticator_plugins.py
vidyanambiar: @AlexSCorey Good catch. `configuration_encrypted_fields` is set on `self` in `__init__` for LDAP/TACACS instead of as a class attribute like every other plugin, so `klass.configuration_encrypted_fields` (no instantiation) was silently falling back to `[]`.

Fixed in https://github.com/ansible/django-ansible-base/pull/1137/commits/629087f11e768c97081784821d157e945ecfe023  by declaring it as a class attribute on both, matching the existing pattern (avoids instantiating plugins in the view, which would spin up social-auth strategy objects unnecessarily). Also added tests confirming class-level access works — the existing LDAP test was hardcoding `['BIND_PASSWORD']` instead of using `klass.configuration_encrypted_fields`, which is why it missed this.

I manually verified with `curl -s -u $CREDS http://localhost:44928/api/gateway/v1/authenticator_plugins/ | python3 -m json.tool` that the pattern fields are not returned on BIND_PASSWORD and SECRET in LDAP and TACACS respectively.

AlexSCorey (approved):
Looks good to me.

PR #1138, opened 2026-09-15
AAP-85247 Fix race condition in DABContentType ID assignment

justinc1:
When multiple models need fallback ID assignment in the same batch, they were all calculating max_id + 1 inside the loop, resulting in duplicate IDs and database constraint violations.

This occurred when:
- 2+ models are added to permission_registry in an upgrade
- Their Django ContentType IDs are already taken by other services
- Both hit the fallback path and calculate the same next ID

Example failure:
- shared.organization: max_id=18 → assign id=19 ✓
- shared.team: max_id=18 → assign id=19 ✗ (duplicate key error)

The fix calculates next_available_id once before the loop and increments it for each conflicting entry, ensuring unique IDs.

Reproducer and analysis: attached to
https://redhat.atlassian.net/browse/AAP-85247

## Description

- What is being changed?
- Why is this change needed?

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

## Testing Instructions

https://redhat.atlassian.net/browse/AAP-85247 has attached .tar with docker-compose.yml.
```
docker compose up -d postgresql
docker compose run --rm metrics-init-1
docker compose run --rm metrics-init-2
  # show stack backtrace during post_migration
```

To test fix, update metrics-service-rhel9:latest image with modified create_types.py:
```
cat <<EOF >Dockerfile
FROM registry.redhat.io/ansible-automation-platform-27/metrics-service-rhel9:latest
COPY create_types.py /usr/lib/python3.12/site-packages/ansible_base/rbac/management/create_types.py
EOF

docker build . -t registry.redhat.io/ansible-automation-platform-27/metrics-service-rhel9:latest-fix
# use metrics-service-rhel9:latest-fix in docker-compose.yml.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved identifier assignment for newly created content types.
  - Prevented duplicate identifiers when multiple content types are created together.
  - Ensured fallback assignments skip identifiers already in use or reserved during the current creation process.
  - Preserved identifiers for content types that already have registered matches.
  - Improved identifier consistency during setup and migrations, including cases involving existing records.


ShaiahWren (approved):
Looks good overall, 
The test ordering that CodeRabbit flagged is probably something to fix. 
Needs rebase before merge.

ansible_base/rbac/management/create_types.py
```diff
@@ -25,6 +25,31 @@ def model_class(apps, ct):
...
+        >>> find_next_unreserved_id(5, {5, 6, 7})
+        8
+        >>> find_next_unreserved_id(5, {6, 7})
+        5
+        >>> find_next_unreserved_id(5, set())
+        5
+    """
+    next_id = starting_id
+    while next_id in reserved_ids:
+        next_id += 1
```
AlanCoding: Something is just weird about the way this method is written. And I think it is that `max(reserved_ids)` should be used.

ansible_base/rbac/management/create_types.py
```diff
@@ -25,6 +25,31 @@ def model_class(apps, ct):
...
+        starting_id: The ID to start searching from
+        reserved_ids: Set of IDs that are already reserved
+
+    Returns:
+        The next unreserved ID (>= starting_id)
+
+    Example:
+        >>> find_next_unreserved_id(5, {5, 6, 7})
+        8
+        >>> find_next_unreserved_id(5, {6, 7})
```
AlanCoding:
```python
find_next_unreserved_id(5, {6, 8, 10})
```

should give `11`

ansible_base/rbac/management/create_types.py
```diff
@@ -60,11 +94,16 @@ def create_DAB_contenttypes(
             # of any new entries created here to the id of its corresponding ContentType
             # from the actual contenttypes app, allowing many filters to work
             real_ct = ct_cls.objects.get_for_model(model)
-            if not dab_ct_cls.objects.filter(id=real_ct.id).exists():
```
AlanCoding: I think I see how this is the bug. It checks the database, but a prior item in-loop set a pk to this value.

ansible_base/rbac/management/create_types.py
```diff
@@ -60,11 +94,16 @@ def create_DAB_contenttypes(
...
+            # Check both database AND batch reservations to avoid ID collision
+            if not dab_ct_cls.objects.filter(id=real_ct.id).exists() and real_ct.id not in reserved_ids:
                 ct_item_data['id'] = real_ct.id
+                reserved_ids.add(real_ct.id)
             else:
-                current_max_id = dab_ct_cls.objects.order_by('-id').values_list('id', flat=True).first() or 0
-                ct_item_data['id'] = current_max_id + 1
+                # Skip IDs already reserved in this batch
+                next_available_id = find_next_unreserved_id(next_available_id, reserved_ids)
```
AlanCoding: I think this is where I'm getting all mixed up about your intention.

You shouldn't redefine `next_available_id`, if you're adding it to `reserved_ids` anyway.

Something isn't clear to me.

ansible_base/rbac/management/create_types.py
```diff
@@ -44,6 +69,15 @@ def create_DAB_contenttypes(
...
     content_types = get_local_dab_contenttypes(using, dab_ct_cls)
 
+    # Calculate the next available ID once before the loop to avoid race condition
+    # where multiple entries in the same batch calculate the same max_id + 1
+    current_max_id = dab_ct_cls.objects.order_by('-id').values_list('id', flat=True).first() or 0
+    next_available_id = current_max_id + 1
+
+    # Track IDs reserved in this batch to prevent collision between direct and fallback assignments
+    # A fallback ID could otherwise collide with a later direct assignment from real_ct.id
+    reserved_ids = set()
```
AlanCoding: It would make it make sense if you added all pks of all existing `DABContentType` objects to this list right here.

ansible_base/rbac/management/create_types.py
```diff
@@ -60,11 +94,16 @@ def create_DAB_contenttypes(
             # of any new entries created here to the id of its corresponding ContentType
             # from the actual contenttypes app, allowing many filters to work
             real_ct = ct_cls.objects.get_for_model(model)
```
AlanCoding: we should rename this to `django_ct` for clarity.

ansible_base/rbac/management/create_types.py
```diff
@@ -60,11 +94,16 @@ def create_DAB_contenttypes(
...
-            if not dab_ct_cls.objects.filter(id=real_ct.id).exists():
+            # Check both database AND batch reservations to avoid ID collision
+            if not dab_ct_cls.objects.filter(id=real_ct.id).exists() and real_ct.id not in reserved_ids:
                 ct_item_data['id'] = real_ct.id
+                reserved_ids.add(real_ct.id)
             else:
-                current_max_id = dab_ct_cls.objects.order_by('-id').values_list('id', flat=True).first() or 0
-                ct_item_data['id'] = current_max_id + 1
```
justinc1: Those to lines are bug. We needed to add 2 entries to dab_ct. For the second dab_ct entry we computed the same `current_max_id`, so we got pk/id collision.

ansible_base/rbac/management/create_types.py
justinc1: `reserved` are entries that are not yet in `DABContentType` table, but will be after whole to-be-added `DABContentType` batch is saved.

Current PR would try to have `DABContentType.id` list compact - if `django_ct.id` cannot be used, then use next unused `dab_ct.id`.

Maybe variable `reserved_ids` should be named `dab_ct_reserved_ids`?

Question @AlanCoding : 
> find_next_unreserved_id(5, {6, 8, 10})
> should give 11

There is something wrong if `dab_ct.id=5` would be created? Or it just sounds wrong to have autoincrement id, and then create first 6, 8, 10, then 5.

Honestly, I wanted to just have as small change as possible. I'm not sure what problems would arise if `dab_ct ids` would differ from `django_ct ids`.

justinc1:
@AlanCoding I did ask for re-review, but id not push any code changes yet. I just have some additional questions.

PR #1141, opened 2026-09-17
[AAP-86343] Add performance regression tests for CleanTextMixin

vidyanambiar:
## Description
- **What is being changed?** Adding three performance regression tests to `CleanTextMixin` in `test_clean_text_mixin.py` to guard against accidental performance regressions in the validator.
- **Why is this change needed?** Perf & Scale team feedback recommends testing validation performance "closer to source" as upstream unit tests, rather than only in downstream ATF benchmarks, to isolate this code's cost from DB/serialization/network overhead.
- **How does this change address the issue?** Tests exercise `CleanTextMixin.validate()` directly with realistic payloads (repeated calls, large JSON, long text) to catch linear-time regressions early.

JIRA: https://redhat.atlassian.net/browse/AAP-86343

## Type of Change
- [x] Test update

## Self-Review Checklist
- [x] I have considered performance implications (tests add negligible overhead; <1s total)

### Steps to Test
1. `pytest test_app/tests/lib/serializers/test_clean_text_mixin.py::TestCleanTextMixinPerformance -v`
2. All three tests should pass (repeated-call, JSON payload, long text)
3. Full suite: `pytest test_app/tests/lib/serializers/test_clean_text_mixin.py` (103 tests total)

### Expected Results
All tests pass; no performance regression in existing test suite.

## Summary by CodeRabbit

- **Tests**
  - Added performance regression coverage for text validation.
  - Verifies repeated validations remain efficient across realistic and larger JSON payloads.
  - Adds safeguards for validating substantially longer free-text descriptions without excessive slowdowns.
  - Uses averaged timing checks to improve the reliability of performance measurements and detect scaling regressions.


vidyanambiar:
## Code Review: [AAP-86343] Add performance regression tests for CleanTextMixin

**Verdict**: ✅ READY_FOR_MERGE  
**Score**: 9/10  
**Risk**: LOW (test-only changes)

---

### Summary

This PR adds three wall-clock performance regression tests for `CleanTextMixin.validate()` addressing Perf & Scale team feedback to test "closer to source" as upstream unit tests rather than only in downstream benchmarks.

**All findings from CodeRabbit and autonomous review have been addressed:**
- ✅ Fixed ineffective ratio assertions (now averages over stable iterations, no defeating floor)
- ✅ Added warmup call to stabilize Django model metadata caching
- ✅ Merged adjacent f-strings for cleaner code
- ✅ Added missing docstring to `build_payload` helper
- ✅ Fixed `enable_validation` fixture cleanup (yield + reset for test isolation)

### Test Results

- **All tests passing**: 103/103 in test_clean_text_mixin.py ✓
- **Performance test overhead**: <2 seconds total for all three new tests ✓
- **Black formatting**: Verified ✓
- **CI status**: Passing (1 check complete, 15 running)

### Key Changes

1. **test_repeated_validate_calls_stay_cheap** — Guards against per-call cost regression (1000× calls under 5s)
2. **test_large_realistic_json_payload_scales_linearly** — Guards against quadratic JSON traversal (50 vs 500 entries)
3. **test_long_single_text_field_scales_linearly** — Guards against non-linear text field scaling (2KB vs 100KB)

### Design Notes

Tests use intentionally generous thresholds (order-of-magnitude, not tight) to avoid CI flakiness while still catching realistic performance regressions. Per Perf & Scale feedback, these measure the full `is_valid()` pipeline (not isolated mixin code) to detect regressions in production-like context.

### Recommendation

**Ready to merge.** No blockers. This is a clean test-only addition that strengthens the codebase's regression detection for a security-critical validator, addressing explicit team feedback.

---

[AAP-86343]: https://redhat.atlassian.net/browse/AAP-86343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

larrymou9:
Hello @vidyanambiar Reviewed with an eye toward whether these will actually catch a real performance regression and with the assistance of Claude and I think they will. Nice work.

The three tests map cleanly onto the real algorithmic risk in `CleanTextMixin.validate()`:
      - Per-call overhead (`_classify_fields` walks `model._meta.get_fields()` on every call, uncached) via the repeated-call test
      - Non-linear JSON recursion (`_validate_json_dict`/`_validate_json_list`) via the large-payload test
      - Non-linear regex/decode cost (`_decoded_variants` + `_contains_markup` + the various denylist regexes) via the long-text test

**_Methodology looks solid_**: warmup call before timing, averaged over enough iterations for a stable measurement, and generous multipliers (30x/100x) that leave headroom while still catching an actual quadratic-style regression. The `enable_validation` fixture teardown fix is a good catch too — avoids leaking `ENHANCED_INPUT_VALIDATION_ENABLED` into unrelated tests.

Two optional, non-blocking thoughts (no action needed for this PR to merge):

1. **Tier 1 coverage**: these all exercise the Tier 2 free-text path (`validate_free_text`) and JSON traversal. `validate_resource_name` (Tier 1 name fields) uses the `regex` module with unicode property matching + a sliding-window combining-mark scan, and isn't covered here. Lower risk since it's bounded to 512 chars by the pattern itself, but could be worth a follow-up test if you want full coverage of the mixin.

2. **CI job placement**: these currently run under the default tox env (`py312`/`py312-sqlite`), which uses `pytest -n auto` across all of `test_app/tests` -- so they execute concurrently with many other test processes competing for CPU on the same runner. That's fine for correctness tests, but wall-clock assertions are more exposed to that kind of contention than something like a DB query-count assertion. Worth keeping an eye out for occasional flakiness; if it shows up, a dedicated isolated job (same idea as the `py312-query-count-scale` job added in #1110, which also runs without `-n auto` for a related reason, but kept as its own job rather than combined -- different fixtures/dataset needs and failure domains) would likely fix it. Not blocking now since margins are generous and CI's green.

Nice work, I don't see any concerns blocking this from merging.

ttuffin (approved):

